PR #254, opened 2013-03-21
feat(tooltipProvider): added appendToBody option for $tooltip

### LINKED ISSUE #139: Consider add container option, that solves some bootstrap issues
Hello,

The `container` option was introduced to solve some issues of tootip and popover (like style break in btn-group, popover hidden on modal, etc). So,  IMO, it also could be implemented here.

Reference:
https://github.com/twitter/bootstrap/pull/5768

I would be happy in submit a PR if you agree.

Thanks!

joshdmiller:
Okay, in re #139 I decided to just create a completely independent option called `appendToBody` for now. When `true`, the tooltips will be appended to the body. I do not think this is the best long-term solution, but until someone comes up with something brilliant for containers, this will do.

I intentionally did not update the docs. I think before we unleash the `$tooltipProvider` on the public, we should add a few more options to it, some proper defaults, and a few more detailed tests.

I also did not update the demo as that is not feasible in this case (needs a `config` block). But all the tests pass and I created a [working Plunker](http://plnkr.co/edit/OAtk4gUhqGx8R6ZeYH6T?p=preview).

This doesn't close #139 as that's still an issue, but it provides a good stopgap. 

What do y'all think?

### DIFF
--- a/src/tooltip/test/tooltip.spec.js
+++ b/src/tooltip/test/tooltip.spec.js
@@ -123,4 +123,43 @@ describe('tooltip', function() {
   }));
 });
 
-    
+describe('$tooltipProvider', function() {
+  var elm, 
+      elmBody,
+      scope, 
+      elmScope,
+      body;
+
+  // load the tooltip code
+  beforeEach(module('ui.bootstrap.tooltip'));
+
+  // load the template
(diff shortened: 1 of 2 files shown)


stefano-caporale-wk:
What if I need to append the popover to a custom container?

PR #364, opened 2013-04-27
feat($tooltip): support for custom triggers

### LINKED ISSUE #131: Consider add trigger property to popover
In some cases, I use the popover as a tooltip. Would be nice have a `trigger` property to set the event type, that for now is hardcoded.

joshdmiller:
The `$tooltip` service now has two ways to customize the default triggers.
The `$tooltipProvider` takes a `trigger` option and the `*-trigger`
attribute can be applied to a single element.

The `$tooltipProvider`'s `trigger` option overwrites the default value
but the attribute will overwrite both.

A few logical default triggers are supported out of the box and have an
associated map to determine which hide trigger to use. `mouseenter` ->
`mouseleave`, `focus` -> `blur`, and `click` -> `click`. If any other
trigger is provided, it will be used to both show and hide the tooltip.

Custom hide triggers are not yet supported as they would require some
code trickery due to the way `$observe` works.

Closes #131


krico:
@monkeymonk I want to do the same thing. No luck so far.  :+1:

PR #654, opened 2013-07-11
add feature for the button in the tooltip to hide the tooltip

maxisam:
Here is the case: http://stackoverflow.com/questions/17171361/angular-ui-bootstrap-popover-with-close-button


akhilesh84:
Hi - Did we ever get a closure on this one?

Issue #764, opened 2013-08-05
Proposition for typeahead : Open suggestion view on focus


jdecool:
It could be interresting to automatically open the suggestion view when the input field has focus.

It will allow user to see choices it can make when a filter is used and the list is not editable.

What do you think about ?

Hotell:
I've created this custom combobox directive as a workaround. check it out.
works with latest ui-bootstrap 0.12 and angular 1.3.x
http://embed.plnkr.co/ZtuoTVgPLuMWDT2ejULW/preview

javiermarinros:
Any news on this?

andrewboni:
+1 for getting this in

callicles:
+1

sheymann:
+1, it's really useful on short list of options

piernik:
+1

cristifilip:
+1

scottrogers:
+1

mattvague:
+!

rvanbaalen:
**No more +1 comments please** 

That way we can be notified about comments that actually need attention instead of being flooded with +1's

Enykey:
Can you update Angular-UI doc to reflex the fact that typeahead-min-data cannot be 0 and should be 1 or more, please?

icfantv:
@Enykey please feel free to fork the repo and submit a PR.

shyamal890:
This seems to be one of the most sought after feature yet there is no one who's working on it. WHY?!

wesleycho:
@shyamal890 please exercise patience/professionalism - maintaining a large open source project is difficult, and we have many issues that we have to manage with limited resources. Showing impatience will not get us to consider this faster, and only adds noise since instead of working through the backlog of issues, we are being distracted improperly - we are doing the best we can to get through our backlog of issues and systematically address them, and adding noise will only slow down the project more.

thesalami:
Thank you @chenyuzhcy for answering a 2 year waited question with < 10 lines of code!
We all look forward to the next release.

gandaldf:
@chenyuzhcy is there a way to open it programmatically?
for example... if I set focus like this: `element.find('input').focus();` it doesn't popup.

Kuroky360:
+1

gandaldf:
I created a directive that embeds the Typeahead directive and mimics the functionality of a combobox/select:

    .directive('combobox', ['$compile', '$timeout', function($compile, $timeout) {
    var directive = {
[...]

PR #888, opened 2013-08-26
refactor(typeahead): Add typeahead controller (WIP)

chrisirhc:
Tasks left:

- [x] Clean up duplicated code
- [ ] Add tests
- [ ] See if there are any other additions needed
- [ ] Finalize public API

Current known issues:

- Breaks defaults. Default min length is no longer there.

---

Allows other directives to extend functionality of controller.

Since typeahead-popup is a sibling that is right after the typeahead
input element, the controller can't be shared through the directive's
'require' option. As such, it's shared through the new child scope.

#### Example Typeahead Extension Directive

Here's an example that extends the typeahead directive to open on focus (#764):
http://plnkr.co/edit/UAWfTE
(Note that the example is based on my build which is on my demo branch https://github.com/chrisirhc/bootstrap-1/tree/demo/typeahead-extension , as typeahead needs fixes #886, #887 to get that functionality)

The directive is just:

    .directive('typeaheadOpenOnFocus', function () {
      return {
        require: ['typeahead', 'ngModel'],
        link: function (scope, element, attr, ctrls) {
          element.bind('focus', function () {
            ctrls[0].getMatchesAsync(ctrls[1].$viewValue);
            scope.$apply();
          });
        }
      };
    })

This follows ideas of `ngModel` and its `ngModelController` in AngularJS. The `typeaheadController` provides an API (just another object) that can be overridden, added, or called by other directives.

### DIFF
--- a/src/typeahead/test/typeahead-popup.spec.js
+++ b/src/typeahead/test/typeahead-popup.spec.js
@@ -48,14 +48,13 @@ describe('typeaheadPopup - result rendering', function () {
 
     scope.matches = ['foo', 'bar', 'baz'];
     scope.active = 1;
-    $rootScope.select = angular.noop;
-    spyOn($rootScope, 'select');
+    $rootScope.typeaheadCtrl = jasmine.createSpyObj('typeaheadCtrl', ['_selectActive']);
 
     var el = $compile('<div><typeahead-popup matches="matches" active="active" select="select(activeIdx)"></typeahead-popup></div>')(scope);
     $rootScope.$digest();
 
     var liElems = el.find('li');
     liElems.eq(2).find('a').trigger('click');
-    expect($rootScope.select).toHaveBeenCalledWith(2);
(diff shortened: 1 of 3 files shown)


uriklar:
+1

wesleycho:
There is a lot of good work here, but this is likely very outdated - @chrisirhc any plans on finishing this up?  If not, we can open a separate issue with breaking up typeahead as a task.

chrisirhc:
@wesleycho , let's leave this out since we need some work for Angular 1.3 anyway.
This didn't get in because I couldn't get to writing tests.

chrisirhc:
I'm also hoping to move typeahead out of this repo since it's no longer core to bootstrap.

Kikketer:
I really need this feature... are we simply delayed because we didn't have time to write a test?

chrisirhc:
Yes. Feel free to open a pull request with tests to speed things along.

Kikketer:
Thank you, but I found it was much quicker to find an alternative (AngularStrap in this case).

PR #907, opened 2013-08-28
xml and html5 templates compliance

tomaszkrym:
1. added 'data-' prefix to custom attributes
2. replaced '&', '<' with XML entities &amp; &lt;
3. added "" value to attributes without value


k7sleeper:
HTML5 compliance is a must have for WCAG 2.0 compliance and that's a must have for the US market from december 2015.

pkozlowski-opensource:
@k7sleeper could you please point to the section of WCAG 2.0 which states that "HTML5 compliance is a must"?

k7sleeper:
It seems to me that I read section 4.1.1. of [WCAG 2.0 specification](http://www.w3.org/WAI/WCAG20/quickref/20150226/#ensure-compat-parses) not accurately enough :-( It only requires:

> **4.1.1 Parsing:** In content implemented using markup languages, elements have complete start and end tags, elements are nested according to their specifications, elements do not contain duplicate attributes, and any IDs are unique, except where the specifications allow these features. (Level A)
Note: Start and end tags that are missing a critical character in their formation, such as a closing angle bracket or a mismatched attribute value quotation mark are not complete.

which definitely does not say that one have to be fully HTML compliant.

[^](http://www.paciellogroup.com/blog/2012/02/wcag-2-0-parsing-error-bookmarklet/) gives a good explanation of WCAG conformance in relation to HTML conformance.

You helped me (and my team) much by simply asking that question.
Pls excuse any inconveniences that my statement my have caused.

pkozlowski-opensource:
@k7sleeper the way I read it is that one should write "valid" HTML. It doesn't seem to say that it needs to be XHTML or HTML5.

k7sleeper:
But which "specifications" does the requirement refer to? [This](https://html.spec.whatwg.org/multipage/syntax.html) for example?

PR #1207, opened 2013-10-25
fix(typeahead): support two-way data binding fully

### LINKED ISSUE #981: Typeahead changes input display value with model (array of objects)
I have an array of id-value objects, I want my typeahead to let the user to select a value while having its id in the model (as I can do in select directive with the 'select as label for value in array' expression).
The typeahead displays values correctly while I'm typing, but when I select one from the dropdown list I get its id in the input box:
http://plnkr.co/edit/AMV7OaRMaYkq4UIhL7cP?p=preview

Any help?

rostislav:
This PR implements the full support for two-way data binding in typeahead and fixes the problem of model being shown in the input instead of label when typeahead is used.

Closes #981

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -103,14 +103,14 @@ describe('typeahead tests', function () {
       expect(inputEl.val()).toEqual('Alaska');
     });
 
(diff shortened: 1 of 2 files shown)


wesleycho:
Closing as outdated - feel free to open a more current PR.

PR #1415, opened 2013-12-15
fix(tooltip): update placement on content change

bekos:
Since the recompute of position is a heavy calculation and given
that this happens only when tooltip is visible, ie position
is already calculated, we can reuse it to update the position when
widht/height changes.


rvanbaalen:
Closing this PR due to no activity.

chrisirhc:
Resolved via #3435 .

Issue #1274, opened 2013-11-19
refactor(collapse): use ngAnimate


caitp:
My patch is a WIP, I'll try and push it here tomorrow.

There are some advantages to ngAnimate, such as not needing the `emulateEndTransition()` or whatever it was called, for transition-less browsers.

So yeah, I think it will work out nicely (once it's working at all, I mean).

PR #1444, opened 2013-12-21
refactor(collapse): use ngAnimate

chrisirhc:
Update: After merging #1675, I realized most of that logic can now be moved into collapse, as AngularJS 1.3 supports style options on animation calls. I've refactored it so that accordion and collapse share the same logic.

This is my attempt at #1274 .


chrisirhc:
Updated this PR to move accordion animation logic into collapse directive. Planning to merge this soon.

wesleycho:
This LGTM.

PR #1542, opened 2014-01-08
(WIP) chore(tests): use bower for test lib dependencies

chrisirhc:
**Do not merge!**

Update: I played around with switching version numbers and it looks like the version resolution on Bower is... incomplete. After switching the version number, one still has to go through some prompts in order to get the correct AngularJS version installed. I've opened an issue on bower/bower#1037.

Consider the `npm install` commit but for now I don't think we should use bower.

--- 

Also in this PR, run `npm install` on grunt configuration. I saw this on AngularJS and I think that's how they keep the packages Jenkins up-to-date. I'm thinking it's not that harmful but we can discuss if that shouldn't be in this PR.

Related to package discussion in #1503.


chrisirhc:
Closing in favor of #3461 .

PR #1607, opened 2014-01-18
fix(tooltip): oversize tooltip, arrow position

khusnetdinov:
Good day!  Create issue after solve the problem. If there is need a something additional, please say - I'll add it.


wesleycho:
Closing this as outdated.

This PR needs to be updated, as we cannot assume DOM structure by default since users may be using custom templates.  Feel free to open a new PR that is more current that makes non-breaking changes.

PR #1644, opened 2014-01-22
fix(dropdown): use $animate for adding and removing classes

olostan:
As AngularJS 1.2 was released, there is a need to use $animate for adding and removing classes.

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -41,7 +41,7 @@ angular.module('ui.bootstrap.dropdown', [])
   };
 }])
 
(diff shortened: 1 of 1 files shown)


src/dropdown/dropdown.js
```diff
@@ -54,7 +54,12 @@ angular.module('ui.bootstrap.dropdown', [])
   };
 
   $scope.$watch('isOpen', function( value ) {
-    self.$element.toggleClass( openClass, value );
+    if (value)  {
```
snehatulsi: $animate should have a promise, and the code after that should be inside the resolved function. Else the toggle function would be called before the dropdown is opened.

olostan: This pull request is closed as I see (it was created looong time ago).

PR #1675, opened 2014-01-25
refactor(accordion): use ngAnimate for animations

chrisirhc:
~~Update: Most likely going to change this approach to something similar to what I did for #1772.~~

~~Not ready yet until carousel is fixed with ngAnimate as ngAnimate breaks carousel.~~

- Deprecate collapse module

Consider removing the collapse module after transition module is deprecated as well.

@pkozlowski-opensource , the way I see it, ngAnimate makes animations coupled with the template. I think this makes sense as it's a layer of presentation logic. As such, perhaps having class names only in animations (and templates) might be okay.

TODO:
- [ ] Add in documentation that ngAnimate is now required for animations
- ~~Update when https://github.com/angular/angular.js/pull/5984 is updated~~
- [x] Needs to check animate close event whether it's the correct event.

### DIFF
--- a/src/accordion/accordion.js
+++ b/src/accordion/accordion.js
@@ -1,4 +1,4 @@
-angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
+angular.module('ui.bootstrap.accordion', [])
 
 .constant('accordionConfig', {
   closeOthers: true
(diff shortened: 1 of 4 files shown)


Foxandxss:
@chrisirhc this is old and a bit outdated, I rebased it but there is no animation, I guess I need to learn more about .animation and css animations :P

Foxandxss:
Link to repro: http://plnkr.co/edit/0Mxq8jdqpDSxb7x0tURa?p=preview

chrisirhc:
Rebased and fixed for Angular 1.3. Just needs review.

src/accordion/accordion.js
```diff
@@ -127,4 +127,64 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
...
+ * Animations based on addition and removal of `in` class
+ * This requires the bootstrap classes to be present in order to take advantage
+ * of the animation classes.
+ */
+.animation('.panel-collapse', function () {
+  return {
+    beforeAddClass: function (element, className, done) {
+      if (className == 'in') {
+        element
+          .css({height: '0'})
```
Foxandxss: collapsing does that for you and it seems to be working without that line.

chrisirhc: Indeed, good catch.

Foxandxss:
This fixes #2871 as well.

karianna:
I thought we didn't 'merge' PRs?  Shouldn't this have been cherry picked instead?

chrisirhc:
@karianna This is rebased onto master, there's no Merge commit created. It's just GitHub shows it as "Merged" if the exact commit SHA is preserved after merging.

PR #1754, opened 2014-02-06
refactor(typeahead): add typeahead controller

chrisirhc:
First step to adding typeahead controller. There is some duplicate code but that will be cleaned up in future PRs.


wesleycho:
Closing as typeahead is now in a controller - the controller needs to be refactored still, but now it can be done a little more incrementally.

Issue #1780, opened 2014-02-11
Tooltip show wierdly even if content is whitespaces


thierryx96:
Tooltips show, even if text is white spaces. The content (whitespaces) isn't even displayed, which in my opinion is inconsistent behavior ;) . My fix proposal is checking if the text is empty (as it is already), and whitespace(s) to prevent the tooltip rendering.

![tooltip_normal](https://f.cloud.github.com/assets/4757194/2135150/105ab780-92f9-11e3-8a9f-66039972b526.jpg)
![tooltip_whitespaces](https://f.cloud.github.com/assets/4757194/2135151/105eb556-92f9-11e3-9691-750d9496bb24.jpg)

wesleycho:
Re-opened, issue was never resolved.

PR #1784, opened 2014-02-11
fix(tooltip) : issue #1780, prevent rendering if text is whitespaces

thierryx96:
tooltip : issue #1780, prevent rendering if text is whitespace.

Tooltips show, even if text is white spaces. The content (whitespaces) isn't even displayed, which in my opinion is inconsistent behavior ;) . My fix proposal is checking if the text is empty (as it is already), and whitespace(s) to prevent the tooltip rendering.

Cure : test for empty / blank content.


wesleycho:
Closing this as outdated, feel free to open a new PR based off of current master.

Issue #1812, opened 2014-02-16
Modal + iOS virtual keyboard


stajs:
There is a bug with Bootstrap 3 modals when the virtual keyboard in iOS is displayed:

- http://www.abeautifulsite.net/blog/2013/11/bootstrap-3-modals-and-the-ios-virtual-keyboard/
- https://github.com/twbs/bootstrap/issues/9023

iOS/Safari converts elements that are `position: fixed` to something else (`relative` I think) when the virtual keyboard is shown. I have created a plunk to reproduce the downstream effect on Angular UI Bootstrap:

- http://plnkr.co/edit/Qe7QCn

### Bottom of modal is cut off

1. Open the modal from the link at the top of the page.
2. Tap in to the first input field.
3. Use the virtual keyboard "next input field" control (`>`) to tab to the last input.

![photo 1](https://f.cloud.github.com/assets/2253814/2181710/cc7aff08-9761-11e3-8666-0107a396bb67.PNG)

### Overlay does not cover above modal

1. Open the modal from the link at the bottom of the page.
2. Tap in to input 6.
3. Swipe down (to scroll up).

![photo 2](https://f.cloud.github.com/assets/2253814/2181711/d11dc464-9761-11e3-8956-d7c4e17bc973.PNG)

### Overlay does not cover below modal
1. Open the modal from the link at the top of the page.
2. Tap in to input 7.
3. Swipe up (to scroll down).

![photo 3](https://f.cloud.github.com/assets/2253814/2181712/d578be06-9761-11e3-8cac-eb09b93cc1f4.PNG)

The only workaround that seems to actually work so far involves (ugh) user-agent sniffing. As terrible as that sounds, I have implemented a fix locally and would like to submit a PR for review.

niftylettuce:
@gabesumner thanks for your code!

niftylettuce:
@gabesumner Have you seen this <http://www.abeautifulsite.net/bootstrap-3-modals-and-the-ios-virtual-keyboard/>? /cc @gilluminate @talha-asad @cvrebert @stajs

stajs:
@niftylettuce Yeah, I linked to that blog post when I raised the issue (although he has removed the `/blog/2013/11` from the URL). =)

wesleycho:
I believe that this can be fixed through usage of the `windowClass` and `backdropClass` options, with user agent sniffing in the user's codebase.  I do not believe this is something that UI Bootstrap should address.

[...]

PR #1813, opened 2014-02-17
Candidate fix for #1812 - Modal + iOS virtual keyboard

stajs:
If the user-agent is iOS, then:

1. Add an extra class so that people can choose to opt in to the workaround by adding a style.
2. Adjust the heights for the modal and backdrop.
3. Ensure modal opens inside the viewport by adding top margin.

Style needed to complete the workaround:

```css
.modal-ios {
	position: absolute;
	overflow: visible
}
```

You could add the style inline, but probably better to just offer a hook for people to opt-in to.

Ref: #1812

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -117,8 +117,8 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
     };
   }])
 
-  .factory('$modalStack', ['$transition', '$timeout', '$document', '$compile', '$rootScope', '$$stackedMap', '$userAgent', '$window',
-    function ($transition, $timeout, $document, $compile, $rootScope, $$stackedMap, $userAgent, $window) {
+  .factory('$modalStack', ['$transition', '$timeout', '$document', '$compile', '$rootScope', '$$stackedMap', '$userAgent',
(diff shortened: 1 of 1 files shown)


gluecksmelodie:
Awesome. Is there any chance to upvote this PR to be merged into master soon?

karianna:
@stajs Needs tests and can't be merged

stajs:
Well, it didn't have merge conflicts when I submitted it a year ago ;P

I don't have time in the short-term to investigate and resolve the conflicts, so if someone else feels like digging in that would be great.

@karianna What kind of tests are you looking for? I included some test cases in the referenced issue #1812. It's the kind of visual bug you'll have to run with an iOS device to test, so I included a plunk to reproduce a number of problems.

karianna:
Fair point on the tests (I'll have to look into how we can automate those).

wesleycho:
I do not believe that a user agent sniff is the correct approach in UI Bootstrap here.  What can be done is a user can do the user agent sniffing in their code, and pass in the appropriate strings to `windowClass` and `backdropClass` to add the conditional classes there.

Closing this as invalid.

Issue #1843, opened 2014-02-22
[Proposal] Select directive


bekos:
As I have done with the other directives, I would like to get some feedback (or code review) before submitting the actual PR and as I keep writing more tests.

This is a dropdown select directive that uses dropdown buttons as a placeholder. Some features are:
- Single & multiple selects.
- Group options.
- Keyboard interaction.
- Dynamic option templates.
- More... :-)

My main goal was to keep it's interaction as close as possible to the native select.

Play here: http://plnkr.co/edit/K5SEa6hgdbZXfhR6chE7?p=preview

PR #1849, opened 2014-02-24
feat(dropdownSelect): add select directive

bekos:
Based on #1843.

BTW, I would prefer to have the directive as `prefix-select` :-)


wesleycho:
Going to close as something we won't support due to the internal structure of UI Bootstrap changing quite a bit since this PR, including template urls being different, and the overhead of supporting a new feature.

This is probably best released as a standalone library.

PR #1874, opened 2014-03-02
feat(dateParser): add `dateParser` service

bekos:
Will appreciate if anyone is interested to review or add tests.


cyberwombat:
Shouldn't this provide support to initialize the datepicker with a string value?

PR #1875, opened 2014-03-02
Show / Hide control arrows around timepicker

BenjaminCh:
Add an optionnal parameter on timepicker object in order to control the visibility of the controls arrows.


wesleycho:
The history needs to be cleaned up on this PR.

jnels124:
is this PR ever going to be accepted?

rvanbaalen:
Not sure if we should. Why not add your own custom template or css to achieve this goal?

wesleycho:
It looks like this is already in `master` in a form with `show-spinners`.

PR #1913, opened 2014-03-09
fix(datepicker): works without ng-model

Foxandxss:
This was one of the things we needed to fix on datepicker for the next version. It just needed a guard to check that there is a ngModel present.

For the tests, well, that is a hard part. There is no difference visually on the datepicker without ng-model, the only difference now is that it doesn't throw an exception :P. So I decided to test that the actual month is showed.

Testing for actual date is pretty hard because you can't hardcode anything so I decided to check that the title is the correct one.

I had to improvise a little bit on the year one :P

Ping @bekos

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -81,8 +81,10 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.position'])
     if ( this.element ) {
       this._refreshView();
 
-      var date = ngModelCtrl.$modelValue ? new Date(ngModelCtrl.$modelValue) : null;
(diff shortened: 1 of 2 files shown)


wesleycho:
I'm not sure we should support optional ng-model on datepickers.  This looks harmless enough, but I can't forsee a situation where someone would want a datepicker but have no model attached to it.  Even so, we could just require ng-model be used, even if for dummy purposes.

I feel like a better change would just be to enforce ng-model's presence.

chrisirhc:
I agree with @wesleycho . We should probably require ngModel instead. I also don't think this is a priority for 0.13, so moving this to Backlog.

karianna:
merge conflicts as well.

Foxandxss:
This one won't be merged, i leave it open as a reminder

PR #1922, opened 2014-03-11
feat(datepicker): make widget accessible

bekos:
Tests and some polishing is missing, but I think this is one of the most accessible datepicker's out there! :dancers: 

Added:
- Keyboard navigation in calendar and Escape on popup.
- Focus handling when closing/open popup.
- WAI-ARIA roles

Breaking changes:
- Looping modes is removed. Ie from max mode you don't move to min mode.
- For popup, open on focus is removed, but open with down arrow is added.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -17,9 +17,12 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.position'])
   maxDate: null
 })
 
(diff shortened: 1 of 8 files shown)


rvanbaalen:
Even though this is topic pretty old I completely agree with @jbthoma regarding the `open on focus`. I just spent a little while figuring out what I did wrong because the popup didn't open on focus, which I expected.

I appreciate your efforts for accessibility but you should also keep usability in mind. Opening on focus as default and setting and having the arrow down option as extra / fallback would be way better imho. It's best of both worlds.

PR #1924, opened 2014-03-12
feat(dropdown): add alignment option

troch:
I have added "alignElements" function in position.js for aligning horizontally a target element to a host element, in the same fashion than "positionElements".

I make use of it in dropdown.js for aligning the dropdown menu with the dropdown toggle element.
`<ul class="dropdown-menu" dropdown-align="right">`
Possible dropdownAlign values are:
- left
- right
- left-auto (like left but will prevent overflowing on the right)
- right-auto (like right but will prevent overflowing on the left)
- center

The idea is -if this pull request gets accepted- to apply the same to type-ahead like first attempted in PR #1378

I have not added tests yet, please tell me if they are required.


wesleycho:
Does this only support aligning horizontally and not vertically?

Also, is there any reason why this cannot be done in CSS?

karianna:
@troch - also needs to be rebased with master

troch:
@wesleycho I think you're right, this can be done in CSS.

karianna:
Closing as invalid in that case...

RobJacobs:
@troch Check out this [plunk](http://plnkr.co/edit/XWYE6iIvYmDpwlk8K0A3?p=preview) where I've been playing around with using css to do this.

bholzer:
Am I missing something? I can't add custom classes to the typeahead dropdown unless I use a custom template (which is overkill for me), and I can't add just pass a position property to to directive for where to show the dropdown? And besides, if I am doing this in CSS, I have to use `!important` because I have to override the styles provided to the element by the directive.

I don't understand why this isn't a feature, I'm having to use hacks to work around this

PR #1950, opened 2014-03-18
typeahead: $compile match template after adding to DOM

cherbst:
This allows the template to contain custom directives which use controllers that exists somewhere up in the DOM tree.


wesleycho:
Landed as 03446c56335d5ee0970f9730af215fea1dd53f48, thanks!

PR #1952, opened 2014-03-18
fix(timepicker): Enhanced minutes/hours interaction

Akkuma:
Minutes & hours will now properly validate user input in all cases. Before both only handled 'X1' and now both validate '1X' as well. Additionally, we do not update the minutes/hours model unless needed to via non-keyboard interaction.


wesleycho:
Closing, as this is in need of rebasing on current `master` - feel free to open a new pull request with this feature with all conflicts addressed.

Issue #1837, opened 2014-02-22
Datepicker displays off the bottom of the screen


cgastin:
Since the popup for the datepicker always displays under the input, this becomes a problem when the input is at the bottom of the screen because the popup for the datepicker displays off the bottom of the screen. 

A better solution is to detect the bottom of the screen and re-position the popup above the input. 

I have actually modified v0.10.0 to do this, and will be making a pull request to fix this problem.

gitnik:
:+1: for this functionality. This render the datepicker pretty much unuseable

rvanbaalen:
@cgastin Can you provide a plunkr and confirm this is still an issue in UI Bootstrap 0.13.0?

rvanbaalen:
Closing due to inactivity. Please re-open if you can provide a plunkr and confirm this is still an issue with Angular 1.3 and UI Bootstrap 0.13.

cgastin:
Here is a plunker demonstrating the problem.  

http://plnkr.co/edit/5G45sNCYErtvIqczJRRk

While the scroll bar appears now, and the user can get to the calendar, which I don't believe was the case in earlier versions. When the calendar is going to extend past the bottom boundary of the screen then the calendar should re-position itself to above the input box the same way the datepicker does in Bootstrap.

cgastin:
Since it was closed by the owner, I can not reopen after adding the plunker. Will you reopen, or should I open a new issue.

wesleycho:
The Plunker is invalid - it was lost somewhere it looks like, and needs another reproduction.

icfantv:
Closing in favor of #4762.

PR #1838, opened 2014-02-22
fix(datepicker) - #1837 popup displays off the bottom of the screen

cgastin:
detects the popup off the bottom of the screen, and repositions the
popup above the parent element. This is a fix for #1837

PR #2060, opened 2014-04-14
fix(datepicker): position on screen

mtraynham:
Fixes Bug #1837.  This is jqlite compatible.

This solution alters the suggested positioning with the height/width of the popup element.  To achieve the offsetWidth/offsetHeight of the popup, I used the jquery trick of setting the display to block which generates the offsetWidth/offsetHeight and then set it back after calculation.

I know there is already a pull request (PR #1838) for this, but it was answered from 2 months ago, required jQuery and also didn't compensate for the left/top and right boundaries.


icfantv:
Closing in favor of #4762.

PR #2127, opened 2014-04-30
Foundation for a multi-select directive

nathantsoi:
With this feature, we can build a multi-select like so: http://plnkr.co/v4dOw5p8R5uNPZ0m7cQa

The next step is to move this into a directive.


wesleycho:
Closing due to age

PR #2136, opened 2014-05-03
fix(timepicker): Scrolling over without focus should't change value

### LINKED ISSUE #2046: timepicker: Scrolling over the timepicker without focus on it should't change the value
When you scroll the page and the mouse gets over a timepicker, the page scrolling will stop and instead the value of the timepicker will change.
That happens quite often by accident. In our case we have three timepickers side by side.
I think it would be better if the value of the timepicker only changes when you have focused the timepicker before.

Sanjo:
This fixes #2046. I just added a focused check in the mousewheel/wheel event handlers. In the related four tests I needed to attach the element to the document body to make .focus() work. I added two additional tests that ensure that scrolling will not change the timepicker value when the input element is not focused.


wesleycho:
This needs to be behind a config flag since it is a breaking change.

Closing due to age & needing work - feel free to reopen a PR against current `master` with this addressed.

PR #2142, opened 2014-05-04
fix(datepicker): Convert `init-date` to Date when it's a String

jtomaszewski:
Example, which currently is not working, but this commit fixes it:
```html
<input type="text" datepicker-popup="true" datepicker-options="{'init-date': '2014-02-18'}" />
```


wesleycho:
Closing, as it has been decided that implicit conversion is on the pipeline to be chopped in favor of explicit definitions for dates.

PR #1848, opened 2014-02-24
feat(popover): support templates

chrisirhc:
**Update: This is ready to merge.**
~~Here's a poll for anybody to get involved in voting to name this directive:
http://goo.gl/forms/oxW1B4L81B~~ Thank you. It'll be called `popover-template` and `tooltip-template`.

Update: I finally got to this. This should be ready for testing.

TODO:
- [x] Needs Tests
- [x] Needs documentation
- [x] Doesn't support live updating of the positioning after the template has been reloaded
- [x] Doesn't support changing of the template URL (there's no watcher)

This is a proof of concept to use a tooltipController to transclude a template.

It's based off #1600 and most of the changes are from that. See https://github.com/chrisirhc/angular-ui-bootstrap/commit/0a1510c957bd385db61dcb1a8366fef97fd38e74 for actual changes for the popoverWindow.
record

Tests are missing but I wanted to open this up for discussion or comment.

Fixes #220

PR #2151, opened 2014-05-06
Tooltip :: Enable ngModel to use parent scope

mattslocum:
Tooltip used to use a child scope. So elements that use ng-model with primitives don't update the parent scope. http://jsfiddle.net/mslocum/CEqA8/

Solution:
This makes ngModel still have 2-way data binding with elements. http://jsfiddle.net/mslocum/CEqA8/2

### DIFF
--- a/src/alert/test/alert.spec.js
+++ b/src/alert/test/alert.spec.js
@@ -50,6 +50,16 @@ describe('alert', function () {
     expect(alerts.eq(2)).toHaveClass('alert-warning');
   });
 
+  it('should respect alert type binding', function () {
+    var alerts = createAlerts();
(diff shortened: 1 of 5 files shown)


wesleycho:
@chrisirhc any thoughts on this now that the tooltip refactor is done?

mattslocum:
It looks like #1848 just got merged 9 days ago. I'll have to give it a try, but I doubt my PR is needed anymore.

chrisirhc:
Yep, this shouldn't be required anymore. The refactor has removed the scope definition on the tooltip directive. @mattslocum feel free to open an issue if you find that this isn't resolved, as it's definitely a bug.

PR #2153, opened 2014-05-06
modal component: "resolve" now also handles named services

chessydk:
The documentation for modal - resolve prop states "members that will be resolved and passed to the controller as locals; it is equivalent of the resolve property for AngularJS routes"

But AngularJS routes can resolve a service by name (a string) which the modal component can not. My fix solves it by using $injector.get to resolve it when it is a string.


pkozlowski-opensource:
@chessydk we haven't heard back from you and we can't add code without associated unit tests.... Closing for now, feel free to open another PR with the all the listed items tackled. Thnx!

PR #2177, opened 2014-05-09
fix(datepicker): Fix `init-date`

wleelavivat:
When passing 'init-date' through datepickerOptions, need to create Date object again
Also, this allow passing 'init-date' as string


wesleycho:
This PR has a lot of bit rot - can you update the PR and squash your commits?

wesleycho:
Actually, closing this as I think 98e2bdfc8b4dd108acdf3fac75a795675889bbc7 fixes all of these issues.

PR #2181, opened 2014-05-09
feat(typeahead): input with zero length

OronNadiv:
Supports input with zero length including focus event. Inspired by https://github.com/zacronos/bootstrap


wesleycho:
This PR has a lot of history pollution.

Feel free to open a new PR with more details that explains what is being accomplished better and a cleaned up history.

PR #2232, opened 2014-05-21
feat(carousel): no-wrap attribute

Firfi:
Hey,

I wonder how to implement impossibility to wrap carousel around (i.e. user can't move it back when there's first slide and can't move it forward when there's last slide). This is possible implementation.

P.S. Implementing this I found a problem with and old test 'should hide navigation when only one slide' for carousel. 
If I change scope.slides=[{active:false,content:'one'}]; to scope.slides=[{active:false,content:'one'}, {active:false,content:'two'}]; test doesn't fail. I tried to add another $apply() call but it doesn't seems to work so I leave it as it is.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -123,41 +123,53 @@
  * `on-select-page` is removed since `ng-change` can now be used.
 
   Before:
-
+  
+  ```html
   <pagination page="current" on-select-page="changed(page)" ...></pagination>
-
+  ```
+  
   After:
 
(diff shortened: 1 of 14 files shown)


wesleycho:
This history on this PR is messed up, likely due to usage of `git merge` instead of `git pull --rebase origin master`.

The actual changes does not look to be much - I am closing this due to the history problems, if this is desired, please open a new pull request based off of recent master.

PR #2254, opened 2014-05-24
feat(tabs): it should not select first not active tab as selected

karol-f:
Tabs should behave such way that if you specifically set tab.active to "false" is should not be selected even if it's first tab.

This behavior do not change current way tabs are working. It only adds new behavior if we specifically set active === false.

Test are added to cover new funcionality.


wesleycho:
Landed as 91b5fb62eedbb600d6a6abe32376846f327a903d - thanks!

Closes #2742

Issue #2227, opened 2014-05-20
PaginationController init not called before $watch triggered


pschuegr:
Hi there,

I'm having a minor issue with the PaginationController.  If the $watch on 'totalPages' is triggered before init is called, the ngModelCtrl.$render does not exist:

https://github.com/angular-ui/bootstrap/blob/master/src/pagination/pagination.js#L8

Maybe the watch should be set up in the init?

PR #2257, opened 2014-05-25
fix(pagination): fixes issue when init called after watch triggered

pschuegr:
Sorry for delay.  Tests pass.  Fixes https://github.com/angular-ui/bootstrap/issues/2227


wesleycho:
Thanks for this - this should land in the next release

Issue #2204, opened 2014-05-15
Dropdown button and ngTouch


stha:
Hello,

i'm using the dropdown directive to realize a language selection. I'm also using the ngTouch module to get enhanced touch device support. This works fine on the inner `<li><a ...></a>`-Elements but not on the button. The `ng-click-active` css class is not added like documented in the angularjs documentation. See https://docs.angularjs.org/api/ngTouch/directive/ngClick.

What I'm doing wrong?

```html
<div class="push-right cp-btn-dropdown cp-language btn-group" dropdown is-open="language.isopen" ng-controller="Language">
            <button type="button" class="btn btn-default dropdown-toggle">
               <span class="flag_{{currentLang}}"> </span> <span class="caret middle"></span>
            </button>
            <ul class="dropdown-menu" role="menu">
                <li><a href="" ng-click="changeLanguage('de_DE')"><span class="flag_de_DE"></span></a></li>
                <li><a href="" ng-click="changeLanguage('en_US')"><span class="flag_en_US"></span></a></li>
            </ul>
</div>
```

Thx for your time!

wesleycho:
Closing, as this is a user error.

kotmatpockuh:
@wesleycho same issue...

PR #2263, opened 2014-05-27
added  ng-click attribute to enhance ngTouch support on dropdown buttons

stha:
See #2204.


wesleycho:
This is an auto-generated file - you should modify the correct location in the dropdown README.md file.

PR #2271, opened 2014-05-28
Proposal to delay datepicker template rendering

UNOwen:
I was wondering if it would be possible to delay the compilation of datepicker popup template until it is about to be displayed. The issue I had with instant compilation, is that when adding a bunch of datepicker popups on one page (8 in my case), page rendering lags visibly (~0.5 sec using eyeballing measurement, on a lower-end laptop). 

The way I implemented it seems to work for me, but it does break some of the tests. The commit below is merely a proposal to consider a (rather) minor optimization.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -596,7 +596,7 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
       var $popup;
       function ensurePopupCompiled() {
         if (!angular.isUndefined($popup)) {
-          return;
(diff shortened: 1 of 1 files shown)


rvanbaalen:
(By now) Travis is failing. Can you make it compatible with the latest UI Bootstrap?

RobJacobs:
I think this will get covered with #3666 since most of the time consuming rendering is wrapped in an ng-if.

chrisirhc:
Closing in favor of #3666 .

PR #2274, opened 2014-05-28
Spinner

enkodellc:
I have a simple spinner ui directive I would like to contribute to UI-Bootstrap. It is a numeric spinner (up down). I forked boostrap but I am new to both github, angular and the testing so I was not going to build the tests until I was confident that it was not a waste of time. Please take a look at what I have and let me know if I should pursue finishing the testing and documentation or if this is something someone wants to take on. Thanks.


wrummler:
+1 This would be great to have. I've got a large Angular 1.x project using Angular UI Bootstrap, and I recently had to include both jQuery and jQuery UI simply to use the jQuery UI Spinner. It'd be nice to go back to plain Angular + UI Bootstrap.

wesleycho:
This needs unit tests.

@wrummler this should probably be easy to code oneself - in fact, including jQuery & jQuery UI for what appears to be a simple component sounds like overkill. Those are the exact cases where it is better to homebrew the widget.

enkodellc:
@wesleycho I will add tests if this will be considered to be added to angular-ui. It was my first PR on github over a year ago. Please let me know if I should spend time on it, thanks.

pkozlowski-opensource:
@wesleycho @enkodellc I apprecuate all the worka and welcom all contributions to this project, but I dont think we should be including the proposed directive in this repo. firstly it isn't using bootstrap css so could be a directive in angular-ui but not here. Secondly it implements the functionality of an input of type number that exists nativelly in many browsers, so i would expect that there will be less and less need for this type of directives.

once again, i do appreciate the effort but we need to draw a line somewhere

enkodellc:
@pkozlowski-opensource Thanks for the response. After learning Angular and using Git more I completely agree. I will close this PR. Thanks for you great work.

wesleycho:
I agree too @pkozlowski-opensource, I was just giving the opportunity for response first - glad to see we're all on the same page.

Thanks for the effort though @enkodellc!

Issue #2268, opened 2014-05-28
Popover append-to-body is positioned incorrectly in FireFox


bheithaus:
Due to an issue in $position factory, specifically the offset function, this line here:

[src/position/position.js:75](https://github.com/angular-ui/bootstrap/blob/master/src/position/position.js#L75)
```
top: boundingClientRect.top + ($window.pageYOffset || $document[0].body.scrollTop || $document[0].documentElement.scrollTop),
```

when I click a popover with append-to-body="true", the top property, which should be around 880px, is coming back as 2211px, and the popover is being placed way down below the rest of the page.

{top: "2211px", left: "813.8499755859375px"}

if I understand correctly, the problem here is that $window.pageYOffset is 0, which should be used, but is skipped over because of the ```||``` operator.

so I made a pull request, I hope it is helpful.

BobbieBarker:
It looks like if #2275 get's merged in that this issue will be fixed. I"m going to close this one for now, no need to keep a duplicate open, we can reopen if if #2275 doesn't resolve this issue.

PR #2275, opened 2014-05-28
fix(popover): position#offest - FireFox bug

bheithaus:
fix position factory for FireFox in append-to-body="true" case

this is for issue #2268


wesleycho:
Should note that this doesn't fix #2268, but would help give a mechanism for fixing it it seems.

wesleycho:
We will want unit tests for the two new functions.

wesleycho:
Closing due to age - feel free to open the PR again with accompanying Plunkers that demonstrate the failure in current `master` and the fix in action.

PR #2284, opened 2014-05-30
Add backdrop class to new modals

davidwparker:
Given the ability to add backdrop classes:
https://github.com/angular-ui/bootstrap/pull/1862

When a modal calls another modal, due to some race conditions:
https://github.com/angular-ui/bootstrap/blob/master/src/modal/modal.js#L159
https://github.com/angular-ui/bootstrap/blob/master/src/modal/modal.js#L170
a modal's background may be reused and the new backdrop won't have the new classes.

The new modal should get any new backdropClasses provided.

I wasn't sure how you would expect to test this, so I didn't provide any tests, sorry...


wesleycho:
Closing due to bitrot & work being needed.

Feel free to open a new PR that addresses all of the concerns.

PR #2289, opened 2014-06-02
Modal is incompatible with Protractor

hazeledmands:
Protractor uses Angular's [`$browser.notifyWhenNoOutstandingRequests`](https://github.com/angular/protractor/blob/f23565d5db4fbb102cfec8311ce9dfaee52e9113/lib/clientsidescripts.js#L28) to determine when it's okay to continue sending commands via webdriver.

Right now, when Protractor does an action that results in a modal being shown, `notifyWhenNoOutstandingRequests` gets called immediately, often before the modal is done displaying, resulting in flakey specs.

This commit contains a failing test simulating this behavior.

### DIFF
--- a/src/modal/test/modal.spec.js
+++ b/src/modal/test/modal.spec.js
@@ -1,5 +1,5 @@
 describe('$modal', function () {
-  var $controllerProvider, $rootScope, $document, $compile, $templateCache, $timeout, $q;
+  var $controllerProvider, $rootScope, $document, $compile, $templateCache, $timeout, $q, $browser;
   var $modal, $modalProvider;
 
   var triggerKeyDown = function (element, keyCode) {
@@ -24,7 +24,7 @@ describe('$modal', function () {
     $modalProvider = _$modalProvider_;
   }));
 
-  beforeEach(inject(function (_$rootScope_, _$document_, _$compile_, _$templateCache_, _$timeout_, _$q_, _$modal_) {
+  beforeEach(inject(function (_$rootScope_, _$document_, _$compile_, _$templateCache_, _$timeout_, _$q_, _$modal_, _$browser_) {
     $rootScope = _$rootScope_;
     $document = _$document_;
     $compile = _$compile_;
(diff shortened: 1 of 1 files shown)


wesleycho:
Can you take a look at this again @chrisirhc?

wesleycho:
This appears to still be an issue, likely because $modal is not completely relying on ngAnimate currently for opening the modal.

bjaanes:
Are there any workarounds for this?

icfantv:
@bjaanes not to my knowledge.  You may try reaching out to the StackOverflow community to cast a wider net.  If you go that route, in order for people to want to help you, I would strongly encourage you to show you've done research and know exactly why the problem exists as well as creating a minimally working plunker that demonstrates the issue.

wesleycho:
@icfantv the test makes it obvious what the problem is - we just need to figure out what is going on that is causing it to fail. Incidentally, I did take another look at this in the past few days, but still am not sure what is causing the problem.

This is up for grabs for anyone to fix.

icfantv:
Ok, apologies.

thorn0:
Have a look at the [implementation](https://github.com/angular/angular.js/blob/v1.5.8/src/ng/browser.js#L74) of `$browser.notifyWhenNoOutstandingRequests` and you'll see why the test fails. 
```js
  self.notifyWhenNoOutstandingRequests = function(callback) {
    if (outstandingRequestCount === 0) {
      callback();
    } else {
      outstandingRequestCallbacks.push(callback);
    }
  };
```
If you call this function before starting an async operation (with $timeout or $http), the callback will be just executed immediately. Especially in unit tests where we have the mock `$browser` service, whose [implementation](https://github.com/angular/angular.js/blob/v1.5.8/src/ngMock/angular-mocks.js#L171) of `notifyWhenNoOutstandingRequests` is even simpler:
```js
  notifyWhenNoOutstandingRequests: function(fn) {
    fn();
  }
```
I haven't investigated the Protractor's side of this issue. Just saying that the test is invalid.

ahmehri:
@chrisirhc did this issue have been resolved?

PR #2297, opened 2014-06-04
fix(typeahead): support for complex views and complex models

joshvillbrandt:
This is not in state to be merged, I'm simply sharing / continuing a conversation.

This commit adds back the ability to do object attribute labels when the model is also an object attribute. Three tests are affected by this:

* `typeahead tests initial state and model changes should correctly render initial state if the "as" keyword is used`
 * this is currently broken - discussion below and feedback requested!
* `typeahead tests selecting a match should correctly update inputs value on mapping where label is not derived from the model`
 * the select syntax was updated here from `state as` to `state.name as`
* ` typeahead tests non-regressions tests issue 786 - name of internal model should not conflict with scope model name`
 * I'm not sure if this test applies anymore - I've commented it out

So the initial state of the typeahead is broken at the moment. In order to correctly render the ng-model viewValue, the whole object needs and not just the modelValue. If the modelValue is the object, then this shouldn't be a problem, otherwise the typeahead would need to fetch from the full object, probably from the server. Using the viewValue wouldn't be a sure way of finding the original object, so this would likely be a function passed in through an extra directive attribute for initializing a typeahead with this scenario.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ We are always looking for the quality contributions! Please check the [CONTRIBUT
 You can generate a custom build, containing only needed modules, from the project's homepage.
 Alternatively you can run local Grunt build from the command line and list needed modules as shown below:
 
-```
+```javascript
 grunt build:modal:tabs:alert:popover:dropdownToggle:buttons:progressbar
 ```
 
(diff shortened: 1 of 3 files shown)


wesleycho:
Closing as old - feel free to open a new PR against current `master` with this if there is still an issue.

PR #2306, opened 2014-06-08
fix(datepicker): week count issues

a5sk4s:
This fixes a few issues with week counting

(1) Week counts in January are counted as last years' weeks when the first Thursday is from last year (e.g., 53, 54, 55, ...).  See January 2010, for example.

(2) Week counts in December continue counting the last week(s), which may be part of next year, as weeks from the current year.  See December 2014, for example.

(3) Week counting changes depending on the start day.  The week containing June 7th 2014 is week 22 with the default start day (Sunday), but week 23 with start day Monday.  See examples on http://angular-ui.github.io/bootstrap/ (popup has start day Monday)

The code has been changed so that the Thursday of the week always determines, which week count the week gets assigned, regardless of the start day of the week.


rvanbaalen:
@a5sk4s Sorry that this PR took so long. Can you confirm that it's still relevant in the most recent version? If so, we'll be pulling it in soon.

yanivefraim:
Related: #3321, #3158, #3148

a5sk4s:
I believe it is still relevant - do not have a lot of time at the moment to dive into it again - just fixed a left ddescribe in the spec file, though and merged master into the branch - happy to help, if there are any questions, though.

rvanbaalen:
Thanks for double checking! We'll be pulling in PRs soon and this one looks pretty good. Hold on tight!

---

chrisirhc:
It looks like `getISO8601WeekNumber` is already getting the Thursday of the week [on these lines](https://github.com/angular-ui/bootstrap/blob/0bcd30c4a6d65763e1c4d1b6820a4537a11e95ee/src/datepicker/datepicker.js#L278-L285). Is the logic wrong?

chrisirhc:
Removed my previous wrong comment as I understand the problem better now.

src/datepicker/test/datepicker.spec.js
```diff
@@ -1711,4 +1711,75 @@ describe('datepicker directive', function () {
...
+      it('is saturday', function() {
+        element = $compile('<datepicker ng-model="date" starting-day="6"></datepicker>')($rootScope);
+        $rootScope.$digest();
+  
+        expect(getWeeks()).toEqual(['23', '24', '25', '26', '27', '28']);
+      });
+    });
+
+    describe('first week in january', function() {
+      beforeEach(inject(function() {
```
chrisirhc: This empty beforeEach can be removed

src/datepicker/datepicker.js
```diff
@@ -259,9 +259,10 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
 
         if ( scope.showWeeks ) {
           scope.weekNumbers = [];
-          var weekNumber = getISO8601WeekNumber( scope.rows[0][0].date ),
+          var thursdayIndex = (4 + 7 - ctrl.startingDay) % 7,
+              curWeek = 0,
               numWeeks = scope.rows.length;
-          while( scope.weekNumbers.push(weekNumber++) < numWeeks ) {}
+          while( curWeek < numWeeks ) {scope.weekNumbers.push(getISO8601WeekNumber( scope.rows[curWeek++][thursdayIndex].date ));}
```
chrisirhc: This is probably better re-written as a for loop. It was previously simpler as a while loop but now it makes more sense as a for loop as it's more readable.

chrisirhc:
Good work here. Needs rebasing though.

picodealion:
Great stuff, thanks!

wesleycho:
I think we should try to get this in if there are no other issues other than stylistic issues - we can do the refactoring necessary when merging this in.

chrisirhc:
Merged. Thank you!

PR #2315, opened 2014-06-09
fix(timepicker): Always pad the hours and minutes

### LINKED ISSUE #2314: Always pad hours and minutes in the timepicker
Right now the timepicker only pads the minutes and hours when the template is first loaded. The hours and minutes are not padded when the model changes.

This makes it difficult to create a custom template for the timepicker that uses ```<select>``` elements to let the user choose the ```hour``` and ```minutes```. The problem is shown in [this plunkr](http://plnkr.co/edit/phKW6WEGgIJQolVWV1xd?p=preview). 

The hours and minutes are initially selected correctly. To see the issue, try changing the values to a number less than 10.

mbuttu:
The timepicker only padded the minutes and hours when the template first
loaded. The hours and minutes are now also padded when the model
changes.

Fixes #2314.

### DIFF
--- a/src/timepicker/test/timepicker.spec.js
+++ b/src/timepicker/test/timepicker.spec.js
@@ -714,11 +714,39 @@ describe('timepicker directive', function () {
       expect(getModelState()).toEqual([14, 40]);
     });
 
+    it('always pads hours & minutes', function() {
+      // Make sure hours and minutes are padded in 12 hour format
+      $rootScope.meridian = true;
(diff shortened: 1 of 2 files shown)


wesleycho:
This is a breaking change - this should be an opt-in setting.

wesleycho:
Closing due to issue with implementation not being opt-in.

Feel free to open a new PR based on current `master` with an opt-in flag added.

PR #2333, opened 2014-06-12
feature(typeahead): Added a prefix entry to the typeahead list

tUrG0n:
Assuming the template has a ``prefix`` attribute, searching will always append an entry that says "Search for: ".
This is very handy if you want people to search for things that are not necessarily in your dataset displayed.


rvanbaalen:
Can you fix the PR? Travis is failing.

PR #2361, opened 2014-06-18
Avoid TypeError when accessing matches

oliversalzburg:
Accessing length on matches may cause a TypeError if matches isn't an array.

Maybe I'm using the directive wrong, but occasionally, I'll run into this issue and get `TypeError: Cannot read property 'length' of undefined`.


wesleycho:
It looks like this fix already made it into master in some form - closing.

PR #2377, opened 2014-06-22
feat(modal): added support for autofocus fields on open

nl5887:


wesleycho:
I merged #2945, which should also address the original issue - thanks for the effort though!

PR #2387, opened 2014-06-24
Add popoverHtmlUnsafe directive.

NateRedding:
Popover appears to be missing a directive corresponding to the tooltip-html-unsafe directive. 

I have added tests similar to tooltip-html-unsafe, and updated the documentation. I also  tried to name my commit similarly to how I see other commits named.


wesleycho:
Now that #1848 is merged, this can now be closed as completed :) .

PR #2389, opened 2014-06-25
Date 2014-05-00 should be invalid

gr4b4z:
When I type date in format yyyy-mm-dd, ex. (2014-05-00) , parser give me previous day (2014-04-30). From my perspective it's wrong. This date should be invalid. So I made neccessary modifications.

I'm doing it from a timezone UTC+2.


wesleycho:
Landed as c19b8879e902e8753b4077e7983ec629242436fc, thanks!

PR #2394, opened 2014-06-26
fix missing doc parameter @ Datepicker: toggle-weeks-text

llafuente:
fix missing doc parameter @ Datepicker: toggle-weeks-text


chrisirhc:
Thanks for the PR. However, I believe this option has been removed along with the weeks button.

PR #2395, opened 2014-06-26
fix(datepicker): issue #2369

### LINKED ISSUE #2369: Datepicker manual input forced to US format
Hello,
I'm using datepicker with input (so the user can insert the date manually). If I select in the datepicker the date, it will show up with format dd.MM.yyyy (which is what I want). But if I manually write inside the input, the datepicker popup won't be set to the correct date and will shoow US format date (MM.dd.yyyy).

you can try there:
http://plnkr.co/edit/t9ERna33YcG8ubDqDPgy?p=preview

sphinks13:
modified test date as default dateFormat is 'yyyy-MM-dd'

Fixes #2369 (Edit by @chrisirhc to add ref to issue for link)

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -509,7 +509,7 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
           ngModel.$setValidity('date', true);
           return viewValue;
         } else if (angular.isString(viewValue)) {
-          var date = dateParser.parse(viewValue, dateFormat) || new Date(viewValue);
+          var date = dateParser.parse(viewValue, dateFormat);
           if (isNaN(date)) {
             ngModel.$setValidity('date', false);
             return undefined;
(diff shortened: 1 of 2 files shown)


wesleycho:
This is still a useful idea I think - we're getting some weird cross-browser bug issues due to the fallback since the Date constructor behavior differs between browsers.  I think the true change we need to address the cross-browser issues is to return `null` if it fails the date parsing.  Thoughts?

chrisirhc:
@wesleycho , this should actually accomplish that because it'll always use the return value of the dateParser when a format is specified.

I think this is missing tests to test the changes from this behavior though. We should add tests to check that it now returns undefined whenever the date doesn't parse.

wesleycho:
So looking at this carefully, `dateFormat` must exist by this point currently in the datepicker - this means that this is a duplicate of some other PRs out there.

Closing as a duplicate.

PR #2403, opened 2014-06-28
Tabs: Use fade animation on tab contents

Merec:
Adds fade="true" to tabset.

```html
<tabset fade="true">
	<tab heading="Fading Tab 1">
		<b>First</b> Faded Content!
	</tab>
	<tab heading="Fading Tab 2">
		<i>Second</i> Faded Content!
	</tab>
</tabset>
```

Requires ngAnimate and additional CSS:
```css
.tab-content > .tab-pane.fade.in-add {
  display: block;
  opacity: 0;
}
.tab-content > .tab-pane.fade.in-add.in-add-active {
  opacity: 1;
}
```


khashayar:
Why to have the css animation as an attribute? What if someone else wants, let's say, slide-in animation? As you demonstrated, It could be achieved only with css without any markup / config in directive.

karianna:
@Merec I'm afraid that @khashayar is correct, we unfortunately can't accept this change.

Merec:
I just added this, because the fade effect is offical supported and mentioned by bootstrap
http://getbootstrap.com/javascript/#tabs-examples

PR #2425, opened 2014-07-06
fix(modal): port @thcipriani's solution to modal body shift

MrOrz:
This pull request addresses the scrolling problem that dates back to an old issue https://github.com/twbs/bootstrap/issues/9855, which is also related to #879.

When the modal dialog opens, the content of `body` is shifted due to the absence of the vertical scrollbar.

out

Demonstration of the bug: http://plnkr.co/edit/aEaIphp0kGBC3G9dNw6j

[The solution](https://github.com/twbs/bootstrap/pull/13103) was proposed by @thcipriani in March this year, and is shipped with Bootstrap v3.2.0. This pull request contains a port of @thcipriani 's solution.

Demonstration of the solution: http://plnkr.co/edit/n1OQrpXPx7MTcozZrJA7

For Bootstrap v3.1.1, we need to add the following CSS for the solution to work.

```css
.modal-scrollbar-measure {
  position: absolute;
  top: -9999px;
  width: 50px;
  height: 50px;
  overflow: scroll;
}

.modal-open .modal {
  overflow-x: hidden;
  overflow-y: auto;
}
```

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -121,14 +121,47 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
     };
   })
 
-  .factory('$modalStack', ['$transition', '$timeout', '$document', '$compile', '$rootScope', '$$stackedMap',
-    function ($transition, $timeout, $document, $compile, $rootScope, $$stackedMap) {
+  .factory('$modalStack', ['$transition', '$timeout', '$document', '$window', '$compile', '$rootScope', '$$stackedMap',
+    function ($transition, $timeout, $document, $window, $compile, $rootScope, $$stackedMap) {
 
       var OPENED_MODAL_CLASS = 'modal-open';
+      var MEASURE_SCROLLBAR_CLASS = 'modal-scrollbar-measure';
 
       var backdropDomEl, backdropScope;
       var openedWindows = $$stackedMap.createNew();
       var $modalStack = {};
+      var scrollbarWidth = 0;
+      var bodyEl = angular.element($document[0].body);
+
+      // From twbs/bootstrap modal
+      function checkScrollbar() {
+        if (bodyEl[0].clientWidth >= $window.innerWidth) { return; }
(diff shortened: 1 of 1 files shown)


wesleycho:
Verified this is still an issue in [this Plunker](http://embed.plnkr.co/Is7DoRcA8up9ZICsTnyV/), even with current Bootstrap.

wesleycho:
I created a build based off of current master with this patch, and this seems to have not fixed the problem, as can be seen by [this Plunker](http://embed.plnkr.co/SjNisSMRHvqDzWTHmkZv/preview). One big change that landed that may affect this is that the modal now has ngAnimate support, but this can be reproduced even without ngAnimate loaded.

Can you investigate what has changed that affects this implementation and reimplement this in a way that fixes the issue on current master?

thcipriani:
@wesleycho fwiw, bootstrap recently had a regression that was fixed fairly nicely by @hnrch02 in https://github.com/twbs/bootstrap/pull/15378—didn't look to see what build you're using of bootstrap.

wesleycho:
The build in the Plunker is just 3.3.4, although I tried with 3.1.1 I believe.

chrisirhc:
Unfortunately, as mentioned in the commit that fixed this in https://github.com/twbs/bootstrap/commit/0907244256d923807c3a4e55f4ea606b9558d0ca, it introduces a new bug: https://github.com/twbs/bootstrap/issues/14040 . In light of that, I'm moving this out of 0.13.0 release for later consideration perhaps when a better solution is found.

Klaster1:
In case someone else wants a fix, but not to use other repo/fork, I've made a dirty hack based on @MrOrz's pull request and bootstrap code: https://gist.github.com/Klaster1/b52311698c73e1d83b95. It uses MutationObserver on body, so you'll need polyfill for older browsers. Keep in mind that I didn't test it for all possible scenarios, so issues might happen.

Foxandxss:
This is really old. If this is still an issue, I would love a new issue with an up to date plunker so we can fix it.

Thank you.

PR #2433, opened 2014-07-09
angular.ui.progressbar component - $watch max attribute

shaggyone:
It's common situation, that max value changes in runtime. So I've added watch to it.


wesleycho:
We currently support optional watching of `max`, so this PR is now unnecessary.

Issue #2331, opened 2014-06-12
Unable to use init-date attribute


Mitsukurinadae:
Hey all,

I'm unable to get the init-date working as an attribute set for the datepicker, does anyone have any examples I can pull to see where it's working? I was looking at the demo but for whatever reason this commit removed the only reference to the initDate in the Javascript:
https://github.com/angular-ui/bootstrap/pull/2184

I copied the plunker on the demo page and tried to edit it:
http://plnkr.co/edit/ZGz7rmv09p6SFVVGB9VX?p=preview

I tried sticking it in the datepicker options as initDate: $scope.initDate
and I also tried to just define init-date: initDate on the attributes of the datepicker.

I've browsed through as many issues as I could and Google didn't have an example as the init-date is pretty new.

Thanks for your help!

felixzapata:
hi @kbaltrinic, for me your solution works in my project, so I'm afraid that until your pull request will be accepted, I'm gonna to put direct your changes into my file and load the datepicker from local instead of bower.

Sorry :(

Thanks.

trickpattyFH20:
@felixzapata after you compiled the changes from pull request #2972, how did you implement "init-date"? is it in the dateOptions object like: $scope.dateOptions = { initDate: ''01/14/1997'' }; or was it on the input element with init-date=" '01/14/1997' " or was there some other method/date format that you used?

felixzapata:
for my case, this way:

``` 
$scope.dateOptions = {
            minDate: new Date(2014, 11, 1),
            maxDate: new Date(2015, 1, 28),
            initDate: new Date(2015, 0, 9)
        };

<input name="date" datepicker-options="options" date-disabled="disabled(date, mode)" min-date="options.minDate" max-date="options.maxDate" datepicker-popup="dd/MM/yyyy" placeholder="dd/mm/aaaa" is-open="opened" ng-model="date" readonly="readonly" type="text">

```

felixzapata:
not only this pull request @trickpattyFH20 would be great ;) , I would need another pull request that it was sent a few months ago [about the navigation](https://github.com/angular-ui/bootstrap/pull/2550).

michaellopez:
+1

dciccale:
Please accept #2438 made like a hundred years ago!

rvanbaalen:
Please be just a little more patient. We're actively going through all PRs at the moment.

wesleycho:
[...]

PR #2438, opened 2014-07-10
fix(datepicker-popup) init-date attribute

thomaswinckell:
Fix of the init-date attribute for the datepicker-popup. The init-date attribute, when set in a datepicker-popup element, is now added to the datepicker element and in the datepicker-popup scope, so the datepicker element can retrieve the attribute value. See issue #2331.


dciccale:
LGTM

rvanbaalen:
Could you add a test for this case as well? Ill be pulling this in shortly then.

karianna:
@thomaswinckell as @rvanbaalen mentions, can you add a test for this?

wesleycho:
Landed as c5b63ded0aebd7ae80f238bd0a33f1d9c0746dba with tests.  Thanks!

Issue #2404, opened 2014-06-28
Modal open calls should resolve in order


nlwillia:
Modal instances should be added to the stack in the same order as their $modal.open calls.  Currently, due to template and resolve resolution, modal instances are added in the order that the promises they depend on are fulfilled.  This can be a problem if modals are linked to router states and a series of nested modals each with its own deferred dependencies is launched by a deep state link.  The resulting display layer order should be deterministic, but instead it is dependent on which promises resolve first.

Here is a [demo of the problem in action](http://plnkr.co/edit/zYk9RW2rWi1NgUOPDDLx?p=preview).  The modals are opened in order from 1-3, but the resolve order is 3-1 and modal 1 ends up on top  of the stack instead of the bottom.

When opening a new modal, its promises should be chained off the promise of the previous modal opened so that no modal is added to the stack until all of its predecessors have successfully resolved.

PR #2443, opened 2014-07-11
#2404 Modal open calls should complete in order

nlwillia:
Modal open calls wait on template and resolve promises.  Even though the
resolution order of the promises among different modals is
non-deterministic, the open calls should complete in the same order they
were invoked.


wesleycho:
Is this for having multiple modals open? It should be noted that UI Bootstrap does not support multiple modals being opened at the same time.

nlwillia:
Use case: modals bound to router state.  "main.detail" is a modal detail popup.  "main.detail.edit" is a nested modal on top of that.  If someone follows a deep link to the edit state, then the router will issue modal.open calls in a deterministic order, but without this fix the modal resolves could complete out of order resulting in the wrong modal being on top.

There's no reason why modals shouldn't be allowed to open on top of each other.  Nesting too deeply may not be great UI design, but consider a modal edit dialog with a fancy "are you sure?" exit prompt that also uses a modal.  That would be two modals open at the same time.  In practice, I haven't found a problem with an [arbitrary level of nesting](http://plnkr.co/edit/o4mAa8?p=preview).

src/modal/modal.js
```diff
@@ -358,9 +360,26 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
             var templateAndResolvePromise =
               $q.all([getTemplatePromise(modalOptions)].concat(getResolvePromises(modalOptions.resolve)));
```
chrisirhc: How about getting the top opened window's `opened` promise and appending it here?

chrisirhc:
I think this is a good idea. However, I think your PR doesn't work for more modals. I've made a suggestion above that should also minimize the changes you need to make.

src/modal/modal.js
nlwillia: That's a good suggestion, but I don't think it can work because `$modalStack` isn't updated until `templateAndResolvePromise` is resolved.

nlwillia:
I've committed a change that should fix the problem with more modals and updated the test to cover more scenarios.

src/modal/modal.js
nlwillia: I've reorganized things so that the opened promise is the only one burdened with chaining concerns.  It shuffles things around a little, but it leaves the `templateAndResolvePromise` alone.

src/modal/modal.js
```diff
@@ -390,14 +392,31 @@ angular.module('ui.bootstrap.modal', [])
...
+
             var modalResultDeferred = $q.defer();
-            var modalOpenedDeferred = $q.defer();
             var modalRenderDeferred = $q.defer();
 
+            //add opened, but ensure it resolves in order
+            var previous = modalOpenedPromiseChain;
+            modalOpenedPromiseChain = $q.all([previous || $q.when(undefined), templateAndResolvePromise])
+              .then(function() { return true; }, function(reason) { return reason[reason.length-1]; })
+              ['finally'](function() { modalOpenedPromiseChain = previous; });
```
chrisirhc: This `finally` may call prematurely if 3 modals are opened and the first modal's templateAndResolvePromise fails. In such a scenario, if a 4th modal is opened, the `modalOpenedPromiseChain` would be undefined, but that's not correct.

src/modal/modal.js
```diff
@@ -390,14 +392,31 @@ angular.module('ui.bootstrap.modal', [])
...
+            var templateAndResolvePromise =
+              $q.all([getTemplatePromise(modalOptions)].concat(getResolvePromises(modalOptions.resolve)));
+
             var modalResultDeferred = $q.defer();
-            var modalOpenedDeferred = $q.defer();
             var modalRenderDeferred = $q.defer();
 
+            //add opened, but ensure it resolves in order
+            var previous = modalOpenedPromiseChain;
+            modalOpenedPromiseChain = $q.all([previous || $q.when(undefined), templateAndResolvePromise])
```
chrisirhc: There's no need to do `$q.when(undefined)`, even if `previous` is `undefined`, it'll just be marked as resolved.

src/modal/modal.js
```diff
@@ -390,14 +392,31 @@ angular.module('ui.bootstrap.modal', [])
...
+              .then(function() { return true; }, function(reason) { return reason[reason.length-1]; })
```
chrisirhc: The fail handler here is incorrect. It should return the `templateAndResolvePromise` here on failure because it should tolerate failure on the `previous` promise but still fail if `templateAndResolvePromise` has failed.

chrisirhc:
This is rather complex if `$modalStack` isn't updated until the promise is resolved. Perhaps we need to add it to the $modalStack, or track all the waiting promises until they're resolved. The `finally` call can instead remove it from the list of waiting promises.

nlwillia:
@chrisirhc thank you for your comments (and your patience!).  I should have realized that templateAndResolvePromise needs to remain chained because it controls insertion order onto the stack.  Here's another revision that I think addresses the problems you've raised.  This could still be reworked to coordinate through $modalStack if you prefer, but as is, it seems to resolve and cleanup correctly.

wesleycho:
Can you squash and rebase?

src/modal/test/modal.spec.js
```diff
@@ -24,6 +24,7 @@ describe('$modal', function () {
     $timeout = _$timeout_;
     $q = _$q_;
     $modal = _$modal_;
+	$modalStack = _$modalStack_;
```
wesleycho: Indentation is off.

src/modal/modal.js
```diff
@@ -419,54 +424,61 @@ angular.module('ui.bootstrap.modal', [])
...
-                ctrlInstance = $controller(modalOptions.controller, ctrlLocals);
-                if (modalOptions.controllerAs) {
-                  modalScope[modalOptions.controllerAs] = ctrlInstance;
+                modalOpenedDeferred.resolve(true);
+
+              }, function resolveError(reason) {
+                modalOpenedDeferred.reject(reason);
+                modalResultDeferred.reject(reason);
+              })
+              ['finally'](function() {
```
wesleycho: Couldn't this be just `finally`? It should be noted that we only support 1.3+ currently, so we do not need to worry about IE8 support.

src/modal/modal.js
```diff
@@ -419,54 +424,61 @@ angular.module('ui.bootstrap.modal', [])
             var templateAndResolvePromise =
               $q.all([getTemplatePromise(modalOptions)].concat(getResolvePromises(modalOptions.resolve)));
 
+            // Wait for the resolution of the existing promise chain.
+            // Then switch to our own combined promise dependency (regardless of how the previous modal fared).
+            // Then add to $modalStack and resolve opened.
+            // Finally clean up the chain variable if no subsequent modal has overwritten it.
+            var samePromise;
+            samePromise = promiseChain = $q.all([promiseChain])
+              .then(function() { return templateAndResolvePromise; }, function() { return templateAndResolvePromise; })
```
wesleycho: The second callback is forcing a resolve if there is an error?

nlwillia:
Not sure that rebase commit did what I wanted it to.  I'll have to look at this further in another week when I'm back from vacation and have access to my dev machine.

PR #2461, opened 2014-07-14
fix(modal) Implement W3-compliant tabbing on the modal - no tabbing into the background document

jmwolfe:
This PR enables the bootstrap modal to use tabbing in forms without the tab cycle taking you to the background document. It implements the full standard W3 specification for browser tabbing. Demo is updated to demonstrate tabbing. May want to scale that back if it's "too much about tabbing." But I think it makes for a strong demo and test of the feature. Added a red outline


smsenesac:
Just in case someone was thinking of using this PR, it actually does not work if you have input elements that are hidden or excluded using the ng-hide/show/if directives.

PR #2463, opened 2014-07-14
feat(dropdown): Added dropdownMenu directive to use for context menus

rasmusvhansen:
I have added a feature to dropdown, so it can be used as a context menu. Meaning it will show up at the specified coordinates. See demo.js and demo.html for details.

It is implemented as a dropdownMenu directive. So elements marked with dropdown-menu will register with the dropdown controller, and when dropdown is toggled, be called with the mouse event that triggered the dropdown.

I have also made a change to dropdown-toggle so it stops propagation, since this could e.g. trigger an accordion to toggle if the dropdown-toggle was placed inside an accordion header.

### DIFF
--- a/src/dropdown/docs/demo.js
+++ b/src/dropdown/docs/demo.js
@@ -18,4 +18,20 @@ angular.module('ui.bootstrap.demo').controller('DropdownCtrl', function ($scope)
     $event.stopPropagation();
     $scope.status.isopen = !$scope.status.isopen;
   };
+})
+
+.directive('contextMenuToggle', function () {
+  return {
+    require: '^dropdown',
+    link: function (scope, element, attrs, dropdownCtrl) {
+      element.bind('contextmenu', function (evt) {
+        scope.$apply(function () {
+          console.log('context menu right clicked');
+          evt.preventDefault();
+          evt.stopPropagation();
+          dropdownCtrl.toggle(true, evt);
+        });
+      });
+    }
+  };
 });
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -53,7 +53,8 @@ angular.module('ui.bootstrap.dropdown', [])
       openClass = dropdownConfig.openClass,
       getIsOpen,
       setIsOpen = angular.noop,
-      toggleInvoker = $attrs.onToggle ? $parse($attrs.onToggle) : angular.noop;
+      toggleInvoker = $attrs.onToggle ? $parse($attrs.onToggle) : angular.noop,
+      callbacks = [];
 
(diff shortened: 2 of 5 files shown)


src/dropdown/dropdown.js
```diff
@@ -156,4 +177,54 @@ angular.module('ui.bootstrap.dropdown', [])
...
+      }
+      dropdownCtrl.register(onOpen);
+      function onOpen(coords) {
+        if (!coords) {
+          return;
+        }
+        // Hide it before moving it in to the right place
+        element.css('display', 'none');
+        // Needs to be done in a timeout, otherwise it is not visible and thus has no width/height
+        $timeout(function () {
```
wesleycho: This should support usage of $animate I think, in case the user wants to animate in the menu

wesleycho:
Tests are failing.

This sounds like an interesting feature, although I'm wondering whether it is one we want to support as a part of the core.

Marking needs review, hopefully we have another comment here.

src/dropdown/dropdown.js
```diff
@@ -156,4 +177,54 @@ angular.module('ui.bootstrap.dropdown', [])
...
+            top = top - (totalHeight - docHeight);
+          }
+          if (totalWidth > docWidth) {
+            left = left - (totalWidth - docWidth);
+          }
+          element.css('position', 'fixed');
+          element.css('top', top + 'px');
+          element.css('left', left + 'px');
+        });
+      }
```
RobJacobs: Could the position service be used here instead?

src/dropdown/docs/demo.js
```diff
@@ -18,4 +18,20 @@ angular.module('ui.bootstrap.demo').controller('DropdownCtrl', function ($scope)
...
   };
+})
+
+.directive('contextMenuToggle', function () {
+  return {
+    require: '^dropdown',
+    link: function (scope, element, attrs, dropdownCtrl) {
+      element.bind('contextmenu', function (evt) {
+        scope.$apply(function () {
+          console.log('context menu right clicked');
```
wesleycho: Remove console.log

src/dropdown/dropdown.js
```diff
@@ -135,6 +155,7 @@ angular.module('ui.bootstrap.dropdown', [])
 
       var toggleDropdown = function(event) {
         event.preventDefault();
+        event.stopPropagation();
```
wesleycho: This is a breaking change I believe, since there is a listener on `document` for the click event.

wesleycho:
Closing due to age & lack of response.

It should be noted that we intend on fixing position issues in 1.2

Issue #2326, opened 2014-06-11
Timepicker & seconds


SkyQuant:
How can I have a third input box for seconds editing additionaly to hours and minutes?

barronj:
I also vote for seconds. I just had an issue today where it was complaining that the time I entered was not greater than or equal to the minimum time for my field even though it looked identical. Turns out, the seconds were off. And this is for an application where seconds matter.

Thank you.

jupake:
+1. We also now need seconds to be changeable.

chrisirhc:
Maybe fulfilled via resolving #3312 . Will look into time input types as well.

Foxandxss:
Let's continue in #4272

Issue #1763, opened 2014-02-07
having seconds in the timepicker


bhamde:
Is it possible to add a seconds field in the timepicker? It would be very useful.
Thanks

antoinepairet:
keeping #2326 and closing this one as duplicate

Issue #1231, opened 2013-11-05
[New Feature] Timepicker should support setting the seconds


zsong:
Timepicker should support setting the seconds.

antoinepairet:
@richidd Please see #2326 for tracking this

PR #2467, opened 2014-07-15
feat(timepicker): added support for seconds

papoola:
timpicker can now show seconds as well which is by default hidden. passing `timepickerConfig.showSeconds=true` or `<timepicker show-seconds="true" ...>` will enable this feature. Additional tests and documentation is still missing, and thats why I call to volunteers to help with adding tests and documentation particularly in https://github.com/angular-ui/bootstrap/issues/2326 https://github.com/angular-ui/bootstrap/issues/1763 https://github.com/angular-ui/bootstrap/issues/1231

### DIFF
--- a/src/timepicker/test/timepicker.spec.js
+++ b/src/timepicker/test/timepicker.spec.js
@@ -71,9 +71,9 @@ describe('timepicker directive', function () {
     return e;
   }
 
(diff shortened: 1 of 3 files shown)


Foxandxss:
Seems like we have modern implementations to this, sorry for the delay.

Closing this.

Issue #220, opened 2013-03-13
feat(popover): support template url for partial


zero-:
As requested in #202 by @joshdmiller here's my request for external partial support for popover which would make popover much more versatile and lower the need of duplicate code in the apps. Already working implementation can be found in one of the "competing" Bootstrap implementations at http://mgcrea.github.com/angular-strap/#/popover.

DaveWM:
I got fed up waiting for this feature, so I made the [ngPopoverButton directive](https://github.com/DaveWM/ngPopoverButton). Posting it here in case some of you guys would find it useful.

madhums:
can't believe this thread is still open!

QuentinFchx:
What is preventing the PR to be merged (apart from conflicts) ?
I'd be glad to help fixing the last issues!
It needs to be merged asap because this is becoming hilarious (1 year issue, with 200+ comments, on a 7k stars project, just to allow html in a popover...)

karianna:
@QuentinFchx https://github.com/angular-ui/bootstrap#impt---2015-plans-and-angular-13-support

QuentinFchx:
The [0.12.1 milestone](https://github.com/angular-ui/bootstrap/milestones/0.12.1) only contains the https://github.com/angular-ui/bootstrap/pull/2996 PR tagged as `needs:merging`. How can I help to make this happen ?
This will allow us to move forward toward 0.13.

karianna:
We'll be merging that issue at the next maintainers meeting.  @chrisirhc will be teaching the rest of us new maintainers to cherry pick the fix and perform a full release.  Then we get onto the hard work of triaging the ~120 or so PR's out there :-).

stramel:
@karianna :+1: This is great news for this library.

nnjpp:
That is such an enormous relief. I was under the impression that the project had seriously lost momentum but it looks like I'll soon be able to start recommending it again.

karianna:
Well I'm not going to promise miracles as there are 187 outstanding Pull Requests to get through for starters, but yes you should start seeing some movement over the coming weeks.

Yappli:
Marvelous !

metanav:
Any progress? I understand guys are working purely part-time and unpaid and I appreciate that.
But I think this issue is blocking many of us.

karianna:
Meeting tonight to begin the first steps.

Alos:
Hey guys? Did you got a chance to talk about this?

[...]

PR #2479, opened 2014-07-19
feat(popover): added template as per requests in issue #220, took code from jbruni

calekennedy:
There seemed to be a huge amount of interest in merging in the type of templating @jbruni was using in his fork. However his fork is outdated. This leads to a feature disparity between his fork and the core library. One of my projects requires features from both version, so I took it upon myself to merge them. Hopefully this merge will find its way into the core library where it will help other users who require both sets of functionality. :smile:


chrisirhc:
#1848 is reviewed and ready to merge pending a name poll. Look there if you're interested in getting this feature.

PR #2513, opened 2014-07-25
fix(datepicker): don't fall back to parse using new Date(...)

### LINKED ISSUE #2508: Datepicker in Chrome parses in incorrect format
I configured the dateFormat for the German style: dd.MM.yyyy
Now when I enter "1.2." it gets parsed as 2nd of January. Expected is 1st of Febuary.

Another example that is pretty confusing for a user:
"1.02.2015" -> February 1st, 2015
"01.02.2015" -> January 2nd, 2015

I already browsed the code a bit and the cause is falling back to [new Date(viewValue)](https://github.com/angular-ui/bootstrap/blob/master/src/datepicker/datepicker.js#L512) which behaves very different across browsers.
Firefox: new Date("1.2.") -> undefined
Chrome: new Date("1.2.") -> Tue Jan 02 2001 00:00:00 GMT+0100 (CET)
Safari: new Date("1.2.") -> undefined

This can be reproduced using the [demo datepicker](http://angular-ui.github.io/bootstrap/).

I suggest to remove the fallback and to only rely on the DateParser that's part of bootstrap. Or at least to make the fallback optional.

cdietze:
Fixes #2508

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -509,7 +509,7 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
           ngModel.$setValidity('date', true);
           return viewValue;
         } else if (angular.isString(viewValue)) {
-          var date = dateParser.parse(viewValue, dateFormat) || new Date(viewValue);
+          var date = dateParser.parse(viewValue, dateFormat);
           if (isNaN(date)) {
             ngModel.$setValidity('date', false);
             return undefined;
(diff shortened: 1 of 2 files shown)


jmayday:
+1 (haven't seen this PR and prepared mine https://github.com/angular-ui/bootstrap/pull/3288). After some searching I found a lot of issues/PRs with some solutions. Would be nice to finally have one of them merged into master. 

Cheers.

wesleycho:
This fails tests when rebased off of master.

karianna:
@wesleycho Travis passes?

wesleycho:
That is probably because it was valid when the PR was open.  It does not pass currently though.

chrisirhc:
As I commented on #3294, this needs tests for validity of the date if it doesn't follow a specified format.

PR #2515, opened 2014-07-25
feat(datepicker): add open-on-focus option

greenboxal:
Adds the so requested option to open the datepicker popup when the input gets focus.
I understand that it was removed to improve accessibility but in some cases opening on focus is needed.

Also, I'm not sure on how to test this, so I didn't write any unit tests :/


wesleycho:
Can you give an example where opening on focus would be desired?

Also, the datepicker seems to have changed quite a bit since this PR was filed - can you rebase your branch off of current master?

rvanbaalen:
Closing due to inactivity will be re-opened once @greenboxal chooses to respond.

PR #2524, opened 2014-07-29
feat(modal): Add CSS class(es) to the top modal window

R3VoLuT1OneR:
Adds ability to add CSS class(es) to the top modal window.

Can be used for hiding other modal windows with CSS if there are multiple modal windows opened.


wesleycho:
Can you explain the difference between this and the existing `windowClass` feature?

R3VoLuT1OneR:
@wesleycho the difference is that `windowTopClass` is applied only on the top window ( the first one in the stack of windows ) and `windowClass` is applied for all windows. This class is useful only if multiple popup windows are opened.

One more time. The problem I tried to solve, is the hiding of bottom popup windows, if multiple windows are opened.

wesleycho:
I don't quite understand - we do not support multiple modal windows being opened at the same time. Is what you want a feature for supporting multiple modal windows being opened at the same time?

chrisirhc:
We do support multiple modals through a modal stack. We can consider this but I think we should put this off to merge after the 0.13 release.

wesleycho:
Hm, I'm completely mistaken then, whoops.

karianna:
@R3VoLuT1OneR Can you rebase with master please?

R3VoLuT1OneR:
@karianna rebase done, sorry for slow response

wesleycho:
Question - could the `windowTemplateUrl` option be used to replace the template to currently accomplish what is desired?

wesleycho:
I misunderstood this feature before - I see what this is doing now.

If the class is only to be applied to the top modal, shouldn't the class be removed from the other modals?

Thinking about this, I am not sure we would want this - what are your thoughts @R3VoLuT1OneR?

R3VoLuT1OneR:
@wesleycho I am not sure if I understand you correctly.

windowTopClass - Applied only on the top modal window ( first one in the stack of modal windows ) on other modal windows should not be that class.

Usage: Show only top modal window, and hide others with css. ( .windowClass {display: none;} .windowTopClass {display: block;}  for example).

By default when you opening multiple modal windows they all displayed and it is hard to hide unwanted windows with CSS, only with JS.

wesleycho:
The problem is when a new modal is opened, we would have to go through and remove this class for all of the other modals then. I am just concerned with potential complexity here, but I guess I can accept this use case.

This PR needs to be rebased though, and after rebasing, the commits need to be squashed into one.

R3VoLuT1OneR:
Rebased and squashed.

About removing the "windowTopClass" from all other windows, actually we should remove class only from 1 window ( previous top ) on opening new one and on closing window we should add class to new top window, that what my code doing. There are also tests for this use case.

wesleycho:
Hm, makes sense - this LGTM.

PR #2543, opened 2014-08-04
feat(progressbar): respond to changes in max dynamically

coojam:
We have a use case where the max of the progress bar should change dynamically so this pull request enables that behaviour.


tdhsmith:
Looks good! :+1:

Only thing I'd say is that the docs could use an eye symbol next to `max` to indicate it is watched.

wesleycho:
This looks good - just make sure it is up to date with master.

karianna:
@coojam can you please rebase with master?

coojam:
OK - I believe I have rebased with master.

coojam:
Whoops - wrong button.

src/progressbar/progressbar.js
```diff
@@ -41,7 +48,9 @@ angular.module('ui.bootstrap.progressbar', [])
         transclude: true,
         controller: 'ProgressController',
         require: 'progress',
-        scope: {},
+        scope: {
+            max: '=?'
```
chrisirhc: Since it's a one way binding, use `&` instead

src/progressbar/progressbar.js
```diff
@@ -10,7 +10,10 @@ angular.module('ui.bootstrap.progressbar', [])
         animate = angular.isDefined($attrs.animate) ? $scope.$parent.$eval($attrs.animate) : progressConfig.animate;
 
     this.bars = [];
-    $scope.max = angular.isDefined($scope.max) ? $scope.max : progressConfig.max;
+
+    if (!angular.isDefined($attrs.max)) {
```
chrisirhc: Should be able to just do `if ($attrs.max) {` since the only way it can be falsey (but defined) is if it's set via `max=""` and that's invalid.

src/progressbar/progressbar.js
```diff
@@ -23,6 +26,10 @@ angular.module('ui.bootstrap.progressbar', [])
             bar.percent = +(100 * value / $scope.max).toFixed(2);
         });
 
+        $scope.$watch('max', function( max ) {
```
chrisirhc: Should only add watcher in an if block when `if ($attrs.max) {` since otherwise the max will never change.

wesleycho:
We currently support optional watching of `max`, so this PR is now unnecessary.

PR #2561, opened 2014-08-08
feat(datepicker): ng-disabled attribute added to datepicker

gediminasgu:


src/datepicker/datepicker.js
```diff
@@ -576,16 +577,29 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
...
+          }
+          else {
+            scope.isOpen = false;
+          }
         } else {
           $document.unbind('click', documentClickBind);
         }
       });
 
+      scope.$watch('ngDisabled', function(value) {
```
wesleycho: Use `scope.$watch($parse(attrs.ngDisabled), ...)`

chrisirhc: Should be able to do `scope.$parent.$watch(attrs.ngDisabled`. Need to use parent because this is an isolate scope.

wesleycho:
Closing due to this being superceded by 434c6026a540387fd2504e5b1f53f05a39e21bca

PR #2565, opened 2014-08-08
Allow Accordion Groups to be open by default

malls:
Added a cute & necessary lil feature that allows <accordion-group> tags to have an open-by-default option, which lets them be open by default.

`<accordion-group open-by-default="true">
You'll see this!
</accordion-group>`


wesleycho:
Documentation needs to be added for this feature.

Foxandxss:
with `is-open="true"` it closes perfectly. Perhaps the behavior change in the past last year.

Closing this.

Issue #2616, opened 2014-08-21
Add classes to specific dates by any desired logic


Urigo:
Add the ability to add classes to specific date by an outside logic (a user function).
That way we can change a looks of specific dates with association to a list of events for example.

wesleycho:
As #2619 has been merged, this should be already addressed.

PR #2619, opened 2014-08-21
feat(datepicker): Add custom class to specific days via outside logic

Urigo:
Adding classes to specific dates with outside logic similar to date-disable.
Help with bringing more functionality to the picker that might prevent people from using fullcalendar.

https://github.com/angular-ui/bootstrap/issues/2616


nemanjavuk:
+1

This would be great for next release!

sandervanloock:
+1
Nice!

arjunasuresh3:
+1

wesleycho:
Landed as 0bcd30c4a6d65763e1c4d1b6820a4537a11e95ee, nice work!

JGSimko:
Could someone please provide an example of this getting set up as a "datepicker-options" property? I've tried stepping through the code & my custom-class function gets called but the "ng-class" property is never added to the DOM resulting in no class being added.

PR #2647, opened 2014-08-29
Add start and end events for collapse and expand transitions.

NateRedding:
There are plenty of use cases where one would need to know when a collapse or accordion section is finished expanding/collapsing. This PR adds some simple start and end events that are emitted around the expand/collapse transistions. This will allow users of the collapse or accordion widgets to know when the animation is done so that they can take further action.

### DIFF
--- a/src/collapse/collapse.js
+++ b/src/collapse/collapse.js
@@ -26,6 +26,7 @@ angular.module('ui.bootstrap.collapse', ['ui.bootstrap.transition'])
         }
 
         function expand() {
+          scope.$emit('expand.start');
           if (initialAnimSkip) {
             initialAnimSkip = false;
             expandDone();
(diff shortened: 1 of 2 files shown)


NateRedding:
@chrisirhc I see you added this to the 'Purgatory' milestone. Does that mean it is in limbo and is pretty much being ignored? Any idea when it might get looked at?

karianna:
Hopefully @chrisirhc will chime in shortly - my additional thoughts on this - Is there some documentation that would also need to be updated?  Can that be added to this PR?

chrisirhc:
I put this as Purgatory as I don't see it resolving any existing issues (hence, it's not a priority) and because this introduces a possible breaking API change in the future.

The current plan is to convert the collapse module to use the `$animate` service which would have a `$animate:close` event. This is to make use of the services provided by Angular to keep it consistent with its animation APIs.

I was working on #1444 before, but now it seems that the $animate service has changed since and there might be a more concise way of using the $animate service.

wesleycho:
Since we migrated to ngAnimate for this, I am closing this as superceded - as @chrisirhc mentioned, $animate has events you can hook into if necessary.

Issue #2597, opened 2014-08-18
Typeahead: move focus to text input with arrow up


danemacaulay:
Both Chrome and Twitter's typeahead allow users to arrow up out of the typeahead list in order to refocus on the search box.

In the current implementation, an up keypress cycles through the list, skipping the text input.

Would be happy to do a pull request if there's any interest in this as a feature.

PR #2650, opened 2014-09-02
Typeahead: move focus to text input with arrow up

danemacaulay:
Per issue #2597

Input element included in list of results with an activeIdx of -1 and is the default selection on typing. Arrow keys allow full range of navigation from list to input.

I think its good to no longer default to the list and have user explicitly arrow down to list, thoughts?


wesleycho:
@danemacaulay any update on this?

wesleycho:
Closing due to age - feel free to open a new pull request with comments addressed & rebased off of current `master`.

PR #2672, opened 2014-09-07
fix(progressbar): remove ngClass to avoid conflict with ngAnimate

### LINKED ISSUE #2670: Progressbar type property updates conflicts with ngAnimate causing no animation
When ngAnimate is imported updating a progress bar value and type at the same time results in the progress bar not animating.  Here is an example

http://plnkr.co/edit/uZr9XqsUgyT4sjDTKhLS?p=preview

It seems the best way to fix this is by this by removing ngCass from the bar templates and just update the bar element class directly in the directive.  I will create the pull request if this seems like a reasonable approach.

rgaskill:
fix #2670

### DIFF
--- a/src/progressbar/progressbar.js
+++ b/src/progressbar/progressbar.js
@@ -5,7 +5,9 @@ angular.module('ui.bootstrap.progressbar', [])
   max: 100
 })
 
-.controller('ProgressController', ['$scope', '$attrs', 'progressConfig', function($scope, $attrs, progressConfig) {
+.controller('ProgressController', ['$scope', '$element', '$attrs', '$animate', 'progressConfig', function($scope, $element, $attrs, $animate, progressConfig) {
+    $animate.enabled(false, $element);
+
     var self = this,
         animate = angular.isDefined($attrs.animate) ? $scope.$parent.$eval($attrs.animate) : progressConfig.animate;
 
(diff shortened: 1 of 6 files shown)


karianna:
Hi @rgaskill - Could you please add an accompanying test?

rgaskill:
@karianna will do!

src/progressbar/progressbar.js
```diff
@@ -5,7 +5,9 @@ angular.module('ui.bootstrap.progressbar', [])
   max: 100
 })
 
-.controller('ProgressController', ['$scope', '$attrs', 'progressConfig', function($scope, $attrs, progressConfig) {
+.controller('ProgressController', ['$scope', '$element', '$attrs', '$animate', 'progressConfig', function($scope, $element, $attrs, $animate, progressConfig) {
+    $animate.enabled(false, $element);
```
rgaskill: This is the fix.

src/progressbar/test/progressbar.spec.js
```diff
@@ -140,6 +167,15 @@ describe('progressbar directive', function () {
       expect(barEl).not.toHaveClass(BAR_CLASS + '-success');
       expect(barEl).toHaveClass(BAR_CLASS + '-warning');
     });
+
+    it('should not animate using ngAnimate', function() {
```
rgaskill: new test for the progress bar

rgaskill:
@karianna take a look.  The test turned out to be more interesting than I expected.  Let me know if there is a better way to test that the animation is turned off.

karianna:
@chrisirhc - Can you comment on the choice of purgatory for this PR?

chrisirhc:
Closing this as I have commented in #2670 that this is no longer valid for our next release.

PR #2677, opened 2014-09-09
fix(tab): IE9 disabled attr renders grey text on enabled tab

leonardseymore:
This issue can be seen on the boostrap UI page using IE9.  The second tab looks disabled when it is really enabled:
http://angular-ui.github.io/bootstrap/#/tabs

- IE9 renders elements with 'disabled' property with grey text
- Tabs are rendered with grey text even if they are not disabled
- By renaming attribute to 'disable' instead of 'disabled' tabs with dynamic 'disable' attribute still render correctly


leonardseymore:
What about supporting both attributes and marking "disabled" in the APIs as deprecated?

wesleycho:
This would require a documentation update.

I think this would be a good breaking change, as using the disabled attribute by the directive currently is bad.

Can you add the documentation change required?

chrisirhc:
@leonardseymore I like the idea of supporting both attributes (`disable`, `disabled`) and dropping the `disabled` attribute later on. I think this directive is probably used in many places.

wesleycho:
We are deprecating support for `disabled` then in the meantime? That sounds sensible to me in this case.

chrisirhc:
Yep that's right, we should deprecate the use of `disabled`. We can log a warning whenever `disabled` is used unless it's suppressed, similar to #3497 .

leonardseymore:
I have made some more changes based on the comments.

src/tabs/docs/readme.md
```diff
@@ -26,9 +26,10 @@ AngularJS version of the tabs directive.
  	_(Defaults: false)_ :
  	Whether tab is currently selected.
 
- * `disabled` <i class="glyphicon glyphicon-eye-open"></i>
+ * `disable` <i class="glyphicon glyphicon-eye-open"></i>
  	_(Defaults: false)_ :
  	Whether tab is clickable and can be activated.
+ 	Note that this deprecates the "disabled" attribute
```
chrisirhc: How about:
```md
Note that this was previously the `disabled` attribute.
```

chrisirhc:
Looks mostly good to me. Maybe just documentation wording can be improved.

src/tabs/docs/readme.md
wesleycho: Or
<pre>
Note that this was previously the `disabled` attribute, which is now deprecated.
</pre>

wesleycho:
We can probably merge this in and patch the documentation change we want - no need to hold this up longer I think.

wesleycho:
Thanks for this find and fix - it should go out when we release 0.13.0.

PR #2691, opened 2014-09-12
fix(datePicker): Don't parse without date format

oliversalzburg:
Under certain conditions, parseDate() is invoked before dateFormat is set.
This will cause an error to be thrown.


src/datepicker/datepicker.js
```diff
@@ -509,6 +509,7 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
           ngModel.$setValidity('date', true);
           return viewValue;
         } else if (angular.isString(viewValue)) {
+          if (!dateFormat) return null;
```
RobJacobs: Should initialize var dateFormat = datepickerPopupConfig.datepickerPopup instead.

karianna:
@oliversalzburg See @RobJacobs comments + this no longer merges.

oliversalzburg:
Shocking, given that I wrote the patch half a year ago :p

It's a 1-line change. I'm sure @RobJacobs can manage to put his suggestion into the code himself.

Issue #2235, opened 2014-05-21
Carousel not able to set active slide after assining new slides


kimvanhees:
I have a simple Carousel example. In this example on Plnkr I change the slides in my application. When I set the active slide after I change slides it goes to that slide and then slides out into oblivion or it goes to the first slide instead.

http://plnkr.co/edit/PJg9U4HZ1k5aSTSvbl3k?p=preview

    angular.module('plunker', ['ui.bootstrap', 'ngTouch']);
    function CarouselDemoCtrl($scope) {
        $scope.genderPerson = "men";
        $scope.myInterval = -1;
        $scope.slides = [];

        $scope.$watch("genderPerson", function( newValue, oldValue ) {
            $scope.MakeSlides();
        });

        $scope.MakeSlides = function() {
            var newSlides = [];
            for ( var i = 0; i < 10; i++ ) {
                newSlides[i] = { image: 'http://api.randomuser.me/portraits/' + $scope.genderPerson + '/' + i + '.jpg' };
             }
             $scope.slides = newSlides;
             if ( $scope.slides[6] ) {
                 $scope.slides[6].active=true;
             }
        }
    }

stuartm:
I'm seeing the same issue with the latest release 0.13.2, if I update the change list used by the carousel then no slides are 'active' and the carousel disappears.

shabetya:
0.13.4 version also has that bug again! if redefine slides, active class is not set

dotcom9:
Is there a workaround for this bug in 0.13.4? Unfortunately, I can't upgrade due to project constraints.

icfantv:
@dotcom9 you could try back porting the fix depending on how invasive it is.  I understand all too well the pain of being forced to remain on an old version.

Issue #2062, opened 2014-04-15
Transitions never finish for hidden elements


alexreardon:
``file: transition.js``
If an element is not visible then the ``transitionEnd`` events will never file.

**Quick fix** - uses jQuery ``is:('visible')``

```

// Wrap in a timeout to allow the browser time to update the DOM before the transition is to occur
$timeout(function () {
    if (angular.isString(trigger)) {
        element.addClass(trigger);
    } else if (angular.isFunction(trigger)) {
        trigger(element);
    } else if (angular.isObject(trigger)) {
        element.css(trigger);
    }
    //If browser does not support transitions, instantly resolve
    if (!endEventName) {
        deferred.resolve(element);

        // new
        return;
    }

    /***** NEW:  *******/
    // Invisible elements
    // if element is not visible then transitionEnd will never fire as animation will not occur
    // need to manually resolve the transition
    if (!element.is(':visible')) {
        transitionEndHandler();
    }
});

```

This fix requires jQuery ``.is(':visible')`` however I am sure there would be a jqlite work around.

**Example of problem:**
If you start a collapse on an element that is hidden then the ``collapsing`` class will be added, but it will never be removed and replaced with ``collapse in``.

Issue #1513, opened 2014-01-03
Problem with carousel after hiding and showing


uriva:
Hi there,
Whenever I hide the carousel, wait a couple of seconds and show it, the left and right arrows no longer work and the carousel doesn't auto cycle.

For some reason I can't save a plunker but here's the html:

	<!doctype html>
	<html ng-app="plunker">

	<head>
	  <script src="http://ajax.googleapis.com/ajax/libs/angularjs/1.1.4/angular.js"></script>
	  <script src="http://angular-ui.github.io/bootstrap/ui-bootstrap-tpls-0.9.0.js"></script>
	  <script src="example.js"></script>
	  <link href="//netdna.bootstrapcdn.com/bootstrap/3.0.3/css/bootstrap.min.css" rel="stylesheet">
	</head>

	<body>

	  <div ng-controller="CarouselDemoCtrl">
		<input type="checkbox" ng-model="hide">Click to hide
		<div ng-hide="hide">
		  <carousel interval="2000">
			<slide ng-repeat="slide in slides" active="slide.active">
			  <img ng-src="{{slide.url}}" style="margin:auto;">
			  <div class="carousel-caption">
				<p>{{slide.description}}</p>
			  </div>
			</slide>
		  </carousel>
		</div>
	  </div>
	</body>

	</html>

And js:

	angular.module('plunker', ['ui.bootstrap']);
	function CarouselDemoCtrl($scope) {
	  $scope.slides = [{
		  url: 'http://i.imgur.com/xKE07.jpg',
		  description: 'bla'
		}, {
		  url: 'http://wpmedia.blogs.windsorstar.com/2013/01/cat_shot_by_arrow-1.jpg',
		  description: 'bli'
		}];
	}

Issue #1185, opened 2013-10-19
carousel interval work only one time when transition set none


scarletsky:
see http://plnkr.co/edit/cleBHsUy6xMuumulb7Lw?p=preview
I just set interval to 1000 and add

```
    <style type="text/css" media="all">
        .carousel-inner > .item {
          -webkit-transition: none;
                  transition: none;
        }
    </style>
```

After that, interval only work one time.

PR #2699, opened 2014-09-15
Refactor carousel to work with ngAnimate

### LINKED ISSUE #1350: Carousel strange transition [bootstrap 3]
I've made video that show behaviour (`angular 1.2.3`).

https://dl.dropboxusercontent.com/u/21598387/AngularUI_Bootstrap3_carousel.mov

chrisirhc:
This is a much cleaner solution that makes use of ngAnimate as it was intended.

It requires a few lines of CSS that will be added by default (similar to how AngularJS adds CSS for ng-hide).
I'll add a Grunt task that will provide the CSS in a separate file so that users that have CSP can add the CSS files separately.

Supercedes #1878.

Fixes #1350 #1513 #1185 #2062 #2235.

Things left:
- [x] Move CSS into .css file in carousel directory
- [x] Add Grunt task to compile CSS files into JS that will be loaded when the JS is loaded
- [x] Add Grunt task to concatenate CSS into one file for CSP users
- [x] Add prefixes for transition css


JackMorrissey:
This fixed the numerous issues I was having with Carousel on IOS. Thanks!

karianna:
@chrisirhc - This can't be automerged at present...

douglasduteil:
@chrisirhc I think you should merge it to put an end to it. It is still up to date ?

wgorder:
before it is merged @intellix  comment is applicable and should be added as well.

chrisirhc:
I'll rebase it today

chrisirhc:
Rebased.

wgorder:
@chrisirhc   did you happen to see my comment about noting @intellix  post regarding 1.3 incompatibility? Its a few posts up, but I'll quote it here for posterity (not sure if its related to what you are seeing on the demo site):

```
This fails now in 1.3 as $ngAnimate functionality return promises. So just need to change:

$animate.addClass(slide, directionDirClass, function() {
    slide.removeClass(directionDirClass + ' ' + direction);
    animationDone();
});
to

$animate.addClass(slide, directionDirClass).then(function() {
    slide.removeClass(directionDirClass + ' ' + direction);
    animationDone();
});
I did do this locally and would have done PR but seems i'm using a really old version of this locally.

```

src/carousel/carousel.js
```diff
@@ -23,51 +23,30 @@ angular.module('ui.bootstrap.carousel', ['ui.bootstrap.transition'])
...
-        //We have to create new pointers inside a closure since next & current will change
-        (function(next,current) {
-          $scope.$currentTransition.then(
-            function(){ transitionDone(next, current); },
-            function(){ transitionDone(next, current); }
-          );
-        }(nextSlide, self.currentSlide));
-      } else {
-        transitionDone(nextSlide, self.currentSlide);
+      if ($animate.enabled() && !$scope.noTransition && nextSlide.$element) {
```
wesleycho: The only question I have about this check is that there is the potential for `nextSlide.$element.data('$$ngAnimateState').disabled` to be true - in that case, this would be broken.  Do we want to say that we do not support this use case?

chrisirhc: What do you think if I do this on the `$animate:before` event for any slide element (and attach it in the slide directive)?
Since it should only mark `$currentTransition = true` when the animation has started, and it'll only close if it starts. There's a possible race condition where an animation hasn't started but has been queued, but that shouldn't matter so much I think.

chrisirhc:
@wgorder , Thanks for the heads up! I just saw that, I'll fix that for Angular 1.3 since we're going for Angular 1.3 compatibility in the next release.

src/carousel/carousel.js
chrisirhc: Also, I was just looking at the animation code, I think the close event always even if the animation is disabled for the element.

wesleycho: Looks like you're right - ignore my comment.

chrisirhc:
Fixed compatibility with Angular 1.3

ammar91:
What I would need to do in order to make this change on my side?

wesleycho:
This PR LGTM.

chrisirhc:
Merged via 431b9c7 . Thanks!

PR #2703, opened 2014-09-15
Set timpicker widths on input elements to fix display in inline forms

mbfisher:
When the timepicker component is used in an inline form the hour and minute inputs do not display correctly:
screen shot 2014-09-15 at 10 18 14
A reproduce scenario can be found at http://plnkr.co/oOH1kc447zPemyLhj8cr; you'll need to use a desktop device and possibly view in fullscreen mode, as the issue occurs above a min-width of 768px. 

Currently the widths of the hour and minute inputs are set on the `td` parents:
```html
<td style="width:50px;" class="form-group" ...>
  <input type="text" ng-model="hours" ... />
</td>
```
Bootstrap applies `width: auto` to inline form controls on a device-width above 768px:
```css
@media (min-width: 768px) {
    .form-inline .form-control {
        ...
        width: auto;
        ...
    }
}
```

Moving the fixed widths to the `input` elements fixes the issue. 

Is this the best way to fix this?

M


wesleycho:
Confirmed this is a good fix - landed as 8e89440ba86d26d74ed799843cde7f9f2bf26c0b, thanks!

Issue #2532, opened 2014-07-31
Pagination: add custom page indexes


macluck:
Hey!

I had to implement a pagination, where instead of page numbers, specific strings need to be displayed (first three letters of first and last item in the page). The only way to do it, was to tweak *makePage* function.

Currently the function is: 

```javascript
// Create page object used in template
      function makePage(number, text, isActive) {
        return {
          number: number,
          text: text,
          active: isActive
        };
      }
```

And when it is called, the number is passed as a text:

```javascript
// Add page number links
        for (var number = startPage; number <= endPage; number++) {
          var page = makePage(number, number, number === currentPage);
          pages.push(page);
        }
```

So what I did, is added an extra attribute to the directive, to which I can pass an array with the pagination indexes (separately prepared in a backend, with same size as number of total pages). These values are added when the page is created.

Do you you think it could be a good idea, to add the new attribute, which can be responsible for overriding default number pages? In that case, of course, preparation of valid values for override is a developer's responsibility.

PR #2705, opened 2014-09-16
Add page-label attribute to pagination directive

ashaffer:
The page-label attribute is an expression that returns a label given a page index (in the scope local $page).

This is a fix for issue #2532

### DIFF
--- a/src/pagination/pagination.js
+++ b/src/pagination/pagination.js
@@ -97,6 +97,7 @@ angular.module('ui.bootstrap.pagination', [])
       }
 
       // Setup configuration parameters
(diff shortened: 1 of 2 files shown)


shawnzhu:
This is a good use case for this feature: [add text instead of numbers in pagination, angular-bootstrap](http://stackoverflow.com/questions/29822369/add-text-instead-of-numbers-in-pagination-angular-bootstrap)

t4gedieb:
1+

wesleycho:
Going to close this due to it being outdated - feel free to file a new PR against current `master` with the updates to be in a mergeable state.

PR #2724, opened 2014-09-20
refactor(modal): use ngAnimate for animation

chrisirhc:
Supercedes #1772 

This makes use of the new `$animate:close` event to determine that the animation is done instead of using `$transition`. There are minimal changes to the current code and there's one fix and the removal of some timers.

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -1,4 +1,4 @@
-angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
+angular.module('ui.bootstrap.modal', [])
 
 /**
  * A helper, internal data structure that acts as a map but also allows getting / removing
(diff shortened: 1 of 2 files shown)


Rush:
I'd like to see it merged as well. :)

karianna:
@chrisirhc - it can no longer merge automatically.

chrisirhc:
Should fix #3182 as well. I plan to rebase this once #2699 is merged.

chrisirhc:
Rebased. Now if anyone can take a look, I'll merge this.

src/modal/modal.js
```diff
@@ -207,19 +205,15 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
...
-        if (transitionEndEventName) {
+        if ($animate.enabled()) {
           // transition out
-          var timeout = $timeout(afterAnimating, emulateTime);
-
-          domEl.bind(transitionEndEventName, function () {
-            $timeout.cancel(timeout);
-            afterAnimating();
-            scope.$apply();
+          // TODO: use .one when upgrading to >= 1.2.21
```
wesleycho: We could version sniff (`angular.version.major`, `angular.version.minor`, and `angular.version.dot`) here to handle the differences properly - thoughts?

wesleycho:
Other than that one minor comment (may be a non-issue), this PR LGTM.

src/modal/modal.js
chrisirhc: Ah, since we're requiring 1.3 after this release, .one is already usable. I'll remove this and use .one.

PR #2734, opened 2014-09-23
feat(accordion): optimize accordion to lazy load content

chrishenkel:
I've noticed accordions slowing down the UI when containing a large number of directives. There have been some solutions to this problem, for example:

http://crashthatch.tumblr.com/post/68378038141/lazy-angular-bootstrap-accordions

One of the problems with this approach however, is that it loses the animation, in addition to making your code unnecessarily complicated.

I've added a new directive in this pull request, called accordionBody, that when wrapping content, will lazily load the content on clicking the group link, in addition it will wait for the content to load before expanding a collapsed accordion group.

### DIFF
--- a/src/accordion/accordion.js
+++ b/src/accordion/accordion.js
@@ -54,7 +54,7 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
 })
 
 // The accordion-group directive indicates a block of html that will expand and collapse in an accordion
-.directive('accordionGroup', function() {
+.directive('accordionGroup', function($timeout) {
   return {
     require:'^accordion',         // We need this directive to be inside an accordion
     restrict:'EA',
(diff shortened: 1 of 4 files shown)


src/accordion/accordion.js
```diff
@@ -66,15 +66,23 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
...
+        return $scope.isOpen;
+      };
     },
     link: function(scope, element, attrs, accordionCtrl) {
       accordionCtrl.addGroup(scope);
 
       scope.$watch('isOpen', function(value) {
+        $timeout(function () {
+          scope.isExpanded = scope.isOpen;
+        });
```
wesleycho: Wouldn't this add an extra $digest cycle?

wesleycho:
Closing this as an implementation that is better provided as a third party component if the user desires - this also has suffered bitrot from its age.

Issue #2454, opened 2014-07-12
Use $interval instead of $timeout in the carousel implementation


pkozlowski-opensource:
Based on https://github.com/angular-ui/bootstrap/pull/1630#issuecomment-48760790

PR #2776, opened 2014-10-02
fix(carousel): replaced $timeout with $interval when it was wrong

Gahen:
I had a to tweak a bit two tests since $interval never throws an exception after flushing, hope I did it OK. 

This addresses #2454


thomasthebaker:
Any chance this can be merged for your next release?  This issue is blocking me from using angular-ui as I'm using protractor. Thanks.

PR #2791, opened 2014-10-06
Update modal.js

doranen:
add timeout before setting scope.animate = false; to make sure the modal backdrop's scope gets applied.


wesleycho:
This is likely invalid now with the work that has been done in #2724.

PR #2792, opened 2014-10-06
feat(typeahead): add noResults indicator binding

jtreuting:
Added a binding for typeahead-no-results that is similar to the binding for typeahead-loading.  I wanted to easily display text below the textbox when no results were found and so this seemed like the cleanest way.  Without this, my best solution was to set the flag in my controller each time within the query method when the results are returned.

Usage:

    <div>
        <input ng-model="result" typeahead="item for item in loadMatches()" typeahead-loading="isLoading" typeahead-no-results="isNoResults">
        <div ng-show="isLoading">Loading ...</div>
        <div ng-show="isNoResults">No matches found</div>
    </div>

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -384,6 +384,40 @@ describe('typeahead tests', function () {
       expect($scope.result).toEqual('AL');
       expect(inputEl.val()).toEqual('AL');
     });
+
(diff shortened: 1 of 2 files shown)


wesleycho:
Can you rebase this off of current `master`?

jczerwinski:
Done. See #4068

jtreuting:
Great, glad this got pulled in.  Thanks.

PR #2798, opened 2014-10-07
fix: Replacing timeout with interval for compatibility with Protractor

moses-seeq:
In order to test an application that includes the datepicker using Protractor, $timeout has been replaced with $interval.  In the case of datepicker, this was a small, straightforward change.


vlajos:
I had the same issue with alert. And applied a similar fix: https://github.com/angular-ui/bootstrap/pull/3982
There is a bit more background: https://github.com/angular/protractor/issues/169#issuecomment-49078228

If you agree with this new approach, I can help to port $timeouts to $interval. (Let me know.)

Foxandxss:
We decided that this bug should be fixed in protractor and not here. Closing this.

PR #2801, opened 2014-10-07
fix(timepicker): input attribute

### LINKED ISSUE #2796: Datepicker and Timepicker on single model
Hi,

I want to modify date and time with Datepicker and Timepicker on common model (same model for date and time). In the end I'm sending it to my service. I found some problems though-
Every time I change time, my date is coming back to current date :(

```html
<input datepicker is-open="true" ng-model="date" />
<input timepicker ng-model="date"/>
```

http://plnkr.co/edit/SAdQEV0cnw3IPXVf03q9

On provided plunker model is initialized with some date - change time and date will be set to current date (same behavior when You set model value with Datepicker - timepicker will override date)

Maybe this is some kind of misuse and I should use two separated models - if so, please let me know.

mgibas:
Resolves #2796

### DIFF
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -242,6 +242,7 @@ angular.module('ui.bootstrap.timepicker', [])
     controller:'TimepickerController',
     replace: true,
     scope: {},
+    priority: 1,
     templateUrl: 'template/timepicker/timepicker.html',
     link: function(scope, element, attrs, ctrls) {
       var timepickerCtrl = ctrls[0], ngModelCtrl = ctrls[1];
--- a/src/timepicker/test/timepicker.spec.js
+++ b/src/timepicker/test/timepicker.spec.js
@@ -888,5 +888,29 @@ describe('timepicker directive', function () {
     });
   });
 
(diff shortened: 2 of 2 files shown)


src/timepicker/timepicker.js
```diff
@@ -242,6 +242,7 @@ angular.module('ui.bootstrap.timepicker', [])
     controller:'TimepickerController',
     replace: true,
     scope: {},
+    priority: 1,
```
wesleycho: What is the purpose of this?

mgibas: please take a look at #2796 - there is full explanation.

wesleycho: Setting the priority has a huge code smell, and likely is not the correct solution.  This would be a breaking change.

Even if setting the priority is the approach to go, setting it to 1 is no good - we would need to choose a more appropriate value.

chrisirhc:
If priority fixes this, it's probably an ngModelController change (Angular 1.3) issue. Changing the priority changes the order of the formatter/parser .

wesleycho:
Closing as apparently already resolved & lack of information.

PR #2810, opened 2014-10-09
feat(carousel): Add slideChanged handler.

bgourlie:
I needed this feature for something I'm working on and figured I'd create a pull request with the following intention:

- Get it committed if others might find it useful, or
- based on feedback, determine if there's a better way to accomplish this.

Either way, any feedback would be appreciated.  Thanks!

### DIFF
--- a/src/carousel/carousel.js
+++ b/src/carousel/carousel.js
@@ -60,6 +60,13 @@ angular.module('ui.bootstrap.carousel', ['ui.bootstrap.transition'])
       }
       self.currentSlide = nextSlide;
       currentIndex = nextIndex;
+
+      if($scope.slideChanged){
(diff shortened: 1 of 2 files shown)


src/carousel/carousel.js
```diff
@@ -220,7 +227,8 @@ angular.module('ui.bootstrap.carousel', ['ui.bootstrap.transition'])
     scope: {
       interval: '=',
       noTransition: '=',
-      noPause: '='
+      noPause: '=',
+      slideChanged: '&'
```
wesleycho: Can you explain why this is an `&` binding?

wesleycho:
Closing, as this can be accomplished currently by setting a $watch for the `active` property of a slide, and this appears to have negligible gain.

PR #2814, opened 2014-10-09
feat(popover): added variation of popover with html unsave capabilities

ascheucher:
Hi,

Some people (including me) needed a popover, which is able to use html unsave content.
I implemented it in the same way, the tooltip-html-unsave was done.
Extended the documentation and also copied and adopted the test a bit. This could be better for sure, but it test the same functionality, like on the standard popover.

regards,
Andreas


wesleycho:
I'm going to close this in favor of #1848, which should make it in the 0.13.0 release in the next couple of weeks.

Thanks for the effort though!

PR #2820, opened 2014-10-13
Fix if the openScope doesn´t exist anymore

qstiegler:
If the dropdown is not available in the DOM anymore but with the $document.bind the click event will be called it creates an error as you can´t $apply on a "null" element. This resolves the problem.


wesleycho:
I think this PR is now invalid.  Let us know if this is still an issue.

PR #2822, opened 2014-10-13
Close bug on ng-show when minutes < 10

alexandreDavid:
When ng-hide/ng-show, timepicker's minutes loses it first 0 when exist. In this case timepicker becomes invalid.
When pad function is always used on updateTemplate, 0 doesn't disapear.
So, I also remove useless param on updateTemplate and refresh.

Regards.


wesleycho:
Test is failing.

Also, this should be a configurable value, as some users may want the padding, and some may not.

rvanbaalen:
@alexandreDavid Any update on this?

wesleycho:
Closing due to lack of activity & issues needing resolving with existing work.

PR #2823, opened 2014-10-13
Stringify pad return when value >= 10

alexandreDavid:
In some case, increment function indeeds become timepicker invalid because minutes are invalids.
Solution : stringify value in all case for always return string value when using pad function.


wesleycho:
Landed as 405dab65f72f5bed488eb8f5dfa5a1a848c4a606, thanks!

PR #2825, opened 2014-10-13
fix(tooltip): Use  only once, don't use an empty transclusion fn

Igosuki:
The additional $digest and empty transclusion seem unnecessary in 1.2 and are breaking changes in 1.3 (ng-class breaks).


Igosuki:
Hey @chrisirhc sorry I didn't get back to you on this. #2951 does demonstrate the exact problem.

wesleycho:
Thanks for the find, I am closing this PR as it appears this is no longer an issue with the changes that have been made to the tooltip.

Issue #2789, opened 2014-10-06
datepicker with ng-focus throws $apply error


martinstein:
After the changes in issue #1922, the suggested fix for keeping the old "open on focus" behavior is to use `is-open="dt.open"` and `ng-focus = "dt.open = true"`.

However, using `ng-focus` this way causes the datepicker to break with the error:

    $apply already in progress
    [$rootScope:inprog] http://errors.angularjs.org/undefined/$rootScope/inprog?p0=%24apply

This seems to be a problem for multiple people (in the discussion of #2149):
https://github.com/angular-ui/bootstrap/issues/2149#issuecomment-43695213
https://github.com/angular-ui/bootstrap/issues/2149#issuecomment-44204065
https://github.com/angular-ui/bootstrap/issues/2149#issuecomment-54630847
https://github.com/angular-ui/bootstrap/issues/2149#issuecomment-56773007

So, this looks important for many people, especially since it occurs because of the breaking changes of the #1922 fix.

alexszilagyi:
@chrisirhc: could you please tell me if there's any update on this bug? I'm using lastest version and 1.3.11 and seemts to persist.

wesleycho:
Closing this, as we are moving to Angular 1.3 with 0.13.0.

PR #2845, opened 2014-10-16
Fix for issue #2789 "$apply already in progress" when using ng-focus to open the popup.

pedroxs:
Fix for issue #2789 where it was throwing "$apply already in progress" when using ng-focus to open the popup.


wesleycho:
Closing this, as the original issue is not valid for the upcoming 0.13.0 release.

Issue #2812, opened 2014-10-09
'M' and 'd' date formatter too strict


loicoudot:
* With 'MM' and 'dd' date formatting, date is valid if and only if we have two digits month or day (ex. 01 or 12) => OK 
* With 'M' and 'd' date formatting, the leading '0' is forbidden to validate a date (ex. 1 is valid, 01 is invalid) => OK

But what if we want to tolerate an input with OR without the leading '0'. I think it can be very handy to tolerate the leading '0' of an input.

see the [plunkr](http://plnkr.co/edit/5e7MfSIid3cPZ6nSys81?p=preview)

One could patch the `dateParser` service like this 
```javascript
    'M': {
      regex: '0?[1-9]|1[0-2]',
      apply: function(value) { this.month = value - 1; }
    },
    'd': {
      regex: '0?[1-2]?[0-9]{1}|3[0-1]{1}',
      apply: function(value) { this.date = +value; }
    },
```

ui-bootstrap 0.11.2
angular 1.3.0-rc5

wesleycho:
I do not believe we will support this, as we are paralleling the Angular dateFilter currently - work needs to be done to abstract the date parsing to allow users to use a library like moment.js, at which point this may be a possibility.

PR #2855, opened 2014-10-19
feat(dateparser): optional leading '0' for 'M' and 'd' dateparser format

loicoudot:
Fix for issue #2812

### DIFF
--- a/src/dateparser/dateparser.js
+++ b/src/dateparser/dateparser.js
@@ -30,7 +30,7 @@ angular.module('ui.bootstrap.dateparser', [])
       apply: function(value) { this.month = value - 1; }
     },
     'M': {
-      regex: '[1-9]|1[0-2]',
+      regex: '0?[1-9]|1[0-2]',
       apply: function(value) { this.month = value - 1; }
     },
     'dd': {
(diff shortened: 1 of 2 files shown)


yankee42:
After some mysterious debugging on an application I wrote why on the issue that some people were just not able to input their birthday I figured that what you describe here is the reason. It worked for some people because they entered dates with preceding zeros or used the date picker dialogue while others did not and I did not test this either.

The behaviour feels unnatural and the only solution would be to tell users that they either must or must not enter their birthday with preceding zeros.

I manually applied @chrisirhc patch (except that I changed the regex for `dd` and `mm` rather than `d` and `m`) and it does seem to work perfectly. I agree with @Garios that parsing for `dd` should be the same as for `d`. Because really... who would expect that failing to specify a leading zero is an error. However I'd also be OK with switching to just `d` in all my date fields...

Anyway: Is there still an issue with that patch? So far nobody disagrees with the patch (except for the discussion whether this patch should be extended even further) or did I miss anything. I would really love to see this resolved :-).

wesleycho:
I think this would need a documentation change so that users of the datepicker know what to expect here.

nipec:
Hi, I need this feature for my project - how can I help to get it into the next release?

wesleycho:
Closing, as the implementation is currently flawed - this needs to be opt-in.

Track this feature request in #4805, hoping to have an implementation ready sometime this week.

PR #2859, opened 2014-10-20
fix(collapse): handling when the content has height zero

### LINKED ISSUE #2836: Collapse fails to update its height when toggling collapse model
I have a simple collapse.

    <ul collapse="isCollapsed">
      <li ng-repeat="item in items">
        {{item}}
      </li>
    </ul>
    
Here is the controller:

    angular.module('ui.bootstrap.demo').controller('CollapseDemoCtrl', function ($scope) {
      $scope.isCollapsed = true;
      $scope.items = [];
      $scope.add = function() {
        $scope.items.push(Math.random());
      }
    });
    
Plunker: http://plnkr.co/edit/ZYE3A4gt2I3JZgOEXxhf?p=preview

When I toggle the **isCollapsed** from true to false and add more items to the list, the collapse height is not updated.

The problem is when toggling **isCollapsed** from true to false, if the list has no item (height: 0px), the collapse will not set the height attribute to auto. Thus, it will make the collapse fail to update its height.

It does not happen when **isCollapsed** is initialized with false value (still with empty list), since the height attribute is auto in this case.

ramandv:
Fixes #2836

### DIFF
--- a/src/collapse/collapse.js
+++ b/src/collapse/collapse.js
@@ -31,7 +31,12 @@ angular.module('ui.bootstrap.collapse', ['ui.bootstrap.transition'])
             expandDone();
           } else {
             element.removeClass('collapse').addClass('collapsing');
-            doTransition({ height: element[0].scrollHeight + 'px' }).then(expandDone);
+            if(element[0].scrollHeight === 0) {
+              expandDone();
+            }
+            else {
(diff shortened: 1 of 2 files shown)


porjo:
:+1:

chrisirhc:
Closing this as #1444 will supersede this fix. Thanks! However, we're switching to using ngAnimate.

Issue #2612, opened 2014-08-20
is-disabled doesn't grey out heading when using <accordion-heading>


jcapalbo:
When using <accordion-heading> for the header, is-disabled="true" will not grey out the heading when the accordion group is disabled.  Functionally it is disabled but a user wouldn't know.

Plunker: http://plnkr.co/edit/IjFFumF1aVKyWuDTrf6Z?p=preview

icfantv:
This is working as expected with latest version of libraries:  please see [this plunker](http://plnkr.co/edit/IjFFumF1aVKyWuDTrf6Z).  Closing as fixed.

PR #2866, opened 2014-10-21
Add watcher to isDisabled to style disabled accordion-heading properly

KillaBee1985:
Fixes [#2612](https://github.com/angular-ui/bootstrap/issues/2612)

When you use 'accordion-heading' directive it transcludes content. So we just need to add 'text-muted' class if isDisabled is true.


wesleycho:
This is better accomplished on the user side, as can be seen on the demo in the docs.

PR #2869, opened 2014-10-22
Make toggling accordion groups keyboard accessible

blah238:
http://plnkr.co/edit/Ih2u4SymbKkC4pdwbDMi


petercrona:
Note that If you have a button or something with "ng-click" inside <accordion-heading></accordion-heading>. Then the URL will change when clicking on the button unless you do $event.preventDefault(). Since you didn't need to stop the event before, this can introduce new behavious on websites.

henrihietala:
I just noticed that this workaround doesn't work for IE9-10 in terms of accessibility.

PR #2888, opened 2014-10-26
fix(modal): Fix focus when the dialog is close or cancelled

Tobino:
This can be reproduced on the sample page on http://angular-ui.github.io/bootstrap/#/modal

* In the Dialog section click 'Open me!' button
* Close the Dialog
* Hit tab key

Result: focus is lost in the page

According to http://www.w3.org/TR/wai-aria-practices/#dialog_modal
When the dialog is closed or cancelled focus should return to the element in the application which had focus before the dialog is invoked. This is usually the control which opened the dialog.


wesleycho:
Thanks for this - this is good work.

PR #2890, opened 2014-10-27
Allow $modal resolve to accept plain values

jeffchienzabinet:
Often I find myself writing code like:
$modal.open({...resolve: {object: function() { return obj; }...})
because the caller already knows what object should be.

This could potentially be condensed to:
$modal.open({...resolve: {object: obj; }...})
if resolve accepts plain values, in addition to functions and injectables.


wesleycho:
Closing as per comments here - feel free to open a PR with better error notification for unsupported resolves.

Issue #1688, opened 2014-01-28
Inline-Datepicker doesn't add ng-(in)?valid


Crisfole:
The inline datepicker allows typing in an invalid (disabled) date, and doesn't add the ng-invalid class to the datepicker.

It seems that the dropdown itself is modified to include the ng-invalid but not the actual input...Am I misunderstanding something?

antoinepairet:
@ovaris see the updated README, we are working on angular 1.3 compatibility

Crisfole:
@ovaris: I'm not part of the team.  I have exactly zero say in the `angular-ui`'s compatibility :)

karianna:
https://github.com/angular-ui/bootstrap#impt---2015-plans-and-angular-13-support

kipusoep:
+1

BenjaminCh:
+1

chrisirhc:
Related work: #3020

Please don't +1, we already know this is a priority.

estvmachine:
How alternative resolve this issue?, if i want a datepicker with validation of min and max. Thanks for the response in advance.

avalon12:
Create a custom directive to do the validation for you, if you really need it.. Until a fix is available.

Here is a blogpost describing such a directive.

https://rashimuddin.wordpress.com/2014/08/24/date-range-custom-validation-in-angular-js/

ComeOutToPlay:
Please refer to http://stackoverflow.com/questions/24359128/bootstrap-angular-datepicker-set-ng-model-invalid-not-change/31595743#31595743 at my response (Aquiles)

narangrishab:
Hello Guys, How to do start date end date validation using Angular JS for bootstrap datepicker?

wesleycho:
I believe this is fixed currently, as the homepage works as expected.

@narangrishab: take a look at the documentation for the datepicker, and read up on how validation in Angular works.

Closing as fixed.

dylaninvodo:
As of September 15th, 2015 the datepicker-popup example at https://angular-ui.github.io/bootstrap/ is not correctly setting the error classes on the input element when a date is manually entered that is before the min-date value.

Screenshot: 
<img width="564" alt="screen shot 2015-09-15 at 2 09 27 pm" src="https://cloud.githubusercontent.com/assets/1335029/9887094/74204544-5bb3-11e5-86d5-c8094633fc08.png">

cen1:
Yep, seems broken to me. Invalid not set on the text input.

[...]

PR #2901, opened 2014-10-29
fix(datepicker): Add minDate and maxDate validation for popup.

stevecavanagh:
Add validation to the input associated with the datepicker-popup.  
Closes: <a href = "https://github.com/angular-ui/bootstrap/issues/1688">1688</a>


antoinepairet:
See #3020 instead
rgds

PR #2912, opened 2014-10-31
feat(typeahead): scroll list when traversing it using arrow keys

suyash1208:
Hi,
If you limit the visual area of dropdown-menu (by setting max-height:150px) and want to traverse the list using arrow keys, the highlighted value can go out side the visual area. List doesn't scroll up(if first value of list is selected) and down(if last value of list is selected) when up arrow key or down arrow key is pressed respectively.
Plunker with issue: http://plnkr.co/edit/zF2DMQXX9PdaEZJNfEgd?p=preview
enhancement: http://plnkr.co/edit/2d59mQcKVTIVdmad3vHg?p=preview

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -117,18 +117,17 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
         } else {
           element.attr('aria-activedescendant', getMatchId(index));
         
-        setTimeout(function(){
-                var activeLiElem = document.getElementById(getMatchId(index));
-                if(activeLiElem){
-                    var activeAElem = activeLiElem.getElementsByTagName('A')[0];
-                    if(activeAElem){
-                       activeAElem.focus();
-                       element[0].focus();
-
-                    }
-                }
-            });
         
+          var activeLiElem = document.getElementById(getMatchId(index));
+          if(activeLiElem) {
+              var activeAElem = activeLiElem.getElementsByTagName('A')[0];
+              if(activeAElem) {
+                  setTimeout(function(){
+                      activeAElem.focus();
+                      element[0].focus();
+                  });
+              }
+          }
         }
       });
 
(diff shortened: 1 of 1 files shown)


src/typeahead/typeahead.js
```diff
@@ -116,6 +116,18 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
           element.removeAttr('aria-activedescendant');
         } else {
           element.attr('aria-activedescendant', getMatchId(index));
+        
+        
+          var activeLiElem = document.getElementById(getMatchId(index));
```
wesleycho: You should inject `$document` into here and use `$document[0].activeElement`.

src/typeahead/typeahead.js
```diff
@@ -116,6 +116,18 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+          if(activeLiElem) {
+              var activeAElem = activeLiElem.getElementsByTagName('A')[0];
```
wesleycho: This makes too much assumption about the presence of an anchor tag in this matcher. Need to guard against no result being returned.

src/typeahead/typeahead.js
```diff
@@ -116,6 +116,18 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+              if(activeAElem) {
+                  setTimeout(function(){
```
wesleycho: Why the `setTimeout`?

src/typeahead/typeahead.js
```diff
@@ -116,6 +116,18 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+                      activeAElem.focus();
+                      element[0].focus();
```
wesleycho: Why the second call to focus?

wesleycho:
This needs some work, and is also missing tests.

wesleycho:
Closing due to the numerous issues as mentioned.

PR #2916, opened 2014-10-31
typeahead-focus-first=false option to prevent first match from being focused

### LINKED ISSUE #908: typeahead list - don't auto select / highlight first item
Need a way to be able to not auto "select" / highlight the first match in the typeahead list. There are workarounds for direct use of bootstrap such as below. We can make custom changes in our copy of angular ui bootstrap typeahead but then we can't easily upgrade later.

http://stackoverflow.com/questions/9564977/bootstrap-typeahead-dont-autoselect-first-item

http://bibwild.wordpress.com/2013/04/04/overriding-bootstrap-typeahead-to-not-initially-select/

wkonkel:
Fixes #908.

Should be 100% backwards compatible with existing markup and functionality.

Current functionality (first result is automatically focused as you type):
`<input typeahead="..."  />` 

New functionality (first result is NOT automatically focused as you type):
`<input typeahead="..."  typeahead-focus-first="false" />`

### DIFF
--- a/src/typeahead/docs/readme.md
+++ b/src/typeahead/docs/readme.md
@@ -51,3 +51,7 @@ The typeahead directives provide several attributes:
 * `typeahead-wait-ms` <i class="glyphicon glyphicon-eye-open"></i>
    _(Defaults: 0)_ :
    Minimal wait time after last character typed before typeahead kicks-in
+
+* `typeahead-focus-first` <i class="glyphicon glyphicon-eye-open"></i>
+   _(Defaults: true)_ :
+   Should the first match automatically be focused?
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -80,7 +80,12 @@ describe('typeahead tests', function () {
         this.message = function () {
           return 'Expected "' + this.actual + '" to be opened.';
         };
-        return typeaheadEl.length === 1 && typeaheadEl.hasClass('ng-hide') === false && liEls.length === noOfMatches && $(liEls[activeIdx]).hasClass('active');
+
+        return (typeaheadEl.length === 1 &&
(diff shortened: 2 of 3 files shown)


BooleanOperator:
This is a very sweet feature, however, I think it is missing the action for tab. When nothing is selected pressing enter closes the typeahead, When pressing tab it goes to the next field but does not close the typeahead list. Not sure if this is the right place to note it.

acgentry:
+1 to closing the typeahead onblur of the input, with templates the form could be covered in a large drop down and the user has to tab back after tabbing off and escaping the screen if he doesn't want to select anything.

pnowy:
+1 to closing the typeahead onblur from me too! I have a form when the user can select something from dictionary or enter new value. When user does not select from popup and use tab the popup still appears. Does anyone know how can I prepare byway for this?

lvostinar:
This commit is not really enough for us, not closing the list at tab out is quite a big deal. For now I just disabled the select by tab key, I think tab should be used for navigation only. Our scenario is that the user wants to quickly navigate using tab between input fields but doesn't want to change the value of the typeahead. So tab in/ tab out action should return the control to its initial state.

icfantv:
@lvostinar, commenting on old and closed issues is a really good way for us to miss stuff.

As yours is a support related request, please follow the instructions [here](https://github.com/angular-ui/bootstrap#need-help).

lvostinar:
I commented here because I think this issue needs to be reopen. The fix is merged in your codebase but is not complete.

icfantv:
@lvostinar, you want it to be reopened because this solution is not enough for you (your exact words).  I interpret this to mean that you are wanting an added feature above and beyond what this issue and fix was designed to address.

If you would like us to consider your request, I ask that you please take the time to open a new issue to track this request we will see about addressing it.  You can even reference this issue from the new one.  Help us help you.

PR #2934, opened 2014-11-06
traverse parent scopes to evaluate the expression given in init-date

dkuerner:
If init-date is set and contains a valid angular expression, parent scopes are evaluated for this expression. First $parent may not contain the expression thus this method traverses all parents for the given expression.


wesleycho:
I believe this is now fixed as a result of 98e2bdfc8b4dd108acdf3fac75a795675889bbc7 .  Thanks for the effort here though!

PR #2940, opened 2014-11-07
feat(pagination): adds an "update" attribute for a callback that is invoked when the page is about to change

mjamin:
The callback gets invoked before the view value is set. If the callback returns a promise, the view value is set after the promise resolves. This allows the user to trigger the loading of paged data and let the pagination reflect the new state *after* the data has finished loading.


wesleycho:
I don't think we want to support this - this can be handled appropriately via ng-change and a callback at the appropriate location. This also adds another watcher (needs to be optional, not required as implemented here), which has negative performance implications.

Issue #2659, opened 2014-09-03
Angular 1.3.0 rc0 causes datepicker to display initial date as unformatted string


byudaniel:
The behavior of $viewValue has changed in [this commit.](https://github.com/angular/angular.js/commit/1eda18365a348c9597aafba9d195d345e4f13d1e) Angular 1.3 rc0 sets $viewValue for dates as a string, so date formatting fails.

nespapu:
Many thanks to @DaveWM and @meyerds!

jmayday:
Thanks for this code. 
If someone could just explain me how exactly this works then it would be great. 

So, my question here is why <code>$viewValue</code> gets date like "Feb 17, 2015" and not in specified format? What's a kind of magic for me is that displayed value is in correct format, despite it differs from $viewValue. So, what is the role of $viewValue?

slavafomin:
Hooray! ))

rvanbaalen:
For anyone still using ~0.12 with Angular 1.3, this is a drop-in fix based on @bcronje's comment

```js
.directive('datepickerPopup', ['datepickerPopupConfig', 'dateParser', 'dateFilter', function (datepickerPopupConfig, dateParser, dateFilter) {
            return {
                'restrict': 'A',
                'require': '^ngModel',
                'link': function ($scope, element, attrs, ngModel) {
                    var dateFormat;

                    //*** Temp fix for Angular 1.3 support [#2659](https://github.com/angular-ui/bootstrap/issues/2659)
                    attrs.$observe('datepickerPopup', function(value) {
                        dateFormat = value || datepickerPopupConfig.datepickerPopup;
                        ngModel.$render();
                    });

                    ngModel.$formatters.push(function (value) {
                        return ngModel.$isEmpty(value) ? value : dateFilter(value, dateFormat);
                    });
                }
            };
        }]);
```

neonexus:
Thank you so much for this simple fix @rvanbaalen!

rvanbaalen:
You're welcome @neonexus :+1:

harrychenca:
@rvanbaalen  Thanks! Works perfectly.

BalajiTE:
Hi rvanbaalen,

Thank you for the fix, but for some reason it's not working for me. Here is how I am using it:

I am using AngularJS 1.3.14, ui.bootstrap.min.js 0.12.1 and ui.bootstrap-tps.min.js 0.12.1 versions

[...]

PR #2943, opened 2014-11-08
fix(datepicker): format values received from external model changes

jessedhillon:
When `ngModel.$render` receives a value updated externally (that is, not as a response to a datepicker event, but some other model update) it does not actually apply configured formatting of any kind. It is assumed that the value is a string, but if it is an instance of `Date` then it is implicitly cast to a string and the element's value is set to that. This fix corrects the behavior by parsing it and applying local filters before assigning the element's value.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -546,9 +546,10 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
 
       // Outter change
       ngModel.$render = function() {
-        var date = ngModel.$viewValue ? dateFilter(ngModel.$viewValue, dateFormat) : '';
-        element.val(date);
-        scope.date = parseDate( ngModel.$modelValue );
+        var date = ngModel.$viewValue ? parseDate(ngModel.$viewValue) : null;
+        var display = date ? dateFilter(date, dateFormat) : '';
+        element.val(display);
+        scope.date = date;
       };
 
       var documentClickBind = function(event) {


uberspeck:
Would love to see this merged as well!

uberspeck:
BTW, here's an example of the bug on jsFiddle
http://jsfiddle.net/uberspeck/4ht8y4nw/

uberspeck:
FWIW, @DaveWM has a workaround in this thread...
https://github.com/angular-ui/bootstrap/issues/2659#issuecomment-60750976

Arkh1:
As a side note...  if ng-minlength is added to the input field it breaks the datepicker.  Anybody have any ideas on what might be the cause of that?

spongessuck:
Why hasn't this been merged yet?

karianna:
@jessedhillon are you able to add a test for this PR?

jessedhillon:
Yeah, I'll try to get to it this weekend

jessedhillon:
Question: isn't this issue covered by the test at line 1316 in `src/datepicker/test/datepicker.spec.js`, `custom format updates the input correctly when model changes`?

karianna:
Should be yes, I assume the test still passes?

alvinyao:
Should it be merged now?

karianna:
Yes - there's a Cherry Picking technique involved which I and other new maintainers are going to learn about in tonight's meeting.  I'm hoping that several PR's will start to get 'merged' over the coming weeks.

jessedhillon:
Sorry for the delay in response. Yes, it's passing the test in my branch. Looks like you already figured it out though! :+1:

antoinepairet:
Should be fixed in https://github.com/angular-ui/bootstrap/commit/5f9afe5a86af0e207eaacd6a9b1f202fd78d7009

Closing for now. Thanks to all

Issue #488, opened 2013-05-31
bug(carousel): carousel fails to update model changes


srfrnk:
Using a carousel with a repeater and an ordeBy filter is great...
Only changes to model are not reflected in the order of the slides.
See fiddle... http://jsfiddle.net/xh4TE/1/
Any way around that?

PR #2953, opened 2014-11-11
fix(carousel): #488 Carousel fails to update model changes.

kkruit:
- Uses dom element index instead of internal slides index.
 - Adds order to indicators to match slide order.


wesleycho:
I'm closing this PR, as it seems the original issue is now addressed in current master - let us know if this is still a problem

Issue #274, opened 2013-03-27
Typeahead doesn't support two-way data binding fully...


Siyfion:
See the plunkr here: http://plnkr.co/edit/3ryOYo

If the typeahead is to truly use "the same, flexible syntax as the select directive (http://docs.angularjs.org/api/ng.directive:select)" then it needs to support two-way binding via this method.

szaccaria:
Hello @pkozlowski-opensource I read that this issue is was closed... I've this jfiddle https://jsfiddle.net/qq4gqn6t/12/  but when I change the model the label not change... Can help me?
Thanks

icfantv:
@szaccaria, support-related questions are best suited by posting them to StackOverflow where there's a significantly wider audience to get them answered.

szaccaria:
@icfantv done! thanks for suggest!

szaccaria:
@icfantv I had made a directive... hope that this helps somebody https://jsfiddle.net/qq4gqn6t/91/ and http://stackoverflow.com/questions/35044544/bootstrap-ui-typeahead-doesnt-support-two-way-data-binding?noredirect=1#comment61870083_35044544

PR #2962, opened 2014-11-13
feat(typeahead): access to $item in typeahead-input-formatter and item resolver

miiihi:
This solves problems with input formatting and initialization. IMO it fixes https://github.com/angular-ui/bootstrap/issues/981, https://github.com/angular-ui/bootstrap/issues/274... 

In short, we add access to $item in formatter, if we have an item, if we don't (ie. on initialization and programaticaly setting value), we use callback for item resolving and rerun formatters, when callback returns.


smurugavel:
Hello..
This is a great accomplishment in typeahead for complex data needs. $item suits me well. I was wondering if there are plans to port this code, so it is compatible with angular ui 0.13.3 and angular 1.4?

thanks
Murugavel

pkozlowski-opensource:
Well, this "great accomplishment" doesn't merge cleanly any more and doesn't have associated tests. There is no way we can merge such a substantial patch without associated unit tests.

@miiihi see my comment above - if you are still interested in getting this in, please rebase on top of current master, add tests and submit a new PR. Thnx!

smurugavel:
I see that as an attempt made and being successful. I agree that this doesn't merge well on latest code but it worked in ui-bootstrap-tpls-0.13.0.js. Yes, I understand the latest version is 0.13.3

pkozlowski-opensource:
@smurugavel it is more about lack of unit tests. Once again, we need to have tests to prove that the code does what it claims and avoid regressions in the future.

miiihi:
Well, I'm interested in getting this merged, since we use this in production - but as there was zero response to my pull request I didn't pursue this any further...

Will rebase and prepare unit tests asap.

PR #2972, opened 2014-11-14
Fix 2331

kbaltrinic:
This fixes the various issues reported under issue #2331.  Specifically it resolves the following issues and provides passing tests to cover each:

* initDate property when present on datepicker-options was not being handled correctly by the datepicker dropdown.
* init-date attribute worked when used on a datepicker but not a datepicker dropdown.  It now works the same on both.
* init-date attribute was only evaluated when the datepicker dropdown was initialized.  It is now watched and will impact the date displayed in the dropdown up until the point were a date is selected or the model is assigned a date programatically.

Please let me know if you have any questions, comments or concerns.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -44,7 +44,18 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
 
   $scope.datepickerMode = $scope.datepickerMode || datepickerConfig.datepickerMode;
   $scope.uniqueId = 'datepicker-' + $scope.$id + '-' + Math.floor(Math.random() * 10000);
-  this.activeDate = angular.isDefined($attrs.initDate) ? $scope.$parent.$eval($attrs.initDate) : new Date();
+
+  if(angular.isDefined($attrs.initDate)) {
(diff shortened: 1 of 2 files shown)


src/datepicker/datepicker.js
```diff
@@ -44,7 +44,18 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
 
   $scope.datepickerMode = $scope.datepickerMode || datepickerConfig.datepickerMode;
   $scope.uniqueId = 'datepicker-' + $scope.$id + '-' + Math.floor(Math.random() * 10000);
-  this.activeDate = angular.isDefined($attrs.initDate) ? $scope.$parent.$eval($attrs.initDate) : new Date();
+
+  if(angular.isDefined($attrs.initDate)) {
+    this.activeDate = $scope.$parent.$eval($attrs.initDate) || new Date();
```
wesleycho: Why the change from testing if `angular.isDefined($attrs.initDate)`?  If it is undefined, will it evaluate `$scope.undefined`?  I think in this case, leaving the check is desirable.

wesleycho:
Landed as 98e2bdfc8b4dd108acdf3fac75a795675889bbc7 - this is pretty solid work, thanks a lot!

PR #2977, opened 2014-11-17
fix(modal): fix failure messages on toBeResolved and toBeRejected

onjiro:
Before fix, empty object displayed as actual.
`Expected "{}" to be ...` => `Expected "resolved value or reason" to be ...`


wesleycho:
Landed as ab919f9f89aeaaa336b5d3fe1aed1e0617b5a482, thanks!

PR #2978, opened 2014-11-17
feat(modal): pass reason when opened promise rejected

onjiro:
Pass rejected reason on opened promise rejected, like $routeChangeError event on ngRoute.
Give chance to handle error depends on rejected reason on opened promise.


wesleycho:
Thanks, this was a good catch!

januswel:
thx, wesleycho!!

PR #3005, opened 2014-11-21
Limit progress bar width to 100%

hsubu:
In the event that a progress bar has a value that is over 100% (such as fundraisers exceeding their goal), the formatting breaks.  The current ng-style attribute allows the width to be whatever the percentage is.  These changes will max the width to 100%.


wesleycho:
Thanks, this will be in the next release.

Issue #2156, opened 2014-05-07
Dropdown directives conflict with bootstrap's JS


lijuenc:
The dropdown directives have a restrict value of `CA`. This means that if you also have bootstrap.js referenced as a script file, the directives conflict with Bootstrap's JavaScript hooks into the `dropdown` and `dropdown-toggle` classes.

See this plunker. Notice that it takes two clicks to open a non-directive dropdown.

http://plnkr.co/edit/AYHYLesd4025dUdfLT7i?p=preview

michaeldjeffrey:
Is there anywhere where the breaking changes are documented between version? This one just tripped me up, and wondering if there are any more I don't know about.

karianna:
@michaeldjeffrey Not that I'm aware of - you can look at all of the issues with a angular1.3 label to get a rough idea.

iKronyck:
thanks @chrisirhc it worked for me in laravel and react js

PR #3007, opened 2014-11-21
fix(dropdown): Added support for class directives in the dropdown component

johnsonw:
Added support for class directives in the dropdown component. The component currently supports attribute directives only.


Delapouite:
class directive support has been intentionally removed after this discussion :
https://github.com/angular-ui/bootstrap/issues/2156

wesleycho:
Thanks for the find @Delapouite - closing as addressed by #2156.

PR #3012, opened 2014-11-24
fix(datepicker): disable title button when in max mode

kevinoid:
When the calendar is in the maximum configured mode, clicking the title button has no effect.  To convey this to users, facilitate styling of this state with CSS, make the non-functional button non-clickable, and to take the button out of focus order, I suggest that the title button be disabled when in the maximum mode.  This PR implements that change.

I appreciate your consideration and welcome your feedback.

Thanks,
Kevin


wesleycho:
This looks good, merged

kevinoid:
Great.  Thanks for the review and merging!

PR #3022, opened 2014-11-26
feat(timepicker): add seconds support

rdinicut:


Zielun:
What's the status of this feature change? Any plans to release it?

siebeneicher:
+1

karma.conf.js
```diff
@@ -27,7 +27,7 @@ exclude = [
 // - Safari
 // - PhantomJS
 browsers = [
-  'Chrome'
```
wesleycho: Undo this change.

src/timepicker/timepicker.js
```diff
@@ -3,7 +3,9 @@ angular.module('ui.bootstrap.timepicker', [])
 .constant('timepickerConfig', {
   hourStep: 1,
   minuteStep: 1,
+  secondStep: 1,
   showMeridian: true,
+  showSeconds:true,
```
wesleycho: Default should be false, to agree with current behavior.

wesleycho:
The directive should default to the current behavior, but allow for optional overriding via a config object at the directive level.

In the directive's isolate scope, we could have
<pre>
scope: {
  config: '=?'
}
</pre>
And if it is present, use it to override the default options.

Foxandxss:
Let's continue in #4272

PR #3029, opened 2014-11-28
fix(Modal): Use attribute observe and add a render promise.

zaknuces:
There are two issues with Modal.
1) Modal Directive relays on timeout service to set the animation and focus. The better alternate is to use attribute service's observe method. It is more consistent then $timeout.

2) As of now, there is no way of knowing whether the modal is rendered or DOM ready. Opened promise is pretty useless as it will only notify that template and resolves are resolves successfully or rejected. To solve this, we can have one more promise called 'rendered' which will be settled when modal is rendered. 

Pros:
- User can manipulate DOM (if needed) when rendered promise is settled.
- Avoid any race conditions in the scenarios where the modal content dependents on the dimension of the container. Like setting the maximum height of the modal content element on the basis of modal dialog height etc. 

Short Coming:
- We should reject rendered promise if rendering takes more than allotted time.


wesleycho:
Landed as 99af5f8a369b13b018ec1e592d82a1a140cbb6bb - thanks, this is some good work!

PR #3030, opened 2014-11-28
test: add buttons.spec.js unit test, coverage 100%

cookfront:
add buttons.spec.js unit test, coverage 100%


wesleycho:
Thanks for this, this is good unsexy but important work.

PR #3035, opened 2014-12-01
feat(pagination): Show ellipsis when rotating

distinctdan:
When we have more pages than can be shown and we're rotating, shows ellipsis' like:

Previous, ..., 3, 4, 5, 6, 7, ..., Next

This also adds a handy function in the tests to convert the pager to a string, so that we can intuitively check the state of the pager all at once, instead of having to check each element's text individually.


wesleycho:
This appears to be a breaking change - this should be an opt-in feature.

pkozlowski-opensource:
Closing as this PR wasn't touched for 6 months and the [comment[(https://github.com/angular-ui/bootstrap/pull/3035#issuecomment-85867895) wasn't addressed. It seems like we've got another PR for the same: https://github.com/angular-ui/bootstrap/pull/3565

PR #3045, opened 2014-12-03
feat(dropdown): Make Auto-Close Dropdowns optional. Fix #2218

### LINKED ISSUE #2218: Make Auto-Close Dropdowns optional
Hi, 

a DropDown always closes when clicked somewhere on the document. It would be great if this would be optional - for example if i put a complex dialog in a dropdown, i want to close it manually. For example: a calendar widget. 

Here you can find a Stackoverflow issue which adresses exactly that: 
http://stackoverflow.com/questions/23698316/how-to-prevent-an-angular-bootstrap-dropdown-from-closing-unbind-event-which-wa/23723686?noredirect=1#23723686

The corresponding lines in code can be found here: 
https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L12

It would be great to apply line12 only if the user whished for example by an option.

mariocasciaro:
Introduces an `auto-close` option which - when specified - works as follows:

  * `outsideClick` - closes the dropdown automatically only when the user clicks any element outside the dropdown.
  * `disabled` - disables the auto close. You can then control the open/close status of the dropdown manually, by using `is-open`.


asafdav:
+1

wesleycho:
@chrisirhc I noticed you mentioned in #2218 about being willing to review a PR if someone implements the feature.  This PR looks pretty good to me, but I'd like a second pair of eyes since I'm too tired at the moment.

That this is an optional feature makes this ok IMO.

src/dropdown/docs/readme.md
```diff
@@ -2,3 +2,7 @@
 Dropdown is a simple directive which will toggle a dropdown menu on click or programmatically.
 You can either use `is-open` to toggle or add inside a `<a dropdown-toggle>` element to toggle it when is clicked.
 There is also the `on-toggle(open)` optional expression fired when dropdown changes state.
+By default the dropdown will automatically close if any of its elements it's clicked, you can change this behaviour by setting the `auto-close` option as follows:
```
Delapouite: if any of its elements it's clicked -> if any of its elements is clicked

src/dropdown/docs/readme.md
```diff
@@ -2,3 +2,7 @@
...
+
+  * `outsideClick` - closes the dropdown automatically only when the user clicks any element outside the dropdown.
```
chrisirhc: Should add `always` here and state it's the default behavior.

src/dropdown/dropdown.js
```diff
@@ -85,6 +93,14 @@ angular.module('ui.bootstrap.dropdown', [])
     return self.toggleElement;
   };
 
+  scope.getAutoClose = function() {
+    return autoClose;
```
chrisirhc: Can just return `$attrs.autoClose` here. This will allow support for dynamic autoClose options. Though unlikely to be useful, we can save a variable and add support for that.

chrisirhc:
Great work. I did a review. This looks good. Just styling and a suggestion.

mariocasciaro:
I'll make the edits as soon as I have 5 minutes. Thanks for reviewing!

mariocasciaro:
@chrisirhc Merged with upstream and addressed all the comments.

wesleycho:
Can you clean up the history?  The merge commit should not be there.

mariocasciaro:
Done (and merged with latest commits)

wesleycho:
Thanks for the work here, this will make it in 0.13.0!

PR #3048, opened 2014-12-04
Update rating.html

jr0jas:
Avoid the stop propagation if the parent of the Rating have click event


wesleycho:
This should be something configurable, via attribute such as `rating-config`, with it being an object with a `stopPropagation` key that can be set to true or false.  Default behavior should be true.

This will require unit tests as well.

wesleycho:
Closing due to flawed implementation - feel free to open a new PR that addresses the concerns mentioned here.

PR #3056, opened 2014-12-06
fix(modal): Promise resolution after close animation #3057

nauticalcoder:
This change attempts to fix Issue #3057
This change moves the result promise resolution to after the close animation is finished.

This is to solve the issue where the result promise is currently getting resolved before the animation completes.  If you display a message box in the promise handler, it is displaying before the modal closes.


fiznool:
+1, is there any chance of this being merged in?

I'm currently just setting an arbitrary timeout to stop content inside my modal from changing while the animation runs, and it would be much better to have this fix.

karianna:
@nauticalcoder Are you able to add some tests for this?

nauticalcoder:
I will work on adding some tests for this.  It might be a day or two.

SnackyPete:
It would be better if this fix was designed as a Promise, rather than a callback. :)

wesleycho:
Thanks for the effort, but I believe this is now addressed by @chrisirhc's work on transitioning to using ngAnimate for the modal.

wesleycho:
Hmm I closed this prematurely I believe, but this implementation is flawed - the problem is that it should be opt-in, and hidden behind a configuration option so as not to be a breaking change.

PR #3065, opened 2014-12-08
make it useable in xhtml documents

sashless:
XHTML does not allow empty attributes

Produces errors like  

`
Failed to set the 'innerHTML' property on 'Element': The provided markup is invalid XML, and therefore cannot be inserted into an XML document.
`


Delapouite:
Note: attribute minimization in XHTML is described here : http://www.w3.org/TR/2000/REC-xhtml1-20000126/#h-4.5

karianna:
Are you able to add a test for this please?

PR #2316, opened 2014-06-09
Added support for up/down arrow keys on timepicker

GFoley83:
Added the ability for the user to adjust hours and minutes for the timepicker by using the up/down keyboard arrow keys. This allow the user to adjust time without the need to use the mouse as they can simply tab across the input boxes (as I haven't used `e.preventDefault()` for the `keydown` event).
By default the arrow keys are supported but this can be turned off by using an additional timepicker attribute `arrow-keys="false"`

PR #3068, opened 2014-12-08
feat(timepicker): have up/down arrow keys control time selection

joshfriend:
Similar to #2316, but:
* has tests/docs.
* properly consumes up/down events with `preventDefault`.

### DIFF
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -6,7 +6,8 @@ angular.module('ui.bootstrap.timepicker', [])
   showMeridian: true,
   meridians: null,
   readonlyInput: false,
-  mousewheel: true
+  mousewheel: true,
+  arrowkeys: true
 })
 
 .controller('TimepickerController', ['$scope', '$attrs', '$parse', '$log', '$locale', 'timepickerConfig', function($scope, $attrs, $parse, $log, $locale, timepickerConfig) {
@@ -24,6 +25,11 @@ angular.module('ui.bootstrap.timepicker', [])
     var mousewheel = angular.isDefined($attrs.mousewheel) ? $scope.$parent.$eval($attrs.mousewheel) : timepickerConfig.mousewheel;
     if ( mousewheel ) {
       this.setupMousewheelEvents( hoursInputEl, minutesInputEl );
+    }
+
+    var arrowkeys = angular.isDefined($attrs.arrowkeys) ? $scope.$parent.$eval($attrs.arrowkeys) : timepickerConfig.arrowkeys;
+    if (arrowkeys) {
+      this.setupArrowkeyEvents( hoursInputEl, minutesInputEl );
     }
 
     $scope.readonlyInput = angular.isDefined($attrs.readonlyInput) ? $scope.$parent.$eval($attrs.readonlyInput) : timepickerConfig.readonlyInput;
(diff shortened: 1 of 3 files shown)


src/timepicker/timepicker.js
```diff
@@ -114,6 +120,39 @@ angular.module('ui.bootstrap.timepicker', [])
 
   };
 
+  // Respond on up/down arrowkeys
+  this.setupArrowkeyEvents = function( hoursInputEl, minutesInputEl ) {
+    hoursInputEl.bind('keydown', function(e) {
+      if ( e.which === 38 ) { // up
+        $scope.$apply(function() {
+          $scope.incrementHours();
```
wesleycho: This should be rewritten as
<pre>
$scope.incrementHours();
$scope.$apply();
</pre>

src/timepicker/timepicker.js
```diff
@@ -114,6 +120,39 @@ angular.module('ui.bootstrap.timepicker', [])
...
+        });
+        e.preventDefault();
+      }
+      else if ( e.which === 40 ) { // down
+        $scope.$apply(function() {
+          $scope.decrementHours();
```
wesleycho: Similarly here.

src/timepicker/timepicker.js
```diff
@@ -114,6 +120,39 @@ angular.module('ui.bootstrap.timepicker', [])
...
   };
 
+  // Respond on up/down arrowkeys
+  this.setupArrowkeyEvents = function( hoursInputEl, minutesInputEl ) {
+    hoursInputEl.bind('keydown', function(e) {
+      if ( e.which === 38 ) { // up
+        $scope.$apply(function() {
+          $scope.incrementHours();
+        });
+        e.preventDefault();
```
wesleycho: This should be at the top of the if conditional

src/timepicker/timepicker.js
```diff
@@ -114,6 +120,39 @@ angular.module('ui.bootstrap.timepicker', [])
...
+      }
+      else if ( e.which === 40 ) { // down
+        $scope.$apply(function() {
+          $scope.decrementHours();
+        });
+        e.preventDefault();
```
wesleycho: Same here

src/timepicker/timepicker.js
```diff
@@ -114,6 +120,39 @@ angular.module('ui.bootstrap.timepicker', [])
...
+      }
+    });
+
+    minutesInputEl.bind('keydown', function(e) {
+      if ( e.which === 38 ) { // up
+        $scope.$apply(function() {
+          $scope.incrementMinutes();
+        });
+        e.preventDefault();
```
wesleycho: Same with this if block

src/timepicker/timepicker.js
```diff
@@ -114,6 +120,39 @@ angular.module('ui.bootstrap.timepicker', [])
...
+      }
+      else if ( e.which === 40 ) { // down
+        $scope.$apply(function() {
+          $scope.decrementMinutes();
+        });
+        e.preventDefault();
```
wesleycho: Same here.

src/timepicker/timepicker.js
```diff
@@ -26,6 +27,11 @@ angular.module('ui.bootstrap.timepicker', [])
       this.setupMousewheelEvents( hoursInputEl, minutesInputEl );
     }
 
+    var arrowkeys = angular.isDefined($attrs.arrowkeys) ? $scope.$parent.$eval($attrs.arrowkeys) : timepickerConfig.arrowkeys;
```
wesleycho: This would be better changed to use the isolate scope in the directive.

joshfriend: meaning I should use `$scope.$eval` instead of `$scope.$parent.$eval`?

src/timepicker/timepicker.js
```diff
@@ -26,6 +27,11 @@ angular.module('ui.bootstrap.timepicker', [])
       this.setupMousewheelEvents( hoursInputEl, minutesInputEl );
     }
 
+    var arrowkeys = angular.isDefined($attrs.arrowkeys) ? $scope.$eval($attrs.arrowkeys) : timepickerConfig.arrowkeys;
```
wesleycho: I meant modify the directive scope to set `arrowkeys: '=?'`.  It should be present by the time the init function is run, so you can directly pull the value from `$scope.arrowkeys` in that case.

wesleycho: Actually upon looking at the rest of the source, change this back to $scope.$parent.$eval - I would rather have consistency with the rest of the source for now.

wesleycho:
This LGTM once that change is made back to use $scope.$parent.$eval.

joshfriend:
@wesleycho would you like me to rebase this branch so that 0027763 and 8eb93ce are squashed?

wesleycho:
That won't be necessary, I can do the rebasing if needed.

wesleycho:
Landed as 22961157ad4636db12336abba54cc893554b99d6, thanks!

PR #3085, opened 2014-12-10
Change totalPages watcher to avoid resetting page number before data loa...

synapscommerce:
...ds

If data is being loaded asyncronously but the page number is already known then pagination will update page number to 1 initially before ajax data comes back, and it will stay on page 1 instead of the page originally intended.

e.g. if $scope.currentPage = 2; then <pagination ng-model="currentPage" ng-change="changed"  ....  ></pagination> will update $scope.currentPage to 1 when the view loads initially (no data = 1 page) and fire the change handler, putting the user back on page 1.


src/pagination/pagination.js
```diff
@@ -54,6 +54,8 @@ angular.module('ui.bootstrap.pagination', [])
   });
 
   $scope.$watch('totalPages', function(value) {
+    if(!$scope.totalPages)
```
wesleycho: Change this to
<pre>
if (!value) {
  return;
}
</pre>

wesleycho:
Can you also add a unit test for this?

ovarunendra:
+1

wesleycho:
This should now be fixed in `master` via 30099a0cd080936b22027d37c52b3b32f1fb3ad7

PR #3108, opened 2014-12-15
Modify resources for publishing to npm. 

### LINKED ISSUE #1636: install via npm
Currently, angular-bootstrap is installable via bower and nuget. Would be great to register package in registry at https://npmjs.org/ to let it be installed via `npm install angular-ui-bootstrap`

maxkorp:
This PR resolves #1636 

This adds a .npmignore to reduce the size of a published NPM package (while allowing the built files in ./dist to be published), as well as setting npm's prepublish script to run grunt build, so the files don't have to be built after pulling in from npm.

Alternatively, we could have this be an install script (so once they pull the source from NPM, it compiles locally), but as described on the NPM site that is generally an antipattern, and where possible the ready-to-go concatenated scripts should be published instead. 

With this NPMignore setting, the only files that should be present once published to NPM should be the compiles js in dist, the npmignore, the changelog, the readme, license, and package.json.


rnicholus:
It would be great if angular-bootstrap could be published to npm.  Looks like mostly everything is in place here.  Any idea when this can be published to npm?

maxkorp:
It seems all of the angular-ui stuff using a specific publishing schema. I want to modify that to support publishing to NPM as well, but I jjust havent had the time.

See here: https://github.com/angular-ui/ui-layout/issues/71

rnicholus:
It will probably be easiest if someone just forks this project and publishes it to npm.  I may do that in the near future in fact (unless someone beat me to it).

maxkorp:
Possibly, but there are a number of useful modules on here, and forking a bunch of them just to have publishable versions seems unnecessarily complicated. Short term easiest, yes, but long term certainly not ideal, unless theres pushback to allowing the publishing via the custom platform.

rnicholus:
Hmm, I guess I'm not familiar with this "custom platform" for publishing.

maxkorp:
See here: https://github.com/angular-ui/ui-layout/pull/72#issuecomment-67903371

Here is the repo that handles their publishing code: https://github.com/angular-ui/angular-ui-publisher

maxkorp:
For reference: https://github.com/angular-ui/angular-ui-publisher/issues/3

PR #3109, opened 2014-12-16
Use `ng-if` to hide inactive tabs

Dawil:
Hi there. By using `ng-if` the scope for the tab is destroyed when hidden. If there are a lot of large tabs then this can significantly reduce the length of a $digest cycle. Feel free to close the PR I just thought I'd mention this since I found it gave large performance improvements for my app.

It's also easy for people to add this to their app without forking or modifying the code by simply placing the following code above their application code:

```
angular.module("template/tabs/tabset.html", []).run(["$templateCache", function($templateCache) {
  $templateCache.put("template/tabs/tabset.html",
    "\n" +
    "<div class=\"tabbable\">\n" +
    "  <ul class=\"nav {{type && 'nav-' + type}}\" ng-class=\"{'nav-stacked': vertical, 'nav-justified': justified}\" ng-transclude></ul>\n" +
    "  <div class=\"tab-content\">\n" +
    "    <div class=\"tab-pane\" \n" +
    "         ng-repeat=\"tab in tabs\" \n" +
    "         ng-if=\"tab.active\" \n" +
    "         ng-class=\"{active: tab.active}\"\n" +
    "         tab-content-transclude=\"tab\">\n" +
    "    </div>\n" +
    "  </div>\n" +
    "</div>\n" +
    "");
}])
```

Note that in some cases this could change the behavior of your code and it's not surprising some of the tests failed. If you did want to in some way merge this change you'd probably rather have it as a configurable option or advise in the docs the ability to override the templateCache.

### DIFF
--- a/src/tabs/tabs.js
+++ b/src/tabs/tabs.js
@@ -95,6 +95,7 @@ angular.module('ui.bootstrap.tabs', [])
     link: function(scope, element, attrs) {
       scope.vertical = angular.isDefined(attrs.vertical) ? scope.$parent.$eval(attrs.vertical) : false;
       scope.justified = angular.isDefined(attrs.justified) ? scope.$parent.$eval(attrs.justified) : false;
+      scope.hideTabs = angular.isDefined(attrs.hideTabs) ? scope.$parent.$eval(attrs.hideTabs) : false;
     }
   };
 })
--- a/template/tabs/tabset.html
+++ b/template/tabs/tabset.html
@@ -2,7 +2,8 @@
   <ul class="nav nav-{{type || 'tabs'}}" ng-class="{'nav-stacked': vertical, 'nav-justified': justified}" ng-transclude></ul>
   <div class="tab-content">
     <div class="tab-pane" 
-         ng-repeat="tab in tabs" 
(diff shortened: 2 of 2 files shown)


realityking:
If this is merged, there needs to be a way to opt out. We sometimes use tabs for forms, and form validation wouldn't be performed on the hidden tabs if this was merged as is.

karianna:
Build fails and needs tests

wesleycho:
This needs to be opt-in, not opt-out, as this is a breaking change.

Documentation would also need to be updated about this.

Dawil:
I added two commits via the github interface that demonstrate an example of how to make it opt-in by providing a `hideTabs` flag. I've not tested it nor changed the documentation, this is just an example.

Just to be clear I've already solved this problem in my use case by overriding the `templateCache`, as described in my original post. Don't feel like I'm pressuring you guys to get this feature in.

wesleycho:
The only thing missing here is documentation.

wesleycho:
I am going to close this due to there being adequate workarounds for using ng-if, but keep an eye out for #4836 - we are going to break the current API and fix it.

Issue #3111, opened 2014-12-16
Memory-leak in the typeahead-directive


p4fg:
There exists a memory-leak in the typeahead-directive when used in conjuction with angular 1.2 and jQuery (2.1.0). This bug is related to #2485 that was closed without any action and #2658 that was closed with a patch.

JSFiddle that reproduce the leak: http://jsfiddle.net/wq74xo2t/4/

JSFiddle with patched typeahead: http://jsfiddle.net/wq74xo2t/6/

rvanbaalen:
Please reproduce with UI Bootstrap 0.13.0 and Angular 1.3 to confirm this is still an issue and re-open if it is.

jimmytheneutrino:
Hi, I have this issue with UI Bootstrap 0.13.4 and Angular 1.4.9. 
While the bootstrap version is old I think the reason is the same: 
the two lines starting at
https://github.com/angular-ui/bootstrap/blob/master/src%2Ftypeahead%2Ftypeahead.js#L280

My problem seemed to be fixed by unbinding the two listeners on originalScope.$on('$destroy', ...).

icfantv:
@jimmytheneutrino, this looks like a good catch.  Would you mind please opening a new issue for this and referencing this issue?  Also, if you're feeling up to it, submitting a well-written PR would be accepted.  Thanks.

RobJacobs:
The handlers are being removed in the originalScope destroy: https://github.com/angular-ui/bootstrap/blob/master/src%2Ftypeahead%2Ftypeahead.js#L472

icfantv:
Whoops, I missed that.  Thanks @RobJacobs - @jimmytheneutrino the listeners are only added and removed when `append-to-body` is `true`.

jimmytheneutrino:
Good! Note that there was no removal of listeners in version 0.13.4: https://github.com/angular-ui/bootstrap/blob/0.13.4/src/typeahead/typeahead.js 

This means that an upgrade will fix my issue. Thanks!

PR #3112, opened 2014-12-16
Fix for memory-leak in typeahead

p4fg:
Fix for issue #3111


wesleycho:
Landed as b5a80c08f2364db2170766366cc7433661e737ae, thanks!

I slightly modified it since I was hesitant with the immediate element destruction to destroy it only on $destroy.

Issue #3115, opened 2014-12-16
Datepicker completly prevents usage of certain keys.


oleduc:
Hello,

The application I'm currently working on heavily uses keyboard shortcuts and as such, I cannot have a library that calls event.stopPropagation() and doesn't provide an option to disable it.

In the datepicker controller:
```javascript
  $scope.keydown = function( evt ) {
      console.log('Key downnned!');
    var key = $scope.keys[evt.which];

    if ( !key || evt.shiftKey || evt.altKey ) {
      return;
    }

    evt.preventDefault();
    evt.stopPropagation();

    if (key === 'enter' || key === 'space') {
        console.log('self.isDisabled(self.activeDate)',self.isDisabled(self.activeDate));
      if ( self.isDisabled(self.activeDate)) {
        return; // do nothing
      }
      $scope.select(self.activeDate);
      focusElement();
    } else if (evt.ctrlKey && (key === 'up' || key === 'down')) {
      $scope.toggleMode(key === 'up' ? 1 : -1);
      focusElement();
    } else {
      self.handleKeyDown(key, evt);
      self.refreshView();
    }
  };
```

There should be a way to disable evt.stopPropagation(); and let the application that uses the datepicker decide if it wants to do something the with the key after it has been pressed.

Pull request is on its way.

# To reproduce, press enter without focusing on a datepicker and then press enter while focusing on the datepicker.
Plunker -> http://plnkr.co/edit/aoDoDtkDg7w6dM7JjajQ?p=preview

wesleycho:
This should be addressed now that #3116 is in.

PR #3116, opened 2014-12-17
Fix datepicker event propagation being blocked by default by adding an option (shortcut-propagation).

oleduc:
This would be a fix for issue #3115. This allows the propagation of shortcuts when using the inline datepicker. This is a requirement for applications that makes a heavy usage of contextual shortcuts.


wesleycho:
Landed as 89ab4580cd3d0bbdf63e02fbebe2cb3a9f2616dd, thanks!  I added 13bd516cfcc1b2ee940112675830f7ce3d93ea4f to address @MikeMatusz's comment

Issue #2970, opened 2014-11-14
Modal backdrop broken with bootstrap 3.3.1


LeleDev:
**IMPT** Read bottom message, Bootstrap update has likely fixed this

To be more precise, broken due to this commit:
https://github.com/twbs/bootstrap/pull/14927/files

Temporary workaround: override new css code with older one

https://github.com/angular-ui/bootstrap#impt---2015-plans-and-angular-13-support

GordanBlake:
+1

R0cKET:
+1

staslev:
@allaire thanks!

wcpaez:
+1, allaire temp fix solves problem

sean3z:
+1

gbozee:
+1

bodiddlie:
+1, allaire temp fix solved issue, but I had to add position: fixed; for .modal-backdrop as well

TheWaWaR:
+1, and thanks for allaire's temp fix!

odellcraig:
+1

ClearCloud8:
+1

karianna:
@allaire Are you able to submit a PR (with a test) with your fix?

allaire:
@karianna I'm swamped at work currently, won't have time to do this this week :/

karianna:
No rush! I'm one of the very new maintainers so I'm looking to do some Cat Herding on the various PR's and workarounds like yours over the coming month.

maxisam:
@karianna @allaire I think using CSS to fix this issue won't work in certain browser. That is why bootstrap team handle it in JS. You can check out their discussion. As far as I know, my implementation is just like what they did with jQuery. However, I agree with @pkozlowski-opensource, my solution isn't pretty. Not so ng.

ezzatron:
@bodiddlie Thanks for the `position: fixed;` comment. Without this, the modal backdrop does not work when opening a modal at the bottom of a long page (one with a vertical scrollbar).

IngwiePhoenix:
When I was testing the `bootstrap3-dialog` plugin with 3.3.1, I had two issues:

- The dialog was behind the backdrop.
- The backdrop only covered a felt 50% of the page. The top and bottom part was not covered at all.

I havent gotten to test the fixes displayed in this issue so far - but it'd be great to have this fixed... I already paniced, that it could have been my own CSS stuff ;)

trumbitta:
:+1:

[...]

PR #3117, opened 2014-12-17
Update modal.js to handle backdrop issue in bootstrap 3.3.1

maxisam:
To work with bootstrap 3.3.1. Ref #2970

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -75,7 +75,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
     };
   }])
 
-  .directive('modalWindow', ['$modalStack', '$timeout', function ($modalStack, $timeout) {
+  .directive('modalWindow', ['$modalStack', '$timeout', '$document', function ($modalStack, $timeout, $document) {
     return {
       restrict: 'EA',
       scope: {
@@ -106,6 +106,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
           if (!element[0].querySelectorAll('[autofocus]').length) {
             element[0].focus();
           }
+          angular.element($document[0].querySelectorAll('div.modal-backdrop')).css('height',element[0].scrollHeight + 'px');
         });
 
         scope.close = function (evt) {


twwwt:
+1

skmedia:
+1

tony0918:
Any ETA? Thanks.

karianna:
@maxisam Can you add a comment above the ```$(angular.element...)``` line stating the reason for it.  I think this PR is sane, but it'll allow more experienced folks than myself to see why it was added and enhance/revert as needed.

pkozlowski-opensource:
I'm sorry but we can't merge this one as it assumes certain HTML structure of the backdrop template. We can't do this as people might want (and do) override default templates.

maxisam:
@pkozlowski-opensource I thought about that too. But this project is for Bootstrap 3. It makes sense that we need to use modal-backdrop class in the template because it is in the original Bootstrap 3 and it is actually in the code not in the template.

var angularBackgroundDomEl = angular.element('<div modal-backdrop></div>');

 I do believe that we have different ways to resolve the issue, but I think it would require to change more than one line and one place.

pkozlowski-opensource:
@maxisam we don't need to have one-liner fix. If a more generic solution requires more code so let it be. How about writing a dedicated directive that sets a height of a given element to `scrollHeight` and invoking this directive from a template?

pkozlowski-opensource:
We shouldn't be afraid of creating small, focused directives - ones that are doing one thing well, are easy to test etc. The advantage of this approach is that people then can compose those smaller primitives.

maxisam:
@pkozlowski-opensource  do you mean we create a directive invoked by the backdrop to set its height with the scrollHeight of modalwindow  ?

kadamska:
@allaire your fix doesn't work for me when the modal is invoked after scrolling down... It works fine with no scroll offset.

harsh55:
@kadamska : Just add this
.modal-backdrop{
    bottom: 0;
    position: fixed;
}
It will work for sure.  :+1:

xtreemrage:
@kadamska : it works! :+1:

betoecomp:
Using this approach also @harsh55 :thumbsup:

PR #3119, opened 2014-12-17
Only resetMatches when showing matches.

mwinkels:
Make the dismissClickHandler less aggressive by only resetting matches when matches are being shown. Use $apply insetad of $digest.


wesleycho:
This needs tests around the $apply change.

If that change is undone and $digest is kept, I would be happy to merge this as is though.

wesleycho:
Merging this without $apply change.

PR #3132, opened 2014-12-22
Remove trim() function when compiling tplContent

robertjf:
the trim() function is not supported in IE8 and removing it doesn't seem to affect the template compilation at all...

There are no other changes required to typeahead in order to make it work on IE8 with AngularJS 1.2.x


wesleycho:
We do not officially support IE8 - have you tested this change in IE9 - 11, Safari 7 & 8, Firefox, and Chrome?

wesleycho:
Actually, I am going to close this, as we are fully breaking with IE8 in our next release 0.13.0 soon, and current master already is incompatible with IE8 with Angular 1.3 as a hard requirement.

PR #3148, opened 2015-01-02
reset week counter when crossing the new year.

sean-stanley:
Currently January counts weeks as 52, 53, 54, 55, 56 but should count 52, 1, 2, 3, 4. This simple fix should make the counting correct.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -262,7 +262,9 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
           var weekNumber = getISO8601WeekNumber( scope.rows[0][0].date ),
               numWeeks = scope.rows.length;
           while( scope.weekNumbers.push(weekNumber++) < numWeeks ) {
-            if (weekNumber === 52) weekNumber = 1;
+            if (weekNumber === 52) { 
+              weekNumber = 1; 
+            }
           }
         }
       };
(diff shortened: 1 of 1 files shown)


antoinepairet:
Travis build is failing due to formatting error spotted by jshint:

````
Running "jshint:files" (jshint) task
Linting src/datepicker/datepicker.js ...ERROR
[L265:C36] W116: Expected '{' and instead saw 'weekNumber'.
            if (weekNumber === 52) weekNumber = 1;
````

regards

sean-stanley:
I've got it passing JSHint now.

diablomarcus:
This looks good to me and I don't see any issues popping out of it. :+1:

japherwocky:
I fixed this on a client project by doing:

    while( scope.weekNumbers.push( (((weekNumber++ -1) % 52) +1) ) < numWeeks ) {}

but really just wanted to say that it's pretty awesome to come here and see a pull request already waiting :)

japherwocky:
Actually... this pull request doesn't work correctly, nor does the line I pasted above.  I've had to do this instead:

    if (weekNumber === 53) {
        weekNumber = 2;
    }

because weekNumber is initially set to 52, the if condition isn't checked until it's bumped to 53.  My modulo version is flawed because the first week on the calendar is really both 52 _and_ 1.

japherwocky:
sean -

I'm sorry, I really don't understand how week counting works.  I guess the week starting with January 4th is actually week 1 after all?

sean-stanley:
I admit it took me some thinking about. But I thought it odd that 52 and 1 were the same. According to this: http://www.epochconverter.com/date-and-time/weeknumbers-by-year.php?year=2014

The first week of the year apparently did begin December 29th. Though confusingly they count 2015 as having 53 weeks. I guess it has to do with the number of Monday's in a year. I might revert my last commit unless I can think of a better way to match the counting to the calendar at epochconverter.com

yanivefraim:
Related: #2306, #3321, #3158

karianna:
PR needs some tests.

chrisirhc:
Closing this as #2306 contains tests and also fixes the same issue.

Issue #2886, opened 2014-10-25
Datepicker only accept text that matches format


stephendparker:
If someone is using a datepicker with popup, like here:

http://angular-ui.github.io/bootstrap/#/datepicker

Is it possible to mark the field as invalid until they have formatted the date like provided?  For instance, if I am typed:
10
It treats this as a valid date (october 1, 2001) .  I would rather treat this as invalid, until I completed by typing of my date like:
10/20/2014

jmayday:
Ahh. Seems like I haven't noticed this issue and also created a similar PR for this. https://github.com/angular-ui/bootstrap/pull/3288

wesleycho:
This should now be fixed in `master`.

PR #3152, opened 2015-01-02
fix(datepicker): Datepicker only accept text that matches format

clockworked247:
fixes issue #2886


jmayday:
HI. I just found this PR. Few days ago I prepared mine, very similar one. But I haven't added additional parameter strict-date. For now I would say it's redundant one because is there really a situation when we have some date format defined (and actually we always have, don't we?) but don't want date to be scrict-parsed? I just can't imagine such requirement.

karianna:
@clockworked247 - This fails in the travis build, can you update the PR?

clockworked247:
@karianna - Sure. Will do.

clockworked247:
@jmayday - I believe the changes are quite similar.  I only added the param to preserve existing functionality, so if you think it's not necessary, I'm all for your change.  Looking at the code, you're removing the fallback and relying on dateParser.parse().  That code appears to effectively do the same thing in returning undefined if the date is not valid.  So that looks good to me.

rvanbaalen:
@clockworked247 The build is still failing. Can you please fix this PR?

clockworked247:
@rvanbaalen My understanding was that my PR was a duplicate of #3294, so I'm OK to just close this and let the issue be hashed out there.  It seems it's under debate though.  

I'm not really following what #3417 conflicts with the change.  Considering we need to validate a date entered manually in the associated text field, it only makes sense to add this kind of change.  Otherwise, the Date() fallback will just change the invalid date to some odd number that has nothing to do with what the user entered.  Super confusing...

Closing this, as it's a duplicate.

rvanbaalen:
Thanks for detecting the duplicate @clockworked247 :+1:

Issue #3167, opened 2015-01-08
Tooltip on button doesn't hide if button is disabled on click


Hypercubed:
Reopening #1926.  If a tooltip is attached to a button that is disabled while the tooltip is displayed the hide trigger cannot fire, therefore the tooltip remains open.  See this plunk: http://plnkr.co/edit/gjA46FQ8Ln74zgi8IMR2?p=preview

I propose adding the following to the tooltip directive:

```
attrs.$observe( 'disabled', function ( val ) {
  if (val && ttScope.isOpen ) {
    hide();
  }
});
```

PR #3171, opened 2015-01-08
fix(tooltip): Fix for issue #3167

antoinepairet:
Fix for #3167 based on #3169 with a test.


wesleycho:
Landed as 87a3607632cae03ead57b377bb91ca27610e7730 - thanks!

Issue #2999, opened 2014-11-19
error on typeahead


Tallyb:
in the last line (in getMatchesAsync):
          var onCurrentRequest = (inputValue === modelCtrl.$viewValue);
          if (onCurrentRequest && hasFocus) {
            if (matches.length > 0) {
an error is thrown that matches.length is not defined. 
on the next attempt - it works ok.

antoinepairet:
It does not happen every time. I have the feeling there is a issue of timing. Using your plunkr, if I quickly use the typeahead the error will show up. If I wait long enough, the error will not happen.

That being said, the fix seems obvious and harmless. thanks @unkleara

rkriitr:
if (matches && matches.length > 0)  prevents the undefined error, but first time, the matched dropdown does not show up anyway. Please tell me what needs to be done.

rkriitr:
sorry, in the code
<input ng-model="brandChainName"
			typeahead="name for name in brandChainNames | filter:$viewValue | limitTo:8"
			typeahead-template-url="columnTwo.html" >
brandChainNames  was not intialized. Now it is working. It is not a bug in bootstrap, it is bug on the coder's part.

PR #3178, opened 2015-01-12
fix(typeahead): resolves property length of undefined error #2999

unkleara:
An error is thrown that matches.length is not defined. So we check if matches is truthy first.

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -453,7 +453,7 @@ describe('typeahead tests', function () {
       changeInputValueTo(element, 'c');
       expect(element).toBeClosed();
 
-      deferred.resolve(['good1', 'stuff']);
+      deferred.resolve(['good', 'stuff']);
       $scope.$digest();
       expect(element).toBeOpenWithActive(2, 0);
     });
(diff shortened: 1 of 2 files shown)


antoinepairet:
@unkleara thanks for the PR. Do you think you can write a test for this?

unkleara:
Opps I changed deferred.resolve(['good', 'stuff']); to deferred.resolve(['good1', 'stuff']); to make sure I was logging a correct area. Want me to change that back?

Let me know any adjustments that need to be made. It can be tested other places but it makes most sense testing it within the promise since it's going through that function first.

antoinepairet:
Yes please change it back.
Thanks for the test.
We will review it and merge it asap (I am on mobile now, so I cant review proprly)

stmichael:
:+1: I'm having the same issue. Is there any progress on this issue? I'd really appreciate it, if it could be merged.

johnnyreilly:
Is there any update on this?  Also causing me issues...

wesleycho:
Is there a particular reason why it returns as undefined?

chrisirhc:
There's no reason for that. I'm removing that expect call as I merge.

Issue #3096, opened 2014-12-14
Problem when pressing Esc on a datepicker 


jcstritt:
When you have the cursor on the input field of a datepicker, an  Esc key press is not propagated to the parent component (a modal box for example), so that the box does not close. When you make the change below on the "scope.keydown" function of this directive, all is ok. For me it's a bug in the two last releases :
```
      scope.keydown = function(evt) {
        if (evt.which === 27) {
          evt.preventDefault();
          if (scope.isOpen) { // new check here !!!
            evt.stopPropagation();
          }  
          scope.close();
        } else if (evt.which === 40 && !scope.isOpen) {
          scope.isOpen = true;
        }
      }; 
```

PR #3179, opened 2015-01-12
Don't stop ESC propagation unless dropdown is open

MikeMatusz:
In reference to issue #3096
If the escape key should have other functionality in the context of a form (for example, if datepicker is on a modal) it makes sense to prevent this from happening if escape is used to close the dropdown.  If the dropdown is closed, however, the event should be allowed to propagate.


karianna:
Hi @MikeMatusz - Are you able to add a test case for this?

MikeMatusz:
Sure thing.  Test case added.

PR #3183, opened 2015-01-13
Make datepicker respect dateFormat inside ng-if

### LINKED ISSUE #3124: datepicker re-rendering doesn't work properly
When re-rendering an input with the datepicker-popup directive, ui-datepicker doesn't seem to take into account the previously selected model value.

Please take a look to [this plunker](http://plnkr.co/edit/7JDJhacNMMOpVuLLzYSM?p=preview), pick a date, then toggle the checkbox two times to remove and recreate the datepicker form field. 

You'll notice that the form field value changed (date formatting is lost), which to form validation issues

Opened a [stackoverflow question](http://stackoverflow.com/questions/27460550/re-rendering-angular-ui-boostrap-datepicker-inconsistency)

hamfastgamgee:
Makes the datepicker respect dateFormat inside ng-if.

It turns out this case didn't fail in angular 1.2 but is necessitated
(along with several other fixes) by angular 1.3 not having the
$viewValue still be a proper Date().  So this patch only covers a piece
of the puzzle.

Partially fixes #3124

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -544,11 +544,13 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
         });
       });
 
-      // Outter change
-      ngModel.$render = function() {
-        var date = ngModel.$viewValue ? dateFilter(ngModel.$viewValue, dateFormat) : '';
-        element.val(date);
-        scope.date = parseDate( ngModel.$modelValue );
+      // Outer change
+      ngModel.$render = function () {
+        if (dateFormat) {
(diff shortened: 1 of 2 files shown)


hamfastgamgee:
To be clear: This change is completely safe in angular 1.2, and actually slightly speeds up rendering by skipping some work in ngModel.$render the first time through.  angular 1.3 just happens to choke on the string representation of the date without this and some other fixes, whereas without this change angular 1.2 gets through the first call to ngModel.$render with the wrong value rendered and then fixes itself after attrs.$observe fires and triggers another ngModel.$render.

wesleycho:
Landed as 2c2dba6d145c95500efe7f6be6248d56ae1aebee, thanks!

PR #3200, opened 2015-01-17
Fixed a double negative

krico:
no model valie is ~~not~~ specified


chrisirhc:
Merged into https://github.com/angular-ui/bootstrap/commit/2724b46282bf38d63b24a7c366fc68e285758cac

PR #3201, opened 2015-01-19
update gulp build to better align with custom builds

leon:
the dist files should not contains the version since this breaks builds when a new version comes out.

The package.json is also updated to better align with npm

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -57,14 +57,14 @@ module.exports = function(grunt) {
           banner: '<%= meta.banner %><%= meta.modules %>\n'
         },
         src: [], //src filled in by build task
-        dest: '<%= dist %>/<%= filename %>-<%= pkg.version %>.js'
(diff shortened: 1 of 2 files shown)


mchlbrnd:
+1! When can we expect this to be merged?

wesleycho:
@chrisirhc & @Foxandxss, what are your thoughts on this? I believe you two are the ones familiar with the release process.

This is a breaking change, but one that may be sensible to do since the file names should just be fixed.

chrisirhc:
We need to rework the publishing process. Right now, the contents of the bootstrap-bower repository should be what's used for publishing to npm. That repository doesn't contain version numbers.

Foxandxss:
I don't think this PR matches what we are doing with our releases. We are using an external repo and all the stuff we generate into /dist is merely internal only.

If you think there is an issue yet, please open an issue for proper discussion.

Thanks.

Issue #1903, opened 2014-03-07
problem when data saving in typeahead with ng-blur and typeahead-on-select event


asithw:
if typeahead use with ng-blur event(when user need additional word after picking from dropdown ) also have to use  typeahead-on-select event (data picking).if some one need to write additional word have use ng-blur event for data saving.

so when selecting item from drop down both events fire with two DIFFERENT model values.so it is better to dissable ng-blur when data picking

http://plnkr.co/edit/8gu3J8jyDSGqBDTYqfuz?p=preview

check with  console log

intensite:
Has this issue been resolved?  Is there a work around?

The problem appears when the user clicks on an item after partially typing content in the typeahead.  The blur event is fired with only the portion that was typed and not the full item selected, thus causing serious validation issues.

ahmedmdmohiuddin:
Any resolution for this issue ?

stefanhuber:
This is still a problem! Not resolved!

Ahmed-Saleh:
I worked on a workaround. but still this an issue here as well.

ceolaflannery:
Any solution for this/work around that anyone has found?

legumanP2T:
A solution could be to have a callback function on the typeahead close event. But there is not…

JulieMarie:
If this issue is closed and all associated issues are closed as well, how does this get fixed?  This is still unresolved as of 2.0.0.

PR #3211, opened 2015-01-21
Optional autofocus on typeahead selection

angi-:
There's an issue with typeahead: when you select an item from the list the input is autofocused. While this is desirable for some users, I clear the input value on focus leaving me with an empty input after selection. That's why I added a parameter that tells the typeahead not to autofocus after selection. This parameter is optional so it will not affect users that don't specify it.

This also adresses the issue #1903


karianna:
Hi @angi- Are you able to add a test for this fix?

angi-:
Done, I added tests for both cases: normal functionality should autofocus, specifying typeahead-autofocus="false" should not autofocus.

wesleycho:
The history needs to be cleaned up here - there should be no merge commit, and it should be squashed into one.

wesleycho:
Closing due to being superceded by #4212.

I apologize that this was not able to get in, but the history was an issue in this PR.

PR #3212, opened 2015-01-21
fix(dropdown): keyboard navigation

### LINKED ISSUE #1228: Arrow keys don't work in DropdownToggle
In the dropdown toggle, the arrow keys to navigate through the items in the list don't work.

Reproducable by going to http://angular-ui.github.io/bootstrap/#/dropdownToggle and selecting "Edit in plunker".

yarl:
Add option to use up and down arrows to navigate through list.
Fixes #1228

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -51,13 +51,14 @@ angular.module('ui.bootstrap.dropdown', [])
   };
 }])
 
-.controller('DropdownController', ['$scope', '$attrs', '$parse', 'dropdownConfig', 'dropdownService', '$animate', function($scope, $attrs, $parse, dropdownConfig, dropdownService, $animate) {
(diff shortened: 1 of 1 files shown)


karianna:
Hi @yarl - Thanks for the PR - are you also able to add some test(s) around it?

porjo:
+1

src/dropdown/dropdown.js
```diff
@@ -90,6 +91,27 @@ angular.module('ui.bootstrap.dropdown', [])
       self.toggleElement[0].focus();
     }
   };
+  
+  scope.selectOption = function( evt ) {
+    var elems = angular.element(self.$element).find('.dropdown-menu a');
```
wesleycho: It would be better to create a directive where when a value is true, then it focuses on the element.

wesleycho:
As mentioned, this is in need of tests currently.

bleggett:
[Have a PR] (https://github.com/angular-ui/bootstrap/pull/3685) that adds this functionality. Started with yarl's fix but made it an optional flag, added some tests, and removed the apparent dependency on jQuery.

PR #3217, opened 2015-01-22
add debounced resize and placement breakpoints for tooltip

imcnally:
Hey Angular UI team,

I'm working on a project where we want to show feature hints, so I'm using UI tooltip to programatically open and close the tooltip (demo [here](http://codepen.io/imcnally/pen/Pwzbrg)). That means that the tooltip can be open on a window resize or orientation change. The problem there is that the tooltip is positioned when it's created. To work around this, I added an optional resize handler, that simply calls `positionTooltip`, debounced, on window resize.

As a compliment, I added the option to specify breakpoints on the `placement` attribute, so you can do:
```html
<tooltip placement="{768 : 'left', 1024 : 'right', default : 'bottom}"></tooltip>
```

So if, on resize, `matchMedia('max-width: ' + resolution + 'px').matches` returns true, the corresponding placement value is updated.


karianna:
PR does need a test :-)

imcnally:
thanks @karianna , added that in 6641375

src/tooltip/tooltip.js
```diff
@@ -130,6 +148,17 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
...
 
+            var repositionTooltip = function() {
+              prepPlacement();
+              positionTooltip();
+            };
+
+            var repositionOnResize = function () {
+              var resizeAttributes = $parse(tAttrs.tooltipResize)() || {};
+              var debounceTime = resizeAttributes.debounce || 250;
+              $window.addEventListener('resize', debounce(repositionTooltip, debounceTime));
```
wesleycho: This needs to be garbage collected.

src/tooltip/tooltip.js
```diff
@@ -279,7 +313,42 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
...
+                  });
+                }
+                // first matched breakpoint sets the placement
+                for (var i = 0, j = placementBreakpoints.length; i < j; i++) {
+                  if (window.matchMedia('only screen and (max-width: '+placementBreakpoints[i].resolution+'px)').matches) {
+                    return ttScope.placement = placementBreakpoints[i].placement;
+                  }
+                }
+                // if no breakpoints match, use the value 'default'
+                return ttScope.placement = val['default'];
```
wesleycho: Should be `val.default`

src/tooltip/tooltip.js
```diff
@@ -130,6 +148,17 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
               tooltip.css( ttPosition );
             };
 
+            var repositionTooltip = function() {
+              prepPlacement();
+              positionTooltip();
+            };
+
+            var repositionOnResize = function () {
```
wesleycho: Use `function repositionOnResize() {`

src/tooltip/tooltip.js
```diff
@@ -130,6 +148,17 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
               tooltip.css( ttPosition );
             };
 
+            var repositionTooltip = function() {
```
wesleycho: Use `function repositionTooltip() {`

wesleycho:
I am not sure how I feel about the `matchMedia` logic. I think that should be broken up into a different PR.

karianna:
@imcnally - Can you rebase with master please?

src/tooltip/tooltip.js
imcnally: default's a reserved word. isn't it better this way?

imcnally:
thanks @karianna. up to date now.

RobJacobs:
I could see debounce and throttle methods being useful outside of the tooltip directive.  Should these be added to a helper service similar to how the position service is architected?

imcnally:
@RobJacobs I'd be happy to them if you all thought it'd be useful.

wesleycho:
I think we need to handle this in a more general way - this is something that could be useful when positioning any object.

However, one can implement this on the user's side with the current tooltip API through the usage of the `tooltip-is-open` feature.

I am going to close this due to age, but we have some big work on the $position service that should help make positioning much more improved and flexible so that it is seamless on the user side.

PR #3224, opened 2015-01-26
fix(accordion): Using angular.noop instead

jrgleason:
This is a rather small change, I would be more than happy to go a quick grep for other instances but should be a fairly small change. 

According to Angular docs, etc this seems to be the desired way to handle a noop function.

https://docs.angularjs.org/api/ng/function/angular.noop


karianna:
Hi @jrgleason Can you also add/modify the test for this?

jrgleason:
I will look into it but no tests failed so I assumed there was no breaking change. Can you point me toward the test you think needs to change?

jrgleason:
I double checked and I am not sure there is a test that needs to change anywhere as this should be covered by other tests, and existing angular core tests. We would be testing that angular.noop == function(){}.

chrisirhc:
Thank you @jrgleason , I've reworded the commit to say "refactor" as this doesn't change any functionality, and hence, also doesn't require any tests.

PR #3235, opened 2015-01-28
Inner tag holding tooltip content

rlogiacco:
The idea is to provide support to something like

```
<span tooltip="@myTag">
  This is the span text<myTag>this is the tooltip text</myTag>which can be interpolated
</span>
```

I didn't write unit tests as I'm not sure this is something you might want to add: I find it mandatory because I want to allow CMS editors to manipulate the tooltip text via WYSIWYG editors.

I added a default tag name `tooltip` to be used whenever the only provided value is the `@` char: convention over configuration....

### DIFF
--- a/src/tooltip/test/tooltip2.spec.js
+++ b/src/tooltip/test/tooltip2.spec.js
@@ -133,4 +133,26 @@ describe('tooltip directive', function () {
     expect(fragment).not.toHaveOpenTooltips();
   });
 
+  it('should retrieve content from default inner tag', function () {
+    var fragment = compileTooltip('<span tooltip="@">Trigger here<tooltip>Tooltip text</tooltip></span>');
+
+    var tt = fragment.find('span');
+    tt.trigger( 'mouseenter' );
+    expect( tt.text() ).toBe( 'Trigger here' );
+
+    var tooltipScope = tt.scope().$$childTail;
+    expect( tooltipScope.content ).toBe( 'Tooltip text' );
+  });
+
+  it('should retrieve content from named inner tag', function () {
+    var fragment = compileTooltip('<span tooltip="@myTag">Trigger here<myTag>Tooltip text</myTag></span>');
+
+    var tt = fragment.find('span');
+    tt.trigger( 'mouseenter' );
+    expect( tt.text() ).toBe( 'Trigger here' );
+
+    var tooltipScope = tt.scope().$$childTail;
+    expect( tooltipScope.content ).toBe( 'Tooltip text' );
+  });
+
 });
--- a/src/tooltip/test/tooltip2.spec.js
+++ b/src/tooltip/test/tooltip2.spec.js
@@ -155,4 +155,48 @@ describe('tooltip directive', function () {
     expect( tooltipScope.content ).toBe( 'Tooltip text' );
   });
 
+  it('should retrieve content from default interpolated inner tag', function () {
+    var fragment = compileTooltip('<span tooltip="@">Trigger here<tooltip>Tooltip text</tooltip>!!!</span>');
+
+    var tt = fragment.find('span');
+    tt.trigger( 'mouseenter' );
+    expect( tt.text() ).toBe( 'Trigger here!!!' );
+
+    var tooltipScope = tt.scope().$$childTail;
+    expect( tooltipScope.content ).toBe( 'Tooltip text' );
+  });
+
+  it('should retrieve content from attribute if non matching inner tag', function () {
+    var fragment = compileTooltip('<span tooltip="@">Trigger here <b>Tooltip text</b></span>');
+
+    var tt = fragment.find('span');
+    tt.trigger( 'mouseenter' );
+    expect( tt.text() ).toBe( 'Trigger here Tooltip text' );
+
+    var tooltipScope = tt.scope().$$childTail;
+    expect( tooltipScope.content ).toBe( '@' );
+  });
+
+  it('should retrieve content from attribute if non matching inner tag', function () {
+    var fragment = compileTooltip('<span tooltip="@myTag">Trigger here <b>Tooltip text</b></span>');
+
+    var tt = fragment.find('span');
+    tt.trigger( 'mouseenter' );
+    expect( tt.text() ).toBe( 'Trigger here Tooltip text' );
+
+    var tooltipScope = tt.scope().$$childTail;
+    expect( tooltipScope.content ).toBe( '@myTag' );
+  });
+
+  it('should retrieve content from attribute if no inner tag', function () {
+    var fragment = compileTooltip('<span tooltip="@tooltip">Trigger here</span>');
+
+    var tt = fragment.find('span');
+    tt.trigger( 'mouseenter' );
+    expect( tt.text() ).toBe( 'Trigger here' );
+
+    var tooltipScope = tt.scope().$$childTail;
+    expect( tooltipScope.content ).toBe( '@tooltip' );
+  });
+
 });
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -265,11 +265,10 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
                 if (tag) {
                   ttScope.content = tag.html();
                   tag.remove();
+                  return;
                 }
               }
-              if (!ttScope.content) {
-                ttScope.content = val;
-              }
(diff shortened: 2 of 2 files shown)


karianna:
@rlogiacco - The automatic merge failed, you'll need to resolve conflicts and repush

rlogiacco:
Ok, after trying a couple of solutions I believe I'm not getting it right. Any help is appreciated...

karianna:
@rlogiacco - It was actually a test that failed - see the details link on the Travis notice

```
Firefox 31.0 (Linux) tooltip directive basic scenarios with default options should update tooltip when its content becomes empty FAILED
	Expected "<div style="top: 0px; left: 0px; display: block;" tooltip-popup="" title="" content="some text" placement="top" animation="animation" is-open="isOpen" class="tooltip top in fade" ng-class="{ in: isOpen(), fade: animation() }">
	  <div class="tooltip-arrow"></div>
	  <div class="tooltip-inner ng-binding" ng-bind="content">some text</div>
	</div>" to have "1" opened tooltips.
	@/home/travis/build/angular-ui/bootstrap/src/tooltip/test/tooltip2.spec.js:82:7
	
..........................................................................
Firefox 31.0 (Linux): Executed 663 of 663 (1 FAILED) (32.286 secs / 31.341 secs)
```

rlogiacco:
@karianna I understand it's a failing test and I know it is somewhat related to testing the tooltip doesn't show up if it's empty, but I don't really get how I did break the test and what should I do to fix it...
If you, or anybody else, could provide some direction that would help :smile:

karianna:
I'd try one of the options here: https://github.com/angular-ui/bootstrap#support and also the Gitter chat attached to this project.  I'm actually *not* a Javascript developer (I'm a Java/JVM hacker), so I can't advise more than that I'm afraid...

rlogiacco:
Managed to fix the test issue

rlogiacco:
The tool-tip attribute is matched against an inner tag if the content starts with an `@`: if there's no matching inner tag the attribute still defines the tool-tip content.
This is in order to ensure an higher degree of backward compatibility.

rlogiacco:
@karianna is anybody going to merge this request, and the other ones for that matter?

rvanbaalen:
Please be patient with us @rlogiacco we're a small team of unpaid volunteers spending as much time as we have on maintaining and improving this project. But there is a huge backlog of PR's where we need to go through.

We're working on it. Thanks for you efforts on this particular feature.

rlogiacco:
I didn't want to pretend more work from any of you @rvanbaalen and I apologize if I came across like that. I was trying to understand if there is a stall (I see there are a number of pending pull requests) and eventually contribute some of my own time if needed.
I'm personally running a patched version happily :smile:

src/tooltip/tooltip.js
```diff
@@ -260,6 +260,14 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
              * Observe the relevant attributes.
              */
             attrs.$observe( type, function ( val ) {
+              if (val && val.length >= 1 && val.charAt(0) == '@') {
```
wesleycho: <pre>
if (val && val.charAt(0) === '@') {
</pre>
is sufficient.  `String.prototype.charAt` will return an empty string if val is an empty string.

src/tooltip/test/tooltip2.spec.js
```diff
@@ -133,4 +133,70 @@ describe('tooltip directive', function () {
...
+    var tt = fragment.find('span');
+    tt.trigger( 'mouseenter' );
+    expect( tt.text() ).toBe( 'Trigger here' );
+
+    var tooltipScope = tt.scope().$$childTail;
+    expect( tooltipScope.content ).toBe( 'Tooltip text' );
+  });
+
+  it('should retrieve content from named inner tag', function () {
+    var fragment = compileTooltip('<span tooltip="@myTag">Trigger here<myTag>Tooltip text</myTag></span>');
```
wesleycho: `<myTag>` does not seem correct - browsers will transform this tag to `<mytag>`.

src/tooltip/tooltip.js
```diff
@@ -260,6 +260,14 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
...
+                var tag = element.find(val.length > 1 ? val.substring(1) : 'tooltip');
+                if (tag.length == 1) {
+                  ttScope.content = tag.html();
+                  tag.remove();
+                  return;
+                }
```
wesleycho: This is all better broken up as
<pre>
var tag;
if (val.length > 1) {
  tag = element.find(val.substr(1));
  ttScope.content = tag.html();
  return tag.remove();
}

tag = element.find('tooltip');
</pre>

wesleycho:
I hesitate to merge this because I'm not quite sure I understand why this would have to be done this way and cannot be done with something like `tooltip-html-unsafe`.  Can you explain the technical reasons why this is impossible more in depth?

chrisirhc:
Indeed, this seems like a low frequency use case which may turn out to be quite confusing to use.
There can be multiple inner tags (`<span tooltip="@i">H<i>e</i><i>y</i></span>`) and the behavior isn't defined in that case. The attribute syntax is not as ambiguous.

rlogiacco:
First of all, I don't understand how `tooltip-html-unsafe` can possibly provide the feature this patch tries to add: for what I've understood that is used to allow HTML content within the tooltip, while this is intended to use a tag to provide the tooltip content (HTML safe or unsafe).
With regards to the multiple tags my intended behaviour was to pick the first one, but I'm not strongly opinionated on this.
While I agree the attribute syntax is better in most cases it is not always desirable, like in my context where I want to allow CMS users edit those tooltips. By using a tag I can allow tooltip editing through FCKEditor and similar, without it I can't let those users edit the tooltips (no, HTML source exposure is not an option).

In the end, while I understand my case is not the most common, I don't understand why you are not willing to add support for the feature.

src/tooltip/tooltip.js
rlogiacco: There's no check for the '@' character, which is mandatory to distinguish between a tooltip content from a tooltip content reference...

src/tooltip/test/tooltip2.spec.js
rlogiacco: I run these tests successfully... are they failing in your environment?

src/tooltip/tooltip.js
rlogiacco: The error is in checking for `=> ` as it should rather be `>`

Thanks for catching this

wesleycho:
The implementation is what we are hesitant about, not necessarily supporting WYSIWYG editors.

We wouldn't mind supporting your use case, but it has to be done in a way that is well defined.

src/tooltip/test/tooltip2.spec.js
wesleycho: It's not that they would fail - the tests would not catch this mistake.  Unless the element is an SVG element, the browsers will transform all HTML tags to lower case, as per HTML spec.

chrisirhc:
One other thing, this API means that any tooltip that begins with the `@` character will be broken.

Perhaps we can support your use case through allowing you to provide custom content by modying the tooltip controller. I'll look into that.

I think modifying the tooltip's default behavior is risky (as abovementioned) and also complicates the tooltip logic.

rlogiacco:
I believe the implementation I'm suggesting is well defined and I tried to summarize it in the initial pull request, but if you believe it is not then it's up to you to reject my request: so far I've seen stylish improvements in the implementation details and a possibly never failing test (apologies, I'm at work and unable to verify).

rlogiacco:
@chrisirhc that's incorrect: if the tooltip attribute has value `@me` and there is no `me` nested tag than `@me` will be the tooltip content.
The issue raises if there is an inner tag named `me` but you wanted to use `@me` as tooltip content: in that case adding a trailing space will be sufficient to resolve the clash.

Anyway, while I'm willing to discuss this, it is clear to me you are not interested into this specific implementation, which means I'll keep it for myself and will eventually switch to yours if and when you'll have one.

rlogiacco:
Quite a long time have passed since this pull request was opened: should I consider it rejected? Have you come out with some alternative and better implementation?

wesleycho:
Closing, due to age and to concerns that are in need of addressing.

This feature is wanted, feel free to open an issue first so implementation can be hashed out before work is committed prematurely.

Issue #3228, opened 2015-01-27
Problem with $transition on recent Chrome version


warpdesign:
$transition figures out the transitionEnd event name this way:

    var transitionEndEventNames = {
        'WebkitTransition': 'webkitTransitionEnd',
        'MozTransition': 'transitionend',
        'OTransition': 'oTransitionEnd',
        'transition': 'transitionend'
      };

Problem is, with recent Chrome versions (tested with 40.0.2214.93 m (win64), and 42.0.2288.4 canary (64-bit)), $transition uses *webkitTransitionEnd* but Chrome triggers a *transitionend* event instead.

As a result, some Bootstrap directives (eg. carousel) are broken in Chrome versions since end transition is not detected.

For example, you cannot goto next carousel slide more than once since next/prev buttons are disabled until transition end event has been received:

    //Prevent this user-triggered transition from occurring if there is already one in progress
    if (!$scope.$currentTransition) {
      return self.select(slides[newIndex], 'prev');
    }

This may be a bug in Chrome, but meanwhile, replacing above definition with this one fixes the problem:

    var transitionEndEventNames = {
        'transition': 'transitionend',
        'WebkitTransition': 'webkitTransitionEnd',
        'MozTransition': 'transitionend',
        'OTransition': 'oTransitionEnd'
      };

knorrus:
Experiencing same issues. Chrome Version: 40.0.2214.93 m, Win 7 (x64)
Also found bug report: https://code.google.com/p/chromium/issues/detail?id=451756

zackiles:
I'm experiencing the same issues. A few related things are broken in chrome as well which makes me think this is a very recent (within the last few days) bug in chrome. It''s effecting the accordion as well. When clicking the accordion headers some groups with lots of content either fail to display or display after a very long time (15 seconds +). You can find an example at http://liftcannabis.ca/points with the accordion near the bottom of the page in the most recent version of Chrome.

zackiles:
For an update, regarding the accordion issue i mentioned, removing -webkit-transition-duration: .35s; from the ".collapsing" class in bootstraps css fixes the issue with accordions not opening and closing properly for me.

warpdesign:
I have submitted a simple fix: https://github.com/angular-ui/bootstrap/pull/3238

wesleycho:
Is this still an issue in Chrome 41?

zackiles:
@wesleycho from what I've seen, no, this issue seems to have resolved itself now.

[...]

PR #3241, opened 2015-01-29
fix for transitions break in Chrome 40+

warpdesign:
fix for transition break, see #3228


karianna:
Thanks! PR just needs the tests updated :-).

mhaligowski:
@karianna What tests need updating? Travis says it's ok?

Aaike:
I also have no idea what tests would need updating.
But i just experienced this bug and applying this fixes it. please accept the PR.

topikachu:
It seems new chrome made a change that only fires 'transitionend'  if both  'transitionend' and 'webkitTransitionEnd' are bound to dom. Please see my test at http://jsfiddle.net/kgtba8gv/4/

Some library for example ngAnimate uses "transitionend" so ui.bootstrap transitions has no chance to be called.

This PR really helps a lot of if some code need to work with ngAnimate.

wesleycho:
This seems like a fragile fix - it is relying on the order the keys are in on the for var-in loop, which is not reliable.

wesleycho:
I am closing this, as we have decided that we want to move to full ngAnimate support for 0.13, and we are going to be removing this service.  We are close to completion for usage of ngAnimate instead of $transition.

PR #3244, opened 2015-01-30
wrap ngModelCtrl.$modelValue with scope.$eval

szaccaria:
Hello,
 at the rows 22 and 62, I had add "scope.$eval(-----)" to wrap ngModelCtrl.$modelValue because sometimes my ngModelCtrl.$modelValue has the string "true" instead true value... and the check box in first not work as expect.
Thank to all!

### DIFF
--- a/src/buttons/buttons.js
+++ b/src/buttons/buttons.js
@@ -19,7 +19,7 @@ angular.module('ui.bootstrap.buttons', [])
 
       //model -> UI
       ngModelCtrl.$render = function () {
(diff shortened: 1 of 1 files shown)


karianna:
@szaccaria Build fails, please update PR with tests

wesleycho:
Rejecting this as invalid - it is up to the user to ensure that the data used is uniform. If it is desired for it to be a string, then you should create an intermediary layer where the data is typed appropriately when consumed.

PR #3245, opened 2015-01-30
Do not select disabled tab.

SaikiranDaripelli:


karianna:
@SaikiranDaripelli Can you please add a test, thanks

src/tabs/tabs.js
```diff
@@ -45,6 +45,9 @@ function TabsetCtrl($scope, $element) {
     if (tab.active && tabs.length > 1) {
       //If this is the last tab, select the previous tab. else, the next tab.
       var newActiveIndex = index == tabs.length - 1 ? index - 1 : index + 1;
+      if(tabs[newActiveIndex].disabled){
+        newActiveIndex=newActiveIndex-1;
```
wesleycho: Please add spaces - as a general rule of thumb, the code should have the same style as the rest of the file.

wesleycho:
Closing, as one can easily override this on the user's side, and this would be a breaking change. Selection by manual call should allow selecting a disabled tab, but a user clicking the tab should not.

No response or activity in 9 months either.

PR #3257, opened 2015-02-04
fixes #3242: Dropdowns broken on iOS mobile/tablet.

### LINKED ISSUE #3242: Dropdowns broken on iOS mobile/tablet
I was experiencing an issue on 0.12.0 & 0.11.2 where the dropdowns were not closing when clicking outside the element unless I clicked the toggle element directly.

Before fix: All desktops and kindle fire worked, ipad mini and air did not
After fix: All desktops, kindle, and ipads worked

This happens even on the examples page:
http://angular-ui.github.io/bootstrap/#/dropdown

This issue is that on iOS mobile devices the following lines are not triggered:
$document.bind('click', closeDropdown);
$document.unbind('click', closeDropdown);

These reside here:
https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js Line 12 & 26

Fix:
$document.bind('click touchstart', closeDropdown);
$document.unbind('click touchstart', closeDropdown);

Reason explained in this thread:
http://stackoverflow.com/questions/7892863/jquery-click-not-working-with-ipad
tl:dr iOS doesn't fire events for elements that do not have an event directly attached to them unless they are link node types.

Unfortunately I am unable to create a pull request due to company policy.

Best Regards,
Steven

jsdevel:
I tested this locally and verified that it works just fine.  Thanks to @stevokk for the fix.

How soon can this get published on npm?


stevokk:
@jsdevel are you sure this works with links as children of the dropdown? My final solution was mentioned in the referenced issue, maybe that is what is required instead of the touchstart solution.

jsdevel:
@karianna I'm not sure how we can add tests for this as Travis isn't running any sort of touch enabled emulator.  Can you give me some ideas here?

karianna:
@jsdevel I wasn't actually aware of that - so I'm happy to avoid the need for a test for this :-).

jsdevel:
Good to hear @karianna!  Yea, I think device emulators are a major feature request of travis right now.  Are you guys using sauce?  Migrating over to that would be cool, but then you'd have to use protractor or selenium for testing as I'm sure you're already know.

karianna:
Haven't looked into that all yet, I'm still very much new to all of this!

stevokk:
@karianna @jsdevel  I don't think this should be merged, I tried this approach and children on the dropdown never received the click event, see original issue angular-ui/bootstrap#3242. My final fix was a workaround to force iOS to emit events on the window.

```
  // Fix iOS click event dispatching for bootstrap dropdown
  angular.element(window).bind('load', function () {
    var iOS = (navigator.userAgent.match(/(iPad|iPhone|iPod)/g) ? true : false);
    if (iOS) {
      angular.element(document.body).css('cursor', 'pointer');
    }
  });
```

See: http://www.quirksmode.org/blog/archives/2010/10/click_event_del_1.html

karianna:
@jsdevel Can you check and update your PR accordingly?

jsdevel:
@stevokk does the change proposed right now break existing behavior?  Or does it simply not handle all the use cases?

stevokk:
@jsdevel If you check this site: http://www.pokerstars.com/en/blog/

The top right language menu on desktop uses the dropdown, when using 'touchstart' it interrupted the site links click event and just closed the dropdown.

I'd say the current library behaviour is wrong as you can't close the dropdown, but this pull request isn't the right way to solve that as it breaks other functionality. The real solution is the cursor:pointer as shown above but may be considered a hack fix - and so it is.

Maybe an investigation into binding the event differently (bubble vs capture) is the way forward here.

jsdevel:
@stevokk I see what you mean now.  Should the README be updated instead with this information?  Or should the real `hack|fix` get merged into `angular-bootstrap`?  I worry that this is a super global hack that may not be welcomed by users.

wesleycho:
I am closing this, as it seems this does not properly fix the issue based on @stevokk's comments.

We are open to a proper fix though, so if there is one that truly works, feel free to open another PR.

PR #3265, opened 2015-02-06
feat(modal): add the ability to close / dismiss with multiple parameters

PierreReliquet:
Add the ability to close ($close) or dismiss ($dismiss) a modal window with multiple parameters.

This PR has been made because for now if a user wants to pass down multiple parameters from the modal to the promise, it has to wrap the values in an object : 
```javascript
$modalInstance.close({
  param1: 'foo',
  param2: 'bar'
});
```

It would be more straightforward to do it like : 
```javascript
$modalInstance.close('foo', 'bar');
```

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -268,18 +268,26 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
         body.addClass(OPENED_MODAL_CLASS);
       };
 
-      $modalStack.close = function (modalInstance, result) {
+      $modalStack.close = function (modalInstance) {
+        // Only one parameter is mandatory the modalInstance since the other one
+        // are recovered dynamically through arguments object
         var modalWindow = openedWindows.get(modalInstance);
         if (modalWindow) {
-          modalWindow.value.deferred.resolve(result);
+          // Here we slice after first element since it is the modalInstance
+          var args = Array.prototype.slice.call(arguments, 1);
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
I'm afraid that won't happen as you can't resolve / reject promises with multiple values. This would be like having multiple values returned from a function or multiple exceptions thrown.

Sorry, we don't want users to be confused about how promises work.

PierreReliquet:
Oh yes I believe you are right in order to simplify my code I just forgot the promise specification...
However I should investigate on why the test I submit inside the PR passed, because it seems not logical.
Thanks for your response anyway!

pkozlowski-opensource:
Yeh, I must say that I'm also surprised by the tests passing!

PierreReliquet:
I gave a look at why the test did pass and the answer is located in the helper defined in the test helper named `toBeResolvedWith`. The helper only validates the first argument of the promise and ignore others so the second arguments which was `undefined` was never tested against the value `'another parameter'` and resulted in a passing test. However, I do not know if this deserves a PR.

The source code of the toBeResolvedWith helper :
```javascript
toBeResolvedWith: function(value) {
  var resolved;
  this.message = function() {
    return 'Expected "' + angular.mock.dump(this.actual) + '" to be resolved with "' + value + '".';
  };
  this.actual.then(function(result){
    resolved = result;
  });
  $rootScope.$digest();

  return resolved === value;
}
```

PR #3270, opened 2015-02-08
modalClass option for $modal

Paldom:
Adding ```modalClass``` as a new optional parameter for ```$modal``` service. The aim of this param is to specify additional CSS class(es) to be added to a modal-content template.

E.g. ```...modalClass: 'panel panel-danger'... ```, and a template
```
...
<div class="modal-header panel-heading">
    <h3 class="modal-title">Error!</h3>
</div>
...
```
makes the modal look "danger type".

### DIFF
--- a/src/modal/docs/readme.md
+++ b/src/modal/docs/readme.md
@@ -13,6 +13,7 @@ The `$modal` service has only one method: `open(options)` where available option
 * `keyboard` - indicates whether the dialog should be closable by hitting the ESC key, defaults to true
 * `backdropClass` - additional CSS class(es) to be added to a modal backdrop template
 * `windowClass` - additional CSS class(es) to be added to a modal window template
+* `modalClass` - additional CSS class(es) to be added to a modal-content template
 * `windowTemplateUrl` - a path to a template overriding modal's window template
 * `size` - optional size of modal window. Allowed values: `'sm'` (small) or  `'lg'` (large). Requires Bootstrap 3.1.0 or later
 
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -90,7 +90,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
       link: function (scope, element, attrs) {
         element.addClass(attrs.windowClass || '');
         scope.size = attrs.size;
-        scope.modalClass = attrs.modalClass;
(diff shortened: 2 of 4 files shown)


karianna:
@Paldom Could you also add some tests around this please?

Paldom:
@karianna Sure, I've also added doc for modalClass.

RobJacobs:
What does this offer that the windowClass option doesn't?  You can already target the modal-content or any of the other modal components using the windowClass option.

Paldom:
@RobJacobs You're right, you can target modal-content from windowClass, but I think you couldn't simply add e.g. `.panel-danger` class to a modal. With windowClass you need to specify a custom class for that, eg. `.myCustomWindowClass .modal-content { ... copied content of .panel-danger comes here ... }`

Paldom:
For example with modalClass, you simply need to specify `panel panel-danger`:
panel-danger

RobJacobs:
@paldom Thanks for explaining that.  I typically handle that scenario by wrapping my modal template in a div and adding the panel class there.  Example:

    <div class="panel panel-danger" style="margin-bottom: 0px;">
        <div class="modal-header panel-heading">
            <h3 class="modal-title">Error!</h3>
        </div>
        <div class="modal-body panel-body">
    
        </div>
        <div class="modal-footer panel-footer">
    
        </div>
    </div>

Paldom:
@RobJacobs I've tried this one, but it needs a little work around to handle double border. It's not critical, I just decided to make this more simple :)

wesleycho:
Seems to be a lot of custom class stuff people are requesting, all of these requests are going to make the API more confusing for someone first using the modal.

One solution would be to hook into `windowScope` and use a custom modal template that overrides the UI Bootstrap one.

I'm not necessarily against merging this though, I would like other thoughts here.

The commits need to be squashed and this needs to be rebased off of current master first.

pkozlowski-opensource:
I'm with @wesleycho here - we need to think harder about the API. I don't believe adding more "switches" / toggle attributes to the modal is doing any good to it.

We've got a use-case here: I want to be able to change modal window decoration on a modal-window instance basis. So we need to solve it in a way that people can add any stuff they might need. Adding a param for a class would be an work-arround only.

Closing since this PR wasn't updated for few weeks and this is not the direction we want to take.

PR #3272, opened 2015-02-09
Allowing dynamic max value

raananw:
If max was changed after the progress bar was created, the value of max in the progress bar was not updated.
This update will allow the user to dynamically change the max value and update the progress bar correctly.


raananw:
Will reopen after running further tests.

PR #3273, opened 2015-02-09
Allowing (optional) dynamic max in progress bar

RaananW:
If max's value was changed after the progress bar was created, the value of max in the progress bar was not updated.
This update will allow the user to dynamically change the max value and update the progress bar correctly.
Max is treated as an optional variable and is being initialized using the default value (100) if not provided.

### DIFF
--- a/src/progressbar/progressbar.js
+++ b/src/progressbar/progressbar.js
@@ -54,6 +54,7 @@ angular.module('ui.bootstrap.progressbar', [])
         require: '^progress',
         scope: {
             value: '=',
+            max: '=',
             type: '@'
         },
         templateUrl: 'template/progressbar/bar.html',
@@ -71,6 +72,7 @@ angular.module('ui.bootstrap.progressbar', [])
         controller: 'ProgressController',
         scope: {
             value: '=',
+            max: '=',
             type: '@'
         },
         templateUrl: 'template/progressbar/progressbar.html',
--- a/src/progressbar/progressbar.js
+++ b/src/progressbar/progressbar.js
@@ -10,7 +10,7 @@ angular.module('ui.bootstrap.progressbar', [])
         animate = angular.isDefined($attrs.animate) ? $scope.$parent.$eval($attrs.animate) : progressConfig.animate;
 
     this.bars = [];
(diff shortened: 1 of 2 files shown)


karianna:
@raananw - Can you add some tests for this?

raananw:
Sure, I'll extend the progressbar tests.

raananw:
The test added is the only thing that didn't work. Hope it's enough.

src/progressbar/progressbar.js
```diff
@@ -54,6 +54,7 @@ angular.module('ui.bootstrap.progressbar', [])
         require: '^progress',
         scope: {
             value: '=',
+            max: '=?',
```
wesleycho: Would an @ binding be enough here?

RaananW: I believe you wouldn't be able to set the max to be dynamic then. I thought using a variable from a different scope would be helpful here. 
Or did I understand you wrong?

wesleycho: the @ binding is a one way data binding - scope.max would be changeable, and would be used like `max="{{max}}"`

RaananW: Yep, if it is used like that it would probably work. I can test that later today.
But isn't the "&" a one-way binding and @ a kind of "string-only" binding?

src/progressbar/test/progressbar.spec.js
```diff
@@ -116,6 +116,13 @@ describe('progressbar directive', function () {
     it('transcludes "bar" text', function() {
       expect(getBar(0).text()).toBe('22/200');
     });
+    
+    it('adjust the valuemax when it changes', function() {
```
wesleycho: `adjusts`, not `adjust`

src/progressbar/progressbar.js
wesleycho: You're right - this is fine.

RaananW:
Just fixed the adjust to adjusts.

wesleycho:
Already did it, landed as 7ccff028ff5ed19af63d9cf1f5e078223924f3a0, thanks :) .

Issue #2069, opened 2014-04-16
Datepicker and Timepicker use $modelValue instead of $viewValue.


AlexHi:
It seems that all angular controls use $viewValue in the $render function, so we are able to convert model value(using $formatters) before displaying.

Brovert:
+ 1

Brovert:
+1

Brovert:
Will this be implemented in any future versions of ui-boostrap? We just ran into this issue as well and I noticed the problem before seeing this forum.

MoeBDC:
+1

AstrasyA:
+1

karianna:
Hi @cleftheris - Can you submit a PR with an accompanying test for your change?

cleftheris:
Hi @karianna I will try to make a PR as soon as possible. 
I am currently having trouble with the $viewValue in datePicker in angular 1.3.x. 
As it seems there is a default formatter in there put by the framework that converts a javascript Date object to a string using the toString() method in order to render the value to the view. I have hacked my way around it by removing it conditionally.

Will post an update soon

Brovert:
Do we know if this will be fixed in an upcoming version?

antoinepairet:
This commit should have fixed it: 5f9afe5

regrads

cleftheris:
@antoinepairet this is not completely compatible with angular version 1.3.x.
Lets say for example that I have a converter directive named "number-to-date-converter" and I need the raw value passed to the datepicker as popup to remain a number. When I update the raw model value the input gets updated __but not the popup calendar!!__.
here is the directive I am using for testing. (It could be a converter from anything to date and back. I am just using number 2 date for simplicity).

    angular.module('ui.bootstrap.demo').directive('numberToDateConverter', function () {
      return {
        require: 'ngModel',
        restrict: 'A',
        link: function (scope, element, attr, ngModel) {
                //view -> model (runs last)
                ngModel.$parsers.unshift(function (date) {
                    return date.getTime();
                });
                //model -> view (runs last)
                ngModel.$formatters.unshift(function (date) {
                    return angular.isNumber(date) ? new Date(date) : date;
                });
            }
      };
    });

and [here is the plunker](http://plnkr.co/edit/AVMJhWd4NoaR9yzZnzbf?p=preview) demonstrating the problem. 

We should always use the $viewValue inside the render method (check the assignment of scope.date). This fixes the problem.
    
    // Outer change
    ngModel.$render = function() {
[...]

PR #3279, opened 2015-02-10
datepicker popup removes angularjs basicText formatter from ngModel.$formatters

### LINKED ISSUE #2899: ControllerAs with DatePicker binds strings, not Dates
This may be related to another 1.3 issue, but when I use code to assign a date to the DatePicker, I get a date string (and it assigns my binding variable to undefined.)

ALSO: dates are messed up -- added 1 day!

See plunkr at http://plnkr.co/edit/TBClYQ7CHKmPL0H3SbB3?p=info

click on the link .. it will show 3-4 date pickers in a loop .. bad date binding....

cleftheris:
fixes #2899 and #2069 in angular 1.3.x.

As it seems there is a default formatter in there put by the framework that converts a javascript Date object to a string using the toString() method in order to render the value to the view. I have managed to get around it by inspecting the formatters and just removing the default if found.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -523,7 +523,16 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
         }
       }
       ngModel.$parsers.unshift(parseDate);
-
+      // get rid of the default angularjs text formatter that messes with the view value.
+      for (var i = 0; i < ngModel.$formatters.length; i++) {
+          var now = new Date();
+          var fmtFn = ngModel.$formatters[i];
+          var isDefault = fmtFn(now) === now.toString();
+          if (isDefault) {
+              ngModel.$formatters.splice(i, 1);
+              break;
+          }
+      }
       // Inner change
       scope.dateSelection = function(dt) {
         if (angular.isDefined(dt)) {


karianna:
@cleftheris Can you add tests please?

cleftheris:
@karianna there are 8 tests failing __before__ my fix related to datepicker if I use AngularJS v1.3.11 plus updated Angular Mocks in "datepicker.spec.js". 

But my fork of the master branch uses angular.js v1.2.16 and not 1.3.x.
These tests are 
* 200 datepicker directive setting datepickerPopupConfig changes date format
* 202 datepicker directive as popup initially to display the correct value in input
* 207 datepicker directive as popup initially opened updates the input correctly when model changes
* 218 datepicker directive as popup custom format to display the correct value in input
* 220 datepicker directive as popup custom format updates the input correctly when model changes 
* 221 datepicker directive as popup dynamic custom format to display the correct value in input
* 223 datepicker directive as popup dynamic custom format updates the input correctly when model changes 
* 224 datepicker directive as popup dynamic custom format updates the input correctly when format changes

Should I also commit a change in the version of the angular lib used for tests "misc/test-lib/angular.js" & "misc/test-lib/angular-mocks.js"?

Please advise

cleftheris:
@karianna  also keep in mind that I run the tests with my fix in place in angular 1.2.x and all is good there too.

antoinepairet:
@cleftheris Thanks.
Another PR was merged. 
This commit should have fixed it: https://github.com/angular-ui/bootstrap/commit/5f9afe5a86af0e207eaacd6a9b1f202fd78d7009

see the working plunkr: http://plnkr.co/edit/0EEK4JpvW7GXqbhAgVwJ?p=preview

regrads

cleftheris:
@antoinepairet @karianna  Thanks for the update. Its true that my fix is not needed. But I got an additional bug related to commit 5f9afe5 please check my [answer here](https://github.com/angular-ui/bootstrap/issues/2069#issuecomment-75743026). It is related to the viewValue and parsers/formatters.

cleftheris:
@antoinepairet I would like to point out IMHO that parsing the viewValue into a date and then filtering back in order to get the correct string to put in the input element seems like an overkill to me. 

The root cause of the problem is as I point out in my PR description a default $formatter put by angular core that does not apply to datepicker popup directive and should be removed once upon initialization. The accepted solution puts unnecessary overhead to the digest cycle.

Regards

PR #3281, opened 2015-02-10
refactor(dropdown): register module at EOF. Update comments.

alettieri:
I used this directive in one of our internal projects. I needed to update the structure (to match our code base) and commenting for documentation purposes. Thought this might be helpful for the project as well?

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -1,161 +1,339 @@
-angular.module('ui.bootstrap.dropdown', [])
+/**
+ * Drop Down Service
+ * This service is used to manage showing and hiding dropdowns.
+ * It's job is to listen to dom events for keydowns and document click events to hide open dropdowns
+ * It will also ensure that it closes dropdowns in the event that a new one is opened.
+ * 
+ * @param $document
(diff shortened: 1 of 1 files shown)


karianna:
Needs tests at the least :-)

alettieri:
@karianna these changes shouldn't be breaking the previous test cases. It's mostly just changing the code structure and adding comments. I ran these against the current test cases and they still pass. Nothing new was introduced.

wesleycho:
I'm closing this as invalid, this does not match the structure of the rest of the project, making it more difficult to follow for anyone contributing or perusing the source code.

alettieri:
@wesleycho fair point on structure. Where is it publishing methods on window?

wesleycho:
The compiled version is not wrapped in an IIFE, so any globals would be globals on window.

alettieri:
@wesleycho ah, yes, I see what you men. I didn't consider that the compiled version wouldn't be wrapped in an IIFE. Anyways, thanks for clarifying and addressing the pull request. Will keep the style/structure in mind if I ever have something else to contribute.

PR #3288, opened 2015-02-11
Datepicker bugfixing

jmayday:
I removed code allowing values which do not match format being parsed as dates.


antoinepairet:
@jmayday Could you squash the commits to make only one. Thanks in advance

jmayday:
Squashed commits are available under branch https://github.com/jmayday/bootstrap/tree/datepicker-bugfixing-squashed and PR is here https://github.com/angular-ui/bootstrap/pull/3294

jmayday:
I've had a look at https://github.com/antoinepairet/bootstrap/commit/4d97d7cf2a97dae487e7cce4bdc53d414ba262e8 and it doesn't work so well with change I proposed (having call to <code>new Date(viewValue)</code> removed). Of course I don't mean that content of commit https://github.com/antoinepairet/bootstrap/commit/4d97d7cf2a97dae487e7cce4bdc53d414ba262e8 is broken. I would just like to clarify some working combination of these two fixes. 

So, let's describe it this way:

- when my change is kept then date can't be initialized
- when my change is ignored (we keep calling <code>new Date(viewValue)</code>) then we still get values like "1" parsed to model as date which is not necessarily expected

wesleycho:
As mentioned in #2513, this fails some tests when based off of current master, and needs to be updated.

chrisirhc:
Closing as #3294 is a newer version of this PR.

PR #3294, opened 2015-02-13
Do not allow non-dates being parsed as dates

jmayday:
I removed code allowing values which do not match format being parsed as dates.


jmayday:
This is just squashed https://github.com/angular-ui/bootstrap/pull/3288

wesleycho:
This does not look like it would be a valid fix.

I am hesitant to merge this, especially since I have a fix that will likely be merged later this week [here](https://github.com/angular-ui/bootstrap/pull/3417).

wesleycho:
After thinking about this, instead of the new Date constructor, it would be fine to have the value set to null. A test would need to be added for this that it turns invalid date strings/objects into null - this should be accomplishable by essentially doing `element.controller('ngModel').$formatters[0]` to grab the formatter unshifted into the array (if there isn't test code already doing so).

It also should not return undefined, but null IMO. The if-else conditional there can be refactored to then do only one return after $setValidity is called.  This would also require testing

chrisirhc:
I'm moving this out of 0.13.0 because:
1) this is a breaking change. Users maybe relying on this behavior.
2) this needs tests for validity of the date, so there's still some work needed here.

rvanbaalen:
@jmayday Please update your PR so it can be merged when the time comes.

jmayday:
So, to wrap things up (because I'm little confused right now) - should we abandon this PR because of https://github.com/angular-ui/bootstrap/pull/3417 like @wesleycho suggested or should we keep it? (null approach also suggested by @wesleycho)?

Issue #3299, opened 2015-02-16
Preventing event from a button in an accordion header refreshes the page


bgever:
Version 0.12 introduced a problem with my use case. I have an accordion, and in its header I have several buttons. When clicking these buttons, the *default behavior* is to execute the action of that button *and* toggling the body content.

In order to prevent this, I capture the `$event` in the binding, and call its `stopPropagation()` in the button handler.

**As of version 0.12 this causes a refresh of the page.**
See the [bug reproduced in Plunker](http://plnkr.co/edit/NNKBrVxp7nfFmQhuA2Iy?p=preview).

I've found the problem to be originating from commit 992b23297cd100ab4e236fba469e3a70566a4163, where a `href` attribute was added to the header, in order for the keyboar navigation to focus on the link element. However, in turn, the browser will hook up a default handler to navigate to the empty URL. This navigation happens when the $event is stopped propagating, like in my example.

I've found the solution to be to attach a noop to the href attribute; `href="javascript:void(0)"`. This keeps the keyboard navigation working, and when stopping the event, still has the expected behavior of just cancelling the accordion toggle.

chrisirhc:
@bgever You need to add a `$event.preventDefault()` call if you want to `stopPropagation`. That prevents the default action of the click (to navigate, in the case of an anchor element with href). By default, Angular does this for you but you've done `stopPropagation` so it can't do it.

See:
http://plnkr.co/edit/CJ30XW83dqC7U7u7twUM?p=preview

chrisirhc:
Closing as it's resolved via #3300 which has been merged.

bgever:
@chrisirhc I've avoided the `$event.preventDefault()` call, because this might imply the wrong behavior. The nested button click itself should not be cancelled (think of multiple event handlers).

Thanks for pointing this solution, it looks to work great with versions before 0.13.

noullet:
Hi,

The issue is back in 0.13.3 and above.

It seems that it has been brought back by fixing #4104, but this issue seems to be back too.

Here are the examples updated to 0.14.2
* #3299 (current issue): http://plnkr.co/edit/PnF4rRNWBvFP8l8QBKmX?p=preview
* #4104: http://plnkr.co/edit/gEA8EvsgZeLUXmDG0WkR?p=preview

For the record, I want to do exactly the same thing as in #4104: when the user clicks on a button, the accordion is not opened, the page is not refreshed to the current location and the user is redirected to a new location.

[...]

PR #3300, opened 2015-02-16
fix(accordion): noop header link to avoid page refresh with nested buttons

bgever:
When nesting buttons in the header, and calling `$event.stopPropagation()` in their click handlers would cause a page refresh, because the default empty URL would be executed. By changing this into a noop, the page refresh is prevented. #3299

### DIFF
--- a/template/accordion/accordion-group.html
+++ b/template/accordion/accordion-group.html
@@ -1,7 +1,7 @@
 <div class="panel panel-default">
   <div class="panel-heading">
     <h4 class="panel-title">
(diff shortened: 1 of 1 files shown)


karianna:
@bgever Are you able to add a test for this?

maximo72:
+1

wesleycho:
Landed as 9ca4ec399be0e0e8f6c3fe6fd924a7d94ce669b5, thanks!

bgever:
Awesome, thanks! @karianna I'm not sure how to write the test for this issue.

wesleycho:
No worries, this isn't something you really can test - it is fine as is.

PR #3322, opened 2015-02-21
fix(datepicker): `ng-model` value can be a timestamp

### LINKED ISSUE #2345: Timestamp treated as invalid model value in datepicker
The datepicker directive says [ng-model can be "a number of milliseconds since 01.01.1970"](https://github.com/angular-ui/bootstrap/blob/master/src/datepicker/datepicker.js#L73). However [`parseData` will not treat such an input as valid](https://github.com/angular-ui/bootstrap/blob/master/src/datepicker/datepicker.js#L504).

khashayar:
Accept a number of milliseconds since 01.01.1970 as a valid value
for `ng-model`

Closes #2345

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -502,6 +502,11 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
       }
 
       function parseDate(viewValue) {
+        if (angular.isNumber(viewValue)) {
+          // presumably timestamp to date object
+          viewValue = new Date(viewValue);
+        }
+
         if (!viewValue) {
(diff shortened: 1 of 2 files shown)


antoinepairet:
Could you add a test to prevent regression?
Once done, please squash as 1 commit.
rgds

khashayar:
Test has been added.

rvanbaalen:
Is this a fix or a feature?

---

khashayar:
It's a fix for #2345

wesleycho:
Landed as d253208bd7ebbd6209c6a25f70e010123198fcd7, thanks!

PR #3329, opened 2015-02-23
Add ARIA accessibility to dropdown demo

kkirsche:
Add ARIA accessibility to dropdown demo per http://getbootstrap.com/javascript/#dropdowns-usage


karianna:
Is it possible to add sensible tests for this?

kkirsche:
I'm not sure how you would test things like roles which are passive and I'm assuming the aria ones can't be as Bootstrap doesn't have tests for this in their own framework. If you have an idea of how it would be done I can but I'm not sure where to start with adding those tests

icfantv:
Closing in favor of #4772.

PR #3333, opened 2015-02-24
feat(tooltip):  Support for tooltip-class configuration

### LINKED ISSUE #3126: Feature Request: Allow custom classes to be applied to tooltips
I would like away to apply custom styling to tooltips using CSS classes. Currently the only way to distinguish between different tooltips in CSS is the direction of the tooltip. Styling the tooltip contents using tooltip-html-unsafe="" works fine, but doesn't allow styling of the tooltip itself.

nonplus:
Closes #3126.


wesleycho:
Landed as d784354a53fa40597cb8c124405ea9a90b6f0b8e, thanks!

PR #3337, opened 2015-02-25
More responsive opening of modals

strille:
Opening a modal now causes one digest loop less, by the following change: The modalBackdrop directive now schedules switching its animate flag using a $timeout **which does not cause a digest loop**.

If we assume a modalBackdrop is always followed by a modalWindow directive in the DOM, then we are guaranteed a digest loop by modalWindow's $timeout (which will happen after modalBackdrop's $timeout since it was added after modalBackdrop's $timeout). At least that is my understanding, and it seems to work in my tests.

I think it's worth trying to reduce the number of excessive digest loops since it's very easy (and common) for the digest loop to take 50-100ms when you have a relatively complex app.


wesleycho:
Can you add a comment about this in the code?  If the $timeout changes in modalWindow, then this information will be lost on whoever is debugging.

wesleycho:
Landed as 7940fd05f01634cd9536379d7dd86a2aae8c2d0a, thanks!

wesleycho:
Scratch my previous comment, this seems to fail the tests - can you rebase off of current master and try to fix the test?

PR #3340, opened 2015-02-26
fix(typeahead): close dropdown on tab with focus-first="false"

dlukez:
When an item isn't selected and you hit "tab", the dropdown stays open even though the focus moves to the second field. An alternative solution might handle this using the 'blur' event. Happy to discuss this.

See this plnkr: http://plnkr.co/edit/zR0HYGVSlMdLgRvCEIZv?p=preview


fernandoneira:
This is good stuff. I'd actually add the same behavior for the Enter key press, since it is basically the same issue in my opinion, i.e. the user has no intent in modifying the input anymore, either because he changed to a different input by pressing Tab or because he wants to submit the form by pressing Enter.

Either you can add this or I'll do a pull request with that change later on if my opinion is not shared.

src/typeahead/typeahead.js
```diff
@@ -275,7 +275,14 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
         }
 
         // if there's nothing selected (i.e. focusFirst) and enter is hit, don't do anything
-        if (scope.activeIdx == -1 && (evt.which === 13 || evt.which === 9)) {
+        if (scope.activeIdx == -1 && evt.which === 13) {
```
wesleycho: Use `===`.

src/typeahead/typeahead.js
```diff
@@ -275,7 +275,14 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+          return;
+        }
+
+        // if there's nothing selected (i.e. focusFirst) and tab is hit, clear the results
+        if (scope.activeIdx == -1 && evt.which === 9) {
```

wesleycho:
This is mostly there, just fix those two things up :) .

dlukez:
@wesleycho I deliberately left the == comparisons as they were. I'm not sure of the ramifications of changing them... Perhaps somewhere they are being set as strings and the coercion is necessary? I don't think they should be changed as a part of this PR, since this PR is concerned with fixing the issue mentioned above rather than refactoring existing code which will carry more risk of introducing a bug.

dlukez:
@fernandoneira I agree with you, however the comment suggested this was by design. If others are in support, I'd be happy to make it work like this for the "enter" keypress too.

wesleycho:
I will be merging this shortly, along with the change to `===` - I did an investigation and it looks like it is only being set to integers, so this should be safe.

PR #3341, opened 2015-02-26
Made accordion heading tab-able for IE9-10

henrihietala:
Having an empty href tag works with modern browsers but not with IE9 or IE10. Adding tabindex="0" makes the accordion heading tab-able without mixing up the tab order.


Foxandxss:
Yo!

This is a small patch, shouldn't be troublesome to merge, not a priority right now (we need 1.3 right now).

@wesleycho Do you have experience with this?

wesleycho:
Landed as 6abad509cd4d44c3ca432f2f21c9ecea0a206b53

Issue #3311, opened 2015-02-18
Something wrong with 'ng-invalid' and 'ng-invalid-parse' classes if 


dimylik:
So, imagine you paste 'Arizona' in to input text, and you have drop-down with one 'Arizona' item coming as preloaded items, if you select that one item with mouse or just click Enter on input, value gets selected, but bootstrap classes 'ng-invalid' and 'ng-invalid-parse' stays on input field, so it has red background.
But if you paste 'Arizon' without last letter and do the same, everything is fine and valid

Looks like onChange() function does now occurs if text and selected item are the same.

My workaround, manual removing 'ng-invalid' and 'ng-invalid-parse' with ng-class:

<input class="form-control search-field"
		type="search"
		ng-model="ngModel"
		placeholder='Search terms ...'
		typeahead="term as (term.name) for term in searchTerms($viewValue) | filter:$viewValue"
		typeahead-min-length="2"
		typeahead-wait-ms="300"
		typeahead-editable="false"
		ng-class="{'ng-invalid' : !ngModel, 'ng-invalid-parse' : !ngModel}"
		autocomplete="off">

Please fix !

karianna:
plunkr please.

juffee:
http://plnkr.co/edit/3nkVcTAT89ik9l6s97G0?p=preview
Here's a plunkr but it seems to happen here only if for example you type 'Ohio', backspace/delete 'o', type 'o' again, then 'tab'. It gets stuck with a 'ng-invalid-parse'. But if you type 'Ohio' in one go, its ok.

angular 1.3.3
angular-ui-bootstrap 0.12

rvanbaalen:
@juffee Can you verify that PR #3352 fixes the problem you described?

juffee:
@rvanbaalen Yes it does fix it.

Issue #3166, opened 2015-01-07
Typeahead bug when selecting a "perfect match" from the dropdown


andertun:
See this [plunker](http://plnkr.co/edit/4H8lZyokFOqmEbkYgqS3?p=preview)

If the input value is exactly the same as one of the items in the dropdown, and you select that item. The input gets an incorrect state (`$invalid = true`).

In our case, this prevents the user to submit the form, and it's hard for them to understand why the field is validated as "has error" even though they did nothing wrong.

A "hack" that can be used until this gets fixed would be much appreciated!

zv3:
Seems that we both ran into the same issue, I fixed this by using the "typeahead-on-select" directive and a controller like below:

```javascript
  .controller('TypeaheadCtrl', ['$scope', function ($scope) {
     $scope.onSelect = function ($item, $model, $label, formInput) {
      formInput.$setValidity('parse', true);
    };
  }])
```
and then in the html view, within the typeahead input element i added the following directives:
```
typeahead-on-select="onSelect($item, $model, $label, yourFormName.yourTypeaheadInputName)" ng-controller="TypeaheadCtrl"
```
I believe this is an angular 1.3.x thing, have not tested it yet with angular 1.2.x

rpdmiranda:
I also ran into this issue, it seems that the parser is not called when the selected value is the same as the one in the input field.

court-jus:
@zv3 solution ignores the parser totally. Is there a way to call the parser manually ?

acgentry:
+1

dlukez:
The parser shouldn't need to be called if the view value doesn't change. On the surface I don't see a problem with setting the 'parse' validation key as valid when an item is selected, considering that the same thing is done with the 'editable' validation key also

PR #3352, opened 2015-03-02
fix(typeahead): run validators manually on select

sagens42:
fix for #3311

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -625,6 +625,18 @@ describe('typeahead tests', function () {
       $(match).click();
       $scope.$digest();
     });
+
+    it('issue #3311 - should trigger validation if filter value equal to list value', function() {
+
+      var element = prepareInputEl('<form name="form"><input name="input" ng-model="result" typeahead="item for item in source | filter:$viewValue"></form>');
+
+      changeInputValueTo(element, 'foo');
+
+      var match = $(findMatches(element)[0]).find('a')[0];
+      $(match).click();
+
+      expect($scope.form.input.$error.parse).toBeUndefined();
+    });
   });
 
   describe('input formatting', function () {
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -250,6 +250,7 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
         model = parserResult.modelMapper(originalScope, locals);
         $setModelValue(originalScope, model);
         modelCtrl.$setValidity('editable', true);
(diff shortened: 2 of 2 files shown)


src/typeahead/typeahead.js
```diff
@@ -251,6 +251,11 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
         $setModelValue(originalScope, model);
         modelCtrl.$setValidity('editable', true);
 
+        if (modelCtrl.$$lastCommittedViewValue === modelCtrl.$viewValue) {
+          // angular will check if model value changed to run validators, so we need to run it manually
+          modelCtrl.$$runValidators(undefined, modelCtrl.$modelValue, modelCtrl.$viewValue, angular.noop);
```
chrisirhc: These are undocumented API. Is there any way to this with the documented API (prefixed with a single `$`)?

sagens42: Actually I think it's enough to run `$setValidity` for `parse` key

src/typeahead/test/typeahead.spec.js
```diff
@@ -625,6 +625,18 @@ describe('typeahead tests', function () {
...
+    it('issue #3311 - should trigger validation if filter value equal to list value', function() {
+
+      var element = prepareInputEl('<form name="form"><input name="input" ng-model="result" typeahead="item for item in source | filter:$viewValue"></form>');
+
+      changeInputValueTo(element, 'foo');
+
+      var match = $(findMatches(element)[0]).find('a')[0];
+      $(match).click();
+
+      expect($scope.form.input.$error.parse).toBeUndefined();
```
wesleycho: You should add a check for the $error before the click as well.

sagens42: `parsed` error was introduced in 1.3, so it's impossible to test it in current environment ( angular 1.2.16 )

rvanbaalen: @sagens42 we've updated the project to 1.3 recently

sagens42:
Fixed in #3166 Closing

katerberg:
It looks like this is still an issue with the new `$validators`. If I have a custom validator added and I type the exact item in a list, it does not fire that validator again after selecting an item. Should I create a new issue for this?

icfantv:
@katerberg - yes, please.  [Here are the instructions](https://github.com/angular-ui/bootstrap/blob/master/CONTRIBUTING.md#you-think-youve-found-a-bug) for doing so, thanks.

katerberg:
@icfantv Created #5136. Let me know if I am missing something since this is my first time contributing a bug to this. Thanks for pointing me in the right direction!

PR #3357, opened 2015-03-03
docs(dropdown): Added examples for issue 2904

aiguofer:
Including example for sub-menu as well as preventing menu close when clicking inside the menu.


wesleycho:
I am not sure this is something that should be in the documentation.  It is an example that does not really demonstrate how to use the directive apart from what is already present in the dropdown documentation.

Thoughts?

chrisirhc:
I agree. The documentation shouldn't contain code that isn't essential for the operation of the directive. Also, I personally wouldn't encourage the usage of `stopPropagation` anywhere unless the developer knows exactly what they are doing.

PR #3358, opened 2015-03-04
fix(datepicker): Parse date from $viewValue instead of $modelValue

decafdennis:
This patch allows the use of `datepicker` in conjunction with an `ngModelController` that has custom formatters and parsers to translate between `$modelValue` and `$viewValue`.

In my particular use case, my model is a custom date value object. I'd like to bind it directly to the `datepicker` and handle the translation between the custom date value format and a `Date` object, with a formatter and a parser that's added to the `ngModelController`. This might be a somewhat obscure use, but it is valid and saves doing manual `$scope.$watch`'s and so on.

Here's a demonstration of my use case: http://jsfiddle.net/dy8vvyey/7/

The `datepicker` directive, being the view, should really use the `$viewValue`, and it should be up to the user of the directive what the `$modelValue` looks like.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -63,8 +63,8 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   };
 
   this.render = function() {
-    if ( ngModelCtrl.$modelValue ) {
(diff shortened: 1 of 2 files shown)


wesleycho:
This makes a lot of sense - good catch!

Landed as 0ecf7faad3b340e171e5c8ca17a17597fa6e6596.

Issue #3354, opened 2015-03-03
Tooltip for disabled input does not disappear


dimylik:
Imagine you have input (checkbox or text, or probably any other disabled input)
you show tooltip on mouse over, fine,
than you have some logic to disable input, also fine,
and you can not close tooltip, as mouseleave action does not work on disabled inputs

http://plnkr.co/edit/sgwAvb6fvchYAjeoNhBK?p=preview

Workaround: 
- add ng-click or mouseleave action on tooltip itself to hide tooltip

icfantv:
Closed per #3362.

PR #3362, opened 2015-03-06
fix(tooltip): Tooltip for disabled input does not disappear #3354

ddomingues:
Fixes the issue #3354


chrisirhc:
Looks good. However, we're trying to avoid adding watchers (when the tooltip isn't displayed) as much as possible as there can be many tooltips on a page.

How about adding this watcher only when the tooltip is shown and removing it when it's hidden?

ddomingues:
For me, looks good @chrisirhc.

chrisirhc:
Nice, looks good. Just needs to be rebased.

ddomingues:
@chrisirhc branch rebased, but when I did it, I observed that was did a commit similar to the issue #3354. Take a look https://github.com/ddomingues/bootstrap/commit/f5503279157893bd31dd52e989c569bb7b6f9e44#diff-e8665eb306ddd26399e5b80f00dae5bbR277.
However, the code doesn't evaluate the expression in a ng-disabled instead, it evaluates an attribute "disabled". I think it forces us to use this attribute in a model. What do you think?

src/tooltip/tooltip.js
```diff
@@ -276,6 +284,14 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
               ttScope.title = val;
             });
 
+            function watchNgDisabled() {
+              unbindWatcher = scope.$watch(attrs.ngDisabled, function (newValue, oldValue) {
```
wesleycho: Is there any reason why you cannot do `attrs.$observe('disabled', ...)` here or something similar?

chrisirhc: I think that would work too. Though I haven't tested what $observe does for boolean attributes like this.

chrisirhc: I stand corrected, `$observe` is a much better idea, more performant (less watchers on scope) and direct to the point.

Thanks for mentioning this @wesleycho , I finally went to dig up how attribute observers work. They're only called on `$set` so it's great. :smile:

ddomingues:
I think so. $observe is a much better solution. So I guest this pull request can be closed @chrisirhc, because it was implemented in an previous commit. Thanks for to mention @wesleycho :wink:

Issue #3372, opened 2015-03-10
tooltip-trigger is not working with expression


brunolellis:
Hi,

I am trying to use tooltip-trigger in an input with an expression to show/hide tooltip, but it is not working as I expect. It looks like expression is not being evaluated.

I've created a plunker to simulate the problem based on ui.tooltip samples: http://plnkr.co/edit/yB7GIKoKnnOWPsKDWrXw

Please, take a look at "Is formexp valid?" section.
Click "Simulate submit" and hover over inputs:
- "tooltip-trigger expression" doesn't show the tooltip message but is contains 'has-error' class;
- "tooltip-enable" works.

Should I use tooltip-enable? It works, but it is not documented, so I am not sure.

Thanks in advance.

karianna:
https://github.com/angular-ui/bootstrap#support

rvanbaalen:
First of all -- If `tooltip-enable` works for you, use that.

Take a look at https://github.com/angular-ui/bootstrap/blob/master/src/tooltip/tooltip.js#L290
You can see that there is no two way binding setup between the directive and the `tooltip-trigger` value. You are programatically changing it's value after the tooltip initialised, which is not supported.

I'm closing this issue now.

brunolellis:
@karianna after some more investigation, I realized that it works until version ui-bootstrap-tpls-0.11.0.js (including this one). Versions 0.12.0 and 0.12.1 are not working.

It may help further investigations.

rvanbaalen:
Good find.  I see that back then, the attributes where all being watched using `attrs.$observe`.
It seems that this was removed in favour of performance.

rvanbaalen:
To be more specific, @chrisirhc removed the observers in this commit:

https://github.com/angular-ui/bootstrap/commit/a65bea95338802b026fd213805b095b5a0b5b393

brunolellis:
@rvanbaalen ok, thanks!
I will use tooltip-enable :)

Are there any guidelines to improve docs?

rvanbaalen:
Just follow the general contribution guidelines.

cm325:
It appears that when the value changes from true to false in tooltip-enable, currently shown tooltip isn't hidden. Is this an oversight?

karianna:
@cm325 Please open a new issue with a plunkr

PR #3373, opened 2015-03-10
docs(tooltip-enable): Added description and sample

brunolellis:
Some explanation about https://github.com/angular-ui/bootstrap/issues/3372


Psykar:
If you have no triggers set, this will actually enable/disable the default trigger of mouseover.
That is, if disabled, mouseover will not work.
If enabled, mouseover will work.

It *won't* just show the tooltip immediately.

brunolellis:
@Psykar hm... I see.

`tooltip-enable`: Is it enabled? It will enable or disable the configured tooltip-trigger.

Is it better?

Psykar:
Sounds better to me :)

chrisirhc:
Made a few adjustments, thanks!

chrisirhc:
Merged via 9a26b368038967ee049fc60e8490a2beebe4648a

PR #3377, opened 2015-03-11
Bug fix for European date format

dbillingham:
Fixed bug whereby the incorrect date was being returned when the datepicker-popup attribute was set to 'dd.MM.yyyy', and an invalid date string was entered.


RobJacobs:
@dbillingham Thanks for the pull request but we have several other pull requests dealing with the same issue: #2395, #2691, #3152, #3294.  There hasn't been a decision made which approach is the best fix as it's tangled with some other fixes.  Closing this for now.

PR #1915, opened 2014-03-10
refactor(datepicker): use ng-if to increase performance

Foxandxss:
It is going to be fun, someone said. Damn, I never felt this much pain on a refactor.

Adding an extra scope in the middle of this directive was way problematic.

Adding that ng-if needed to modify all the popup tests to manually open the datepicker (earlier, you could run your tests vs a hidden datepicker but not if it is not there).

On the other hand, I needed to use the `dot.rule` for the date so it would work as expected.

On the `$destroy` `$popup` is the comment of the ng-if, so you have to destroy the sibling element.

What I am not happy about is the close button. Since there is a new parent scope, you need an extra `$parent`. Can I use there the `dot.rule`? Sure but that would need an extra $watch.

Let me explain: If I change everything like: s/scope.isOpen/scope.popup.isOpen/ it would work, but I still need to `$watch` if the `isOpen` changed outside.

Said this. Feel free to review it and suggest changes.

PR #3382, opened 2015-03-12
fix(datepicker): child scope problem with isOpen

### LINKED ISSUE #3180: The datepicker doesn't work correctly inside a ng-if directive
The datepicker shows only one time, and after that it doesn't appear never more, when I change the ng-if by a ng-show this work correctly

newmanw:
Changed isOpen to and object, now called popupState.  Objects work in 2-way data binding when in a child scope.
Should fix issues with datapicker multiple open/close in ng-if, ng-repeat, etc.

fixes #3180

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -436,7 +436,7 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
     restrict: 'EA',
     require: 'ngModel',
     scope: {
-      isOpen: '=?',
+      popupState: '=?',
       currentText: '@',
       clearText: '@',
       closeText: '@',
@@ -448,6 +448,12 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
           appendToBody = angular.isDefined(attrs.datepickerAppendToBody) ? scope.$parent.$eval(attrs.datepickerAppendToBody) : datepickerPopupConfig.appendToBody;
 
       scope.showButtonBar = angular.isDefined(attrs.showButtonBar) ? scope.$parent.$eval(attrs.showButtonBar) : datepickerPopupConfig.showButtonBar;
+
+      if (!scope.popupState) {
+        scope.popupState = {
+          open: false
+        };
+      }
 
       scope.getText = function( key ) {
         return scope[key + 'Text'] || datepickerPopupConfig[key + 'Text'];
@@ -537,7 +543,7 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
         ngModel.$render();
 
         if ( closeOnDateSelection ) {
-          scope.isOpen = false;
+          scope.popupState.open = false;
           element[0].focus();
         }
       };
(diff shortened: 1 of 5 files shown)


RobJacobs:
This change is included in pull request #1915. Closing unless you feel otherwise.

newmanw:
@RobJacobs is that really the same problem?  I am not so sure.  isOpen is still a primitive, I think you will still have issues with child scopes, ie placing date picker inside ng-if or ng-repeat.

newmanw:
@RobJacobs please re-open.  I pulled #1915 and built it.  Threw the contents into a plunk here:

http://plnkr.co/edit/remf5q4ltkQzM55553tB?p=preview

Same problem, open datepicker, close, and you cannot re-open.

#1915 addressed a performance issue.  There is still an issue with child scopes with the native isOpen variable.

RobJacobs:
@newmanw I apologize for prematurely closing this pull request, I didn't review it closely enough.  The work done in #1915 deals with moving the datepicker date to an object property, so it has some relation what you are doing with the open property.  If we are going to have an object to hold primitives, I'd like to see us use one (popupState.open vs popup.date).

I implemented what your pull request suggests in your date.directive (don't bind to primitives) and it seems to work now (also changed to the 0.12.1 release), see this [plunk](http://plnkr.co/edit/iwocfLRkaTS8SxTNoPNc?p=preview)

rvanbaalen:
Why don't I see the actual patched version being used in the Plunkr? I'm looking at a custom datepicker directive?

From what I can tell in the implementation of this PR, this needs a breaking change.

newmanw:
@rvanbaalen it is definitely not backwards compatible if that is what you mean by breaking change.

IMHO this is a pretty big problem, due to the fact that as it exists you cannot use the datepicker directive inside a child scope (ng-if, ng-repeat, etc...).  This is because the 2-way data bound variable for isOpen is a primitive.  I think the right way to do this is use an object, hence breaking backwards compatibility.

Plunk showing problem in simple ng-if:
http://plnkr.co/edit/QAp0ODDgOhbZrwcuRItB?p=preview

Plunk with my PR version showing working multiple open/close of dp:
http://plnkr.co/edit/OJ2xGxcIhZx27rb1RK5f?p=preview

rvanbaalen:
That is indeed what breaking change means. Thanks for the plunks. Will look into it asap

RobJacobs:
@newmanw @rvanbaalen As I demonstrated in my plunker above, this is a non-issue if you don't bind to primitives in your controller. See this [plunker](http://plnkr.co/edit/bbEI4Lp5P056iky1kPyC?p=preview) for another demo. This scenario is also mentioned in the project [wiki](https://github.com/angular-ui/bootstrap/wiki/FAQ) in the 'My input / select element doesn't work as expected inside the tab / accordion directive' section where it states: 'As a quick rule of thumb you should always have a dot as part of your  ngModel  expression.'

rvanbaalen:
@RobJacobs please contact me to join the (active) team members in determining the upcoming roadmap. 

So basically this pr can be closed since its not an issue as long as you dont bind to primitives?

newmanw:
@RobJacobs @rvanbaalen  $%&*, sorry about ;)  Missed your solution.  It did seem a little weird that you bound the scope to an entire new object, but as you mention as long as I use and object, its the same thing.

IE

```javascript
  $scope.popupState = {
    opened: false
  }
```

```html
          <input type="text" class="form-control" datepicker-popup="MM/dd/yyyy 00:00:00" ng-model="dt" is-open="popupState.opened" close-text="Close" />
```

Truly apologize again that I missed that.  Also really appreciate you guys looking into this in detailed manner.  Love this lib and seems that the team is just as good ;)

PR #3390, opened 2015-03-15
feat(modal): Add new windowScope option

### LINKED ISSUE #3384: windowTemplateUrl how to acess $scope
I am trying to use windowTemplateUrl with a custom template to reference a controller so I can toggle a class on the top level div such that I can use ng-class with the template. I tried various ways to try to get the template to use a controller without success. 

template:
```HTML
<div tabindex="-1" role="dialog" class="modal fade" ng-class="{in: animate}" ng-style="{'z-index': 1050 + index*10, display: 'block'}" ng-click="close($event)">
    Hello: {{test}}
    <div class="modal-dialog" ng-class="{'modal-sm': size == 'sm', 'modal-lg': size == 'lg'}"><div class="modal-content" modal-transclude></div></div>
</div>
```

JS
```javascript
myModal = $modal.open({
      templateUrl: '/static/partials/myModal.html',
      controller: 'MyModalCtrl',
      backdrop: 'static',
      keyboard: 'false',
      windowTemplateUrl: '/static/partials/template_myModal.html'
    });

********

.controller('MyModalCtrl', ['$scope', '$modalInstance', function MyModalCtrl($scope, $modalInstance) {
  $scope.test = ", world!";
```

I would expect my modal to show: "Hello, world!" , but it isn't and i cannot find any documents as how to do so.

wesleycho:
- Expose new `windowScope` option for people using a custom template for the modal window

Fixes #3384


src/modal/modal.js
```diff
@@ -80,7 +80,8 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
       restrict: 'EA',
       scope: {
         index: '@',
-        animate: '='
+        animate: '=',
+        windowScope: '=?'
```
chrisirhc: This can probably be just `=` since we don't have a use of modal-window that doesn't declare the `window-scope` attribute. Even if windowScope is null, it'll still be bound fine.

chrisirhc:
I'm a little afraid this encourages pollution of the $rootScope. How about a `windowInit` option that's an Angular expression that's run in the window's scope?
That way, a user can do `windowInit: 'myTemplateVar = "x"; myFn = $root.x'`.

wesleycho:
It looks like it shouldn't pollute $rootScope, as far as my reading of https://github.com/wesleycho/bootstrap/blob/feat/window-scope/src/modal/modal.js#L279-L289 and https://github.com/wesleycho/bootstrap/blob/feat/window-scope/src/modal/modal.js#L397 goes - is this incorrect?

chrisirhc:
It doesn't pollute the rootScope itself, but in order to use this feature, users would need to place variables in the $rootScope, as the `=` binding is bound to the rootScope. As such, it encourages the user to pollute the rootScope. Sorry, wasn't clear before.

jareddlc:
i did a build of this branch, and I cannot get modals to display additionally i get the error: 

    Syntax Error: Token 'Object' is unexpected, expecting []] at column 9 of the expression [[object Object]]     starting at [Object]]

wesleycho:
@jareddlc try it now - made a silly mistake with the first version.

@chrisirhc users would not have to place a variable in $rootScope to use this - they would just need to set the `windowScope` key to the object of their choice.  This would not pollute $rootScope at all, as the $scope used has $new() applied to it to avoid pollution (seen [here](https://github.com/wesleycho/bootstrap/blob/feat/window-scope/src/modal/modal.js#L399)).  I edited my branch to clearly put my intention (which was not there beforehand) - check out the updated diff to see what I had originally intended.

jareddlc:
@wesleycho I cannot seem to get it working, maybe I am using it incorrectly.

```
//template
<div tabindex="-1" role="dialog" class="modal fade" ng-class="{in: animate, myClass: myState}" ng-style="{'z-index': 1050 + index*10, display: 'block'}" ng-click="close($event)">
    test: {{test}} [[test]]
    <div class="modal-dialog" ng-class="{'modal-sm': size == 'sm', 'modal-lg': size == 'lg'}"><div class="modal-content" modal-transclude></div></div>
</div>
```

```
//modal
var myScope = {test: 'hello, world!'};

myModal = $modal.open({
  templateUrl: '/static/partials/modal_myModal.html',
  controller: 'myModalCtrl',
  backdrop: 'static',
  keyboard: 'false',
  windowTemplateUrl: '/static/partials/template_myModal.html',
  windowScope: myScope,
  scope: $scope
});
```

```
.controller(myModalCtrl, ... {
 //myState changes here
});
```

inside the controller I have state changes that will cause the template to change its class.

antoinepairet:
@jareddlc Please provide a plunkr and not just copied pasted code. thx

wesleycho:
After discussing with @chrisirhc on Slack last week, the ticket and PR may be invalid.  Waiting for response in the issue thread.

marshall007:
@wesleycho since the primary use-case for this and related issues appears to be dynamically adding classes to the modal window, could we just have a `$toggleClass(...)` function injected into the modal content scope just like `$close()` and `$dismiss()`?

readme42:
:+1: 

Same issue. Would like to define different modal wrapper by using windowTemplateUrl like this:

```html
<div class="modal-content">
    <div class="modal-header">
        <h3 class="modal-title">I'm a modal!</h3>
    </div>
    
    <div class="modal-body" modal-transclude></div>

    <div class="modal-footer">
        <button class="btn btn-primary" ng-click="$close(vm.selected)">OK</button>
        <button class="btn btn-warning" ng-click="$dismiss('dont need it')">Cancel</button>
    </div>
</div>
```

and then using different templateUrls for different content.

wesleycho:
@marshall007 couldn't one just override the template to add dynamic classes?

marshall007:
@wesleycho yeah and that's how I've been working around it. Just seems like overriding the template is a lot of work for a fairly common use-case.

wesleycho:
I'm not sure it is a common enough use case - a $toggleClass method would likely not do what is necessary, since it would need a hook up the tree, which would require a binding & thus would require 1+ $watch. That would not be ideal.

wesleycho:
Closing due to nebulous knowledge of situation.

Issue #3375, opened 2015-03-11
ngAnimate breaks fade-in animation of popover


zdila:
After including ngAnimate module to my application it breaks fade-in animation of popover.

Demo: http://plnkr.co/edit/3ECNoqBqfUcbwTq3MnVa

karianna:
versions of angular ui / bootstrap?

zdila:
From linked plunk:
angularjs/1.3.9/angular.js
angularjs/1.3.9/angular-animate.js
ui-bootstrap-tpls-0.12.1.js
bootstrap/3.3.2/css/bootstrap.min.css

wesleycho:
To clarify, it looks like it is breaking when the popover is entering - is this accurate?

zdila:
@wesleycho, exactly.

wesleycho:
Can you try based on current master?  The popover now supports ngAnimate

wesleycho:
Closing, as the popover was refactored to support ngAnimate.

Let us know if this is still an issue.

zdila:
Unfortunately it is still broken. Using `ui-bootstrap-tpls-0.13.0-SNAPSHOT.min.js` from today's clone. You can check it out at http://plnkr.co/edit/94yNAYcGjWcndpIH9AbL (`ui-bootstrap-tpls-0.13.0-SNAPSHOT.js` is there loaded from google drive).

wesleycho:
Posting more useful [Plunker](http://plnkr.co/edit/4IFABsZFdDy4zDSdwnW7?p=preview) for reference

chrisirhc:
Reopening as this isn't fixed on master at the moment. Ref: #3506

PR #3391, opened 2015-03-16
fix(tooltip): Fix animation when ngAnimate is present

wesleycho:
- This disables ngAnimate's animation on the tooltip element due to $animate interfering with Bootstrap's transitions

Plunker of fix in action [here](http://plnkr.co/edit/6Hqw29jlscL9k9HH6CSW?p=preview)

This addresses #3375.


chrisirhc:
Note that we need to upgrade the rest of angular test libs to the same version otherwise we might encounter weirdness in the tests.

Also, instead of disabling the animation on the tooltip completely, we can use the .css approach in https://github.com/angular-ui/bootstrap/pull/2699 , with carousel.css setting up transition values of 0. That's another approach that allows users use other animations if they want.

chrisirhc:
Also, we're using lowercase for module names: fix(tooltip) . That's just for grunt-conventional-changelog to pick things up correctly.

wesleycho:
Ahh, I was not aware of that work - that would be a lot nicer.  I had meant for this to be a stop gap solution that at least fixes the breaking problem until we got to making it truly ngAnimate compatible.

I'll review #2699 when I get home.

wesleycho:
Closing this, will revisit with a more ngAnimate compatible approach.

PR #3395, opened 2015-03-17
fix(timepicker): Make sure ng-model is pristine on initialization

### LINKED ISSUE #3160: Timepicker triggers dirty flag inside ngRepeat
Using just Angular 1.3.3 and Angular-UI 0.12.0, when the following code loads:

JS:
```js
$scope.groups = [];
$scope.groups.push(
  {when: new Date(2014,9,1,10,15)}
);
```
HTML:
```html
<div>
  <div ng-repeat="group in groups">
    <timepicker ng-model="group.when"></timepicker>
  </div>
</div>
```

The timepicker element is immediately marked dirty/triggers the change event. This is problematic when attempting to check for changes on page unload as there will always be detected changes. See here:
http://plnkr.co/Ic2HA390OI2xheT38Izz

This does not happen outside ngRepeat, or with a simple text input inside the ngRepeat. See here:
http://plnkr.co/s0KwGrZkHaog4efyjGBK

I have managed to work around the issue by modifying the directive, but I'm not sure that my change is safe. The modification is to the refresh() function:

```js
  function refresh( keyboardChange ) {
    makeValid();
    if (ngModelCtrl.$viewValue.getTime() != new Date(selected).getTime()) {
      ngModelCtrl.$setViewValue(new Date(selected));
    }
    updateTemplate( keyboardChange );
  }
```

wesleycho:
- On initialization, it is possible for the timepicker ng-model to become dirty due to the update to the hours and minutes values when the model value is parsed. This change guarantees the model is pristine on first render.

Working Plunker [here](http://plnkr.co/edit/OvS27MeT4o5k9sO2aHLW?p=preview).

Fixes #3160.


chrisirhc:
Actually, this isn't quite right either. `updateHours` and `updateMinutes` shouldn't be called when updating the ngModel, as they're in the ng-change expression. The ng-change expression should only fire when the user interacts with the input.

This is actually happening because of the change in `type="text"` handling in AngularJS 1.3 . We use `type="text"` fields with an model with number values.

We can either:
1. Change the hour and minute fields from `type="text"` to `type="number"`. (but note that Chrome / Webkit might add ticker buttons on these fields, which imho is fine)
2. Store the hour and minute as strings. Probably a bad idea, but putting it out there anyway.

wesleycho:
I didn't really like this either - what is happening is `updateHours` and `updateMinutes` are called in order to update validity, which is fine.  The problem seems to be with the refresh function, where it is calling `ngModel.$setViewValue` - perhaps a more correct fix is to move that out of the refresh function then and check for all functions calling it?

chrisirhc:
Hm I think you're misunderstanding me, these functions shouldn't be called in these cases.
There's a check where the model value != view value because entering hours = 10 gets checked with 10 === '10' which is false. This is because the numbers are being converted into strings.

In angular 1.3, the ngModel forces the format of text input fields to be string while before, it could be a number.

wesleycho:
Oh, didn't realize that was happening - would adding a formatter that converts it into a string help any here?

chrisirhc:
Yes, we could also change the `timepicker.html` template so that the input is a `type="number"` as I mentioned above.

wesleycho:
I took a look at turning it into a number, but it turns out all of the timepicker unit tests assume support for string/date values.

The formatter approach might be the correct approach here - it turns out there is some logic in the render method that should actually be in the formatter.

I will write a unit test for this and open a new PR.

wesleycho:
Superceded by #3427

PR #3400, opened 2015-03-17
fix(modal): Modal backdrop does not fade out on close #3182

### LINKED ISSUE #3182: Modal backdrop does not fade out on close
See http://getbootstrap.com/javascript/#modals for expected close animation. Whereas with http://angular-ui.github.io/bootstrap/#/modal, the modal itself animates on close, but the backdrop simply disappears suddenly.

AlicanC:
We start removing the backdrop after the modal gets removed which results in backdrop being removed instantly after modal does its closing animation.

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -186,6 +186,9 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
         //clean up the stack
         openedWindows.remove(modalInstance);
 
+        //start removing the backdrop with the modal
+        checkRemoveBackdrop();
+
         //remove window DOM element
         removeAfterAnimate(modalWindow.modalDomEl, modalWindow.modalScope, 300, function() {
           modalWindow.modalScope.$destroy();


AlicanC:
@chrisirhc @pkozlowski-opensource this is your code and I'm not sure if I'm fixing it right.

karianna:
@AlicanC Build failed and this needs tests

AlicanC:
@karianna My PR is more of a proof-of-what's-wrong than a complete fix. I don't have time to figure out the whole logic behind the modal code, but I think it's now a few minutes of work for someone who knows what's going on in there.

If no one comes up with a proper fix, I will dig in later when I have more time.

wesleycho:
Can you provide a [Plunker](http://plnkr.co/) reproduction of the issue?

chrisirhc:
I've just merged ngAnimate fixes for the modal, so this may not longer be invalid on master.

rvanbaalen:
@AlicanC Any update on this?

AlicanC:
The current project I'm working on does not involve UI Bootstrap so it is very hard for me to find the time to complete this fix.

rvanbaalen:
At least a plunkr reproduction of your issue perhaps?

AlicanC:
@rvanbaalen This was an attempt to fix #3182 and related information is there.

I think I've misread @chrisirhc's post. I thought he changed something that would "make progress" on this issue, but it looks like he fixed the issue. I will post on #3182 if updating UI Bootstrap doesn't solve the issue on our end.

rvanbaalen:
Thanks for the info @AlicanC.

fernandojunior:
I still have this error. [Here](http://plnkr.co/edit/xQOL58HDXTuvSDsHRbra) a plunker to simule the erros. The plunker is based on that modal example in getting started page of UI Bootstrap.
Changes I did:
* angular version 1.3.13 to  1.4.1;
* import angular animate  1.4.1;
* set app to use ngAnimate dependence.

theider:
I started having this problem after upgrading AngularJS in my application and starting to use "smart table" (not sure which came first but these two things changed).

I was able to fix it (i.e. get the modal backdrop to go away) by doing this:

In angular-bootstrap 0.13.0 ui-bootstrap-tpls.js line 2165 in removeAfterAnimate():

      function removeAfterAnimate(domEl, scope, done) {
        // Closing animation
        scope.animate = false;

        if (domEl.attr('modal-animation') && $animate.enabled()) {
          // transition out
          $rootScope.$evalAsync(afterAnimating); // <-- calling this directly makes the backdrop go away.
// this domEl.one does not resolve properly and so the critical afterAnimating method does not get executed.
//          domEl.one('$animate:leave', function closeFn() {
//            $rootScope.$evalAsync(afterAnimating);
//          });
        } else {
          // Ensure this call is async
          $timeout(afterAnimating);
        }

        function afterAnimating() {
          if (afterAnimating.done) {
...

How can we turn this into a fix?  I'm not sure of the impact of not using the .one approach through $animate.close.  BTW $animate.close does not seem to be a method; I see enter and leave but I don't fully understand $animate yet.

PR #3401, opened 2015-03-18
fix(modal): backspace with modal dialog should be disabled #3208

### LINKED ISSUE #3208: backspace with modal dialog should be disabled
When modal dialogue is displayed, backspace should be disabled.

Now, the user can still press "backspace" and the page **behind** the modal moves, but the modal itself stays.

antoinepairet:
Fixes #3208 

When the option "keyboard" is set to true (default), both ESC and BACKSPACE close the modal window.

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -243,11 +243,18 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.transition'])
         var modal;
 
         if (evt.which === 27) {
+          closeOnKeydown('escape key press');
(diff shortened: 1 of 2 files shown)


wesleycho:
We should make sure we want this feature first.  Merging of this is conditional on what comes out of that thread, and this does not appear to be breaking behavior.

wesleycho:
Closing as per https://github.com/angular-ui/bootstrap/issues/3208#issuecomment-126495125

PR #3406, opened 2015-03-19
fix(typeahead): reset 'parse' validation key

### LINKED ISSUE #3166: Typeahead bug when selecting a "perfect match" from the dropdown

dlukez:
When typeahead-editable="false" and we select a perfect match from the results list, the $parsers aren't called (due to no change in the view value), which results in the 'parse' error key not being reset.

As with the 'editable' key, we should reset this once an item is selected because we know we have a valid model matching the view value.

Closes #3166


wesleycho:
Landed as c0a9c707903fa3dfc662887cd348ee2b0bf13f85, thanks!

PR #3409, opened 2015-03-19
feat(popover): support template url for partial

jbruni:
Updated old solution with improvements:
- Very simple but effective implementation.
- Watches `popover-template` attribute. The template can change at any time.
- Bindings and positioning working fine.
- Zero effects on current functionality.
- (In fact, zero modifications in all about popovers, excepts it injects the compiled template into it, when the `popover-template` is used.)

Code in action at http://codepen.io/anon/pen/XJodJr

How to use:
- simply specify the HTML template in the `popover-template` attribute


karianna:
@jbruni Are you able to add some tests around this please?

src/popover/popover.js
```diff
@@ -16,4 +16,25 @@ angular.module( 'ui.bootstrap.popover', [ 'ui.bootstrap.tooltip' ] )
 
 .directive( 'popover', [ '$tooltip', function ( $tooltip ) {
   return $tooltip( 'popover', 'popover', 'click' );
+}])
+
+.directive( 'popoverTemplatePopup', [ function () {
+  return {
+    restrict: 'EA',
+    replace: true,
```
bcherny: replace is deprecated - https://github.com/angular/angular.js/blob/master/src/ng/compile.js#L270

src/popover/popover.js
```diff
@@ -16,4 +16,25 @@ angular.module( 'ui.bootstrap.popover', [ 'ui.bootstrap.tooltip' ] )
...
+    scope: { title: '@', template: '=', placement: '@', animation: '&', isOpen: '&' },
```
bcherny: why not use 1-way binding instead of text binding so `title` and `placement` can also be updated?

chrisirhc:
Happy for people to make a comparison between this and #1848 .

src/popover/popover.js
jbruni: I'm faithfully following what already is in place:
https://github.com/angular-ui/bootstrap/blob/master/src/popover/popover.js#L11

src/popover/popover.js
jbruni: Both this and the other comment truly relates to the current `popover.js` implementation, not exactly to this small effort to make `popover-template` available...

Here, again, I'm just making it behave exactly the same as the normal popover:
https://github.com/angular-ui/bootstrap/blob/master/src/popover/popover.js#L12

jbruni:
@karianna - I'm trying to learn how to make tests in my little free time... already made my first steps... hopefully I can have something as soon as possible.

karianna:
NP:  We're happy to wait - there's lots of PRs we're still looking at.

antoinepairet:
https://github.com/angular-ui/bootstrap/pull/1848 was merged.
@jbruni thanks for your efforts while working on this.
regards,

Closing the issue.

PR #3410, opened 2015-03-19
Added missing ng-change

hitautodestruct:
This documentation was missing the `ng-change` event.

### DIFF
--- a/src/pagination/docs/readme.md
+++ b/src/pagination/docs/readme.md
@@ -5,6 +5,10 @@ A lightweight pagination directive that is focused on ... providing pagination &
 
 Settings can be provided as attributes in the `<pagination>` or globally configured through the `paginationConfig`.
 
+ * `ng-change`
+ 	:
+ 	Call a function on any pagination change.
+
  * `ng-model` <i class="glyphicon glyphicon-eye-open"></i>
  	:
  	Current page number. First page is 1.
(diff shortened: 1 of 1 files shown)


wesleycho:
I'm not seeing anything where ng-change is being used in the pagination directive.  I do not believe this is valid.

chrisirhc:
Hm yea, I think `ng-change` can be used as a typical use case because `ng-model` supports `ng-change` on the same element.

Perhaps instead of mentioning it as a setting, we can mention that ng-change *can be used* on the pagination element (along with any ngModel formatter directive)?

hitautodestruct:
@wesleycho @chrisirhc There seems to be no reference to access any way to get an event. I'm using the ng-change and can provide a demo if you want to see it working. The `ng-change` comes coupled with `ng-model`.

wesleycho:
That sounds like a different issue, one worth filing an issue about.

hitautodestruct:
It seems like there is a `selectPage(page.number)` method that is set as ng-click on every element. But again this is not documented.

wesleycho:
Those sound like different issues - in that case, you should file an issue about them. @chrisirhc asked to make a modification to the wording for this PR, get that fixed up and we can merge this :) .

hitautodestruct:
Done :)

wesleycho:
Landed as df0bc00522c264963db9e086f758d42247d35070, thanks!

PR #3411, opened 2015-03-19
Feature/dropdown add to body

### LINKED ISSUE #1030: Add add-to-body option for dropdown
Just as tooltip (and now datepicker via #998) support an add-to-body option, it would be handy to have it for dropdown-toggle.

rysilva:
Fixes #1030 

Look for dropdown-append-to-body attribute on the dropdown directive.  If it's there, append the dropdown-menu to the body and show/hide/position it when the dropdown toggles.

### DIFF
--- a/src/dropdown/docs/demo.html
+++ b/src/dropdown/docs/demo.html
@@ -42,6 +42,20 @@
       </ul>
     </div>
 
+    <!-- Single button using append-to-body -->
+    <div class="btn-group" dropdown dropdown-append-to-body>
+      <button type="button" class="btn btn-primary dropdown-toggle" dropdown-toggle>
+        Dropdown on Body <span class="caret"></span>
+      </button>
+      <ul class="dropdown-menu" role="menu">
+        <li><a href="#">Action</a></li>
+        <li><a href="#">Another action</a></li>
+        <li><a href="#">Something else here</a></li>
+        <li class="divider"></li>
+        <li><a href="#">Separated link</a></li>
+      </ul>
+    </div>
+
     <hr />
     <p>
         <button type="button" class="btn btn-default btn-sm" ng-click="toggleDropdown($event)">Toggle button dropdown</button>
--- a/src/dropdown/docs/readme.md
+++ b/src/dropdown/docs/readme.md
@@ -2,8 +2,13 @@
 Dropdown is a simple directive which will toggle a dropdown menu on click or programmatically.
 You can either use `is-open` to toggle or add inside a `<a dropdown-toggle>` element to toggle it when is clicked.
 There is also the `on-toggle(open)` optional expression fired when dropdown changes state.
+
+Add `dropdown-append-to-body` to the `dropdown` element to append to the inner `dropdown-menu` to the body.
+This is useful when the dropdown button is inside a div with `overflow: hidden`, and the menu would otherwise be hidden.
+
 By default the dropdown will automatically close if any of its elements is clicked, you can change this behavior by setting the `auto-close` option as follows:
 
   * `always` - (Default) automatically closes the dropdown when any of its elements is clicked.
   * `outsideClick` - closes the dropdown automatically only when the user clicks any element outside the dropdown.
   * `disabled` - disables the auto close. You can then control the open/close status of the dropdown manually, by using `is-open`. Please notice that the dropdown will still close if the toggle is clicked, the `esc` key is pressed or another dropdown is open.
+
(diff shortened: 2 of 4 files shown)


rysilva:
I'm attempting to remove the dropdown menu from the body when the main dropdown element fires the `$destroy` event.  However, on the demo page the $destroy event never gets fired.  I've been able to see the event get fired in my own application using this code, though.

In the tests, I do in fact see that the `$destroy` event gets fired if I call `element.remove()` and the code successfully calls `dropdownMenu.remove()`. However, in the test browser window the element never gets removed from the page.  The remove call seems to do nothing.

Any ideas why?

jessicawangdt:
+1!!!

src/dropdown/docs/demo.html
```diff
@@ -42,6 +42,20 @@
       </ul>
     </div>
 
+    <!-- Single button using append-to-body -->
+    <div class="btn-group" dropdown dropdown-append-to-body>
```
wesleycho: Should this be `dropdown-append-to-body="true"`?

wesleycho:
This needs a test to fix the garbage collection.

rysilva:
@wesleycho do you mean it needs a fix to test the garbage collection?  

I had a test that looked like this:

    it('removes the menu when the dropdown is removed', function() {
      element.remove();
      $rootScope.$digest();
      //Here, I am able to verify that line 80 (self.dropdownMenu.remove();) is in fact called but it has no effect in the test environment.
    });

What I was referring to in my comment was that I was unable to verify the outcome of `self.dropdownMenu.remove()` because it remained in the DOM in the test runner's browser, even though I could verify that `self.dropdownMenu` was defined and referred to the correct element.  I'll give it another shot, but I was hoping there might be some testing quirk that I didn't know about.

wesleycho:
@rysilva if you're able to do so, I'm willing to review this again and merge when ready.

Perhaps `element.remove()` did not properly garbage collection the `$scope` and emit the `$destroy` event?

jessicawangdt:
@rysilva, @wesleycho : looking forward to this fix, thanks to both of you.

andyperlitch:
@rysilva and @wesleycho, I believe I understand what is happening: the element is not being cleaned up after each test. In other words, `element.remove()` needs to be called at the end of his test. This is what triggers the removal of the dropdown from the body. I would actually recommend calling `element.remove()` in an `afterEach` block at the top of this whole spec. A lot of tests are doing it in a one-off way anyway, and I see no reason to keep elements around after each test.

I have [issued a pull request](https://github.com/rysilva/bootstrap/pull/1) to @rysilva that not only fixes all this, but also resolves the conflicts with the current master preventing automatic merge. @wesleycho, assuming @rysilva merges my PR to his branch, please review again.

rysilva:
Thanks, I'll take a look

On Apr 1, 2015, at 3:22 AM, Andy Perlitch <notifications@github.com> wrote:

@rysilva <https://github.com/rysilva> and @wesleycho
<https://github.com/wesleycho>, I believe I understand what is happening:
the element is not being cleaned up after each test. In other words,
element.remove() needs to be called at the end of his test. This is what
triggers the removal of the dropdown from the body. I would actually
recommend calling element.remove() in an afterEach block at the top of this
whole spec. A lot of tests are doing it in a one-off way anyway, and I see
no reason to keep elements around after each test.

I have issued a pull request <https://github.com/rysilva/bootstrap/pull/1>
to @rysilva <https://github.com/rysilva> that not only fixes all this, but
also resolves the conflicts with the current master preventing automatic
merge. @wesleycho <https://github.com/wesleycho>, assuming @rysilva
<https://github.com/rysilva> merges my PR to his branch, please review
again.

—
Reply to this email directly or view it on GitHub
<https://github.com/angular-ui/bootstrap/pull/3411#issuecomment-88376956>.

rysilva:
I'm not sure why adding an afterEach would solve the issue, because you're already calling `element.remove()` in the test anyway.  So for that test the afterEach should have no effect, right?

I was going to run locally again but I'm currently getting an error running grunt:

    Running "karma:continuous" (karma) task
    [2015-04-01 08:07:20.330] [ERROR] config - Invalid config file!
     [ReferenceError: module is not defined]

wesleycho:
You may need to run `npm update`.

Also the history needs to be cleaned up - the merge commit should not be there, and all of the commits should be squashed into one.

rysilva:
Squashed all the commits.

samb90:
Cannot wait for this feature to be merged! It fixes all my problems.

wesleycho:
Merged - thanks for the good work @rysilva and @andyperlitch !

andyperlitch:
all @rysilva... thanks @wesleycho!

jessicawangdt:
@andyperlitch , @rysilva , @wesleycho : thanks a lot!!!!!!!!!!!!!!!!!!!!!

Issue #3413, opened 2015-03-20
Rounding issue with star-rating


bodiam:
When using the star rating, the rating doesn't handle fractions correctly, causing ratings like 7.1 to be displayed as 8 stars, instead of applying a rounding. 

I'm not sure how to fix it, but the problem is caused by the rating.html, in this code:

    $index < value

When applying rounding here, this would work, but I'm not sure how I can inject Math into this template. If someone can explain me how, I don't mind creating a pull request with appropriate tests.

PR #3415, opened 2015-03-21
feat(rating): Added rounding logic to rating value

moderndegree:
I added logic that rounds off decimals to display a more accurate number of icons (i.e. stars).
This should close out issue #3413.


src/rating/rating.js
```diff
@@ -58,6 +58,12 @@ angular.module('ui.bootstrap.rating', [])
   this.render = function() {
     $scope.value = ngModelCtrl.$viewValue;
   };
+
+  $scope.$watch('value', function(value) {
```
wesleycho: This seems to be an unnecessary & unperformant approach.

A better approach would be to create a formatter, that way it only gets run when the model value changes.

moderndegree: I'll take a look

src/rating/rating.js
```diff
@@ -12,6 +12,13 @@ angular.module('ui.bootstrap.rating', [])
   this.init = function(ngModelCtrl_) {
     ngModelCtrl = ngModelCtrl_;
     ngModelCtrl.$render = this.render;
+    
+    ngModelCtrl.$formatters.push(function(value) {
+      if (typeof value !== 'undefined' && value % 1 !== 0) {
```
wesleycho: Is there any reason why we can't just do
<pre>
if (angular.isNumber(value) && value === value) {
</pre>
here?  It seems you are trying to test for whether it is a number, this would be a better check.

moderndegree:
Apparently, `angular.isNumber` uses `typeof` so I am using it. Guess I shouldn't make such assumptions in the future.

src/rating/rating.js
```diff
@@ -12,6 +12,13 @@ angular.module('ui.bootstrap.rating', [])
   this.init = function(ngModelCtrl_) {
     ngModelCtrl = ngModelCtrl_;
     ngModelCtrl.$render = this.render;
+    
+    ngModelCtrl.$formatters.push(function(value) {
+      if (angular.isNumber(value) && value % 1 !== 0) {
```
chrisirhc: `-1 % 1 !== 0` How about doing integer truncation so it'll be `value << 0 === value` instead?

chrisirhc:
@moderndegree Actually, the change to use `isNumber` has more to do with the enhancing the readability of the code. Using `angular.isNumber` also avoids the use of a constant `'number'` which can be confusing to someone new to JavaScript.

src/rating/rating.js
moderndegree: wouldn't it be `value << 0 !== value`?

moderndegree:
@chrisirhc Makes sense. I've also swapped out the modulo for bitwise logic.

src/rating/rating.js
chrisirhc: Indeed, sorry for that typo.

src/rating/test/rating.spec.js
```diff
@@ -74,6 +74,20 @@ describe('rating directive', function () {
...
+    expect(element.attr('aria-valuenow')).toBe('2');
+    
+    $rootScope.rate = 2.5;
+    $rootScope.$digest();
+
+    expect(getState()).toEqual([true, true, true, false, false]);
+    expect(element.attr('aria-valuenow')).toBe('3');
+  });
+  
+  it('rounds off the number of stars shown with decimal values', function() {
```
chrisirhc: I think you meant for this to be the title of the new test ^

chrisirhc:
Just saw something else, didn't see that earlier. Otherwise, looks good.

moderndegree:
@chrisirhc Good catch. I blame the scotch.

chrisirhc:
Thank you! I've squashed the commits and cleaned up some whitespace.
Take note next time to clean up that nasty trailing whitespace!

PR #3416, opened 2015-03-21
fix(carousel): respect the order of the slides

### LINKED ISSUE #488: bug(carousel): carousel fails to update model changes

ayakovlevgh:
I've added a new parameter (index) to the slide directive which is used to order slides:
```html
<slide ng-repeat="slide in slides" index="$index">
	<!-- slide content here -->
</slide>
```

You can see it on plunker:

* [carousel without these changes] (http://plnkr.co/erR4SBeLFF6MWewtQzx6)
* [carousel with changes] (http://plnkr.co/krVqsu6n42vaxvrJ6Asi)

Fixes #488


template/carousel/carousel.html
```diff
@@ -1,6 +1,6 @@
 <div ng-mouseenter="pause()" ng-mouseleave="play()" class="carousel" ng-swipe-right="prev()" ng-swipe-left="next()">
     <ol class="carousel-indicators" ng-show="slides.length > 1">
-        <li ng-repeat="slide in slides track by $index" ng-class="{active: isActive(slide)}" ng-click="select(slide)"></li>
+        <li ng-repeat="slide in slides | orderBy:'index' track by $index" ng-class="{active: isActive(slide)}" ng-click="select(slide)"></li>
```
wesleycho: Can you extend this so that it can support functions as well?  It should support all of the possibilities of [`orderBy`](https://docs.angularjs.org/api/ng/filter/orderBy) IMO.

ayakovlevgh: Sorry, but I'm not sure that I understand what you mean. In this peace of code we're just ordering the carousel indicators by the slide index parameter. Do you mean that we should do something like this:
```html
<li ng-repeat="slide in slides | orderBy:orderIndicators track by $index" ng-class="{active: isActive(slide)}" ng-click="select(slide)"></li>
```
and then in controller:
```html
$scope.orderIndicators = function(slide) {
    return slide.index;
};
```
What's the purpose of such extension?

wesleycho: So my concern is if the user is doing something like
```
<slide ng-repeat="slide in slides | orderBy: customOrderFn">
...
</slide>
```
I'm not sure whether that custom order would be respected by this implementation.

ayakovlevgh: it would be, please check this plunker: http://plnkr.co/edit/pmCFIH0GCpBjNBWY8g1D?p=preview.

wesleycho:
Landed as b5f220fa8483f5743ba6ab3610f5064bf5c71be7, thanks for the work here!

Issue #3159, opened 2015-01-06
Datepicker malfunctioning in Firefox: fails to update the model when manually updating the date


darkbls:
Manually changing the date value, after the it was inserted, fails to update the model on Firefox.
It seems to deletes the content (?).

Check the fiddle for a proper example, open it on chrome and firefox.
http://jsfiddle.net/HB7LU/9779/

darkbls:
Anything I can do to help more?

glorat:
Changing to 
$scope.format = 'yyyy-MM-dd';

fixes it. It seems that it is a date editor, not a datetime editor. One issue (related?) in doing this is you'll find dates are internally expressed as datetimes at midnight local timezone. Remote the format filter in your display to see how it looks

karianna:
@darkbls I assume this was with Angular 1.3?

darkbls:
@karianna yes it was
@glorat  thank you, is there any suggestion that would allow one to keep using the hour and minute?

wesleycho:
Transferred to [Plunker](http://plnkr.co/edit/4WK9g3BNjWNiKzJROTcR?p=preview) for working version.

I will investigate.

wesleycho:
Interestingly, this might have to do with how the browsers interpret dates.  The object set in the view for both Chrome and Firefox is a date when the user clicks on the date.

When the user deletes a character, say from `2015-03-18 00:00:00` to give `2015-03-1 00:00:00`, Chrome thinks that `new Date('2015-03-1 00:00:00)` is a valid date string, while Firefox disagrees.

This [new Plunker](http://plnkr.co/edit/CDx8gATkoz5lBPugbMrc?p=preview) illustrates this.

In this case, Firefox's behavior looks to be more strict than Chrome's.  It turns out that Safari also agrees with Firefox here, and Chrome is the outlier.  I will investigate how IE behaves, but I believe that Chrome's behavior is actually the strange one here.  I am not sure we should support this behavior.

newmanw:
Not sure if it helps but, this format works in FF.  Not entirely sure why

```javascript
    $scope.format = 'MM/dd/yyyy HH:mm:ss';
```

Also FWIW I think it should be fixed.  Not sure what to do about new Date('2015-03-1 00:00:00), but seems like once you type in a valid date, ie add a 7 to the end new Date('2015-03-17 00:00:00) the input should re-validate to a correct date, but it doesn't.

I will look into it more to see what else I can find.

[...]

PR #3417, opened 2015-03-22
feat(dateParser): Add support for HH, H, mm, m, ss, s formats

wesleycho:
- Add support for `HH`, `H`, `mm`, `m`, `ss`, `s` formats from Angular's `dateFilter`
- Add support for `:` character in format expression

Working [Plunker](http://plnkr.co/edit/FSY7GHeGwLAUAq4h94Uu?p=preview)

This addresses #3159.

### DIFF
--- a/src/dateparser/dateparser.js
+++ b/src/dateparser/dateparser.js
@@ -1,6 +1,8 @@
 angular.module('ui.bootstrap.dateparser', [])
 
 .service('dateParser', ['$locale', 'orderByFilter', function($locale, orderByFilter) {
+  // Pulled from https://github.com/mbostock/d3/blob/master/src/format/requote.js
+  var SPECIAL_CHARACTERS_REGEXP = /[\\\^\$\*\+\?\|\[\]\(\)\.\{\}]/g;
 
   this.parsers = {};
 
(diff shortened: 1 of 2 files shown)


src/dateparser/dateparser.js
```diff
@@ -81,7 +105,7 @@ angular.module('ui.bootstrap.dateparser', [])
       return input;
     }
 
-    format = $locale.DATETIME_FORMATS[format] || format;
+    format = $locale.DATETIME_FORMATS[format] || format.replace(/:/g, '\\:');
```
chrisirhc: I think tests will fail if a DATETIME_FORMAT from $locale containing `:` is used.
Should probably add some of those formats in the dateparser.spec below (e.g. a format of `short`).

chrisirhc:
Did a review. The main issue I see is the fix for `:` may not be adequate.

src/dateparser/dateparser.js
wesleycho: Should I do a replace on the next line instead then?

chrisirhc: I think a better fix might be to escape all special characters in the regex before doing `new RegExp` on the `createParser` function.

d3 has a nice regexp to do this. See: https://github.com/mbostock/d3/blob/master/src/format/requote.js

wesleycho:
Made the change requested - can you re-review this?

src/dateparser/dateparser.js
```diff
@@ -82,6 +108,7 @@ angular.module('ui.bootstrap.dateparser', [])
     }
 
     format = $locale.DATETIME_FORMATS[format] || format;
+    format = format.replace(SPECIAL_CHARACTERS_REGEXP, '\\$&');
```
chrisirhc: We should instead should do this right before the `new RegExp()` call inside `createParser` as that shows why this character replacement is needed. There's also less weird characters to process in the `createParser`.

chrisirhc:
Did another review heh

wesleycho:
Linking related issues for reference #2508, #2513, #3159, #3288, #3294

src/dateparser/dateparser.js
wesleycho: Can you explain the less weird characters part?  Just moving this inside the `createParser` function isn't enough, there seems to be some specificity required as to when the parsing happens since the regex gets some additional characters through the iteration over `formatCodeToRegex`.  The loop adds an open and closed set of parentheses to the regex.  This means that it would be too late to modify the format string at that stage.

chrisirhc: Hm yea, I'm referring to replacing the characters right before the parentheses are added.

With the less weird characters, I'm referring to that if given a format of `hh:mm`, the `createParser` will now receive `hh\:mm`as the format when you do this here. It doesn't need that extra `\` when looking for `hh` or `mm` (in the `angular.forEach(formatCodeToRegex…` block), it only needs the extra `\` right before creating the `regex` property when feeding it into the `RegExp` constructor with the parentheses.

wesleycho: As per discussion between @chrisirhc and I on Slack, this turns out to be the best spot for modifying the format given how the iteration in `formatCodeToRegex` is implemented.

PR #3427, opened 2015-03-24
fix(timepicker): move render logic to formatter

### LINKED ISSUE #3160: Timepicker triggers dirty flag inside ngRepeat

wesleycho:
- Moves render logic converting model values into date or null object into a formatter, which allows the render function to more correctly render using the viewValue

Working Plunker [here](http://plnkr.co/edit/vm9IgKtXmAYaXERLORpf?p=preview).

Fixes #3160 properly.


src/timepicker/timepicker.js
```diff
@@ -20,6 +20,10 @@ angular.module('ui.bootstrap.timepicker', [])
     ngModelCtrl = ngModelCtrl_;
     ngModelCtrl.$render = this.render;
 
+    ngModelCtrl.$formatters.unshift(function (modelValue) {
+      return (modelValue) ? new Date( modelValue ) : null;
```
chrisirhc: Extra parentheses around the first `modelValue`

chrisirhc:
LGTM . This resolves the issue ? I thought it had something to do with the input fields' type .

wesleycho:
It fixes the issue I think due to the `render` function incorrectly using the modelValue instead of the viewValue and formatting translating the modelValue to the viewValue.

wesleycho:
Note: this PR also would close out #2069.

chrisirhc:
Merged. I left #2069 open as there's still #3494 .

Issue #1098, opened 2013-09-30
Datepicker shows two October 19, 2013 dates! Live at AngularJS UI website


jbruni:
Guys... my client told me datepicker was showing two October 19 2013 dates...

To my surprise, not only he is correct, but also had nothing to do with my particular app...

Just go to http://angular-ui.github.io/bootstrap/#/datepicker and move the month forward (we are currently in September, as I write this), and you will see October 19 both as Saturday **and repeated** as Sunday!

I call "urgency" because we are about to launch this app... and because we can easily see the issue while still in September.

PR #3433, opened 2015-03-25
fix(datepicker): time zone bug

### LINKED ISSUE #3079: [Datepicker] Repeated months
There is an issue with repeated months in some years in Russian time zones (Moscow UTC +3, Novosibirsk UTC +6). 
The issue occurrs only in Firefox (version 33.0, 34.0). In Google Chrome everything is OK.
Angular bootstrap version: 0.12.0.
Screenshots are attached.

datepicker2
datepicker1

ayakovlevgh:
This bug is quite difficult to reproduce because it depends on OS, OS updates, browser, current timezone and even current date.
As I understand, it's connected with daylight savings.
The bug was discussed here: #1098.
The problem with repeated days seems to be fixed here: 7f4b40eb8bc57f03bd306330a6ad227790c33f27 (line 145) . But the problem with repeated months and years still exists.

Some screenshots:
1. OS: Windows 7; current day: December 10, 2014; current time zone: +6 GMT; browser: Firefox 36.0.4
*Wrong January:*
winmonth
*Wrong 2014:*
winyear
2. OS: Debian Jessie / Ubuntu 14.04; current day: December 10, 2014 / March 24, 2015; current time zone: +3 GMT / +6 GMT; browser: Firefox 31.5.0, 33.0, 34.0.
*Wrong April:*
linuxmonth

You can see the reason of the bug in this piece of code:
```javascript
var date = new Date(2014, 0, 1);
console.log(date.getFullYear(), date.getMonth(), date.getDate(), date.getHours());
```
The expected result is *"2014 0 1 0"*, but in some cases it will be *"2013 11 31 23"*.

One more example. I printed the dates which is used to display months in monthpicker:
```javascript
// February: correct
{
	"date":"Date {Wed Feb 01 1984 00:00:00 GMT+0600 (OMST)}",
	"label":"February",
        . . .
}
// March: correct
{
	"date":"Date {Thu Mar 01 1984 00:00:00 GMT+0600 (OMST)}",
	"label":"March",
        . . .
}
// April: error, 1 hour is missing
{
	"date":"Date {Sat Mar 31 1984 23:00:00 GMT+0600 (OMST)}",
	"label":"March",
        . . .
}
// May: correct
{
	"date":"Date {Tue May 01 1984 00:00:00 GMT+0700 (OMSST)}",
	"label":"May",
        . . .
}
```

Discussions of this bug can be found on the Internet. Some of them:
http://stackoverflow.com/questions/5859959/dst-problem-storing-date-in-javascript-date-object
https://code.google.com/p/chromium/issues/detail?id=417640

I've implemented a possible workaround. We should determine whether a date is created with the bug. If the bug arises then we just add 2 hours to the date (adding one hour has no effect in this case). It should guarantee the correct handling of all the bug cases.

Fixes #3079

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -134,7 +134,8 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   // console.log(date.getFullYear(), date.getMonth(), date.getDate(), date.getHours());
   // can result in "2013 11 31 23" because of the bug.
   this.fixTimeZone = function(date) {
-    date.setHours(date.getHours() === 23 ? date.getHours() + 2 : 0);
+    var hours = date.getHours();
+    date.setHours(hours === 23 ? hours + 2 : 0);
   };
 
   $scope.select = function( date ) {
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -127,6 +127,16 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
     return arrays;
   };
 
+  // Fix a hard-reprodusible bug with timezones
+  // The bug depends on OS, browser, current timezone and current date
+  // i.e.
+  // var date = new Date(2014, 0, 1);
+  // console.log(date.getFullYear(), date.getMonth(), date.getDate(), date.getHours());
+  // can result in "2013 11 31 23" because of the bug.
+  this.fixTimeZone = function(date) {
+    date.setHours(date.getHours() > 12 ? date.getHours() + 2 : 0);
+  };
+
   $scope.select = function( date ) {
     if ( $scope.datepickerMode === self.minMode ) {
       var dt = ngModelCtrl.$viewValue ? new Date( ngModelCtrl.$viewValue ) : new Date(0, 0, 0, 0, 0, 0, 0);
@@ -237,10 +247,11 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
       }
 
       function getDates(startDate, n) {
-        var dates = new Array(n), current = new Date(startDate), i = 0;
-        current.setHours(12); // Prevent repeated dates because of timezone bug
+        var dates = new Array(n), current = new Date(startDate), i = 0, date;
         while ( i < n ) {
-          dates[i++] = new Date(current);
+          date = new Date(current);
+          ctrl.fixTimeZone(date);
+          dates[i++] = date;
           current.setDate( current.getDate() + 1 );
         }
         return dates;
@@ -339,10 +350,13 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
 
       ctrl._refreshView = function() {
         var months = new Array(12),
-            year = ctrl.activeDate.getFullYear();
+            year = ctrl.activeDate.getFullYear(),
+            date;
 
         for ( var i = 0; i < 12; i++ ) {
-          months[i] = angular.extend(ctrl.createDateObject(new Date(year, i, 1), ctrl.formatMonth), {
+          date = new Date(year, i, 1);
+          ctrl.fixTimeZone(date);
+          months[i] = angular.extend(ctrl.createDateObject(date, ctrl.formatMonth), {
             uid: scope.uniqueId + '-' + i
           });
         }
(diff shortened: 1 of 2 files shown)


src/datepicker/datepicker.js
```diff
@@ -127,6 +127,16 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
...
   };
 
+  // Fix a hard-reprodusible bug with timezones
+  // The bug depends on OS, browser, current timezone and current date
+  // i.e.
+  // var date = new Date(2014, 0, 1);
+  // console.log(date.getFullYear(), date.getMonth(), date.getDate(), date.getHours());
+  // can result in "2013 11 31 23" because of the bug.
+  this.fixTimeZone = function(date) {
+    date.setHours(date.getHours() > 12 ? date.getHours() + 2 : 0);
```
wesleycho: This does not look like a valid fix - can you explain this further?

ayakovlevgh: Today reproduced it on Win 7 with GMT-0300 (Brasilia) using Firefox 36.0.4.
Code:
```javascript
var date = new Date(2016, 9, 16);

console.log('date: ', date);
console.log('hours:', date.getHours());
console.log('timestamp:', date.getTime());

date.setHours(date.getHours() + 1);

console.log('date: ', date);
console.log('hours:', date.getHours());
console.log('timestamp:', date.getTime());

date.setHours(date.getHours() + 2);

console.log('date: ', date);
console.log('hours:', date.getHours());
console.log('timestamp:', date.getTime());
```
Result:
```
date: Date {Sat Oct 15 2016 23:00:00 GMT-0300}
hours: 23
timestamp: 1476583200000

date: Date {Sat Oct 15 2016 23:00:00 GMT-0300}
hours: 23
timestamp: 1476583200000

date: Date {Sun Oct 16 2016 01:00:00 GMT-0200}
hours: 1
timestamp: 1476586800000
```
So we should make sure that hours are not equal to 23. If they are, we need to add two hours to the date (as you can see adding one hour doesn't work in this case).
If this part misleads you: `date.getHours() > 12 ` I can change to `date.getHours() === 23 ` .

I think behavior of this fix is quite predictable and should deal with all the cases of this issue.

Sorry if I'm not clear enough.

pnuckowski:
Instead of manage date/time problem manually, maybe it would be a good decision to use moment.js?

src/datepicker/datepicker.js
wesleycho: Yeah, an explicit check for 23 would be better here

ayakovlevgh: Done.

src/datepicker/datepicker.js
```diff
@@ -127,6 +127,16 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
...
   };
 
+  // Fix a hard-reprodusible bug with timezones
+  // The bug depends on OS, browser, current timezone and current date
+  // i.e.
+  // var date = new Date(2014, 0, 1);
+  // console.log(date.getFullYear(), date.getMonth(), date.getDate(), date.getHours());
+  // can result in "2013 11 31 23" because of the bug.
+  this.fixTimeZone = function(date) {
+    date.setHours(date.getHours() === 23 ? date.getHours() + 2 : 0);
```
wesleycho: A micro-optimization here would be to do
<pre>
var hours = date.getHours();
date.setHours(hours === 23 ? hours + 2 : 0);
</pre>

wesleycho:
This is interesting - at first I was skeptical about the solution, but thinking about it some more in the context to the datepicker automatically setting the hours to 0, this might be a sufficient fix.

The only question is whether this is sufficient for us - this would be a breaking change if a developer is supporting the entering of a time.

chrisirhc:
Interesting. This problem arises because the calling the `new Date()` hits a non-existent time in the local timezone. Hence, the time is adjusted forward/backwards to pick a time that exists.

I think this is a good reason to refactor and remove the need to even create these `Date` objects. It looks like they're just created so that they can be used in the click handler. There's no need to use `Date` objects in the month picker since there's always 12 months and it's always Jan - Dec. We can instead add a method that allows you to set the month / year of the current model object instead of providing a `Date` object to replace it with.
By removing the need for `Date` objects, we should isolate the timezone issues to the `select` method.
May need to think about this more, but this is what I've come up with.

---

With regards to this solution, this could work. But we'll want to remove this code soon because of the reasons outlined above.

wesleycho:
I think breaking changes should wait until we split the datepicker into a different repository.  I would rather the datepicker just delegated this behavior to moment.js, it would address most of these woes.

wesleycho:
I plan on merging this today, but one thing I plan on doing with the datepicker is modifying it to allow usage of a custom parser (i.e. moment.js), with the fallback being the current behavior.

wesleycho:
Oops, forgot the PR # in my rebasing...this has landed as f14120149b9f4ff67e8abb8907b5ab973fe3987f

Issue #375, opened 2013-05-01
bug(popover): overflowing window borders


yangit:
Popovers are overflowing over window borders.
http://plnkr.co/edit/q2Zj1F
Solution might be to add "Auto" positioning option.

ilovett:
:+1:

rzschech:
+1

knorrus:
+1

drewbeck:
Still need this! Were I JS ninja sufficient I would whip it up. Alas :(

cleversprocket:
A workaround is to add another directive to the element and bind a mouseenter event or whatever is being used to fire the tooltip and adjust the position:
```javascript
angular.module('app')
    .directive('positionTooltip', ['$timeout', function ($timeout) {
        return {
            restrict: 'A',
            scope: false,
            link: function (scope, element, attrs) {
                var el,
                    arrow,
                    offset;
                element.bind('mouseenter', function () {
                    $timeout(function () {
                        el = element.next();
                        if (el[0].offsetLeft < 0) {
                            offset = el[0].offsetLeft;
                            el.css('left', 0);
                            arrow = el[0].querySelector('.tooltip-arrow');
                            arrow = angular.element(arrow);
                            arrow.css('margin-left', (offset + -5) + 'px');
                        }
                    })
                });
            }
        }
    }]);
```

thegreatmichael:
Does https://github.com/twbs/bootstrap/pull/12328 end up in here somehow?

jpmnteiro:
+1

efosmark:
+1

RobJacobs:
I have a couple of ideas I've been working on for this issue I'd like to share.  I need to pull them together in a plunk, then will post them shortly.

chrisirhc:
@thegreatmichael thanks for mentioning that issue, we can use that as a reference implementation.

rufinitsw:
+1

pyella:
+1

rvanbaalen:
It's already clear that this is a wanted feature. 

#### Please stop all the +1 responses

We'd like to keep this open for on-topic discussion instead of having to receive notifications for all +1 responses.

pads:
[...]

PR #3435, opened 2015-03-25
feat(tooltip): update position dynamically

### LINKED ISSUE #96: fix(tooltip): update placement on content change
Right now, the `tooltip` attribute is observed and so is properly bound. But the position of the tooltip is not updated when the attribute changes, and so then appears off-center.

If you go to the demo page, click in the "Dynamic Tooltip Popup" input, hover over the "dynamic" tooltip, and then proceed to change the value, you can see the issue.

This is very minor right now, but one of the features to still be ported from Twitter's jQuery version will keep the tooltip open, and thus the artifact would be visible.

### LINKED ISSUE #1109: Position service (tooltips, popovers) jump away from element if a dom element is added above the element
For example, see this screencast:

http://screencast.com/t/yflYrdBP

You can also reproduce this bug by adding DOM elements above popover / tooltip elements on the demo site.  The problem is that position is absolute, calculated on show, instead of elements being anchored.  Possible solutions are anchoring to element on which tooltip / popover is applied, or listening for changes in height and adjusting accordingly.

chrisirhc:
Reposition on each digest, this ensures that the tooltip is always positioned
correctly no matter how its content changes.

Fixes #96
Fixes #1109
Closes #2816

Prerequisite of #1848

### DIFF
--- a/src/tooltip/test/tooltip.spec.js
+++ b/src/tooltip/test/tooltip.spec.js
@@ -446,6 +446,58 @@ describe('tooltipWithDifferentSymbols', function() {
 
 });
 
+describe( 'tooltip positioning', function() {
+  var elm, elmBody, elmScope, tooltipScope, scope;
+  var $position;
+
+  // load the tooltip code
+  beforeEach(module('ui.bootstrap.tooltip', function ( $tooltipProvider ) {
+    $tooltipProvider.options({ animation: false });
+  }));
+
+  // load the template
+  beforeEach(module('template/tooltip/tooltip-popup.html'));
+
+  beforeEach(inject(function($rootScope, $compile, _$position_) {
+    $position = _$position_;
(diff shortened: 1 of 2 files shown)


RobJacobs:
Looks like this will also fix #375.

wesleycho:
Yuck...this will likely be pretty unperformant I think.

We can revisit this though after releasing 0.13 and seeing what we can do to improve the performance.

chrisirhc:
Yep indeed. There could be some optimizations to be made but I think it should be okay for now unless users of the directive have a lot of open tooltips at the same time.

chrisirhc:
@RobJacobs unfortunately, this won't resolve #375 . That issue requires changes to the way `$position` works.

venkikalva:
is overflowing window borders issue is fixed?

PR #3441, opened 2015-03-25
Update CHANGELOG.md

ReedD:
In previous versions you needed to add `dropdown-toggle` to menu items that didn't leave the current page or else the dropdown would remain open even after a menu click. This is no longer the case and extra `dropdown-toggle`s will cause the menu to immediately close after opening.


rvanbaalen:
Changelog is auto-generated -- any changes to it will be gone on next release.

PR #3442, opened 2015-03-25
fix(timepicker): remove ng-mousewheel binding

RobJacobs:
The timepicker template has an ng-mousewheel
binding on the hours input element and the
TimepickerController also binds the elemeent
to mouse wheel events.  Removed ng-mousewheel
binding from the template.


chrisirhc:
Small thing, could you change the commit message to be of `refactor` type since this isn't meant to change any behavior?

Issue #2613, opened 2014-08-21
datepicker: watchers count


porjo:
I've been looking at ways to improve performance in my app. I've read that reducing number of unnecessary watchers is something that can help.

Using a simple [watch counter](http://stackoverflow.com/a/19071095) I see that the number of watchers increases by 296 when a datepicker is introduced into the page. See this [plunker](http://plnkr.co/edit/pDTyIq4STsXoAmUkxfog?p=preview) for a demo. AFAIK this is due to the html template using ng-repeat to create the table of days.

Some questions:
1. Is this a real issue i.e. is my analysis correct?
2. If it is a problem, can anything be done?

YaManicKill:
I was looking into this problem (because we have it too).

One solution would be to do an ng-if on the popup. My thought was to have it just do ng-if="isOpen", but that doesn't work as it loses the data stored in the model then, for some reason.

The other option is to create another variable that a user can pass in to say when they want it to be initialised. This would be bound to ng-if and would mean that once it is opened the first time, it is brought into the html and the watchers are assigned.

This would help people who have multiple, as it ends up being a big problem.

However, when trying to do it, I have 5 tests failing. I reckon these are because of the slight performance hit that you get when using ng-if meaning that the popup isn't ready for when it tries to click a date. I'll work on this a bit later tomorrow, but it's at least something that *can* be done.

RobJacobs:
@YaManicKill See this pull request for the ng-if implementation: https://github.com/angular-ui/bootstrap/pull/1915

YaManicKill:
Ah, yeah that's exactly what I was working on. Any plans to actually merge it in?

daric81:
+1
Is there any update on this?

karianna:
https://github.com/angular-ui/bootstrap#impt---2015-plans-and-angular-13-support

cpaof:
+1
28212 watchers for me ;)

RobJacobs:
There's simply too many watchers on the templates.  Doing a quick look through of the popup.html, the ng-repeat that is building the day/week/cell grid, for each day displayed: id={{dt.uid}} = 1 watch, aria-disabled={{!!dt.disabled}} = 1 watch, ng-class="{'btn-info': dt.selected}" = 1 watch, ng-class="{'text-muted' : dt.secondary = 1 watch, {{dt.label}} = 1 watch which adds up to 5 watches per cell in the grid.

cpaof:
[...]

PR #3443, opened 2015-03-25
refactor(datepicker): implement one time bindings

RobJacobs:
Reducing the number of watches created for the 
datepicker by implementing one time binding
on the datepicker templates. Removing aria-disabled
attribute as the ngAria module will add that
to the button with ng-disabled.


template/datepicker/day.html
```diff
@@ -1,20 +1,20 @@
...
-      <th colspan="{{5 + showWeeks}}"><button id="{{uniqueId}}-title" role="heading" aria-live="assertive" aria-atomic="true" type="button" class="btn btn-default btn-sm" ng-click="toggleMode()" ng-disabled="datepickerMode === maxMode" tabindex="-1" style="width:100%;"><strong>{{title}}</strong></button></th>
+      <th colspan="{{5 + showWeeks}}"><button id="{{::uniqueId}}-title" role="heading" aria-live="assertive" aria-atomic="true" type="button" class="btn btn-default btn-sm" ng-click="toggleMode()" ng-disabled="datepickerMode === maxMode" tabindex="-1" style="width:100%;"><strong>{{title}}</strong></button></th>
       <th><button type="button" class="btn btn-default btn-sm pull-right" ng-click="move(1)" tabindex="-1"><i class="glyphicon glyphicon-chevron-right"></i></button></th>
     </tr>
     <tr>
-      <th ng-show="showWeeks" class="text-center"></th>
-      <th ng-repeat="label in labels track by $index" class="text-center"><small aria-label="{{label.full}}">{{label.abbr}}</small></th>
+      <th ng-show="::showWeeks" class="text-center"></th>
```
Foxandxss: Not sure if one.time on ng-show is gonna work.

RobJacobs: I should be using ng-if, but then tests fail.

wesleycho:
Is removing aria-disabled a good idea? Users of ngAria would want that, no?

RobJacobs:
Partially fixes #2613, should compliment #1915

realityking:
From my understanding of the [ARIA spec](http://www.w3.org/WAI/PF/aria/roles#gridcell), the `aria-disabled` should stay as this indicated that the griddle is disabled, not just the button within it.

BTW, for buttons, ngAria is not needed. All modern browsers map the accessible information correctly for disabled button. Unfortunately ngAria will add unnecessary attributes to the DOM anyway.

RobJacobs:
@wesleycho @realityking I should have done a better job of explaining why I think the aria-disabled binding should be removed on the td element.  Here is a [plunk](http://plnkr.co/edit/nhQOpKnWjeYvzElzuxPb?p=preview) with the latest build from master, Angular 1.3, and ngAria loaded.  You will notice the rendered html for the day/month/year picker buttons when disabled looks like:

    <td aria-disabled="true" ng-repeat="dt in row track by dt.date">
      <button tabindex="-1" disabled="disabled" aria-disabled="true"

Using ng-disabled directive on the button element will cause the ngAria service to add the aria-disabled attribute for you.  I don't think it's necessary to have the aria-disabled attribute on both the td and button elements.  

The rendered html for an enabled button doesn't have the disabled or aria-disabled attributes at all where the enabled td element aria-disabled attribute is rendered as aria-disabled="false".  That leads me to believe the correct implementation for using aria-disabled is to remove or add the attribute versus setting the value.

realityking:
I've spent some time with the ARIA spec, but I'm far from an expert.

Still there are a few things worth clarifying:

1. Unlike the `disabled` attribute, `aria-disabled` (and other aria states) is not a [boolean](http://stackoverflow.com/questions/4139786/what-does-it-mean-in-html-5-when-an-attribute-is-a-boolean-attribute) attribute whose presence or absence determines what's going on, the value actually matters. (Otherwise we'd write `aria-hidden="aria-hidden"`). It's a confusing mess, but that's what you get from an evolving system.

2. In my opinion, the `aria-disabled`attribute is correct, because it describes the state of the gridcell, not that of the button within it.

The ARIA Best Practices document has a section on datepicker [date picker](http://www.w3.org/TR/wai-aria-practices/#datepicker), it doesn't mention `aria-disabled` explicitly but the linked implementation in Dojo does use it.

chrisirhc:
I like the change to remove aria-disabled, and leave it to ngAria to handle that for us since we're moving to AngularJS 1.3 anyway.

hamfastgamgee:
I've been running with a closely-related variant of this and #1915 in my local fork for a long time, so I'm going to add a couple of suggestions to the patch if you don't mind. :)

(For reference, we have one data entry page that has probably 50 datepickers on it.  Without this and #1915, let's just say that page would not be loading in under 20 seconds...)

RobJacobs:
@hamfastgamgee I decided to err on the side of caution, I can't imagine why someone would want to change the show weeks or start day attributes while the datepicker is visible, but you never now.

hamfastgamgee:
Fair enough.  Might warrant further discussion as to whether those are actually desired, though, since at least in the start date case it's one extra watcher per day visible.  (The showWeeks stuff is one per picker, so not as big of an issue.)

wesleycho:
@RobJacobs what's the state of the PR? Since we released 0.13.0, we are in the position to merge this if this is ready.

RobJacobs:
@wesleycho It's good to go as far as I'm concerned.  Rebased and tests still passing.

wesleycho:
Alright, I'll review it tonight

BobbieBarker:
I approve

src/datepicker/test/datepicker.spec.js
```diff
@@ -59,10 +59,10 @@ describe('datepicker directive', function () {
     return element.find('thead').find('tr').eq(1);
   }
 
-  function getLabels() {
+  function getLabels( dayMode ) {
```
wesleycho: Remove spaces around `dayMode`

wesleycho:
LGTM minus one code style comment.

PR #3445, opened 2015-03-26
feat(typeahead): implement 'select on blur' option.

DSI-Entreprises:
This PR adds an option 'select on blur' on typeahead. When enabled, this option selects the current highlighted value (in popup list) when the widget is losing focus. In short : clicking outside the widget is identical to hitting the 'tab' key.

By default, this option is disabled.

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -364,6 +364,32 @@ describe('typeahead tests', function () {
       expect(element).toBeClosed();
     });
 
+    it('should not select any match on blur without \'select-on-blur=true\' option', function () {
+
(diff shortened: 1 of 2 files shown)


karianna:
Needs tests and it currently fails the build...

DSI-Entreprises:
Ok, I'm working on it... Thanks!

DSI-Entreprises:
I have added two tests. Let me know if they are good enough. Thanks.

wesleycho:
I like this - scheduling this to be merged for the next release (we've locked down 0.13.0 right now for Angular 1.3 compatibility).

Can you add documentation?

DSI-Entreprises:
Thanks a lot !

PR #3459, opened 2015-03-28
chore(transition): remove transition module

chrisirhc:
BREAKING CHANGE:
The `transition` module along with its `$transition` service are no longer
available. We're now using `$animate` service for animations and would
encourage users to do the same.


wesleycho:
Should this be in the backlog or in 0.13.0? Given that $transition usage is gone, we might as well just delete it as a breaking change from 0.12 to 0.13.

chrisirhc:
I was thinking of merging it next major release, say 0.14, as there might be users depending on it and it's not essential to drop it now. Just want to mark it as deprecated first.

wesleycho:
Might be worth opening a PR marking it as deprecated in the comments for the service or when users consume the service via `$log`.

chrisirhc:
Good idea, will do

karianna:
@chrisirhc Weirdly this needs rebasing

wesleycho:
Don't worry about this, the rebase would be easy when it comes time to do it.

wesleycho:
I merged this in dfa4c97cb6ef7c83d88ffcfe7d31cdd851a58162, but looks like I forgot to just close this particular PR :(

PR #3460, opened 2015-03-28
Removed need for empty dropdown attributes.

palaslet:
By adding "restrict: 'EAC'" to the directives, we no longer need to add the empty dropdown and dropdown-toggle directives.
This need for duplicity (both class and attribute) contributes to more clutter in the html.


chrisirhc:
We've removed class restrict intentionally. See https://github.com/angular-ui/bootstrap/issues/2156 for discussion. The class attribute causes unintentional conflict.

wesleycho:
Closing this, due to the explicit removal of support as mentioned in the above referenced issue.

PR #3461, opened 2015-03-28
chore(tests): use angular installed via npm

chrisirhc:
- Avoid maintaining / copying AngularJS source code on our repo.
- Ensures that Travis checks out the latest Angular version and tests
  against it.

Note, will do the same for jQuery later but that requires some changes to test code.


chrisirhc:
Rebased

Issue #3397, opened 2015-03-17
Carousel - Add "wrap"-option


lippoliv:
Hey there,
on bootstrap you can stop an carousel from endless cycling by set "wrap" to "false".

Looking into the sources here https://github.com/angular-ui/bootstrap/blob/master/src/carousel/carousel.js#L82 I think this is currently not possible for angular js bootstrap ui, isn't it?

Would be nice to have such an Option.

wesleycho:
To clarify, you mean that it will stop on the last slide instead of going back to the first?

lippoliv:
Yes, sorry for beeing unclear ... :)

Digambar14:
When this feature will be available?

PR #3462, opened 2015-03-29
feat(carousel): adding noWrap option to prevent re-cycling of slides

ChewTeaYeah:
Pausing the carousel timer function after the last slide is rendered if the noWrap option on the scope is present and set to a truthy value. Addresses #3397


src/carousel/carousel.js
```diff
@@ -219,7 +223,8 @@ angular.module('ui.bootstrap.carousel', [])
     scope: {
       interval: '=',
       noTransition: '=',
-      noPause: '='
+      noPause: '=',
+      noWrap: '=?'
```
chrisirhc: Since this is readonly, use `'&'` instead and read it using `$scope.noWrap()`. This saves on a watcher.
We'll eventually change the rest of these scope variables to use `'&'` for performance. I'll create another issue for that.

ChewTeaYeah:
OK I've made the noWrap property one-way bound now instead of two-way. I'd also like some feedback on my unit test. I'm new to unit testing angular, and I'm not sure that I'm binding properties to the directive controller's scope properly, i.e. via the compiled element

src/carousel/test/carousel.spec.js
```diff
@@ -73,6 +73,42 @@ describe('carousel', function() {
       expect(indicators.length).toBe(3);
     });
 
+    it('should stop cycling slides when noWrap is truthy', function () {
+      scope.noWrap = function () {
```
chrisirhc: should be `scope.noWrap = true`;

src/carousel/carousel.js
```diff
@@ -219,7 +223,8 @@ angular.module('ui.bootstrap.carousel', [])
     scope: {
       interval: '=',
       noTransition: '=',
-      noPause: '='
+      noPause: '=',
+      noWrap: '&?'
```
chrisirhc: Just `&` is adequate

ChewTeaYeah:
Updated based on feedback from @chrisirhc

src/carousel/carousel.js
```diff
@@ -75,6 +74,11 @@ angular.module('ui.bootstrap.carousel', [])
   $scope.next = function() {
     var newIndex = (self.getCurrentIndex() + 1) % slides.length;
 
+    if ((newIndex === 0) && $scope.noWrap()){
```
chrisirhc: The parentheses around `newIndex === 0` isn't needed.

chrisirhc:
This PR looks good to me. The above is just a style comment that isn't important.
We can look to merge this in after the 0.13 release.

wesleycho:
Looking at this PR and thinking about it some more, this should also support preventing going from a slide of index 0 to the last slide - can you update this PR with that change too?

ChewTeaYeah:
Sorry for the late reply. I can definitely update this PR with that change. Should I handle the documentation updates as well?

karianna:
@ChewTeaYeah Rebase & docs yes please!

ChewTeaYeah:
I've made the changes and modified my unit tests as well, but I'm not sure where I should make the docs change. Carousel has a short blurb about its functionality, but does not explicitly specify its parameters. Should I add to that blurb?

ChewTeaYeah:
@wesleycho can you provide some guidance on the documentation updates per my last comment? This is ready to go other than documentation and I'd love to see it merged in soon :-)

wesleycho:
Yes, adding the blurb for the carousel element would be good.

ChewTeaYeah:
@wesleycho I've updated the docs and added an option to the demo to enable noWrap on the carousel's slides. I also rebased and fixed conflicts, so this is ready to go!

wesleycho:
There you go :) - it will make it out in the next release.

ChewTeaYeah:
Thanks! This was my first contribution to anything open source, ever. Hope its the first of many!

wesleycho:
Keep at it, there's lots to learn and much to do :)

Issue #2884, opened 2014-10-24
Typeahead use bind-html-unsafe


albertein:
The typeahead directive uses bind-html-unsafe for the template of matches, https://github.com/angular-ui/bootstrap/blob/master/template/typeahead/typeahead-match.html.

The property on the directive doesn't indicate on any way that it can be dangerous, it should either:

  a) Remove html-bind-unsafe entirely 
  b) Having typeahead use html-bind-unsafe should be a parameter set explicitly.

I would be happy to do any of those.

fernando-sendMail:
Hello:
I found this issue too, and since I need to create a workaround for this I want to work out this this weekend. Is anybody working on this currently or is it done yet?

If not, let me know so I can do it, and do a pull request.

On a side note I've never made a contribution, do you have any additional guidelines besides the ones posted on github, say for this project in particular?

Thanks for your time!

karianna:
Nothing special, except for make sure it's covered by tests :-) go for it!

nschloe:
This is also popping up for our project; a fix for this would be appreciated. :)

erik-slack:
Same here.

Frondor:
bindHtmlUnsafe is now deprecated. Use ngBindHtml instead (Google Chrome console warning)

andrzejpiasecki:
Same here.

kikar:
+1

bambooyzj:
same here

wesleycho:
This is now removed in `master`, and should be resolved in the next update.

PR #3463, opened 2015-03-29
fix(typeahead): Typeahead uses ngSanitize when present

fernando-sendMail:
Hey there!

This pull request addresses the issue #2884 which mentions the use of the attribute of `bind-html-unsafe` attribute in the default `template/typeahead/typeahead-match.html`.

So basically, I changed a little the testing process, considering what @chrisirhc said about the fix, here's the implemented solution:

1. Added the `angular-sanitize.js` dependency to the `misc/test-lib/` folder (the `angular-sanitize.js` is the same version as the `angular.js` included).
2. Changed the `karma.conf.js` file to add this dependency.
3. Added a line in the demo file to ship it with the `dist` folder.
4. Changed the `typeaheadHilghlight` filter:
  * Checks if `$sanitize` is available to use, if not then a warning is printed to the console using the `$log` service, this will happen when the filer is requested by angular. The `ngSanitize` can be imported as part of the module that's using the `typeahead` directive. The reason of why the `ngSanitize` module is not imported directly on the `ui.bootstrap.typeahead` module is because it will add another dependency to the project, needless to say it will not work if not present.
  * When `$sanitize` is present it will use it to sanitize the string itself if it contains any sequence of `<` followed by `any combination of chars`, and an ending `>`, if not, then the string is wrapped in an object that `ng-bind-html` need to be displayed with the `$sce.trustAsHtml` method.
  * And finally the `<strong>` tags are added to the matched characters.

5. On the file `src/typeahead/test/typeahead-highlight.spec.js` the returned value of the filter need to be unwrapped.
6. Created another spec file (`src/typeahead/test/typeahead-highlight-ngsanitize.spec.js`) that will include the `ngSanitize` module and test that tags and attributes are being escaped.

And that's it, I think that's all.

I'm looking forward to your feedback and comments. Thanks!

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -18,6 +18,7 @@ module.exports = function(config) {
       'misc/test-lib/jquery-1.8.2.min.js',
       'node_modules/angular/angular.js',
       'node_modules/angular-mocks/angular-mocks.js',
+      'node_modules/angular-sanitize/angular-sanitize.js',
       'misc/test-lib/helpers.js',
       'src/**/*.js',
       'template/**/*.js'
--- a/misc/demo/index.html
+++ b/misc/demo/index.html
@@ -14,6 +14,7 @@
     <script src="//ajax.googleapis.com/ajax/libs/angularjs/<%= ngversion %>/angular.min.js"></script>
     <script src="//ajax.googleapis.com/ajax/libs/angularjs/<%= ngversion %>/angular-animate.min.js"></script>
     <script src="//ajax.googleapis.com/ajax/libs/angularjs/<%= ngversion %>/angular-touch.min.js"></script>
+    <script src="//ajax.googleapis.com/ajax/libs/angularjs/<%= ngversion %>/angular-sanitize.js"></script>
     <script src="ui-bootstrap-tpls-<%= pkg.version%>.min.js"></script>
     <script src="assets/plunker.js"></script>
     <script src="assets/app.js"></script>
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "devDependencies": {
     "angular": "^1.4.3",
     "angular-mocks": "^1.4.3",
+    "angular-sanitize": "^1.4.3",
     "grunt": "^0.4.5",
     "grunt-contrib-concat": "^0.5.1",
     "grunt-contrib-copy": "^0.8.0",
--- a/src/typeahead/test/typeahead-highlight-ngsanitize.spec.js
+++ b/src/typeahead/test/typeahead-highlight-ngsanitize.spec.js
@@ -1,0 +1,17 @@
+describe('Security concerns', function() {
+  var highlightFilter, $sanitize, logSpy;
+
+  beforeEach(module('ui.bootstrap.typeahead', 'ngSanitize'));
+
+  beforeEach(inject(function (typeaheadHighlightFilter, _$sanitize_, $log) {
+    highlightFilter = typeaheadHighlightFilter;
+    $sanitize = _$sanitize_;
+    logSpy = spyOn($log, 'warn');
+  }));
+
+  it('should not call the $log service when ngSanitize is present', function() {
+    highlightFilter('before <script src="">match</script> after', 'match');
+    expect(logSpy).not.toHaveBeenCalled();
+  });
+
+});
--- a/src/typeahead/test/typeahead-highlight.spec.js
+++ b/src/typeahead/test/typeahead-highlight.spec.js
@@ -1,38 +1,52 @@
-describe('typeaheadHighlight', function() {
-  var highlightFilter;
+describe('typeaheadHighlight', function () {
+
+  var highlightFilter, $log, $sce, logSpy;
 
   beforeEach(module('ui.bootstrap.typeahead'));
-  beforeEach(inject(function(typeaheadHighlightFilter) {
+
+  beforeEach(inject(function(_$log_, _$sce_) {
+    $log = _$log_;
+    $sce = _$sce_;
+    logSpy = spyOn($log, 'warn');
+  }));
+
+  beforeEach(inject(function (typeaheadHighlightFilter) {
     highlightFilter = typeaheadHighlightFilter;
   }));
 
-  it('should higlight a match', function() {
-    expect(highlightFilter('before match after', 'match')).toEqual('before <strong>match</strong> after');
+  it('should higlight a match', function () {
+    expect($sce.getTrustedHtml(highlightFilter('before match after', 'match'))).toEqual('before <strong>match</strong> after');
   });
 
-  it('should higlight a match with mixed case', function() {
-    expect(highlightFilter('before MaTch after', 'match')).toEqual('before <strong>MaTch</strong> after');
+  it('should higlight a match with mixed case', function () {
+    expect($sce.getTrustedHtml(highlightFilter('before MaTch after', 'match'))).toEqual('before <strong>MaTch</strong> after');
   });
 
-  it('should higlight all matches', function() {
-    expect(highlightFilter('before MaTch after match', 'match')).toEqual('before <strong>MaTch</strong> after <strong>match</strong>');
+  it('should higlight all matches', function () {
+    expect($sce.getTrustedHtml(highlightFilter('before MaTch after match', 'match'))).toEqual('before <strong>MaTch</strong> after <strong>match</strong>');
   });
 
-  it('should do nothing if no match', function() {
-    expect(highlightFilter('before match after', 'nomatch')).toEqual('before match after');
+  it('should do nothing if no match', function () {
+    expect($sce.getTrustedHtml(highlightFilter('before match after', 'nomatch'))).toEqual('before match after');
   });
 
-  it('should do nothing if no or empty query', function() {
-    expect(highlightFilter('before match after', '')).toEqual('before match after');
-    expect(highlightFilter('before match after', null)).toEqual('before match after');
-    expect(highlightFilter('before match after', undefined)).toEqual('before match after');
+  it('should do nothing if no or empty query', function () {
+    expect($sce.getTrustedHtml(highlightFilter('before match after', ''))).toEqual('before match after');
+    expect($sce.getTrustedHtml(highlightFilter('before match after', null))).toEqual('before match after');
+    expect($sce.getTrustedHtml(highlightFilter('before match after', undefined))).toEqual('before match after');
   });
 
-  it('issue 316 - should work correctly for regexp reserved words', function() {
-    expect(highlightFilter('before (match after', '(match')).toEqual('before <strong>(match</strong> after');
+  it('issue 316 - should work correctly for regexp reserved words', function () {
+    expect($sce.getTrustedHtml(highlightFilter('before (match after', '(match'))).toEqual('before <strong>(match</strong> after');
   });
 
-  it('issue 1777 - should work correctly with numeric values', function() {
-    expect(highlightFilter(123, '2')).toEqual('1<strong>2</strong>3');
+  it('issue 1777 - should work correctly with numeric values', function () {
+    expect($sce.getTrustedHtml(highlightFilter(123, '2'))).toEqual('1<strong>2</strong>3');
   });
+
+  it('should show a warning when this component is being used unsafely', function() {
+    highlightFilter('<i>before</i> match after', 'match');
+    expect(logSpy).toHaveBeenCalled();
+  });
+
 });
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -1,4 +1,4 @@
-angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap.bindHtml'])
+angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position'])
 
 /**
  * A helper service that can parse typeahead's syntax (string provided by users)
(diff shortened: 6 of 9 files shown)


src/typeahead/typeahead.js
```diff
@@ -399,13 +399,30 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
 
     function escapeRegexp(queryToEscape) {
-      return queryToEscape.replace(/([.?*+^$[\]\\(){}|-])/g, '\\$1');
+      return queryToEscape.replace(/([.?*+^$[\]\\(){}|-])/g, '\\$1'); // Regex: capture the whole query string and replace it with the string that will be used to match the results, for example if the capture is "a" the result will be \a
     }
 
     return function(matchItem, query) {
-      return query ? ('' + matchItem).replace(new RegExp(escapeRegexp(query), 'gi'), '<strong>$&</strong>') : matchItem;
+      if($sanitize) { // Is the $sanitize service available?
+        if(/<.*>/g.test(matchItem)){
```
chrisirhc: I think this check isn't required. It's safe to $sanitize something even if it doesn't contain any HTML tags.

src/typeahead/test/typeahead-highlight.spec.js
```diff
@@ -1,39 +1,50 @@
...
+    logSpy = spyOn($log, 'warn');
+  }));
+
   beforeEach(inject(function (typeaheadHighlightFilter) {
     highlightFilter = typeaheadHighlightFilter;
   }));
 
   it('should higlight a match', function () {
-    expect(highlightFilter('before match after', 'match')).toEqual('before <strong>match</strong> after');
+    expect(highlightFilter('before match after', 'match').$$unwrapTrustedValue()).toEqual('before <strong>match</strong> after');
```
chrisirhc: Use `$sce.getTrustedHtml(highlightFilter('before match after', 'match'))` instead

src/typeahead/typeahead.js
```diff
@@ -399,13 +399,30 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+    try {
+      $sanitize = $injector.get('$sanitize'); //tries to inject the sanitize service
+    }
+    catch(e) {
+      $log.warn('Unsafe use of typeahead please use ngSanitize'); // Warn the user about the danger
+    }
 
     function escapeRegexp(queryToEscape) {
-      return queryToEscape.replace(/([.?*+^$[\]\\(){}|-])/g, '\\$1');
+      return queryToEscape.replace(/([.?*+^$[\]\\(){}|-])/g, '\\$1'); // Regex: capture the whole query string and replace it with the string that will be used to match the results, for example if the capture is "a" the result will be \a
```
chrisirhc: I think this comment can be moved to right before the declaration of function:
```
    // Regex…
    function escapeRegexp
```

src/typeahead/typeahead.js
```diff
@@ -399,13 +399,30 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
     }
 
     return function(matchItem, query) {
-      return query ? ('' + matchItem).replace(new RegExp(escapeRegexp(query), 'gi'), '<strong>$&</strong>') : matchItem;
+      if($sanitize) { // Is the $sanitize service available?
+        if(/<.*>/g.test(matchItem)){
+          matchItem = $sanitize(matchItem); // If it's present the string gets sanitized and packed in a $sce object.
+        }
+      }
+      matchItem = query? ('' + matchItem).replace(new RegExp(escapeRegexp(query), 'gi'), '<strong>$&</strong>') : matchItem; // Replaces the capture string with a the same string inside of a "strong" tag
```
chrisirhc: The output from the `replace` isn't sanitized. Perhaps do this before calling `$sanitize` on the item.

chrisirhc:
Did a review. Good work here. Also, PR needs rebasing.

fernando-sendMail:
Hi @chrisirhc 
Thanks for your feedback, I polished the code according to your comments. I already pushed the commit.

karianna:
Can't be merged at present

fernando-sendMail:
@karianna I'm sorry, I'm kind of new to this, can I or should I do something about that?

karianna:
@fernando-sendMail you'll need to rebase, e.g. do a `git pull angular-ui/bootstrap master` (I'm assuming you've aliased the angular-ui bootstrap remote), resolve any conflicts and `git push` into your remote branch again.

fernando-sendMail:
@karianna Thanks for the info!

karma.conf.js
```diff
@@ -18,6 +18,7 @@ module.exports = function(config) {
       'misc/test-lib/jquery-1.8.2.min.js',
       'node_modules/angular/angular.js',
       'node_modules/angular-mocks/angular-mocks.js',
+      'misc/test-lib/angular-sanitize.js',
```
chrisirhc: Should change this to use `node_modules/angular-sanitize/angular-sanitize.js` and modify `package.json` to include angular-sanitize as a module since we changed this in https://github.com/angular-ui/bootstrap/commit/5536e9660e6653ad5e780ce99692789d7bed6fcc so that we don't need to include the angular files in this repository.

misc/demo/index.html
```diff
@@ -14,6 +14,7 @@
     <script src="//ajax.googleapis.com/ajax/libs/angularjs/<%= ngversion %>/angular.min.js"></script>
     <script src="//ajax.googleapis.com/ajax/libs/angularjs/<%= ngversion %>/angular-animate.min.js"></script>
     <script src="//ajax.googleapis.com/ajax/libs/angularjs/<%= ngversion %>/angular-touch.min.js"></script>
+    <script src="//ajax.googleapis.com/ajax/libs/angularjs/<%= ngversion %>/angular-sanitize.js"></script>
```
chrisirhc: I think you missed adding ngSanitize into the `demo/app.js`. Is this change needed for the demo to work? If not, we can just leave out this change for now and add it later when needed.

src/typeahead/typeahead.js
```diff
@@ -399,13 +399,29 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+      // Regex: capture the whole query string and replace it with the string that will be used to match 
+      // the results, for example if the capture is "a" the result will be \a
       return queryToEscape.replace(/([.?*+^$[\]\\(){}|-])/g, '\\$1');
     }
 
     return function(matchItem, query) {
-      return query ? ('' + matchItem).replace(new RegExp(escapeRegexp(query), 'gi'), '<strong>$&</strong>') : matchItem;
+      matchItem = query? ('' + matchItem).replace(new RegExp(escapeRegexp(query), 'gi'), '<strong>$&</strong>') : matchItem; // Replaces the capture string with a the same string inside of a "strong" tag
+      if($sanitize) { // Is the $sanitize service available?
+        matchItem = $sanitize(matchItem); // If it's present the string gets sanitized and packed in a $sce object.
```
chrisirhc: There's no need to call `$sanitize` here as ng-bind-html will automatically call it if it's not a trusted value.

src/typeahead/typeahead.js
```diff
@@ -399,13 +399,29 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
 
     function escapeRegexp(queryToEscape) {
+      // Regex: capture the whole query string and replace it with the string that will be used to match 
+      // the results, for example if the capture is "a" the result will be \a
       return queryToEscape.replace(/([.?*+^$[\]\\(){}|-])/g, '\\$1');
     }
 
     return function(matchItem, query) {
-      return query ? ('' + matchItem).replace(new RegExp(escapeRegexp(query), 'gi'), '<strong>$&</strong>') : matchItem;
+      matchItem = query? ('' + matchItem).replace(new RegExp(escapeRegexp(query), 'gi'), '<strong>$&</strong>') : matchItem; // Replaces the capture string with a the same string inside of a "strong" tag
```
chrisirhc: Sorry, on further thought, I realized your initial logic to check for html tags should be used here like:
```js
if (containsHtml(matchItem) && !$sanitize) {
    $log.warn('Unsafe use of typeahead please use ngSanitize');
}
```

This way, the user will only be warned when using typeahead with labels that contain HTML.

src/typeahead/typeahead.js
```diff
@@ -399,13 +399,29 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
 
-  .filter('typeaheadHighlight', function() {
+  .filter('typeaheadHighlight', ['$sce', '$injector', '$log', function($sce, $injector, $log) {
+
+    var $sanitize;
+    try {
+      $sanitize = $injector.get('$sanitize'); //tries to inject the sanitize service
+    }
+    catch(e) {
+      $log.warn('Unsafe use of typeahead please use ngSanitize'); // Warn the user about the danger
```
chrisirhc: Shouldn't log here, instead, log below only when the matchItem contains HTML. Users should only see this message if the labels contain HTML tags.

chrisirhc:
Thanks @fernando-sendMail for your work on this, I did some further review.

chrisirhc:
Also note, should remove dependency on `ui.bootstrap.bindHtml` module as it's no longer used after this change ( :+1: hooray!)

fernando-sendMail:
@chrisirhc Thanks for the review, I'll be working on this over the week.

fernando-sendMail:
Hey @chrisirhc I committed the latest stuff with the fixes and modifications to the test.

src/typeahead/typeahead.js
```diff
@@ -399,13 +399,34 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
     };
   }])
 
-  .filter('typeaheadHighlight', function() {
+  .filter('typeaheadHighlight', ['$sce', '$injector', '$log', function($sce, $injector, $log) {
+
+    var $sanitize;
+    try {
+      $sanitize = $injector.get('$sanitize'); //tries to inject the sanitize service
```
chrisirhc: Since you're just checking for the existence, you can do `$injector.has('$sanitize')`. Also, probably can call this `hasSanitize` instead.

chrisirhc:
Took another look. I saw that you updated `package.json` but the `angular-sanitize.js` is still in your PR and `karma.conf.js` hasn't been updated to use the angular-sanitize.js from node_modules.

fernando-sendMail:
Hey @chrisirhc I think it's more convenient to name it ` isSanitizePresent ` given the use it has in this context.

I made a commit with the other changes that you noticed.

chrisirhc:
Looks good to me. I'm scheduling this for 0.13.1 because we already have a lot of changes in 0.13.0 .

src/typeahead/test/typeahead-highlight-ngsanitize.spec.js
```diff
@@ -0,0 +1,19 @@
+describe('Security concerns', function() {
+  var highlightFilter, 
+  $sanitize,
+  logSpy;
```
chrisirhc: Indentation

fernando-sendMail:
Hey @chrisirhc It's been a while since I committed this.

Do I need to do something else?

wesleycho:
This just needs to be rebased off of current `master` I believe - this LGTM, lot of good work here.

fernando-sendMail:
Hey @wesleycho I'm having some trouble rebasing, I'm running:

`git rebase master`

and I resolve all the conflicts but, seems like it's the same result every time, is there a step that I'm missing?

wesleycho:
Your merge commit might have messed things up a bit here - I will try to fix this locally and get this in.

wesleycho:
Can you fix the broken test first?

wesleycho:
@fernando-sendMail can you open a new PR with fixed history? It seems like something is broken with the history in this one, and I am having a bit of trouble untangling the conflicts.

fernando-sendMail:
@wesleycho Sure, I'll work on that.

fernando-sendMail:
Hey @wesleycho I think I resolved all the conflicts, the PR is ready for review again

wesleycho:
Can you squash into one commit @fernando-sendMail ?

package.json
```diff
@@ -1,38 +1,37 @@
 {
   "author": "https://github.com/angular-ui/bootstrap/graphs/contributors",
   "name": "angular-ui-bootstrap",
-  "version": "0.13.4-SNAPSHOT",
+  "version": "0.13.0-SNAPSHOT",
```
wesleycho: Change this back to 4

wesleycho: Revert all changes to this file excepting the addition of `angular-sanitize`

package.json
```diff
@@ -1,38 +1,37 @@
...
   "repository": {
     "type": "git",
     "url": "https://github.com/angular-ui/bootstrap.git"
   },
   "devDependencies": {
-    "angular": "^1.4.3",
-    "angular-mocks": "^1.4.3",
+    "angular": "^1.3.15",
+    "angular-mocks": "^1.3.15",
+    "angular-sanitize": "^1.3.15",
```
wesleycho: Bump up to `^1.4.3`

template/typeahead/typeahead-match.html
```diff
@@ -1 +1 @@
-<a href tabindex="-1" bind-html-unsafe="match.label | typeaheadHighlight:query"></a>
+<a tabindex="-1" ng-bind-html="match.label | typeaheadHighlight:query"></a>
```
wesleycho: Keep the empty `href` - this is for accessibility.

wesleycho:
With these minor reverts/changes, this should be good to merge. Squashing the commits makes it easier to replay the change on top of `master` when we merge it, which is very useful for us.

fernando-sendMail:
Hey @wesleycho I tried using `git rebase -i <commit to squash>` but that didn't worked for me, it created an aditional commit instead, am I doing something wrong?

wesleycho:
You can do `git rebase -i HEAD~[number of commits without brackets]`

wesleycho:
Keep in mind once you do this, you will have to force push to the remote to overwrite the history with the new version.

fernando-sendMail:
@wesleycho Finally I've made it, it's ready for review again.

wesleycho:
Thanks a lot for sticking with it through the time, this is good work!

fernando-sendMail:
No problem, thanks a lot for the support guys!

PR #3464, opened 2015-03-29
chore(rating): remove redundant validation and add unit tests

imagentleman:
Changes:

1. Unit tests for the onKeyDown method were missing (https://github.com/angular-ui/bootstrap/blob/master/src/rating/rating.js#L62) in order to get a 100% coverage.

2. Angular's HTML compiler throws an error (https://code.angularjs.org/1.2.28/docs/error/$compile/ctreq) if the required controllers are not found (the rating directive has ngModel listed as the second one on https://github.com/angular-ui/bootstrap/blob/master/src/rating/rating.js#L73). The error prevents the link method from being executed (i.e ngModel is not provided on the HTML) and the ngModelCtrl variable can never be falsy.

Working Plunker with the changes (line 3122 of ui-bootstrap.js shows the validation removed):
http://plnkr.co/edit/68xfeULwQhaHeq3W5mD1?p=preview

Plunker with the current version of the rating directive:
http://plnkr.co/edit/XBR4r3OuDSknWwgvpjPE?p=preview

Both behave the same (with and without the validation).


pkozlowski-opensource:
The quality PRs like this make me smile. Thnx @imagentleman !

imagentleman:
Wow, thanks Pawel!

realityking:
Alternatively, ngModelController could be made optional. In the past I've found that less surprising for developers.

wesleycho:
Thanks for this @imagentleman  - as @pkozlowski-opensource said, this is good work.

PR #3465, opened 2015-03-30
fix(modal): fix closing modal with click backdrop in ie8

kotaroshima:
Closing modal directive by clicking on backdrop doesn't work in IE8, since it uses `event.currentTarget` which IE8 doesn't support.

I know that this project does not officially support IE8, but my project consumes this directive and needs to support IE8.


karianna:
Unfortunately we're not supporting IE8

PR #3469, opened 2015-03-30
chore(tests): update to jasmine 2.x

### LINKED ISSUE #3458: Upgrade to Jasmine 2.2
Now using old version of karma-jasmine because our tests need some work to update to Jasmine 2.2 .

Foxandxss:
Jasmine updated to its latest version, mostly changes on the matchers.

This closes #3458


src/typeahead/test/typeahead.spec.js
```diff
@@ -620,8 +636,6 @@ describe('typeahead tests', function () {
 
       // Note that this bug can only be found when element is in the document
       $document.find('body').append(element);
-      // Extra teardown for this spec
-      this.after(function () { element.remove(); });
```
Foxandxss: Jasmine 2 doesn't allow after hooks inside `it` but they are not needed.

chrisirhc: They need to be removed somehow, so perhaps setting it on a global variable and removing it on afterEach if the variable is set?

src/buttons/test/buttons.spec.js
```diff
@@ -201,7 +201,7 @@ describe('buttons', function () {
         expect(btns.eq(1)).not.toHaveClass('active');
 
         btns.eq(0).click();
-        expect($scope.model).toEqual(undefined);
+        expect($scope.model).toBeNull();
```
chrisirhc: Shouldn't this be toBeUndefined? Or is it actually null?

src/typeahead/test/typeahead.spec.js
Foxandxss: I put both test inside a describe to do that, I got the same behavior (tests passing).

src/buttons/test/buttons.spec.js
Foxandxss: It says "undefined is not equal to null" so it is null. I guess Jasmine is more picky with undefined/null.

src/typeahead/test/typeahead.spec.js
chrisirhc: Yep, tests will always pass but this is a DOM leak, so if you keep the tests running a long time, the browser will crash.

Foxandxss: sounds good to me, I'll fix that

src/buttons/test/buttons.spec.js
chrisirhc: Yep, it is. Ok, I hope this isn't a bug. :smile:

misc/test-lib/helpers.js
```diff
@@ -1,25 +1,33 @@
...
-      this.message = function () {
-        return 'Expected \'' + angular.mock.dump(this.actual) + '\' to have focus';
-      };
-
-      return document.activeElement === this.actual[0];
+    toHaveFocus: function(util, customEqualityTesters) {
+      return {
+        compare: function(actual, expected) {
+          return {
+            pass: util.equals(document.activeElement, actual[0], customEqualityTesters)
```
chrisirhc: This is a reference check so we should probably use `===` instead of the utils equals.

src/modal/test/modal.spec.js
```diff
@@ -27,60 +27,91 @@ describe('$modal', function () {
...
-        var modalDomEls = this.actual.find('body > div.modal');
-        return modalDomEls.length === noOfModals;
-      },
+            var result = {
+              pass: modalDomEls.css('display') === 'block' && contentToCompare.html() === content
+            };
 
-      toHaveBackdrop: function() {
+            if (!result.pass) {
+              result.message = '"Expected "' + angular.mock.dump(modalDomEls) + '" to be open with "' + content + '".';
```
chrisirhc: There needs to a message for when the pass is true because the message is displayed when the condition is used with `.not`. Like if `.not.toHaveModalOpenWithContent` fails.

src/position/test/position.spec.js
```diff
@@ -16,18 +16,26 @@ describe('position elements', function () {
...
-      toBePositionedAt: function(top, left) {
-        this.message = function() {
-          return 'Expected "('  + this.actual.top + ', ' + this.actual.left +  ')" to be positioned at (' + top + ', ' + left + ')';
+    jasmine.addMatchers({
+      toBePositionedAt: function(util, customEqualityTesters) {
+        return {
+          compare: function(actual, top, left) {
+            var result = {
+              pass: util.equals(actual.top, top, customEqualityTesters) &&
+                      util.equals(actual.left, left, customEqualityTesters)
```
chrisirhc: Probably shouldn't convert these to `util.equals` either. Since they're supposed to be normal equality comparisons. I'm not sure if there's a good reason to use `util.equals` in these cases.

Foxandxss:
I updated with some of the changes. Will update the others as well, tomorrow I guess.

src/position/test/position.spec.js
Foxandxss: All the util.equals are working perfectly tho. I thought it was the recommended way by jasmine.

chrisirhc:
Good stuff. Thanks!

src/modal/test/modal.spec.js
Foxandxss: Will check. When I researched this new custom matchers for my book, I didn't see the utility on the "true" messages.

src/buttons/test/buttons.spec.js
wesleycho: IIRC, `toEqual` is the equivalent of `==`, so it could just be a not specific enough test.

Foxandxss:
I updated the matchers. About the util.equal, they work correctly using the jasmine algorithm.

src/buttons/test/buttons.spec.js
chrisirhc: Yep it is, I'm aware of that. I'm just questioning the intention of the test.

misc/test-lib/helpers.js
```diff
@@ -1,25 +1,57 @@
...
+        compare: function(actual, expected) {
+          var result = {
+            pass: actual.hasClass(expected)
+          };
 
-      return this.actual.hasClass(cls);
+          if (result.pass) {
+            result.message = 'Expected ' + actual + ' not to have the ' + expected + ' class.';
+          } else {
+            result.message = 'Expected ' + actual + ' to have the ' + expected + ' class.';
```
chrisirhc: Fyi, you can do `result.pass ? 'not ' : ''` if you want to save some lines (as above, in the original test). Not important here.

src/position/test/position.spec.js
chrisirhc: I think `util.equals` is better if it's comparison of an things that are of unknown types. It allows the user to specify custom equality testers and use them.
However, since we don't rely on any custom equality handling, I wonder if it's just a level of unnecessary indirection. @wesleycho any thoughts?

misc/test-lib/helpers.js
Foxandxss: Less lines but harder to read something like:

```result.message = 'Expected ' + actual + result.pass ? ' not ' : ' ' + to have the ' + expected + ' class.';```

chrisirhc: With parentheses and whitespace, I think it's reasonable. See the original message above.
Also note that there were quotes (`'`) around the `actual` and `expected` before this change.

chrisirhc: Again, this isn't important. Just commenting on style. If you think it's more readable, I'm fine with that. A little bit more concerned about the quotes, but also not a big issue.

wesleycho:
Can you update with current master and see if the new stuff needs to be converted over as well?

I am willing to do another review today and merge this in once it is up to date.

Foxandxss:
All is fine.

src/position/test/position.spec.js
wesleycho: It took me a bit of searching to figure this out, since it isn't on the introduction page for Jasmine.

Looking into it a little more, I think this is ok. My first instinct was that this should be changed to the simpler version, but in practice, users who are contributing wouldn't be modifying this. The Jasmine documentation in the source code mentions that this is a modification of Underscore's `isEqual` function at its heart.

wesleycho:
I'm ok with this PR as is - @chrisirhc, thoughts?

chrisirhc:
Looks good to me.

PR #3470, opened 2015-03-30
chore(travis): use latest node

Foxandxss:
Let's update travis like other libraries does :)


wesleycho:
Just to be clear - this will run everything in iojs & node simultaneously?

Foxandxss:
Yup, that would make travis test using a node 0.12 machine and a iojs machine.

realityking:
Since ui-bootstrap doesn't produce any node modules, what is the advantage of running the tests in two different node environments?

This will double the build time, which might be undesireable.

Foxandxss:
Yeah, that is right. I changed it just to bump node version. Even when we don't release node modules, we have lot of dev dependencies which could break with 0.10.

PR #3476, opened 2015-03-31
chore(Gruntfile): Add commonjs package manager support

oznu:
Adds support for commonjs/browserify when installed using NPM and allows injecting the module into an angular application like this:

````
var angular = require('angular');
angular.module('myApp', [
  require('angular-animate'),
  require('angular-ui-router'),
  require('angular-bootstrap'),
]);
````
Implements same method used in https://github.com/angular/angular.js/pull/10732 and angular-ui-router.

### CODE REVIEW

File: Gruntfile.js [hunk 1]
```diff
@@ -58,7 +62,7 @@ module.exports = function(grunt) {
     concat: {
       dist: {
         options: {
-          banner: '<%= meta.banner %><%= meta.modules %>\n',
+          banner: '<%= meta.banner %>\n<%= meta.modules %>\n',
```
wesleycho: Is there any particular reason for this change?

Gruntfile.js
```diff
@@ -66,7 +70,7 @@ module.exports = function(grunt) {
       },
       dist_tpls: {
         options: {
-          banner: '<%= meta.banner %><%= meta.all %>\n<%= meta.tplmodules %>\n',
+          banner: '<%= meta.banner %>\n<%= meta.all %>\n<%= meta.tplmodules %>\n',
```
wesleycho: Ditto

wesleycho:
This looks good, but upon discussion, we are scheduling this for 0.13.1 - we are close to releasing 0.13.0.

chrisirhc:
Before merging this, we should think about whether to publish the templates as a separate module (e.g. ui-bootstrap-tpls). Otherwise, with this change, users can't specify in the code whether they need the templates or not.

It may also make sense instead to publish the default ui-boostrap module with templates included and add another module called ui-boostrap-notpls .

eranimo:
Any update on this? I'm surprised this package, given it's popularity, already doesn't have Browserify support

ankurp:
Lets merge this to support browserify

nl0:
browserify ftw!

oznu:
This has been resolved with https://github.com/angular-ui/bootstrap-bower/commit/c04f8784b8558a821f55c5ac6a8a6c205148f9f1

```
npm install angular-ui-bootstrap
```

```javascript
var angular = require('angular');
angular.module('myApp', [
  require('angular-animate'),
  require('angular-ui-router'),
  require('angular-ui-bootstrap'),
]);
```

PR #3478, opened 2015-03-31
feat(datepicker): Allow specifying next and previous button icon classes

maxfierke:
I'm a fan of Font-Awesome over Glyphicons, so I use Font-Awesome everywhere else in one of the apps I work on but found it detrimental to the user (many of whom are mobile) to download an extra set of icons and CSS when they'd only be used in a few components (datepicker and timepicker).

I'm wondering if this is a feature that would be interesting to anyone else. I realize I could do the same thing by providing different templates, but I'd prefer not to have to maintain a seperate set of datepicker templates for a simple class difference.

*Note: I've got a similar patch lined up for Timepicker too, but I wanted to get feedback on this implementation first and clean it up as needed, then redo the timepicker patch similarly.*

### DIFF
--- a/src/datepicker/test/datepicker.spec.js
+++ b/src/datepicker/test/datepicker.spec.js
@@ -1818,8 +1818,10 @@ describe('datepicker directive', function () {
   });
 
   describe('with empty initial state', function () {
-    beforeEach(inject(function() {
+    beforeEach(inject(function(datepickerConfig) {
       $rootScope.date = null;
+      $rootScope.nextIconClass = datepickerConfig.nextIconClass;
+      $rootScope.prevIconClass = datepickerConfig.prevIconClass;
       element = $compile('<datepicker ng-model="date"></datepicker>')($rootScope);
       $rootScope.$digest();
     }));
@@ -1830,6 +1832,17 @@ describe('datepicker directive', function () {
 
     it('is shows rows with days', function() {
       expect(element.find('tbody').find('tr').length).toBeGreaterThan(3);
+    });
+
+    it('has next and previous icons', function () {
+
+      expect($rootScope.nextIconClass).toEqual('glyphicon glyphicon-chevron-right');
+      expect($rootScope.prevIconClass).toEqual('glyphicon glyphicon-chevron-left');
+
+      var nextIconClassSelector = $rootScope.nextIconClass.split(' ').map(function (c) { return '.' + c; }).join('');
+      var prevIconClassSelector = $rootScope.prevIconClass.split(' ').map(function (c) { return '.' + c; }).join('');
+      expect(element.find('thead').find(nextIconClassSelector).length).toBe(1);
+      expect(element.find('thead').find(prevIconClassSelector).length).toBe(1);
     });
 
     it('sets default 00:00:00 time for selected date', function() {
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -15,7 +15,9 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   yearRange: 20,
   minDate: null,
   maxDate: null,
-  shortcutPropagation: false
+  shortcutPropagation: false,
+  nextIconClass: 'glyphicon glyphicon-chevron-right',
+  prevIconClass: 'glyphicon glyphicon-chevron-left'
 })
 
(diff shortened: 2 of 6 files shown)


src/datepicker/test/datepicker.spec.js
```diff
@@ -1832,6 +1834,17 @@ describe('datepicker directive', function () {
       expect(element.find('tbody').find('tr').length).toBeGreaterThan(3);
     });
 
+    it('has next and previous icons', function () {
+
+      expect($rootScope.nextIconClass).toEqual('glyphicon glyphicon-chevron-right');
+      expect($rootScope.prevIconClass).toEqual('glyphicon glyphicon-chevron-left');
+
+      var nextIconClassSelector = $rootScope.nextIconClass.split(' ').map(function (c) { return '.' + c; }).join('');
+      var prevIconClassSelector = $rootScope.prevIconClass.split(' ').map(function (c) { return '.' + c; }).join('');
```
maxfierke: The Rubyist in me likes this, but the every other part of me thinks this is gross.

rvanbaalen:
Why don't you just overwrite the day, month and year picker templates to achieve your goal? 
The power of directives with external templates is that you can overwrite them and customize them to your own needs.

src/datepicker/test/datepicker.spec.js
```diff
@@ -1832,6 +1834,17 @@ describe('datepicker directive', function () {
       expect(element.find('tbody').find('tr').length).toBeGreaterThan(3);
     });
 
+    it('has next and previous icons', function () {
+
+      expect($rootScope.nextIconClass).toEqual('glyphicon glyphicon-chevron-right');
+      expect($rootScope.prevIconClass).toEqual('glyphicon glyphicon-chevron-left');
+
+      var nextIconClassSelector = $rootScope.nextIconClass.split(' ').map(function (c) { return '.' + c; }).join('');
```
rvanbaalen: Wouldn't this be much more readable?

```
var nextIconClassSelector = '.' + $rootScope.nextIconClass.split(' ').join('.');
```

maxfierke: I think that depends upon how you feel about `map` :stuck_out_tongue:. That being said, I think the best for readability would to just use `".glyphicon.glyphicon-chevron-right"` despite the repeatition.

maxfierke:
@rvanbaalen That definitely is a solution, but I'd prefer to not have a seperate copy of some templates floating around in my project for a simple class replacement.

rvanbaalen:
As much as I appreciate your efforts for creating this PR + corresponding tests, I must say that is part of [the philosophy of this project](https://github.com/angular-ui/bootstrap/wiki/Project-Philosophy#customisability) to keep it simple and provide customisability for users by letting them overwrite our default provided templates.

Therefore I'm afraid this PR won't make it to master. I will let this cross the minds of the rest of the team first though. @Foxandxss and @chrisirhc what do you think?

Foxandxss:
That is why we let users override the templates, so they can swap it with their own version if they need it.

If we start adding opinions to the templates, we will grow a monstrosity.

On the other hand, please, datepicker is enough complicated to push more options to it ;( :P

src/datepicker/test/datepicker.spec.js
rvanbaalen: Indeed, that would be better in my opinion.

Regarding the readability / preference; let's let the benchmarks speak for themselves:
http://jsperf.com/map-vs-split-and-join

:wink:

maxfierke:
@rvanbaalen @Foxandxss Fair enough. I'll close this out then. Thanks for taking the time to review it! :smile:

rvanbaalen:
No problem @maxfierke thanks for taking the time and effort to create a PR

chrisirhc:
I agree. Just wanted to add: If the changes can be applied purely through the template, it can actually standalone as its own module in a gist or its own repository, sort of like a plugin.

rvanbaalen:
Thats actually a good idea @chrisirhc. @maxfierke if you could wrap this up in a module and make it for instance bower installable, others who want to use font awesome can easily install your customization.

maxfierke:
@rvanbaalen @chrisirhc ahhh! Good idea. I'll look into doing that.

maxfierke:
[maxfierke/ui-bootstrap-fontawesome](https://github.com/maxfierke/ui-bootstrap-fontawesome) for interested parties.

rvanbaalen:
That was fast ;-)

PR #3482, opened 2015-04-02
fix(popover): make it work with ngAnimate

### LINKED ISSUE #3375: ngAnimate breaks fade-in animation of popover

chrisirhc:
Fixes #3375

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -421,6 +421,27 @@ function ($animate ,  $sce ,  $compile ,  $templateRequest) {
   };
 }])
 
+// Apply animate class without animating itself
+.directive('tooltipAnimateClass', function () {
+  return {
+    restrict: 'A',
+    link: function (scope, element, attrs) {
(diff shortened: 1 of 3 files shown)


wesleycho:
The previous value in the template was a binding - is this change ok as an immediately run class add?

Does this work when the user does not use ngAnimate?

chrisirhc:
Yep it works, it didn't work as a binding because ngAnimate attempts to animate the animation class (`fade`) along with the `in` class at the same time. This PR ensures that the `fade` class is already there when the `in` class is added.

chrisirhc:
Actually, good question, on further thought, I'll detect that if ngAnimate doesn't exist, skip adding the animation classes.

wesleycho:
This good for merging then?

chrisirhc:
Yep it is. Does fix the popover issues.

PR #3488, opened 2015-04-04
fix(pagination): paging buttons stay focused when page is changed #3486

### LINKED ISSUE #3486: pagination focusing on page that hasn't been clicked
I've noticed that when working with the pagination directive that the focus style will be applied to a page button that hasn't been clicked. Versions I'm using are below.

- angular 1.2.27
- ui-boostrap 0.12.1
- bootstrap 3.3.4

This only happens when you have max-items set to less than the number of total pages. You'll see it when you click a page that shifts the visible range.

To demonstrate the issue I have a Plunk. 
http://plnkr.co/edit/z8a5lNkvtR7zo9gVqU9e?p=preview

Click page 2 and you should see that it shifts and highlights page 2 as being active but page 1 is now focused (cause it'll go away if you click outside the pager).

image

benthepoet:
This fixes #3486. Just needed to pass in the $event to blur the page button after it has been clicked. Page numbers were sticking upon selection because "track by $index" with ngRepeat keeps the DOM from destroying and recreating elements most of the time. Direction links were also affected along with pager template as well.


karianna:
Hi @benthepoet, could you add some tests for this?

src/pagination/pagination.js
```diff
@@ -32,8 +32,11 @@ angular.module('ui.bootstrap.pagination', [])
     $scope.page = parseInt(ngModelCtrl.$viewValue, 10) || 1;
   };
 
-  $scope.selectPage = function(page) {
+  $scope.selectPage = function(page, evt) {
     if ( $scope.page !== page && page > 0 && page <= $scope.totalPages) {
+      if (evt && evt.target) {
```
wesleycho: Will this ever be called without an event object?

benthepoet: Yes, below is the only spot where it gets called without an event.

https://github.com/benthepoet/bootstrap/blob/pagination-fix/src/pagination/pagination.js#L59-L68

wesleycho:
Thanks for this work - this will make it in the next release.

PR #3494, opened 2015-04-05
fix(datepicker): datepicker-popup compatibility with ngModelOptions

### LINKED ISSUE #3349: Datepicker selection off by one selection for display from ngModel with ng-model-options
So I have an issue with the datepicker popup when I have `ng-model-options` present on the element, I have been able to replicate in a plunker [here](http://embed.plnkr.co/AI5SJ9WaqpbSixtvBFJi/preview) and able to confirm that removing the ngModelOptions does indeed restore normal display of the actual active date.

What seems to be happening on a auto-closing datepicker is that the model is getting updated after the datepicker leaves scope, because of the way the mgModelOptions are delaying the model update until it is too late(the datepicker is closed already).

I think it is the `default blur` or the `debounce` I have provided, but I don't see any reason why I wouldn't be able to have them on there for when a user manually types a date in. I found that just wrapping the `self.activeDate` assignment like so: `$timeout(function () { self.activeDate = date; });` in `$scope.select` function in the `DatepickerController` does seem to ensure that the displayed selection stays in sync on every change, since it is delaying the update of the active selection until the digest cycle that the model update is happening in, the demo shows it broken, but I included a version of the datepicker.js with the fix, so you can switch and confirm that it does fix it. But there are probably a couple of other options as well to ensuring that the displayed date gets properly updated.

image

chrisirhc:
Fixes #3349

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -476,9 +476,15 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
         return scope[key + 'Text'] || datepickerPopupConfig[key + 'Text'];
       };
 
-      attrs.$observe('datepickerPopup', function(value) {
-          dateFormat = value || datepickerPopupConfig.datepickerPopup;
-          ngModel.$render();
+      dateFormat = attrs.datepickerPopup || datepickerPopupConfig.datepickerPopup;
+      attrs.$observe('datepickerPopup', function(value, oldValue) {
+          var newDateFormat = value || datepickerPopupConfig.datepickerPopup;
+          // Invalidate the $modelValue to ensure that formatters re-run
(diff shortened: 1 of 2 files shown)


src/datepicker/datepicker.js
```diff
@@ -540,28 +548,47 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
...
+        if (!value) {
+          return true;
+        } else if (angular.isDate(value) && !isNaN(value)) {
+          return true;
+        } else if (angular.isString(value)) {
+          var date = dateParser.parse(value, dateFormat) || new Date(value);
+          if (isNaN(date)) {
+            return false;
+          } else {
+            return true;
```
wesleycho: This can be simplified by just doing `return !isNaN(date);`

wesleycho:
Excepting my one comment, this LGTM.

PR #3496, opened 2015-04-06
feat(tooltip): add tooltip-html, deprecate tooltip-html-unsafe

chrisirhc:
- [ ] Update documentation


darabos:
It would be nice to have documentation for `tooltip-html`. Thanks!

PR #3509, opened 2015-04-08
fix(popover): animations with ngAnimate

### LINKED ISSUE #3375: ngAnimate breaks fade-in animation of popover

### LINKED ISSUE #3506: tooltip - Issues with PR #3126
PR #3126 introduced the ability to support defining a class for the tooltip as an attribute.  This is causing issues with how the popover adds the placement and animation classes, this [plunker](http://plnkr.co/edit/hp2hIKSiZXP0iRU6lhkM?p=preview) demonstrates the issue.  You will notice the popover does not display the arrow correctly and the animation doesn't run when it opens and closes.  I've narrowed down the problem to the following line in the tooltip provider where the template variable is declared:

    'class="'+startSym+'class'+endSym+'" '+

The classes on the outer popover div are not getting added correctly.  The outer popover div should have:

    class="popover right fade in"

but instead have:

    class="popover popover in"

We should also reconsider using 'class' as a variable as that is a reserved word in ECMA 6.

chrisirhc:
- Animations didn't work because the class attribute was mangled during
  compilation due to the way class attributes are merged on directives with
  `replace: true` (happened 

- Rename attribute to popup-class

- Refactor to rename variables named "class" as it's a keyword
  and also looks weird in editors

Fixes #3375
Fixes #3506


wesleycho:
LGTM

chrisirhc:
Yep it's ready

Issue #3511, opened 2015-04-08
Issue in dropdown after angularjs update


snehatulsi:
On upgrading angularjs to latest version, dropdown toggle function behaves wrong.
The on-toggle function of dropdown should be called after the dropdown is toggled. But the opposite happens.

Reason, is the animate function in dropdown controller. 

```js
  scope.$watch('isOpen', function( isOpen, wasOpen ) {
    $animate \[isOpen ? 'addClass' : 'removeClass'\](self.$element, openClass);
      if (angular.isDefined(isOpen) && isOpen !== wasOpen) {
          toggleInvoker($scope, {
              open: !!isOpen
          });
      }
  });
```

Here the toggleInvoker should be inside the resolve function of animate, as animate happens in the next digest cycle.

snehatulsi:
To test this issue, take the dropdown example code in the official website , update the angularjs, and in the toggle function, print the width of the dropdown container. As the function is called before the dropdown opens, it will print zero.

karianna:
Pretty sure this is fixed on Master

snehatulsi:
I guess not, i took the latest stable version just now. Also the official website says, angularjs bootstrap is tested in angularjs 1.2.x, and angularjs is updated to 1.3.15. So bootstrap angular also needs an update. It would be great.

karianna:
@snehatulsi - by master I mean the HEAD of the master branch (the upcoming 0.13.0 release which will support angular 1.3)

chrisirhc:
Also, a Plunker to reproduce this would be very helpful for investigration.

snehatulsi:
A plunker for the same: 
http://plnkr.co/edit/HPUEsVcSMu2Z7A2bDVBB?p=preview

The example is from the official site. In the newer angularjs version, open state has the width as 0, and closed state has the width to 2xx px.
Change the angularjs version to 2.1.x, then the opposite happens , which is the correct behavior.

chrisirhc:
I see, you're saying that the toggle doesn't run at the correct time. It should run after the animation rather than before the animation. That sounds reasonable.

Adding your PR, #3513 for reference.

snehatulsi:
Yes thats the behavior in the previous versions. Also, after the dropdown is toggled in DOM, it should call the on-toggle function.

karianna:
@chrisirhc In the related PR you mentioned this should be fixed for the next release, but we've got the PR in the 0.13.x milestone....

PR #3512, opened 2015-04-08
Issue #3511 fixed

snehatulsi:
On upgrading angularjs to latest version, dropdown toggle function behaves wrong.
The on-toggle function of dropdown should be called after the dropdown is toggled. But the opposite happens.

Reason, is the animate function in dropdown controller.

scope.$watch('isOpen', function( isOpen, wasOpen ) {
$animate [isOpen ? 'addClass' : 'removeClass'](self.$element, openClass);
if (angular.isDefined(isOpen) && isOpen !== wasOpen) {
toggleInvoker($scope, {
open: !!isOpen
});
}
});
Here the toggleInvoker should be inside the resolve function of animate, as animate happens in the next digest cycle.


snehatulsi:
This needs some correction

PR #3513, opened 2015-04-08
Issue #3511 fixed

snehatulsi:
Call on-toggle function of dropdown in resolve function of $animate


chrisirhc:
It looks like with this change, you need to do `$browser.defer.flush()` in the dropdown.spec for the failing tests because the `toggleHandler` is now deferred.

rvanbaalen:
Since this related issue, #3511, is labeled Angular 1.3 I've added the Angular 1.3 label here as well.

chrisirhc:
@rvanbaalen not exactly, it's actually an ngAnimate issue but nonetheless we should try to have this fixed for next release.

chrisirhc:
@snehatulsi , could you please add `$browser.defer.flush()` in the test code so that the tests pass?

snehatulsi:
I don't know where is the test, and where can I add the code. Can you
please help me out.

chrisirhc:
No problem, the tests are located at: https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/test/dropdown.spec.js

You can see the failures when you run `grunt` on your command line once you've [prepared your environment](https://github.com/angular-ui/bootstrap/wiki/Development#prepare-your-environment).

snehatulsi:
@chrisirhc I tried adding the code. I got the error $browser not defined. I tried with modifications, but failed. I don't know the test specs, I have never done testing this way. Please mention where to inject $browser dependency, and where to insert code for $browser.defer.flush()

src/dropdown/test/dropdown.spec.js
```diff
@@ -1,12 +1,13 @@
 describe('dropdownToggle', function() {
-  var $compile, $rootScope, $document, element;
+  var $compile, $rootScope, $document, element, $browser;
 
   beforeEach(module('ui.bootstrap.dropdown'));
 
-  beforeEach(inject(function(_$compile_, _$rootScope_, _$document_) {
+  beforeEach(inject(function(_$compile_, _$rootScope_, _$document_, $browser) {
```
chrisirhc: Use `_$browser_` here

chrisirhc:
Added a comment. You need to add them just before the expect calls on the lines that they're failing.

snehatulsi:
@chrisirhc The test passed

karianna:
@snehatulsi Can you update this PR, it can no longer be merged against master

wesleycho:
Can you squash your commits & rebase off of master?

Issue #2688, opened 2014-09-11
appendTo option for modal dialog


mhaertwig:
What about a "appendTo" option for the modal dialog? 

In my application I would like to use the modal dialog in an element that can be set to fullscreen mode. However, since the modal dialog is appended to the body by default, the dialog disappears every time application's main container is set to fullscreen. Therefore an option to configure the modal dialog's parent element  (e.g. "appendTo") would be nice.

YidingW:
+1

icfantv:
Linking in #4772 but leaving this open for now.

wesleycho:
This has landed as 16d854ca4c70aefb6b03faeb2a20105ce2bea8ec in `master`.

PR #2942, opened 2014-11-07
feat(modal): Adds `defaultContainer` to `$modalProvider`

Lalem001:
Includes basic tests, with adjustments to `toHaveModalsOpen` and `toHaveBackdrop`.

Fixes: #2688

PR #3516, opened 2015-04-08
feat(modal): Adds `defaultContainer` to `$modalProvider`

Lalem001:
Includes basic tests, with adjustments to toHaveModalsOpen and toHaveBackdrop.

Fixes: #2688 
Replaces PR: #2942

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -215,22 +215,22 @@ angular.module('ui.bootstrap.modal', [])
 
       function removeModalWindow(modalInstance, elementToReceiveFocus) {
 
-        var body = $document.find('body').eq(0);
         var modalWindow = openedWindows.get(modalInstance).value;
+        var container = modalWindow.container;
 
         //clean up the stack
         openedWindows.remove(modalInstance);
 
         removeAfterAnimate(modalWindow.modalDomEl, modalWindow.modalScope, function() {
-          body.toggleClass(OPENED_MODAL_CLASS, openedWindows.length() > 0);
+          container.toggleClass(OPENED_MODAL_CLASS, openedWindows.length() > 0);
           checkRemoveBackdrop();
         });
 
-        //move focus to specified element if available, or else to body
+        //move focus to specified element if available, or else to modal container
         if (elementToReceiveFocus && elementToReceiveFocus.focus) {
           elementToReceiveFocus.focus();
         } else {
-          body.focus();
+          container.focus();
         }
       }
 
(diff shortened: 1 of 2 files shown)


marlonmleite:
+1

src/modal/modal.js
```diff
@@ -454,6 +456,7 @@ angular.module('ui.bootstrap.modal', [])
                 backdropClass: modalOptions.backdropClass,
                 windowClass: modalOptions.windowClass,
                 windowTemplateUrl: modalOptions.windowTemplateUrl,
+                container: $modalProvider.defaultContainer,
```
wesleycho: Looking at the original issue this PR references, any particular reason for the choice of `container` over `appendTo`? I think the latter makes more sense semantically, but am open to other options.

src/modal/test/modal.spec.js
```diff
@@ -73,9 +73,11 @@ describe('$modal', function () {
       toHaveModalOpenWithContent: function(util, customEqualityTesters) {
         return {
           compare: function(actual, content, selector) {
-            var contentToCompare, modalDomEls = actual.find('body > div.modal > div.modal-dialog > div.modal-content');
+            var container = (actual === $document) ? actual.find('body') : angular.element(actual);
```
wesleycho: Shouldn't `actual` already be a jqLite wrapped element? I believe `angular.element(actual)` should just be `actual` here.

src/modal/test/modal.spec.js
```diff
@@ -94,7 +96,9 @@ describe('$modal', function () {
       toHaveModalsOpen: function(util, customEqualityTesters) {
         return {
           compare: function(actual, expected) {
-            var modalDomEls = actual.find('body > div.modal');
+            var container = (actual === $document) ? actual.find('body') : angular.element(actual);
```
wesleycho: Ditto

src/modal/modal.js
Lalem001: `container` is just coming off of my use of `defaultContainer`. Would you want `appendTo` and `defaultContainer`? Or change both to `appendTo` and `defaultAppendTo`?

Or perhaps something else entirely?

wesleycho: It should be `appendTo` and there should be no `defaultContainer` reference.

wesleycho:
Should be noted, we now have an `appendTo` feature via 16d854ca4c70aefb6b03faeb2a20105ce2bea8ec - this feature should interop with the new implementation. Given that it takes a DOM node in the $modal service, I think this feature then becomes invalid. If one needs a sensible default, I recommend creating a modal abstraction on top of $modal to pass the desired defaults.

PR #3529, opened 2015-04-11
fix(dropdown): update unit tests

leyanlo:
Changing the value of dropdownConfig.openClass should not break the unit tests


chrisirhc:
Thank you! I adjusted the commit message a little as it only affects tests.

PR #3532, opened 2015-04-12
Update dropdown.js

redmagic:
Close dropdowns also on touchstart. Current version won't close dropdowns in webview.


karianna:
Is it possible to add a test for this?

redmagic:
Hallo Martijn,

Testing the touch event would be an e2e test, I'm not really comfortable 
writing those.

I could look into the current test to see how the test for the document 
click is done, but time is a bit lacking on my side.

Maybe we should convert this pull request to an issue of the dropdowns 
not closing on touch devices.

Regards,

Arjen

src/dropdown/dropdown.js
```diff
@@ -10,6 +10,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
   this.open = function( dropdownScope ) {
     if ( !openScope ) {
       $document.bind('click', closeDropdown);
+      $document.bind('touchstart', closeDropdown);
```
chrisirhc: You can instead add touchstart to the above binding like: `$document.bind('click touchstart', closeDropdown);`

chrisirhc:
Adding e2e testing would be a lot of work. Instead, for now, you can duplicate this test (https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/test/dropdown.spec.js#L59-L64) and trigger a `touchstart` event there to test that the dropdown closes when a `touchstart` event is triggered.

redmagic:
Closed, better PR incoming

Issue #2639, opened 2014-08-27
Timepicker Input - 00


Gils14:
When a user tries to input "00" into the Timepicker minutes field, it only allows one "0" to be typed.  While the 0 does then revert to 00 when the field is no longer in focus, this was frustrating for some of our users. The object we used is below

<code>&lt;timepicker class="time-picker" ng-model="order.appointment.datetime" hour-step="1" minute-step="15"required&gt;&lt;/timepicker&gt;</code>

wesleycho:
A PR for optional support for this would be welcome.

chrisirhc:
Just wondering, after merging #3533, is there still any action needed here for this issue?

karianna:
@Gils14 - Can you please try master and reopen if needed.

PR #3533, opened 2015-04-12
feat(timepicker): always pad minutes

### LINKED ISSUE #1598: Timepicker - leading zeros should not be deleted in the minutes field
This is bad for user experience - If I type 10:05, I look up and only see 10:5, and think i missed the 0, go back, try to add it, nothing happens.

Same thing when typing 10:00 - it should display that way in the field when I type.

wesleycho:
- Always pad minutes as a UX improvement

[Plunker](http://plnkr.co/edit/Dm1mcSn0XC7RuxKrp80p?p=preview)

Fixes #1598
Related to #2639

### DIFF
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -247,7 +247,9 @@ angular.module('ui.bootstrap.timepicker', [])
     }
 
     $scope.hours = keyboardChange === 'h' ? hours : pad(hours);
-    $scope.minutes = keyboardChange === 'm' ? minutes : pad(minutes);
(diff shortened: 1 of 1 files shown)


chrisirhc:
@wesleycho This creates an issue where I can't type "10:50" as the input now. Two digit minutes can't be typed into the display.

I think instead, the fix/feature should be something along:
```
    if (keyboardChange !== 'm') {
      $scope.minutes = pad(minutes);
    }
```

wesleycho:
Good catch - made change as suggested, works for both use cases.

PR #3540, opened 2015-04-14
feat(tabs): disallow deselect of tab in onDeselect

ljantzen:
Replaces pull request 3088

We had the need to disallow changing tabs according to business rules. This pull request implements the functionality where the return value (true/false) of onDeselect() determines if switching tabs is allowed or not.

### DIFF
--- a/src/tabs/tabs.js
+++ b/src/tabs/tabs.js
@@ -14,14 +14,25 @@ angular.module('ui.bootstrap.tabs', [])
       tabs = ctrl.tabs = $scope.tabs = [];
 
   ctrl.select = function(selectedTab) {
+	var allow=true;
+	var prevSelected;
     angular.forEach(tabs, function(tab) {
       if (tab.active && tab !== selectedTab) {
+		prevSelected = tab;
         tab.active = false;
-        tab.onDeselect();
+        var result = tab.onDeselect();
+        if (typeof result != 'undefined') {
+			allow = result;
+		}
       }
     });
-    selectedTab.active = true;
-    selectedTab.onSelect();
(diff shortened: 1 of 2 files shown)


src/tabs/tabs.js
```diff
@@ -14,14 +14,25 @@ angular.module('ui.bootstrap.tabs', [])
       tabs = ctrl.tabs = $scope.tabs = [];
 
   ctrl.select = function(selectedTab) {
+    var allow=true;
```
wesleycho: Fix spacing

src/tabs/tabs.js
```diff
@@ -14,14 +14,25 @@ angular.module('ui.bootstrap.tabs', [])
...
+    var prevSelected;
     angular.forEach(tabs, function(tab) {
       if (tab.active && tab !== selectedTab) {
+        prevSelected = tab;
         tab.active = false;
-        tab.onDeselect();
+        var result = tab.onDeselect();
+        if (typeof result != 'undefined') {
```
wesleycho: Needs a better check here.

wesleycho:
This also needs documentation change. This would be a breaking change, since a lot of users probably do not return any value, which would be `undefined` by default. Some may even return a value that may end up evaluating to false.

chrisirhc:
I don't think this should go into the per-tab logic. It should instead be on the tabset. I'm also not sure this is a good feature in core.

wesleycho:
I don't believe we should support this - it should also currently be possible to work around, and I believe the concerns I brought up are strong enough to discourage this being the default behavior.

Closing as per comments

Issue #3350, opened 2015-03-02
Invalid iteration parsing for typehead


vitaly-t:
From the library code below, RegEx that's used restricts your data binding to attributes only:

````
/**
 * A helper service that can parse typeahead's syntax (string provided by users)
 * Extracted to a separate service for ease of unit testing
 */
  .factory('typeaheadParser', ['$parse', function ($parse) {

  //                      00000111000000000000022200000000000000003333333333333330000000000044000
  var TYPEAHEAD_REGEXP = /^\s*([\s\S]+?)(?:\s+as\s+([\s\S]+?))?\s+for\s+(?:([\$\w][\$\w\d]*))\s+in\s+([\s\S]+?)$/;

  return {
    parse:function (input) {

      var match = input.match(TYPEAHEAD_REGEXP);
      if (!match) {
        throw new Error(
          'Expected typeahead specification in form of "_modelValue_ (as _label_)? for _item_ in _collection_"' +
            ' but got "' + input + '".');
      }

      return {
        itemName:match[3],
        source:$parse(match[4]),
        viewMapper:$parse(match[2] || match[1]),
        modelMapper:$parse(match[1])
      };
    }
  };
}])
````

I have dictionaries of data to be linked against using the standard angular syntax of:
````
"(key, value) as value in data"
where
data={
  key1: "val1",
  key2: "val2",
  etc...
}
````

The above restriction makes it very akward, so I need to transform dictionary values somehow before I can use it. There is no reason why dictionaries cannot be used for typeheads. We just need to fix the code above, somehow.

Any idea how to make it work?

karianna:
plunkr please

VitalyTomilov:
A plunk is only good to show when something is broken. Binding to a dictionary in a typehead never worked to begin with because the RegEx that's used doesn't allow for the attribute to begin with (key,value). So i'm not sure what example you are expecting to see. Please clarify.

karianna:
My bad, changing labels.

dmitriz:
The string "(key, value) as value in data" indeed is rejected as invalid expression - just added this test and it passes:

      prepareInvalidDir = function () {
        prepareInputEl('<div><input ng-model="result" typeahead="(key, value) as value in data"></div>');
      };
      expect(prepareInvalidDir).toThrow();

vitaly-t:
@dmitriz Shouldn't this be the other way round, the dictionary should work and there should be a positive test?

[...]

PR #3541, opened 2015-04-14
test(typeahead): extra invalid expressions

dmitriz:
Added tests to catch for invalid expressions
that are similar to valid and ones from ngOptions.


dmitriz:
More tests to reduce chance to run into bugs by someone trying a valid or similar string to one by `ngSelect` (`ngOptions`) but invalid for `typeahead`.

Related to issue #3350 (https://github.com/angular-ui/bootstrap/issues/3350)

PR #3544, opened 2015-04-15
fix(dateparser)HandleInvalidFormatsBetter

stevecavanagh:
This PR addresses some issues with the dateParser and its handling of certain invalid dates/formats that were mentioned in #3020.


chrisirhc:
I think the `createrParser` method should throw an exception when given an an invalid format. It knows when there are multiple occurrences of the same field (`yyyyy` in this case) as it loops through the characters in the format.

The case that when a valid format is used but the input is invalid should be handled separately.

stevecavanagh:
Thanks @chrisirhc.  I have submitted PR #3568 incorporating your suggestion.

Issue #3542, opened 2015-04-14
Running tests only for a module?


dmitriz:
I would like to run TDD tests only for a selected module. Is there any Grunt task for it?

Development instructions in
https://github.com/angular-ui/bootstrap/wiki/Development
only mention `grunt watch`, which is massive and slow.

For specific modules, only `build:module` is mentioned, 
(whereas `watch:module` ignores the `module`),
with further instructions somewhat sketchy:
"Check the Grunt build file for other tasks that are defined for this project."

I couldn't tell how to do it from looking to `Gruntfile.js` but I'm no "Grunt expert" (using Gulp instead, which I prefer by far).

chrisirhc:
You can edit karma.conf.js to include files only from the modules you need. Otherwise, you can modify the tests to use `fdescribe` for the particular tests you want to run.

dmitriz:
I was hoping for a cleaner solution, don't want getting merge conflicts when pulling updates from upstream.

I would prefer to use a separate custom karma config but not sure how to make it run.

There are also subtleties when modifying karma:
- one does not forget the dependencies of modules, e.g. `bindHtml` and `position` for typeahead:
- the following elegant brace matching globs will NOT watch for file changes (see https://github.com/karma-runner/karma/pull/1251, which is not fixed after all):

    // list of files / patterns to load in the browser
    files: [
       ...
      'src/{typeahead,bindHtml,position}/**/*.js',
      'template/{typeahead,bindHtml,position}/**/*.js'
    ],

One has to use the more verbose globs without braces instead.

dmitriz:
The other solution with `fdescribe` unfortunately doesn't work (without additional hacking) because `fdescribe` breaks the validators:

    $ grunt watch
    Running "enforce" task

    Running "ddescribe-iit:files" (ddescribe-iit) task
    >> src/typeahead/test/typeahead.spec.js has `fdescribe` at line 1
    Warning: Task "ddescribe-iit:files" failed. Use --force to continue.

    Aborted due to warnings.

dmitriz:
See working config here:

https://github.com/dmitriz/bootstrap/commit/f5e4dcbf009945a566142e33e97f3daab4b0eb30

chrisirhc:
To get around the ddescribe tests, just use `--force` like:
```shell
grunt watch --force
# or
grunt --force
```

I don't think we should change anything in the code to support this as we do want *all* the tests to run by default.

[...]

PR #3548, opened 2015-04-15
Custom Karma configuration to only run tests for typeahead module

dmitriz:
Added:

// karma-custom-typeahead.conf.js

Custom Karma configuration to only run tests for typeahead module autoWatch is switched to `true`, so Karma will rerun all tests upon file changes
 
 run as:
   
     karma start karma-custom-typeahead.conf.js

Related to Issue https://github.com/angular-ui/bootstrap/issues/3542


pkozlowski-opensource:
I don't think we should be maintaining separate configs for each directive as proposed in 46f14d1
. This would be code / maintenance duplication while simpler solutions exists (changing globs in karma conf and / or using fdescribe / ddescribe).

dmitriz:
Exactly my point, why I was asking for a better way:
 https://github.com/angular-ui/bootstrap/issues/3542#issuecomment-93247785

However, none of the other proposed ways seems to be better:

- Changing globs in karma is exactly what I did, however, doing it manually every single time,
after every single refresh, is impractical manual work. Further, there are subtleties (as per comments in the file), it took me time to find out and I don't want others to waste it again.

- Fdescribe leads to errors as I mention in the issue, can't be used with current `grunt watch`.

Imo, the only clean way would be have a task like `grunt watch:typeahead`, similar to `build:typeahead`.

Maybe I am missing something, then I'll be happy to learn.

pkozlowski-opensource:
Then we should fix fdescribe / ddescribe issue instead.

dmitriz:
The `ddescribe` warning/rejection is actually for good reason -- it safeguards you from forgetting to switch those `fdescribes` back before a pull request. Which is a good thing. 

But having to run through your code and __manually__ edit them, doesn't feel like a good thing.

Unless a convenient command like `grunt watch:typeahead` is enabled, having a dedicated clearly marked folder with custom Karma config files sounds like the next worse compromise I can think of.

dmitriz:
Close for now as not pressing

PR #3551, opened 2015-04-16
fix(modal): skipping ESC handling for form inputs; fixes #2544

### LINKED ISSUE #2544: [modal] ESC key capture too aggressive
https://github.com/angular-ui/bootstrap/blob/master/src/modal/modal.js#L224-L232

it captures that globally without checking if `event.target` is an user input (`<input type="text|number|checkbox|...">`, `<textarea>`, `<select>`)

stryju:
also added `scripts` section in `package.json` to speed up figuring out which task is for testing ;-)


chrisirhc:
This breaks expected behavior as I mentioned in ##2544

wesleycho:
Hey @stryju, can you make the requested change(s) in #2544? Happy to merge it then.

stryju:
done

src/modal/modal.js
```diff
@@ -251,6 +251,10 @@ angular.module('ui.bootstrap.modal', [])
       }
 
       $document.bind('keydown', function (evt) {
+        if (evt.isPropagationStopped() || /^(?:input|textarea|select)$/i.test(evt.target.nodeName)) {
```
wesleycho: Is this supposed to be `evt.defaultPrevented`? https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented

stryju: ah, true!
also, forgot to remove the form input check

on it

stryju: went with `evt.isDefaultPrevented()`

wesleycho:
This LGTM, based on the comments in the issue.

wesleycho:
I had to do a few tweaking since there have been updates since this PR, but got the tests passing & did some minor cleanup.

Thanks!

stryju:
:beers:

Issue #3554, opened 2015-04-16
Expose $modalstack openedWindows


lc-nyovchev:
For various reasons I'd like to have the map of openedWindows (and not only the top one) easily accessible. Currently it can be (cleanly) achieved through a $delegate that wraps open/close/dismiss functions and keeps track of the instance or through some nasty hacks like using some service to wrap any call $modal.open. However, to avoid such ugliness, it'd be nice if we could expose openedWindows.

pkozlowski-opensource:
> For various reasons

Could you name one or two of those reasons to convince us to expose "private parts"?

lc-nyovchev:
Yes, in the app I am currently developing, I have a generic error handling thing which creates a different modal based on the type of error it encounters. 

I want to cater for cases where I can have more than one modals opened at the same time. I also want various components to be able to make decisions upon inspecting a modal whether to close it or not. 

For example, let's say I have a service A that opens a modal whenever it encounters an error on a page. It is the one that created the modal and consequently has the handle for it and can close it if it sees fit. But if I have another service B that makes a modal when some email arrived or whatever, but in order to show them, it is interested that there are not any modals opened from A, but can show its modal if there are modals from services C and D.

If such a simple use case, and can be achieved by keeping some external 'registry' of all popups, but why should I invent such a system if $modalStack already has what I want. If top is exposed, I don't see why the lowest should also not be, or the N-th. 

There are other use cases, like close all modals of a specific type from a component that doesn't know how they were created. If I can put extra data in the modal, why wouldn't I be able to leverage that in making my decisions whether I should open or close them.

lc-nyovchev:
For those interested, here is a plunkr demonstrating the $delegate way. Imo it is way uglier than just exposing the opened modals.

http://plnkr.co/edit/yKU2oObUaejuimMWxoDA?p=preview

wesleycho:
As per discussion in #3556, I'm closing this issue as not to be supported.

PR #3556, opened 2015-04-16
Pull req ISSUE #3554 (Expose $modalstack openedWindows)

lc-nyovchev:


karianna:
Can you add a test for this please.

pkozlowski-opensource:
I would argue that we shouldn't merge this. As described in #3554 there is a way of doing things with a decorator, without extending the public API surface. Exposing internal data structure for not-so-common use-case (for which an elegant solutions with a decorator) will limit our ability to change internal data structures in the future.

karianna:
@pkozlowski-opensource - Thanks for the review!  Closing as wontfix.

lc-nyovchev:
Here is a plunkr that demonstrates the $delegate approach:

http://plnkr.co/edit/yKU2oObUaejuimMWxoDA?p=preview

PR #3563, opened 2015-04-19
feat(tooltip/popover): fix usage with $sce

### LINKED ISSUE #3558: Change popover/tooltip-template to evaluate expression
**Must do this before 0.13.0 release**, as it would be a breaking change if we do this after release.
Before:
```
popover-template="template.html"
```
After
```
popover-template="'template.html'"
```

The reason for this is to support $sce trusted urls. In order to pass a trusted URL, one needs to use an expression.
```
In JS:
$scope.templateUrl = $sce.trustAsResourceUrl('http://good.com/template.html');

In HTML:
popover="templateUrl"
```

chrisirhc:
Closes #3558

### DIFF
--- a/src/tooltip/docs/demo.html
+++ b/src/tooltip/docs/demo.html
@@ -20,7 +20,7 @@
       <a href="#" tooltip-animation="false" tooltip="I don't fade. :-(">fading</a>
       at elementum eu, facilisis sed odio morbi quis commodo odio. In cursus
       <a href="#" tooltip-popup-delay='1000' tooltip='appears with delay'>delayed</a> turpis massa tincidunt dui ut.
-      <a href="#" tooltip-template="myTooltipTemplate.html">Custom template</a>
+      <a href="#" tooltip-template="'myTooltipTemplate.html'">Custom template</a>
       nunc sed velit dignissim sodales ut eu sem integer vitae. Turpis egestas
     </p>
 
--- a/src/popover/docs/demo.html
+++ b/src/popover/docs/demo.html
@@ -14,7 +14,7 @@
     </div>
     <button popover="{{dynamicPopover.content}}" popover-title="{{dynamicPopover.title}}" class="btn btn-default">Dynamic Popover</button>
 
-    <button popover-template="{{dynamicPopover.templateUrl}}" popover-title="{{dynamicPopover.title}}" class="btn btn-default">Popover With Template</button>
+    <button popover-template="dynamicPopover.templateUrl" popover-title="{{dynamicPopover.title}}" class="btn btn-default">Popover With Template</button>
 
     <script type="text/ng-template" id="myPopoverTemplate.html">
         <div>{{dynamicPopover.content}}</div>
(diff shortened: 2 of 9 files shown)


karianna:
@chrisirhc Tests fail...

chrisirhc:
Thanks for checking @karianna , I marked it as "WIP" because of the failing tests. Just updated it.

src/tooltip/docs/demo.html
```diff
@@ -20,7 +20,7 @@
       <a href="#" tooltip-animation="false" tooltip="I don't fade. :-(">fading</a>
       at elementum eu, facilisis sed odio morbi quis commodo odio. In cursus
       <a href="#" tooltip-popup-delay='1000' tooltip='appears with delay'>delayed</a> turpis massa tincidunt dui ut.
-      <a href="#" tooltip-template="myTooltipTemplate.html">Custom template</a>
+      <a href="#" tooltip-template="'myTooltipTemplate.html'">Custom template</a>
```
wesleycho: Can you explain this change? This seems a little awkward for use.

wesleycho:
This PR LGTM, just had a question about a change.

src/tooltip/docs/demo.html
chrisirhc: This is because if a user needs to use a trusted resource url, the user can now do the following.

In controller:
```
$scope.templateUrl = $sce.trustAsResourceUrl('http://someurl.com/template.htm');
```

In template `tooltip-template="templateUrl"`.

Before, the if the user did `tooltip-template="{{templateUrl}}`, the URL gets converted into a plain string after interpolation and hence, loses its *trusted context*. Then there's no way for the user to specify a trusted URL. This works fine for URLs on the current security origin but will fail when the user specifies URLs from another security origin (different domain/protocol/port).

dmitriz:
This might be some related bug in the new version:

I don't see "Hello, World!" prepopulated with the new version

      angular-ui.github.io/bootstrap/ui-bootstrap-tpls-0.13.0-SNAPSHOT.js

http://plnkr.co/edit/aBBXQrotwkxCotFUAGgC?p=preview

Here is the identical plnkr with the old version that works:

      angular-ui.github.io/bootstrap/ui-bootstrap-tpls-0.12.1.js

http://plnkr.co/edit/NXTjiPyRJMxusFe1Zjzg?p=preview

---

Another way to reproduce it is to view the Demo on your localhost and click "Edit in Plunker" for the `popover` directive.

chrisirhc:
Both those Plunkers aren't using the popover directive. Did you press "Save" to make sure that the Plunker updated?

dmitriz:
Sure, the bug is not related to `popover`. I have isolated the change to the `ng-model` on `input` element, whose behaviour appears different in the two versions.

chrisirhc:
Please open another issue if this isn't related to this particular PR. We can talk about it there.
Let's keep discussion on PRs strictly about the changes in the PR itself.

dmitriz:
Agreed, but when I found it first, I actually thought it was related here...

Issue #2924, opened 2014-11-03
It would be nice if Pagination showed ellipsis's when there are too many pages


wtc-dstone:
Please correct me if I'm wrong on this, but I haven't found an option to enable showing ellipsis's when you have more pages than "max-size". This incorrectly gives the impression that there are less pages than there actually are, especially when you're only on the first page. Now, I could disable rotation, which causes ellipsis's to appear, but rotation is pretty standard and it feels weird to turn it off.

Here's what I've got now:

First, Previous, 3, 4, (5), 6, 7, Next, Last

And here's what I'm going for:

First, Previous, ..., 3, 4, (5), 6, 7, ..., Next, Last

So, is there an easy way to do this already?

PR #3564, opened 2015-04-19
feat(pagination): Add force-ellipses option to display ellipses when rotate is true

OzzieOrca:
When we have more pages than can be shown and we're rotating, ellipses can be shown. Example:

Previous, ..., 3, 4, **5**, 6, 7, ..., Next

Based off https://github.com/angular-ui/bootstrap/pull/3035 and makes that an opt-in feature
Addresses issue https://github.com/angular-ui/bootstrap/issues/2924


OzzieOrca:
Added this functionality to referenced pr.

Issue #3562, opened 2015-04-18
Cleanup Demo Inconsistency and Readability


dmitriz:
There is some inconsistency and lack of clear explanation for a few components on

http://angular-ui.github.io/bootstrap/

- The descriptions for **Button** and **Collapse** do not say what the actual directives are.

- Some directives are used as Elements and some as Attributes. It would be good to state the actual rules rather than guess them from demo examples.

- Typeahead description does not document the API `match.label` and `typeaheadHighlight:query` used inside examples' templates.

- Is there Controller namespace clashing issues? None is mentioned on the page.

Also in the Demo the UIB-components get blended among other "noise" created by the bootstrap markup, so it is hard to tell which one is which.

I would suggest an easy solution to this problem - highlight all new markup related to UIB-components.
This would **really** help the readability.

If there is agreement, I could make the corrections (once I find how to view demo pages locally, that is :)

karianna:
@dmitriz PR welcome!

dmitriz:
Ok  :)

Any easy way to build and view the doc locally?

Couldn't find in Grunt but I'm no Grunt (I'm Gulp :-)

chrisirhc:
When you run `grunt` it builds the documentation in the `dist` folder. You can run a http server (such as `http-server` on npm) and point your browser to `dist/index.html` .

dmitriz:
Thanks! 
Didn't expect the demo pages "hidden" inside `dist`. :)

PR submitted!

dmitriz:
Unfortunately, I could not find an easy way to <mark>highlight</mark> the demo code as I previously proposed. Too bad as that would really improve the readability. 

The main problem seems to be that the same files are used as actual code and again to show the markup. Which is a good precaution against putting messy code on the Demo. But this makes it very hard to add markup without affecting the functionality.

If anyone has a solution - please let me know.

rvanbaalen:
I think it will be too much to highlight the demo code where actual UI Bootstrap code is used. It's expected from users that they have some sort of experience with using Twitter Bootstrap and by having that, I think it's fairly clear to read the code examples and know what's 'new' (e.g. from UI Bootstrap).

wesleycho:
Assigned to @Foxandxss and moved the milestone back - he has been slowly doing some work in this area

PR #3567, opened 2015-04-20
Cleanup Demo Inconsistency and Readability

dmitriz:
Cleaning up Inconsistency and Readability in Demo pages as suggested in
https://github.com/angular-ui/bootstrap/issues/3562

- Some additional comments are taken from the actual source comments.
- Added more description for the `timepicker` configuration `timepickerConfig`. It wasn't clear how to use this config. (It is clear from the code but the description should try to be as self-contained as possible.) The proposed new description is probably not ideal and can be further improved.

Rules applied:
- Consistent wording "the <name> directive" (instead of randomly confusing general English words with directive names).
- Provided directive and service names are always surrounded by backticks.
- All directives used in Demos are to be previously described in the Readme (they were not).


dmitriz:
One more thing is added:

The new `Gulpfile.js` for live reloading. 
Would be a nightmare for me to do all the work without it.
I still had to manually run `grunt after-test` after each edit (but no more manual browser reload).
I would prefer to automate both but could not get the `gulp-grunt` plugin work with current `Gulpfile.js`,
and submitted this issue for now.

https://github.com/gratimax/gulp-grunt/issues/20

rvanbaalen:
Looks like some solid work. @chrisirhc and @Foxandxss what do you think?

src/modal/docs/readme.md
```diff
@@ -1,4 +1,4 @@
-`$modal` is a service to quickly create AngularJS-powered modal windows.
+The `$modal` is an AngularJS `service` (a `provider`) to quickly create AngularJS-powered modal windows.
```
chrisirhc: A service isn't the same as a provider, so I think the parenthesis can be left out. Also "service" shouldn't need to be code (between `).

src/alert/docs/readme.md
```diff
@@ -1,5 +1,5 @@
-Alert is an AngularJS-version of bootstrap's alert.
+The `alert` directive is an AngularJS-version of bootstrap's alert.
 
-This directive can be used to generate alerts from the dynamic model data (using the `ng-repeat` directive);
+It can be used to generate alerts from the dynamic model data (using the `ng-repeat` directive).
 
-The presence of the `close` attribute determines if a close button is displayed
+The presence of the `close` attribute determines if a close button (cross) is displayed inside the alert.
```
chrisirhc: Might want to use `&times;` (&times;)

chrisirhc:
Hm.. I think the documentation changes are great. However, I think that the live reload can also be done in grunt. I don't like the idea of introducing yet another dev dependency for watching for changes to documentation changes.

I think the gulp changes should be in another so that this PR can just focus on documentation changes.

chrisirhc:
Also, I saw that `index.html` was added in this PR, I don't think it should be added. The new documentation will be generated when we make a new release.

pkozlowski-opensource:
+1 for @chrisirhc [take on Gulp as part of this PR](https://github.com/angular-ui/bootstrap/pull/3567#issuecomment-94449138). Nothing against switching to Gulp (quite the contrary!) - we just shouldn't have both.

dmitriz:
Thanks for the feedback guys!

All remarks - critical and not - are greatly appreciated!

About Gulp -- totally see your point! 
I did think about it myself but felt too tired and lazy after all the editing :)

I can surely take it out of this PR.

I am puzzled how `index.html` managed to slip ... should have been ignored ... will try to take it out of the history without ruining the rest.

Will try to rework it and the 2 notes (will comment separately) and resubmit as new PR.

dmitriz:
I am also happy to hear some support for Gulp :)
Are there plans for switching?

I even tried to add the live reloading to directly to the Gruntfile but was the Grunt won the battle :(
I find it quite hard to read and comments are scarce.
It seems to be hard to get Livereload that doesn't suck...

I'll be happy to submit my Gulpfile as a separate PR if there is any support for it. 
It does make my life easier but would make it even easier with proper rebuild and reload,
which I couldn't get working with the `gulp-grunt` plugin unfortunately.

If anyone could add it into Grunt directly, that would help too.
Otherwise this is the best compromise I could come up with for the moment.

It does add 5 new node packages of only 7.5MB in total to the 25 already present ones of 60MB total, so the extra footprint would be still small.

src/alert/docs/readme.md
dmitriz: Good idea!

src/modal/docs/readme.md
dmitriz: All great points! Angular makes it really sweet an easy by avoiding calling a Service two different things :) 

How about:
The `$modal` is an AngularJS Service registered as  `provider` ... ?

dmitriz:
Managed to get Gulp working with LiveReload!
https://github.com/dmitriz/bootstrap/blob/dz-gulpfile-added/Gulpfile.js

Now any edit in demo files automatically updates and reloads the demo pages, even in multiple tabs!
:smile: 

Can submit PR if there is interest.

PR #3568, opened 2015-04-20
fix(dateparser): Throw on invalid date format string.

stevecavanagh:
This PR replaces #3544.  The dateParser will throw an exception for an invalid format string as suggested.

### DIFF
--- a/src/dateparser/dateparser.js
+++ b/src/dateparser/dateparser.js
@@ -95,6 +95,10 @@ angular.module('ui.bootstrap.dateparser', [])
           format[i] = '$';
         }
         format = format.join('');
+        var dupe=format.indexOf(code[0]);
+        if (dupe>-1){
+          throw new Error('Invalid date format string.');
+        }
 
         map.push({ index: index, apply: data.apply });
       }
--- a/src/dateparser/test/dateparser.spec.js
+++ b/src/dateparser/test/dateparser.spec.js
@@ -173,4 +173,12 @@ describe('date parser', function () {
   it('should not parse if no format is specified', function() {
     expect(dateParser.parse('21.08.1951', '')).toBe('21.08.1951');
   });
+
+  it('should not parse if invalid format is specified', function() {
+    expect(function(){dateParser.parse('20.12.20190', 'dd.MM.yyyyy');}).toThrow(new Error('Invalid date format string.'));
+  });
+
+  it('should not parse if invalid value is specified', function() {
+    expect(dateParser.parse('20.12.20190', 'dd.MM.yyyy')).toBeUndefined();
+  });
 });


src/dateparser/test/dateparser.spec.js
```diff
@@ -173,4 +173,12 @@ describe('date parser', function () {
   it('should not parse if no format is specified', function() {
     expect(dateParser.parse('21.08.1951', '')).toBe('21.08.1951');
   });
+
+  it('should not parse if invalid format is specified', function() {
+    expect(function(){dateParser.parse('20.12.20190', 'dd.MM.yyyyy');}).toThrow(new Error('Invalid date format string.'));
+  });
+
+  it('should not parse if invalid value is specified', function() {
```
dmitriz: Two tests with the same name?

src/dateparser/dateparser.js
```diff
@@ -95,6 +95,10 @@ angular.module('ui.bootstrap.dateparser', [])
           format[i] = '$';
         }
         format = format.join('');
+        var dupe=format.indexOf(code[0]);
```
wesleycho: Add spaces around the `=`

src/dateparser/dateparser.js
```diff
@@ -95,6 +95,10 @@ angular.module('ui.bootstrap.dateparser', [])
...
+        if (dupe>-1){
```
wesleycho: Add spaces between the `>` and before `{`

src/dateparser/test/dateparser.spec.js
```diff
@@ -173,4 +173,12 @@ describe('date parser', function () {
   it('should not parse if no format is specified', function() {
     expect(dateParser.parse('21.08.1951', '')).toBe('21.08.1951');
   });
+
+  it('should not parse if invalid format is specified', function() {
+    expect(function(){dateParser.parse('20.12.20190', 'dd.MM.yyyyy');}).toThrow(new Error('Invalid date format string.'));
```
wesleycho: Add spaces, i.e. `function() { dateParser...`

wesleycho:
There is the question whether it is desirable to throw - Angular has a philosophy of allowing some failures to be silent, such as an Angular expression that fails to execute since the function is undefined.

Thoughts?

Balrog30:
As a consumer of angular libraries, I have found that I expect my invalid input to fail silently, and if something is not behaving as expected I watch my own code.

I expect warnings if I'm using something that is deprecated, errors if something fails catastrophically in the architecture of my app, and that's about it.

If user input is invalid, I think that should fail silently unless the developer has set up form validation.

realityking:
Since the format string will be set most of the time by the developer, I think a fatal error is appropriate. If this turns out to be a problem for someone, a configuration option could be added.

Angular's silence on errors is sometimes a real brain teaser, it'd be good not to take this further.

stevecavanagh:
I will clean up the formatting in my PR and resubmit in the next day or so.

---- Rouven Weßling wrote ----

stevecavanagh:
Please close in favor of #3690.

rvanbaalen:
As requested.

stevecavanagh:
Please consider #3691 as a replacement for this PR.

PR #3574, opened 2015-04-21
Tests added to document behaviour of bind-html-unsafe 

dmitriz:
The directive `bind-hmtl-unsafe` is used in Demo pages but never documented.

These specs aim to document it and ensure it works as expected.


chrisirhc:
bind-html-unsafe is being deprecated. Perhaps we should add a deprecating notice.

dmitriz:
What would be the best way to add it? 

It is currently only used in the `typeahead` demo examples and inside `typeahead-match.html`.

But doesn't really have anything to do with `typeahead` and has global availability.
So I'm not certain where would be an elegant way to put it.

PR #3576, opened 2015-04-21
feat(tooltip): now can take the option for the close delay timeout

tmcgee123:
test(tooltip): testing that the close delay works as expected

This is a change to add control over the delay to remove the tooltip from the view. This solution came from the need to delay the closing of the tooltip without overriding bootstrap classes, manipulation of the DOM , or need to create custom triggers that depend on the usage of $timeout. 

The use-case for this was:
1. User hovers over the link for the popover/tooltip
2. The popover/tooltip shows long enough for the user to move the cursor off of the element that triggers the popover and access any link inside of the popover/tooltip content

### DIFF
--- a/src/tooltip/docs/readme.md
+++ b/src/tooltip/docs/readme.md
@@ -20,6 +20,7 @@ will display:
 - `tooltip-animation`: Should it fade in and out? Defaults to "true".
 - `tooltip-popup-delay`: For how long should the user have to have the mouse
   over the element before the tooltip shows (in milliseconds)? Defaults to 0.
+- `tooltip--close-popup-delay`: For how long should the tooltip remained open?
 - `tooltip-trigger`: What should trigger a show of the tooltip? Supports a space separated list of event names.
   Note: this attribute is no longer observable. See `tooltip-enable`.
 - `tooltip-enable`: Is it enabled? It will enable or disable the configured
@@ -63,6 +64,7 @@ methods are available:
   placement: 'top',
   animation: true,
   popupDelay: 0,
+  popupCloseDelay: 500,
   appendToBody: false
   </pre>
 
--- a/src/tooltip/test/tooltip.spec.js
+++ b/src/tooltip/test/tooltip.spec.js
@@ -365,6 +365,60 @@ describe('tooltip', function() {
     }));
   });
 
+  describe('with specified popup close delay', function() {
+    var $timeout;
+    beforeEach(inject(function($compile, _$timeout_) {
+      $timeout = _$timeout_;
+      scope.delay = '1000';
+      elm = $compile(angular.element(
+        '<span uib-tooltip="tooltip text" tooltip-popup-close-delay="{{delay}}" ng-disabled="disabled">Selector Text</span>'
+      ))(scope);
+      elmScope = elm.scope();
+      tooltipScope = elmScope.$$childTail;
+      scope.$digest();
+    }));
+
+    it('should close after timeout', function() {
+      trigger(elm, 'mouseenter');
+      expect(tooltipScope.isOpen).toBe(true);
+      trigger(elm, 'mouseleave');
+      $timeout.flush();
+      expect(tooltipScope.isOpen).toBe(false);
+    });
+    //
+    it('should use default popup close delay if specified delay is not a number and close after delay', function() {
+      scope.delay = 'text1000';
+      scope.$digest();
+      trigger(elm, 'mouseenter');
+      expect(tooltipScope.popupCloseDelay).toBe(500);
+      expect(tooltipScope.isOpen).toBe(true);
+      trigger(elm, 'mouseleave');
+      $timeout.flush();
+      expect(tooltipScope.isOpen).toBe(false);
+    });
+
+    it('should open when not disabled after being disabled and close after delay - issue #4204', function() {
+      trigger(elm, 'mouseenter');
+      expect(tooltipScope.isOpen).toBe(true);
+
+      elmScope.disabled = true;
+      elmScope.$digest();
+
+      $timeout.flush(500);
+      expect(tooltipScope.isOpen).toBe(false);
+
+      elmScope.disabled = false;
+      elmScope.$digest();
+
+      trigger(elm, 'mouseenter');
+
+      expect(tooltipScope.isOpen).toBe(true);
+      trigger(elm, 'mouseleave');
+      $timeout.flush();
+      expect(tooltipScope.isOpen).toBe(false);
+    });
+  });
+
   describe('with an is-open attribute', function() {
     beforeEach(inject(function ($compile) {
       scope.isOpen = false;
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -15,6 +15,7 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
     placement: 'top',
     animation: true,
     popupDelay: 0,
+    popupCloseDelay: 500,
     useContentExp: false
   };
 
@@ -265,7 +266,7 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
               // FIXME: this is a placeholder for a port of the transitions library.
               if (ttScope.animation) {
                 if (!transitionTimeout) {
-                  transitionTimeout = $timeout(removeTooltip, 500);
+                  transitionTimeout = $timeout(removeTooltip, ttScope.popupCloseDelay);
                 }
               } else {
                 removeTooltip();
@@ -321,7 +322,9 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
               ttScope.placement = angular.isDefined(attrs[prefix + 'Placement']) ? attrs[prefix + 'Placement'] : options.placement;
 
               var delay = parseInt(attrs[prefix + 'PopupDelay'], 10);
+              var closeDelay = parseInt(attrs[prefix + 'PopupCloseDelay'], 10);
               ttScope.popupDelay = !isNaN(delay) ? delay : options.popupDelay;
+              ttScope.popupCloseDelay = !isNaN(closeDelay) ? closeDelay : options.popupCloseDelay;
             }
 
             ttScope.contentExp = function() {


src/tooltip/test/tooltip.spec.js
```diff
@@ -277,6 +278,52 @@ describe('tooltip', function() {
...
+  describe('with specified popup close delay', function () {
+
+    beforeEach(inject(function ($compile) {
+      scope.delay='1000';
+      elm = $compile(angular.element(
+        '<span tooltip="tooltip text" tooltip-popup-delay="1000"' +
+        'tooltip-popup-close-delay="{{delay}}">Selector Text</span>'
+      ))(scope);
+      elmScope = elm.scope();
+      tooltipScope = elmScope.$$childTail;
```
dmitriz: Isn't using Angular's internal method `$$childTail` creating a potential for future bugs, should its implementation change in the future?

If I understand it correctly, this is to access child scope, which we should try to avoid to.
Here it is suggested instead to 
> move the model up into the parent...
http://stackoverflow.com/a/14006165/1614973

src/tooltip/test/tooltip.spec.js
```diff
@@ -249,10 +249,11 @@ describe('tooltip', function() {
       scope.$digest();
     }));
 
-    it('should open after timeout', inject(function ($timeout) {
+    it('should open after timeout and set the close delay to the default', inject(function ($timeout) {
 
       elm.trigger('mouseenter');
       expect(tooltipScope.isOpen).toBe(false);
```
dmitriz: Strictly speaking, this doesn't really test that the `tooltip` is closed, only that its scope's `isOpen` is `false`. This **should** be the same, in principle, but only **assuming** it works correctly. 

Would a direct testing of opening/close not be more reliable?

src/tooltip/test/tooltip.spec.js
```diff
@@ -277,6 +278,52 @@ describe('tooltip', function() {
...
+        'tooltip-popup-close-delay="{{delay}}">Selector Text</span>'
+      ))(scope);
+      elmScope = elm.scope();
+      tooltipScope = elmScope.$$childTail;
+      scope.$digest();
+    }));
+
+    it('should open after timeout and close after the close delay timeout', inject(function ($timeout) {
+
+      elm.trigger('mouseenter');
```
dmitriz: Maybe use Angular's native `.triggerHandler()` here?
https://docs.angularjs.org/api/ng/function/angular.element

src/tooltip/test/tooltip.spec.js
tmcgee123: I've used the same setup as the test for the delay of the opening of the tooltip in the file, if you were to look at the rest of the spec it is doing the same. In a test, this is standard practice to access the directive's scope in order to check a value.

src/tooltip/test/tooltip.spec.js
tmcgee123: I am directly testing the opening and closing. The ```isOpen``` variable is bound to the scope and is the main condition that opens and closes the tooltip. It is a boolean, can't get really get any different value other than: true, false, null, undefined.

src/tooltip/test/tooltip.spec.js
tmcgee123: Like I said, if you look at the other tests in this file, they are doing this in the same way. What you are asking is something I would label underneath a style() commit.

src/tooltip/test/tooltip.spec.js
tmcgee123: The only way to "truly" test this would be to set up an interval or timeout and see if it respects the time it takes to close upon the trigger of the closing event of the popover. The reason I wrote the test this way was because of the test written for popup-delay. In order to provide consistent code, I wrote the test in the same manner. If there is a better way to test that actually works, then post your code please :D. Cheers!

src/tooltip/test/tooltip.spec.js
dmitriz: This is just a concern that came up, I know it works for now, but might be helpful for future reference. There might be a better way without depending so deeply on Angular's mechanics.
I  didn't want to add to already huge issue queue, so put it here.

src/tooltip/test/tooltip.spec.js
dmitriz: By direct testing I mean the `display` property of the element. Wouldn't that be more reliable. It is true that this is how it is tested there, but maybe adding such a display test would help to guard against future bugs. Anyway, was just a suggestion.

dmitriz: For the timeout, you surely want to mock it rather then wait for it, if that is what you mean.

src/tooltip/test/tooltip.spec.js
dmitriz: It is correct that `.trigger` is currently used but it is from `jQuery`. The present tests have been written a while ago, but more recently [the Project Philosophy](https://github.com/angular-ui/bootstrap/wiki/Project-Philosophy) sets the goal to get away from the `jQuery` dependency. So for new PRs, it might be a worth contribution towards the new Philosophy to support this trend.

This is actually much more than *style*. There are [subtle differences between two methods](http://api.jquery.com/triggerHandler/). As `Angular` chose the other one, we might want to follow, unless there is a reason not to.

Finally, I totally see your point of following the present conventions, so feel free not to worry about it if you prefer. I may submit a separate PR in the future.

tmcgee123: I was not aware that it was from jQuery. On that note, I think we should create an issue to not use the ```.trigger```. I am most definitely all aboard for removing jQuery as a dependency (I was also unaware that was a goal in the link you provided). 

Thank you for your posts, I always appreciate the opportunity to learn more. If you agree with an issue being created to re-write the tests that are utilizing jQuery, I would be more than happy to create them and then update this file to use ```.triggerHandler()``` as you've suggested.

dmitriz: You are welcome :)

It can be impressive to disable [the presently included `jQuery`](https://github.com/angular-ui/bootstrap/blob/master/karma.conf.js#L18) and see 425 tests out of 755 failing... no small task to suggest!

As an attempt to make it more practical and less daunting, my suggestion would be to state it in the contribution guidelines as *recommendation* for new PRs to consider. It can be always documented in then comments and linked to this or other discussions if that gets too long. 

So sure, are welcome to open issue referring to this and see what the consensus is.

tmcgee123:
I was able to successfully change each ```.trigger``` to ```.triggerHandler``` in this file. Also, I added more assertions for checking whether or not the element is visible and not just "dirty" checking the isOpen scope variable as @dmitriz  has suggested for both.  The only comment I was unable to address was the one pertaining to ```$$childTail``` as the isOpen variable is not accessible publicly and is truly "isolate" to this directive, but is set as a TODO on **line 142** of ```tooltip.js``` where it states "// TODO add ability to start tooltip opened".

src/tooltip/test/tooltip.spec.js
```diff
@@ -33,7 +33,7 @@ describe('tooltip', function() {
   }));
 
   it('should open on mouseenter', inject(function() {
-    elm.trigger( 'mouseenter' );
```
wesleycho: Any particular reason for all these changes from `trigger` to `triggerHandler`?

src/tooltip/tooltip.js
```diff
@@ -319,6 +321,12 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
               ttScope.popupDelay = ! isNaN(delay) ? delay : options.popupDelay;
             }
 
+            function prepPopupCloseDelay() {
+              var val = attrs[ prefix + 'PopupCloseDelay' ];
```
wesleycho: Remove whitespace just after `[` and before `]`.

src/tooltip/tooltip.js
```diff
@@ -319,6 +321,12 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
...
+              var delay = parseInt( val, 10 );
```
wesleycho: Ditto except for the parentheses.

src/tooltip/tooltip.js
```diff
@@ -319,6 +321,12 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
...
+              ttScope.popupCloseDelay = ! isNaN(delay) ? delay : options.popupCloseDelay;
```
wesleycho: Remove extra space after `!`.

wesleycho:
I like this change for it to be more configurable, but this PR seems polluted with some non-relevant changes.

Also for verification, I notice that the account that made the commits is different from the one filing the PR - I assume they are both yours?

src/tooltip/test/tooltip.spec.js
tmcgee123: @dmitriz noted above that ```trigger``` is from the jQuery dependency: "It is correct that .trigger is currently used but it is from jQuery. The present tests have been written a while ago, but more recently the Project Philosophy sets the goal to get away from the jQuery dependency. So for new PRs, it might be a worth contribution towards the new Philosophy to support this trend." Please check the outdated comments above from the changes I made.

wesleycho: I would undo those changes - they can be tackled in a separate PR, but they pollute the feature.

tmcgee123:
@wesleycho can you elaborate on the "pollution" you mentioned above? I'm not sure what you mean by the account that made the commits is different, it's just showing my new avatar and update profile. I'm the only person making these changes if that's what you are asking :D 

Also, if you would've bothered to open the files you've made comments on above, you would see that my spacing is consistent with the old code base. I did not want to make even more changes to just update spacing in the file, especially since everything is trimmed when the file is minified.  My JSHINT is not reading any errors and I don't see a JSCS file either, so if you're going to be implementing standards for spacing then you should make it easier for others to adapt.

src/tooltip/tooltip.js
tmcgee123: @wesleycho I'm staying consistent with the rest of the file, if you want me to update the whole file then I can.

src/tooltip/tooltip.js
tmcgee123: @wesleycho I'm staying consistent with the rest of the file, if you want me to update the whole file then I can.

src/tooltip/tooltip.js
tmcgee123: @wesleycho I'm staying consistent with the rest of the file, if you want me to update the whole file then I can.

src/tooltip/test/tooltip.spec.js
tmcgee123: Will do!

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "author": "https://github.com/angular-ui/bootstrap/graphs/contributors",
   "name": "angular-ui-bootstrap",
-  "version": "0.13.2-SNAPSHOT",
+  "version": "0.13.2-A1.4.3",
```
wesleycho: Undo this change.

src/tooltip/test/tooltip-template.spec.js
```diff
@@ -69,11 +69,10 @@ describe('tooltip template', function() {
...
 
+    $timeout.flush();
     expect( tooltipScope.isOpen ).toBe( false );
 
-    $timeout.flush();
     expect( elmBody.children().length ).toBe( 1 );
   }));
 
 });
-
```
wesleycho: Undo this change.

wesleycho:
I have considered opening code style change PRs and altering the build to enforce it, but it has the downside of instantly merge conflicting almost all PRs - an incremental improvement in style would help. If it is not made, it would likely be made on merging of the PR, which puts more burden on maintainers (it was already painful enough cutting issues down from over 500 to under 330, and open PRs from over 200 to almost 100 the past couple of months).

tmcgee123:
@wesleycho I've made all of the recommended changes that you've left comments on. Please let me know if there are any other changes! Thanks for the help!

src/tooltip/test/tooltip.spec.js
```diff
@@ -853,9 +938,9 @@ describe( '$tooltipProvider', function() {
         elm.trigger('fakeTrigger');
         expect( tooltipScope.isOpen ).toBeTruthy();
         elm.trigger('fakeTrigger');
+        $timeout.flush();
```
wesleycho: Is this change necessary?

src/tooltip/test/tooltip-template.spec.js
```diff
@@ -69,11 +69,10 @@ describe('tooltip template', function() {
     scope.templateUrl = '';
     scope.$digest();
 
+    $timeout.flush();
```
wesleycho: Any particular reason for this change?

wesleycho:
The history needs to be cleaned up - there should not be a merge commit, and it should be squashed into one.

src/tooltip/test/tooltip.spec.js
tmcgee123: Yes, the test was failing on my branch since there is a timeout for closing the popup.

src/tooltip/test/tooltip-template.spec.js
tmcgee123: Yes, the test needs the timeout to flush properly in order to enable the correct state.

tmcgee123:
Ran into a problem rebasing, I'm going to have to re-submit my changes

tmcgee123:
@wesleycho I've updated this again to resolve the merge conflicts, any chance on this getting merged soon? It'd be nice to be able to point to the main repository rather than pointing to my fork. Thanks!!

wesleycho:
I apologize for having to ask this again, but can you rebase off of `master` & adjust the feature one more time? Some major fixes landed with regards to tooltip positioning & $digest errors.

I intend on having this land for 0.14, so the sooner, the better :) .

tmcgee123:
Okay Mr. @wesleycho , I went ahead and cleared my changes and re-applied them, as I was stuck in rebase misery :D still need to learn how to do that properly. Cheers!

src/tooltip/docs/readme.md
```diff
@@ -20,6 +20,7 @@ will display:
 - `tooltip-animation`: Should it fade in and out? Defaults to "true".
 - `tooltip-popup-delay`: For how long should the user have to have the mouse
   over the element before the tooltip shows (in milliseconds)? Defaults to 0.
+- `tooltip--close-popup-delay`: For how long should the tooltip remained open?
```
wesleycho: One dash

src/tooltip/test/tooltip.spec.js
```diff
@@ -365,6 +365,60 @@ describe('tooltip', function() {
...
+    }));
+
+    it('should close after timeout', function() {
+      trigger(elm, 'mouseenter');
+      expect(tooltipScope.isOpen).toBe(true);
+      trigger(elm, 'mouseleave');
+      $timeout.flush();
+      expect(tooltipScope.isOpen).toBe(false);
+    });
+    //
```
wesleycho: Delete comment

wesleycho:
LGTM - I'll merge this soon with the fixes applied.

tmcgee123:
Thanks!!!

ArielAleksandrus:
Does it work with popover? I'm not being able to use it with neither popover nor tooltip. Can I pass a literal value or it should be an angular variable? Is the time in milliseconds?

wesleycho:
It should work with both - this seems to not work as expected though.

tmcgee123:
It looks like in the midst of the many changes that were conflicting with my branch, the functionality stopped working.

For some reason, even though the $timeout is called in the hide function on line 269 in tooltip.js, it is not properly timing out.

```
              if (ttScope.animation) {
                if (!transitionTimeout) {
                  transitionTimeout = $timeout(removeTooltip, ttScope.popupCloseDelay); // here is where $timeout is not working for some reason
                }
              } else {
                removeTooltip();
              }
```

wesleycho:
I noticed this too this past weekend, but I didn't have time to debug since I was out for a large part of it - if you don't find out what happened, I'll probably take a look again sometime later this week.

tmcgee123:
Ahh, it's because of the $evalAsync that is being run before we wait for the animation to completely remove it from the DOM. Since it sets the visibility to false, it is still preset in the DOM, but not visible.

I was able to get the delay to work by changing this:

```
ttScope.$evalAsync(function() {
  ttScope.isOpen = false;
  if (isOpenParse && angular.isFunction(isOpenParse.assign)) {
    isOpenParse.assign(ttScope.origScope, ttScope.isOpen);
  }
});
```

to this:

```
$timeout(function () {
  ttScope.$evalAsync(function() {
    ttScope.isOpen = false;
    if (isOpenParse && angular.isFunction(isOpenParse.assign)) {
      isOpenParse.assign(ttScope.origScope, ttScope.isOpen);
    }
  });
}, ttScope.popupCloseDelay);
```

Also, there is a potential for a slight refactor, since there is shared functionality inside of the callbacks for each $evalAsync:

```
...
ttScope.isOpen = false;
if (isOpenParse && angular.isFunction(isOpenParse.assign)) {
  isOpenParse.assign(ttScope.origScope, ttScope.isOpen);
}
...
```

Maybe we can make a function called assignIsOpenToScope (need better name for this)?

```
function assignIsOpenToScope(isOpen) {
  ttScope.isOpen = isOpen;
  if (isOpenParse && angular.isFunction(isOpenParse.assign)) {
    isOpenParse.assign(ttScope.origScope, ttScope.isOpen);
  }
}
```

RobJacobs:
There is already a timeout for when the removeTooltip() is getting called, I would rather see the hideTooltipBind() restructured to work like the showTooltipBind().  I'll put together a PR for review shortly.

tmcgee123:
I agree, I think it's rather silly that hideTooltipBind simply just calls hide. We can probably just use one or the other.

RobJacobs:
@tmcgee123 PR #4597 has been created to address this issue.

Issue #3573, opened 2015-04-21
Angular bootstrap not compatible with AngularJS 1.2.3


gflandre:
The website states that the project `requires AngularJS 1.2.x`.

However the jqlite AngularJS 1.2.3 embeds does not support the `$element.empty` method (cf. https://code.angularjs.org/1.2.3/docs/api/angular.element) that is used in the latest angular bootstrap version.

This causes the Modal directive to crash.

karianna:
You should be using the latest 1.2.x (1.2.14 I think?)

gflandre:
My point was that the doc is misleading and should be updated.

karianna:
Ah sorry, whereabouts in the docs is this, I can't find that reference..

jodytate:
I think the relevant line is in the site docs, e.g.:

https://github.com/angular-ui/bootstrap/blob/master/misc/demo/index.html#L148
                    
    <li><a href="http://angularjs.org" target="_blank">AngularJS</a> (requires AngularJS 1.2.x, tested with <%= ngversion %>)</li>

Not sure if this is referenced elsewhere.

dmitriz:
`element.empty` is supported by the latest 1.2.x version:

https://code.angularjs.org/1.2.27/docs/api/ng/function/angular.element

Here 1.2.x means the latest version, maybe it should be made more clear?

Foxandxss:
It is misleading indeed. Sometimes it is hard to realize that something you added weren't there on a previous version.

We will fix that for the next version, since we are dropping 1.2.x support.

JobaDiniz:
> since we are dropping 1.2.x support.

This is really sad to hear. Is the latest 0.13.0 version already does not support angular 1.2.28?

Foxandxss:
It is not. The whole ngAnimate library changed a lot from 1.2.x to 1.3.x so we had to drop support for 1.2.x.

There is NO reason to not update an application to 1.2.x.

JobaDiniz:
> There is NO reason to not update an application to 1.2.x
To 1.3.x you meant...

Well, you might be right in theory, but in reality moving to 1.3.x can have a big impact on an already solid project. We have a huge project with a lot of 3rd party libs (not just angular.ui)... truth be told, moving to 1.3 is the right move, but not so simple :grinning:

Foxandxss:
An issue easily mitigated with proper tests :P

PR #3581, opened 2015-04-23
docs(mainpage): add "latest" to Angular 1.2.x

dmitriz:
Clarified Doc, resolves https://github.com/angular-ui/bootstrap/issues/3573


wesleycho:
I don't think this should be merged - `master` only supports 1.3 currently.

dmitriz:
That was meant for the current Demo pages, to avoid the confusion such as in that issue.
Should it go to another branch in that case?

rvanbaalen:
This will be fixed when 0.13 lands. No need to merge it.

PR #3588, opened 2015-04-26
Update transition.js

antonellopasella:
chrome support both styles but triggers only "transitionend" and "animationend" in latest versions


wesleycho:
I'm confused, why should changing the key order matter?

wesleycho:
Also, it should be noted, we are no longer using `$transition` in master - it has been deprecated.

antonellopasella:
- "I'm confused, why should changing the key order matter?"

Because of foreach that catch the first result.

I'm using it in a Phonegap app and stock browser will fire different event than chrome implementation

wesleycho:
It should not matter though - as per the ES spec, you cannot rely on the key order.

In addition, usage of $transition has been removed in `master` & the service has been marked as deprecated. Closing as outdated.

Issue #3450, opened 2015-03-27
Make releases less painful


chrisirhc:
Creating a stub issue here for discussion and thoughts.

Past work to reference: #1666

rvanbaalen:
#3706 related

wesleycho:
It might be worth scheduling this for 0.14.0 - it would be useful maintenance work.

wesleycho:
Closing as superceded by #5715

PR #3591, opened 2015-04-27
Add Meteor.js package support

dotansimha:
Hi AngularUI team,

I'm a package maintainer for Meteor.js, a popular full-stack JavaScript framework.
I'm also a developer on angular-meteor, a library that let's you work with AngularJS on top of Meteor.

A lot of our users use angular-ui-bootstrap and it right now there is a separate package and Github repo that simply manually copies your distribution and publish it to Meteor.
This PR will allow you to directly publish updated versions of angular-ui-bootstrap as they become available. All you have to do is create an account at https://meteor.com/ (click SIGN IN, then Create account). After you've done that, please let me know the name of the account, and I'll add you as a maintainer to the "angularui" organization there.

I've already published the current version of the package on Atmosphere (Meteor's package directory). When you release new versions of angular-ui-bootstrap, you'll be able to publish them to Atmosphere by running `grunt meteor-publish` after updating the version number on the "package.js" file.

Thanks & best regards,
Dotan


chrisirhc:
Our release process is very complex right now, I don't think we should complicate it any further at this point. I think #3450 needs to be tackled first before we can go on with adding more features to the release process.

wesleycho:
Unfortunately, as @chrisirhc mentioned, our release process is complicated & involves manually copying over scripts to https://github.com/angular-ui/bootstrap-bower (the repository we publish from), and I do not believe that we will be able to get to this, as we have a lot of work split between this and the migration to ng2 - this is going to create a simultaneous maintenance issue for us.

We will be glad to entertain supporting publishing to Atmosphere for the ng2 version of Bootstrap though, as we should have an infinitely saner release process.

PR #3600, opened 2015-04-30
feat(typeahead): handles min-length of 0

zacronos:
Correctly handles min-length of 0 (i.e. will keep typeahead popup open even when all input has been deleted).

Still uses a default value of 1 for backward compatibility (when min-length is falsy but not === 0).

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -842,4 +842,13 @@ describe('typeahead tests', function () {
     expect($scope.select_count).toEqual(1);
   });
 
+  describe('minLength set to 0', function () {
+    it('should open typeahead if input is changed to empty string if defined threshold is 0', function () {
+      var element = prepareInputEl('<div><input ng-model="result" typeahead="item for item in source | filter:$viewValue" typeahead-min-length="0"></div>');
+      changeInputValueTo(element, '');
+
+      expect(element).toBeOpenWithActive(3, 0);
+    });
+  });
+
 });
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -41,7 +41,10 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
       //SUPPORTED ATTRIBUTES (OPTIONS)
 
       //minimal no of characters that needs to be entered before typeahead kicks-in
-      var minSearch = originalScope.$eval(attrs.typeaheadMinLength) || 1;
+      var minSearch = originalScope.$eval(attrs.typeaheadMinLength);
+      if (!minSearch && minSearch !== 0) {
+        minSearch = 1;
+      }
 
       //minimal wait time after last character typed before typeahead kicks-in
       var waitTime = originalScope.$eval(attrs.typeaheadWaitMs) || 0;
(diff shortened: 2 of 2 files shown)


src/typeahead/typeahead.js
```diff
@@ -41,7 +41,10 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
       //SUPPORTED ATTRIBUTES (OPTIONS)
 
       //minimal no of characters that needs to be entered before typeahead kicks-in
-      var minSearch = originalScope.$eval(attrs.typeaheadMinLength) || 1;
+      var minSearch = originalScope.$eval(attrs.typeaheadMinLength);
```
dmitriz: Maybe renaming to `minLength` would make it easier to read - similar to the attribute?

zacronos: @dmitriz, I could change that if it is demanded, however note that I did not change that variable name -- `minSearch` was the name used in the original code.

Cleaning up pre-existing issues in the original code seems a rather high burden to place on a 5-line PR (not counting specs) from 1.5 years ago (this is a re-PR, the original was posted in Jan of 2014).

dmitriz: Well it was used only used in 2 places, now would be 5.
Would clean up a bit but no worries.

src/typeahead/test/typeahead.spec.js
```diff
@@ -842,4 +842,13 @@ describe('typeahead tests', function () {
...
   });
 
+  describe('minLength set to 0', function () {
+    it('should open typeahead if input is changed to empty string if defined threshold is 0', function () {
+      var element = prepareInputEl('<div><input ng-model="result" typeahead="item for item in source | filter:$viewValue" typeahead-min-length="0"></div>');
+      changeInputValueTo(element, '');
+
+      expect(element).toBeOpenWithActive(3, 0);
+    });
+  });
```
dmitriz: Should it only allow empty `typeahead` to open, if the `input` is focused?

Otherwise several `typeahead` directives  with empty `input`s would simultaneously open at the start, which is probably not intended.

src/typeahead/typeahead.js
rvanbaalen: Please note that one community members' notes and reviews aren't 'demands' on how you should do things but rather suggestions/tips. There should always be an open discussion :+1:

---

zacronos: @dmitriz, fair enough, you're right that this variable isn't really getting used elsewhere.  I'll add that in.

src/typeahead/test/typeahead.spec.js
zacronos: Yes, that is the current behavior -- see the test starting on line 563, which states an element must be focused for the typeahead to be open.

Also, I believe even on focus the typeahead will not immediately open on focus, regardless of the limit and current input length.  The typeahead only opens when the input changes.

dmitriz: Good catch - you are right!

About immediate opening - this can be useful, for instance, to show most recent searches as soon as the user clicks or taps in the field. Would be a better UX than typing, in my view. That would seem to me the most intuitive behaviour for `minLength = 0`.

zacronos: I actually agree with that -- especially for `minLength = 0`, but really any time, if the user has no idea what a valid input is, it could be a little clunky for them to have to start to type, then delete, in order to see all options.  This would be a nice new feature to add -- an option to open on focus instead of on input change.

mattvague:
What's the status on this?

chrisirhc:
LGTM

PR #3606, opened 2015-05-03
fix(datepicker): MinMaxInputValidation

### LINKED ISSUE #1688: Inline-Datepicker doesn't add ng-(in)?valid

stevecavanagh:
This PR is designed to fix #1688 and replaces my earlier PR #3020.
Enable min and max validation on datepicker popup input.  Revised to only use one formatter and one parser per reviewer suggestions.  Also, modified demo app to avoid scoping issues when binding to primitives.  (See https://github.com/angular/angular.js/wiki/Understanding-Scopes)

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -206,7 +206,9 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
       datepickerMode: '=?',
       dateDisabled: '&',
       customClass: '&',
-      shortcutPropagation: '&?'
+      shortcutPropagation: '&?',
+      minDate: '@',
+      maxDate: '@'
     },
     require: ['datepicker', '?^ngModel'],
     controller: 'DatepickerController',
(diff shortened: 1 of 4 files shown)


src/datepicker/datepicker.js
```diff
@@ -577,6 +579,62 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
         datepickerEl.attr('custom-class', 'customClass({ date: date, mode: mode })');
       }
 
+      function compareDates(date1, date2) {
+        if (isNaN(date1) || isNaN(date2)) {
+          return undefined;
+        }
+        else {
```
wesleycho: Move `else` one line up

src/datepicker/datepicker.js
```diff
@@ -206,7 +206,9 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
       datepickerMode: '=?',
       dateDisabled: '&',
       customClass: '&',
-      shortcutPropagation: '&?'
+      shortcutPropagation: '&?',
+      minDate: '@',
+      maxDate: '@'
```
wesleycho: These new bindings should be optional

src/datepicker/test/datepicker.spec.js
```diff
@@ -1199,8 +1199,12 @@ describe('datepicker directive', function () {
         $document = _$document_;
         $sniffer = _$sniffer_;
         $rootScope.isopen = true;
+        $rootScope.format='yyyy-MM-dd';
         $rootScope.date = new Date('September 30, 2010 15:30:00');
-        var wrapElement = $compile('<div><input ng-model="date" datepicker-popup is-open="isopen"><div>')($rootScope);
+        $rootScope.minimumDate =  new Date('January 1, 1970');
+        $rootScope.maximumDate =  new Date('January 1, 2020');
+        $rootScope.dateDisabledHandler = jasmine.createSpy('dateDisabledHandler');
+        var wrapElement = $compile('<div><input ng-model="date" min-date = "minimumDate" max-date = "maximumDate" date-disabled="dateDisabledHandler(date, mode)" datepicker-popup="{{format}}" is-open="isopen"><div>')($rootScope);
```
wesleycho: This change should be tested separately from the existing test suites.

wesleycho:
Going to close this in favor of #4841 - the approach there seems more elegant.

PR #3623, opened 2015-05-05
Add open-on-focus option to datepicker popup

jbthoma:
This change is based on #2515, except it uses scope.$apply instead of $timeout, and adds datepickerPopupConfig.openOnFocus instead of scope variable @openOnFocus.


karianna:
@jbthoma Can you add tests for this please?

eranimo:
+1

wesleycho:
With the programmatic `is-open` feature, this is something that should be implemented in a custom directive on the user's side.

Closing as something a user should be implementing on their side.

PR #3625, opened 2015-05-05
Validating if markup options are really set in $scope

kreigiron:
Validates if options are really set in $scope when <code> datepicker-options="options"  </code> is given in markup, but no <code> $scope.options </code> exists.

otherwise it will throw an error like:
<code>
TypeError: Cannot read property 'initDate' of undefined
    at link (http://localhost:3000/bower_components/angular-bootstrap/ui-bootstrap-tpls.js:1523:19)
    at http://localhost:3000/bower_components/angular/angular.js:8252:44
    at invokeLinkFn (http://localhost:3000/bower_components/angular/angular.js:8258:9)
</code>
....

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -535,7 +535,7 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
 
       if ( attrs.datepickerOptions ) {
         var options = scope.$parent.$eval(attrs.datepickerOptions);
-        if(options.initDate) {
+        if(options && options.initDate) {
           scope.initDate = options.initDate;
           datepickerEl.attr( 'init-date', 'initDate' );
           delete options.initDate;


311chaos:
I am getting this error "Cannot read property 'initDate' of undefined as well. This only started occurring after updating from 0.12.1 to 0.13.0. What I am trying to understand is if I need to change my values that are getting passed into the directive.

kreigiron:
@311chaos Indeed, this happened to me after updating from 0.12.x to 0.13.x,  and when I have the "datepicker-options="options"" attribute defined, but no $scope.options is defined. It seems that the old version didn't really care about $scope.options existence.

311chaos:
Well, what seems to have changed is the $scope. All of my values are on the $parent scope. So I had:

`data-datepicker-options="SRCtrl.datepicker.dateOptions"`

but it needs to be changed to:

`data-datepicker-options="$parent.SRCtrl.datepicker.dateOptions"`

I can work around this, but it isn't obvious and I don't love referencing the $parent....

wesleycho:
I think a better fix might be `options = scope.$parent.$eval(attrs.datepickerOptions) || {}`.

wesleycho:
I decided to merge as is - this is probably a better check to avoid mutating the value.

Issue #3617, opened 2015-05-05
Datepicker: "...ng-model" value must be..." on Clear action


fifonik:
1. Open http://angular-ui.github.io/bootstrap/
2. Ctrl+Datepicker
3. Open popup datepicker
4. Choose any date
5. Open popup datepicker again
6. Click 'Clear' button
=>
In console you'll see error "Datepicker directive: "ng-model" value must be a Date object, a number of milliseconds since 01.01.1970 or a string representing an RFC2822 or ISO 8601 date."

I have not expected any errors on clearing date (setting it to null).
At least in 0.12.1 I had not have any errors.

Grottkjell:
Having the same issue. Using angular 1.3.15.

chrisfoster78:
I have this problem too and have had to downgrade to an older version to support nullable dates. Any other workaround?

DieterLi:
Same issue with 0.13.0, but good with 0.12.1

anilnatha:
I am running Angular 1.3.16 and get the same error message upon clicking 'clear' in the datepicker modal.

What makes it worse is that I am checking my forms $invalid property which is returning `true` even though the datepicker field in my form is not a required field, so a blank datepicker field should be valid, not invalid.

anilnatha:
Probably worth mentioning the following:

 1. If you open up firebug while viewing the DatePicker demo at [https://angular-ui.github.io/bootstrap/](https://angular-ui.github.io/bootstrap/), it throws the error.
 2. In the example shown at the link, a `clear` function is set in `$scope`; however, in my setup, which is identical to that, the `clear` function is never called. Perhaps because the error is thrown prior to the clear function being called? I thought the clear function is called first if you click the 'clear' button on the datepicker modal, but that doesn't appear to be the case.

anilnatha:
I have created a plunker, with documentation that describes the various problems related to this bug:

http://plnkr.co/edit/Iild15cgR9NXaX0iWjti?p=preview

This example aims to highlight the issue isn't just with the datepicker's model being cleared, but also how it affects form validation.

jackpimbert:
:+1: Same issue here.

askucher:
+1

icfantv:
[...]

Issue #1289, opened 2013-11-24
date picker dd.MM.yyyy


fopsdev:
hi there, i'm testing the datepicker within angular 1.2.2 and bs 3 using our common swiss (and german) date format dd.MM.yyyy
just recognised that entering the date value manually in the textbox does not work as expected. it still uses the first number as the month.

http://plnkr.co/edit/1pdDca5KuWOWh3vPVRvz?p=preview

svarjavand:
Why is this issue and all its referenced ones are closed? DatePicker still don't work with dd/MM/yyyy format when date is entered manually.
Try the plunker on https://angular-ui.github.io/bootstrap/#/datepicker and enter 24/05/2015
current version is: 
angularjs: 1.3.13
bootstrap-ui: 0.13.0

rvanbaalen:
@svarjavand it seems like either this is broken again or @bekos closed it too fast. Reopening.

rvanbaalen:
On a second thought @svarjavand please create a new issue to prevent this one from being polluted. This is closed because the actual issue is fixed; dates with `dd.MM.yyyy` format work.

svarjavand:
@rvanbaalen I can create a new issue but even dd.MM.yyyy format doesn't work.
Try the plunker on the demo page and enter 24.05.2015 in the popup date picker. Selected date will be empty.

rvanbaalen:
Fair enough. Reason enough to keep this open.

realityking:
That's likely the issue described in #3644 (potential fix in #3682)

hendrikebbers:
With version 0.13 the console shows the following error when typing the date "23.12.2007" (pattern: dd.MM.yyyy) directly after typing 23:
Datepicker directive: "ng-model" value must be a Date object, a number of milliseconds since 01.01.1970 or a string representing an RFC2822 or ISO 8601 date.(anonymous function) @ angular.js:11655render @ ui-bootstrap-tpls.js:1071init.ngModelCtrl.$render @ ui-bootstrap-tpls.js:1059ngModelWatch @ angular.js:23423

Tis is working with version 0.12

agriboz:
I would like to know why the datepicker popup date not updating as expected using `dd.MM.yyyy` format. 

http://plnkr.co/edit/7Fmcsz2rRDsi9U5eaeGz?p=preview

PR #3682, opened 2015-05-18
fix(datepicker): Don't parse the date twice

### LINKED ISSUE #3643: datepicker 0.13.0 does not allow model updates after a parse error for certain dates when ng-model-options updateOn: blur is used
Using 0.13.0 of datepicker with ng-model-options="{updateOn: 'blur'}"

When entering a date that causes a parse error (e.g. 'abc'), subsequently entering a 10-character 'mm/dd/yyyy' date (e.g. 01/01/2014) will not update the $modelValue.  

Entering an 8-character date (e.g. 1/1/2014) or selecting the datepicker in the calendar does update $modelValue.

[Plunkr](http://plnkr.co/edit/zY1DiPfEskLmNONddPJF?p=preview)

This seems related to [https://github.com/angular-ui/bootstrap/issues/3609], but I wasn't sure if it was the same issue.  As an aside, the $invalid flag isn't being set on parse errors.

Would be interested if anyone is aware of a workaround.

### LINKED ISSUE #3644: Datepicker has trouble with some dates
This might be related to #3617 and/or #3643.

JSBin: http://output.jsbin.com/pasiweruhe/2/

I have a date field set up with the German date format (`dd.MM.yyyy`). After clearing the field and entering a date where the day part is larger than 12, it will fail and the date will be `undefined`.

The issue seems to be, that the date parser has trouble creating a valid date input when the input is an invalid date input.

realityking:
Fixes #3644, #3617 

Potentially fixes #3643 as well, but I haven't tested that one yet.

Unit tests are missing, as I'm at a loss how to test these issues with karma. I'd appreciate some pointers in this regard. If it's fine to add protractor tests, I'd be happy to do so.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -650,7 +650,7 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
 
       // Detect changes in the view from the text box
       ngModel.$viewChangeListeners.push(function () {
-        scope.date = dateParser.parse(ngModel.$viewValue, dateFormat, scope.date) || new Date(ngModel.$viewValue);
+        scope.date = ngModel.$modelValue;
       });
 
       var documentClickBind = function(event) {


kelvinatorr:
Works for me. Thanks!

rvanbaalen:
Looks good and seems to be fixing some problems. Thanks for the PR! Will look into this soon to merge it.

rvanbaalen:
@realityking Can you please add some tests for this?

realityking:
@rvanbaalen Just to clarify, these tests would need protractor. Is that ok?

rvanbaalen:
My bad. Now I see your comment about Karma tests. Was running an hour of slightly brainless PR and Issue triaging :wink:

realityking:
No problem, but I still don't have an answer ;) Are protactor tests fine?

morgenes:
+1

wesleycho:
This change would have an effect on HTML5 date inputs I believe, since the $viewChangeListener is the only callback that gets fired that parses the date input.

Can you produce a Plunker with an example of this in action for that situation?

Blackbaud-PatrickOFriel:
Any word on the status of this PR?

realityking:
I haven't had time to set up protractor tests, so at least that's still open.

I'd be great if someone could create the test mentioned by @wesleycho as I'm pretty short on time right now.

almeidap:
I can confirm that this fixes some parsing and validation issues (particularly #1289), specially when manually changing the date.

For those using Bower who would like a hotfix, you may fetch AngularUI from here (minified version not updated): https://github.com/almeidap/angular-ui-bootstrap

PR #3685, opened 2015-05-18
feat(dropdown) Add keynav support (fix for #1228)

bleggett:
Adds a new option flag to the dropdown class that enables the use of arrow keys to navigate dropdown list elements, like non-angular bootstrap does. Added tests and updated doc with example. Fix for [#1228] (https://github.com/angular-ui/bootstrap/issues/1228) and based off of [#3212] (https://github.com/angular-ui/bootstrap/issues/3212)

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -10,11 +10,11 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
   this.open = function( dropdownScope ) {
     if ( !openScope ) {
       $document.bind('click', closeDropdown);
-      $document.bind('keydown', escapeKeyBind);
+      $document.bind('keydown', keybindFilter);
     }
 
     if ( openScope && openScope !== dropdownScope ) {
-        openScope.isOpen = false;
+      openScope.isOpen = false;
     }
 
     openScope = dropdownScope;
@@ -24,7 +24,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
     if ( openScope === dropdownScope ) {
       openScope = null;
       $document.unbind('click', closeDropdown);
-      $document.unbind('keydown', escapeKeyBind);
+      $document.unbind('keydown', keybindFilter);
     }
   };
 
@@ -37,7 +37,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
 
     var toggleElement = openScope.getToggleElement();
     if ( evt && toggleElement && toggleElement[0].contains(evt.target) ) {
-        return;
+      return;
     }
 
     var $element = openScope.getElement();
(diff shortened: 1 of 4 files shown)


rvanbaalen:
Please squash your commits.

bleggett:
Squashed.

src/dropdown/dropdown.js
```diff
@@ -112,6 +120,40 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
...
+      case (40): {
+        if ( !angular.isNumber(self.selectedOption)) {
+          self.selectedOption = 0;
+        } else {
+          self.selectedOption = (self.selectedOption === elems.length -1 ?
+                                 self.selectedOption :
+                                 self.selectedOption+1);
+        }
+      }
+      break;
```
wesleycho: Shouldn't this be inside the `case` statement?

src/dropdown/dropdown.js
```diff
@@ -112,6 +120,40 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
...
+      case (38): {
+        if ( !angular.isNumber(self.selectedOption)) {
+          return;
+        } else {
+          self.selectedOption = (self.selectedOption === 0 ?
+                                 0 :
+                                 self.selectedOption-1);
+        }
+      }
+      break;
```
wesleycho: Ditto

src/dropdown/test/dropdown.spec.js
```diff
@@ -213,7 +213,7 @@ describe('dropdownToggle', function() {
       });
 
       return $compile('<li dropdown><a href dropdown-toggle></a>' +
-        '<ul><li><a href="#something">Hello</a></li></ul></li>')($rootScope);
+                      '<ul><li><a href="#something">Hello</a></li></ul></li>')($rootScope);
```
wesleycho: Revert this change

src/dropdown/test/dropdown.spec.js
```diff
@@ -350,8 +350,8 @@ describe('dropdownToggle', function() {
   describe('`auto-close` option', function() {
     function dropdown(autoClose) {
       return $compile('<li dropdown ' +
-        (autoClose === void 0 ? '' : 'auto-close="'+autoClose+'"') +
-        '><a href dropdown-toggle></a><ul><li><a href>Hello</a></li></ul></li>')($rootScope);
+                      (autoClose === void 0 ? '' : 'auto-close="'+autoClose+'"') +
+                      '><a href dropdown-toggle></a><ul><li><a href>Hello</a></li></ul></li>')($rootScope);
```
wesleycho: Revert the indentation change in these two lines

wesleycho:
These are mostly code style issues. @rvanbaalen if you think it's worth merging and fixing these issues before pushing up, this PR LGTM.

src/dropdown/dropdown.js
rvanbaalen: You're saying 

```js
var elems = self.dropdownMenu ? //If append to body is used.
      (angular.element(self.dropdownMenu).find('a')) :
     (angular.element(self.$element).find('ul').eq(0).find('a'));
```

should be copied to each of these `case`'s?

wesleycho: I mean the `break`

src/dropdown/dropdown.js
```diff
@@ -112,6 +120,40 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
...
+      (angular.element(self.$element).find('ul').eq(0).find('a'));
+
+    switch (keyCode) {
+      case (40): {
+        if ( !angular.isNumber(self.selectedOption)) {
+          self.selectedOption = 0;
+        } else {
+          self.selectedOption = (self.selectedOption === elems.length -1 ?
+                                 self.selectedOption :
+                                 self.selectedOption+1);
```
wesleycho: Add spaces before and after the `+`.

Also make the indentation only one level deep, and not align to the `self` in the first line.

src/dropdown/dropdown.js
```diff
@@ -112,6 +120,40 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
...
+        }
+      }
+      break;
+      case (38): {
+        if ( !angular.isNumber(self.selectedOption)) {
+          return;
+        } else {
+          self.selectedOption = (self.selectedOption === 0 ?
+                                 0 :
+                                 self.selectedOption-1);
```
wesleycho: Add spaces before and after the `-`.

Also change indentation to only be one level deep, not align to `self` from the first line.

RobJacobs:
My only concern is with setting focus() on the menu item elements, I'm not sure if this would disrupt the tab flow once the dropdown is closed?  The typeahead has similar functionality but is uses an activeIndex with:

   ng-class="{active: isActive($index) }"

to decorate the element.  Then listens for enter to trigger selection.

src/dropdown/dropdown.js
rvanbaalen: Of course. 

The official C99 specification says the following about the break statement:
> A break statement terminates execution of the smallest enclosing switch or iteration statement.

So I would vote to put these breaks inside the curly brackets as well.

rvanbaalen:
@bleggett Can you incorporate these suggestions?

bleggett:
@rvanbaalen Yep, will do. Will look at what typeahead is doing like @RobJacobs suggested as well.

wesleycho:
Can you squash your commits again? In the future, if you need to sync up with remote master, you can do `git pull --rebase origin master` (assuming origin is pointing to this repository)

bleggett:
@wesleycho Yeah, resquashed. Still getting familiar with git, so forgive the rookie mistakes.

Regarding @RobJacobs suggestion: typeahead creates the menu element dynamically, and so can add attrs to it (like the select and activeIndex) relatively easily.

Dropdown relies on a predefined, user-defined menu element, and I'm less sure about tacking attrs onto that blindly, or expecting users to add this stuff themselves.

Am I way off here?

RobJacobs:
@bleggett Thanks for looking at that, I agree with your assessment.

PR #3653, opened 2015-05-11
fix(modal): IE throws exceptions on body.activeElement.focus

tompi:
fixes #3639

Issue #3639, opened 2015-05-07
Modal close fails in IE if modal was opened from js


Hjertaas:
See stacktrace at the bottom.

Using angular: 1.4.0-beta.5 and angular-bootstrap: 0.13.0
Error seen on IE 9 and 11 (others not tested)

If modal dialog is initiated from javascript (say an Angular controller) there is no activeElement
See line 2209 in ui-bootstrap-tpls.js
var modalOpener = $document[0].activeElement;   // in IE this is undefined

This in turn results in a nullpointer in close() and dismiss() functions. Lines 2262 and 2273 respectively.

We are temporarily rectifying this by running this 
if (!window.document.activeElement) {
    window.document.body.focus();
}
immediately prior to calling $modal.open()

TypeError: Unable to get property 'focus' of undefined or null reference
   at $modalStack.close (http://localhost:9000/ext/angular-bootstrap/ui-bootstrap-tpls.js:2262:11)
   at modalInstance.close (http://localhost:9000/ext/angular-bootstrap/ui-bootstrap-tpls.js:2340:17)
   at fn (Function code:2:191)
   at callback (http://localhost:9000/ext/angular/angular.js:22335:17)
   at Scope.prototype.$eval (http://localhost:9000/ext/angular/angular.js:15236:9)
   at Scope.prototype.$apply (http://localhost:9000/ext/angular/angular.js:15335:11)
   at Anonymous function (http://localhost:9000/ext/angular/angular.js:22340:17)
   at jQuery.event.dispatch (http://localhost:9000/ext/jquery/dist/jquery.js:4664:6)
   at elemData.handle (http://localhost:9000/ext/jquery/dist/jquery.js:4332:5)
   at jQuery.event.dispatch (https://example.com/plugins.3619bc3f.js:3:14958)
   at elemData.handle (https://example.com/plugins.3619bc3f.js:3:9674)

rvanbaalen:
We don't support AngularJS 1.4 yet.

karianna:
Still a valid issue though, putting a 0.13.x milestone on it (e.g. Fix when we get round to looking at 1.4 support)

rvanbaalen:
I get your point @karianna. But I would have to disagree on the part where this would be a 'valid issue'. Yes, this is probably a bug in UI Bootstrap. But like I said -- no one is stating that UI Bootstrap is expected to work with the latest beta AngularJS version.

[...]

PR #3689, opened 2015-05-19
fix(modal): Captures focus in both directions - using tab & shift-tab

### LINKED ISSUE #738: $modal does not stop tabbing on the background
This can be reproduced on the sample page on http://angular-ui.github.io/bootstrap/
- in the Dialog section click 'Open Dialog' button
- keep hitting tab key

result: tab going through background page and data can also be entered.

Hope someone can advise on this problem. Thanks.

Tobino:
A new directive to enforceFocus inside Modal.
This directive works with multiple modal, everything is tested except shift-tab because of karma.
I could test it with protrator.
I also refactor modalOpener.Focus without change in Spec.
fixes #738

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -147,6 +147,65 @@ angular.module('ui.bootstrap.modal', [])
     };
   }])
 
+  .directive('enforceFocus',['$document', '$modalStack', function ($document, $modalStack) {
+    return {
+      link: function($scope, iElm) {
+        var body = $document.find('body').eq(0);
+        var trapFocusDomEl = angular.element('<div tabindex="0"></div>');
+        body.append(trapFocusDomEl);
+
+
+        function currentModal(){
+          var modal = $modalStack.getTop();
+          if (modal && modal.value) {
+            return modal.value.modalDomEl[0] === iElm[0];
+          }
+        }
+
+        //enforceFocus inside modal
+        function enforceFocus(evt) {
+          if (!currentModal()) {
+            return;
+          }
+          if (iElm[0] !== evt.target && !iElm[0].contains(evt.target)) {
+            iElm[0].focus();
+          }
+        }
+        $document[0].addEventListener('focus', enforceFocus, true);
+
+
+        //return lastFocusable element inside modal
+        function lastFocusable(domEl) {
+          var tababbleSelector = 'a[href], area[href], input:not([disabled]), button:not([disabled]),select:not([disabled]), textarea:not([disabled]), iframe, object, embed, *[tabindex], *[contenteditable]';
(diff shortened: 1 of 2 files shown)


Tobino:
This PR will be in conflict with #3639.
I need to remove refactor of modalOpener.Focus.

rvanbaalen:
I'm feeling this should be closed in favour of #3653. But I'm not sure. Therefore, I leave this open with a 'needs: review' label.

Tobino:
This PR fix focus when modal is open and fixe issue https://github.com/angular-ui/bootstrap/issues/738.
https://github.com/angular-ui/bootstrap/pull/3653 fix focus when Modal close.

commit  5ae356f fix a conflict with PR https://github.com/angular-ui/bootstrap/pull/3653

chrisirhc:
This is a lot of overriding of the tab behavior. Is the focusin handler mentioned in my comment https://github.com/angular-ui/bootstrap/issues/738#issuecomment-55193038 inadequate?

Tobino:
I can remove some part if you want, depend on level of a11y you want.
There is 3 parts on this directive [https://github.com/angular-ui/bootstrap/issues/738#issuecomment-48099422]:
* [fix tab](https://github.com/Tobino/bootstrap/blob/a11y-tabbing-modal/src/modal/modal.js#L158..L174)
* [fix shift-tab](https://github.com/Tobino/bootstrap/blob/a11y-tabbing-modal/src/modal/modal.js#L177..L195)
* [no moving to the address bar](https://github.com/Tobino/bootstrap/blob/a11y-tabbing-modal/src/modal/modal.js#L153..L155)

This directive is the angular version of [bootstrap](https://github.com/twbs/bootstrap/blob/master/js/modal.js#L139..L147) and [bootstrap-accessibility-plugin](https://github.com/paypal/bootstrap-accessibility-plugin/blob/master/src/js/modal.js#L15..L22) with support of multiple modals.

PS: [Boostrap](http://getbootstrap.com/javascript/#callout-stacked-modals) doesn't support multiple modals.

e-cloud:
@wesleycho , @paulorbpacheco
**thanks for your effort** on the **traping tab feature on the modal module**. but i have a trouble with the new feature.

i use a tabset in the modal. when i update to **v0.13.2**, my modals without tabs work as expected, but it turn to tabs and modal aggregation, it can only trap the `tab` key on the second tab of the tabset.

here is a plukr that reproduce the problem.
http://plnkr.co/edit/wwgZhXu0fru4L703z5by?p=preview

i'm not sure if i use it on the wrong place or the modal source code has some bugs not concerned yet.
can you guys offset some help?

PR #3690, opened 2015-05-20
fix(dateparser): Throw on invalid date format string.

stevecavanagh:
This PR replaces #3568. The dateParser will throw an exception for an invalid format string.  In addition to fixing some code formatting issues, this PR also handles parsing dates from the datepicker with a datetime-local type.


stevecavanagh:
Closing due to lint failure.  Resubmitted PR #3691.

realityking:
@stevecavanagh Just a FYI, you don't have to open a new pull request if you need to change something. Just push again into the same branch and it will be picked up.

If you want/need to keep it to one commit take a look at [amending](https://de.atlassian.com/git/tutorials/rewriting-history/git-commit--amend/) and [rebasing](https://www.atlassian.com/git/tutorials/rewriting-history/git-reflog)

stevecavanagh:
Sorry, I was under the mistaken belief that I shouldn't amend/rebase after I've pushed.  Thank you, I'll follow that approach going forward.

realityking:
@stevecavanagh That's a good rule for public repositories, otherwise calloboration withothers (or even forks) are in hell. For branches that only serve as the base for a pull request, that doesn't matter as you'll be the only one pushing into it.

PR #3703, opened 2015-05-23
feat(dropdown): Add dropup behavior

### LINKED ISSUE #1317: Make dropdown menu go upwards if it is at the bottom of the browser window
Dear all!

We're having a little issue with dropdowns. We're using dropdown menus tables (context menus). As you can see in the fiddle below, the dropdown menus of the rows at the bottom of the browser window are not completely visible (you would have to use your scrolling wheel to see the whole menu).

http://jsfiddle.net/mhu23/YZx7R/22/

What I learned from bootstrap was that you can use the class "dropup" to open the menus upwards (see code comment in fiddle). But now I'm stuck implementing an intelligent solution. Should I create a directive which dynamically adds the class dropup? But how do I decide if the menu is "rendered" outside the view scope?

Thanks a lot for your help!

Michael

gorork:
Makes dropdown menu go upwards if it's at the bottom of the browser window and is being partially hidden.
Moves DOM logic to directive.

Closes #1317


rvanbaalen:
This is some nice work! I was thinking wouldn't is be best to add this sort of functionality to the $position service in order to allow tooltip and popover to have the same sort of logic?

Also, we need some tests for this beauty.

gorork:
@rvanbaalen Thanks! Moving these calculations to $position service is a great idea. I'd like work on it after creating tests for this one. Would it work?

gorork:
@rvanbaalen I've added tests for this pr. 
If everything looks good here, I'm moving to a new branch for $position service.

src/dropdown/dropdown.js
```diff
@@ -170,9 +162,50 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
...
+            display:    'block'
+          });
+
+          var windowBottom = $window.innerHeight + $window.pageYOffset,
+              dropdownOffset = $position.offset(dropdown).top,
+              dropdownHeight = $position.position(dropdown).height,
+              dropdownMenuHeight = $position.position(dropdownMenu).height,
+              dropdownMenuBottom = dropdownOffset + dropdownHeight + dropdownMenuHeight;
+
+          if ( appendToBody !== undefined ) {
```
wesleycho: Any reason this isn't just `if (appendToBody)`?

src/dropdown/dropdown.js
```diff
@@ -170,9 +162,50 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
         return;
       }
       dropdownCtrl.dropdownMenu = element;
+
+      var dropdownMenu = element,
+          dropdown = element.parent(),
```
wesleycho: This would potentially be a breaking change, if the dropdown element is not the parent element itself.

gorork: @wesleycho Currently, CSS code defines that `.dropdown-menu` **must be** a direct child of `dropdown` element to be displayed:
```
.open > .dropdown-menu {
  display: block;
}
```

src/dropdown/dropdown.js
gorork: Here, if you check for falseness `dropdown.attr('dropdown-append-to-body')` evaluates to `false` even if `dropdown-append-to-body` attribute is applied to dropdown element. 
So we need to specifically target its `undefined` state.

src/dropdown/dropdown.js
wesleycho: That's true to the extent if Bootstrap usage is assumed - I know there are some UI Bootstrap users out there who do not use Bootstrap and use the library, even though the library does not officially support that situation.

It still is worth noting.

src/dropdown/dropdown.js
wesleycho: That is what I am confused about - further up it is set to `dropdown.attr('dropdown-append-to-body')` at the beginning of the linking function, that would presumably be set by the time the watcher is evaluated. As far as I can tell, there is no situation where it would ever be undefined.

src/dropdown/dropdown.js
gorork: @wesleycho Makes sense. I've just pushed changes to accommodate these user cases (test included).

src/dropdown/dropdown.js
```diff
@@ -170,9 +162,53 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
         return;
       }
       dropdownCtrl.dropdownMenu = element;
+
+      var dropdown, appendToBody,
+          dropdownMenu = element;
+
+      scope.$watch(dropdownCtrl.isOpen, function( isOpen ) {
```
wesleycho: Any reason why this $watch isn't in the controller?

gorork: @wesleycho I believe Link function is more appropriate, because most of the logic here is DOM-related (and Link function and Controller share the same scope in this case).

wesleycho:
Can you rebase this off of current `master`?

src/dropdown/dropdown.js
wesleycho: I think I would favor this being in the controller - this adds an additional $watch for what seems to be unnecessary reasons.

wesleycho:
Is there any possible way you can abstract this logic for determining whether to do the up or down to the `$position` service? It could be of immense use to other components for supporting an `auto` setting. It would have to be built in a way that could also entertain the possibility of right vs. left positional rendering as well.

gorork:
@wesleycho Rebased it.
As for $position, I would definitely abstract this. I'd prefer to work on this in a separate branch/pr though.

wesleycho:
Looks like there are conflicts still - also looks like the history got busted - maybe try filing a new PR with the history fixed? You might have rebased when there was a merge commit present - the merge commit causes problems with history, which is why we always rebase.

gorork:
@wesleycho Here is a new PR: https://github.com/angular-ui/bootstrap/pull/4534
Should I close this one?

wesleycho:
Closing as this is superceded by the referenced PR

PR #3708, opened 2015-05-25
dateParser needs to handle invalid base dates

JohannesRudolph:
I noticed that datePicker would fail in the following scenario (reproducible with the official sample): 
 - clear input text
 - enter a valid date according to format
 - observe that the date is parsed as invalid date

The underlying issue seems to be that the dateParser is handled an invalid base date (as in: a Javascript InvalidDate object). This pull request fixes this scenario and should be trivial to review as it contains a failing test which passes with the suggested fix.


bitstorm:
+1 I need it badly :)

src/dateparser/dateparser.js
```diff
@@ -124,8 +124,10 @@ angular.module('ui.bootstrap.dateparser', [])
         results = input.match(regex);
 
     if ( results && results.length ) {
-      var fields, dt;
-      if (baseDate) {
+      var fields, dt,
+          isValidBaseDate = baseDate && !isNaN(baseDate.valueOf());
```
wesleycho: This doesn't look quite right - this should probably be `angular.isDate(baseDate) && !isNaN(baseDate.getTime())`.

wesleycho:
I believe this PR is superceded by #3759 . Thoughts?

rvanbaalen:
It is indeed @wesleycho

chrisirhc:
Closing in favor of #3759

PR #3715, opened 2015-05-26
custom data for days in datepicker

netanelgilad:
Hey guys,
I wanted to add some extra text to selected dates in the date picker. So I have added the logic to datepicker so this would work.
What do you think of the change? if it seems alright, I'll add some tests for it.


rvanbaalen:
Thank you for your efforts. However, we won't be adding new API's that cover very specific use cases. Therefore, I'm closing this issue. Feel free to re-open with good arguments if it should be merged.

PR #3716, opened 2015-05-27
feat(typeahead): Setting to show popup after blur

jhoekx:
We have a typeahead query that in some cases takes a lot of time (20 seconds) due to a very slow backend ouside of our control. If the user does something while waiting, the input element loses focus and no popup is shown. This is frustrating for users.

This PR adds a `typeahead-show-after-blur` attribute that defaults to `false`.

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -579,6 +579,25 @@ describe('typeahead tests', function () {
       expect(element).toBeClosed();
     });
 
+    it('should show the popup after blur if show-after-blur is enabled', function () {
+
+      $scope.items = function(viewValue) {
+        return $timeout(function(){
+          return [viewValue];
+        });
+      };
+      var element = prepareInputEl('<div><input ng-model="result" typeahead="item for item in items($viewValue)" typeahead-show-after-blur="true"></div>');
+      var inputEl = findInput(element);
+
+      changeInputValueTo(element, 'match');
+      $scope.$digest();
+
+      inputEl.blur();
+      $timeout.flush();
+
+      expect(element).not.toBeClosed();
+    });
+
     it('should properly update loading callback if an element is not focused', function () {
 
       $scope.items = function(viewValue) {
--- a/src/typeahead/docs/readme.md
+++ b/src/typeahead/docs/readme.md
@@ -55,3 +55,7 @@ The typeahead directives provide several attributes:
 * `typeahead-focus-first`
    _(Defaults: true)_ :
    Should the first match automatically be focused as you type?
+
+* `typeahead-show-after-blur`
+   _(Defaults: false)_ :
+   Should the popup appear even after the element loses focus?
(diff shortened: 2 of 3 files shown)


rvanbaalen:
Travis failed and I'm not sure if we want to add more API options for every specific use case. Wouldn't it be better if you implemented this specific feature in your own directive that extends UI Bootstrap's typeahead?

jhoekx:
So, the master build is even failing for me with angular 1.3.15. Reverting angular and angular-mocks to 1.3.14 builds it cleanly.

jhoekx:
This specific behavior is quite deeply into the typeahead component. I wouldn't really know how I could skip the attaching of the event handler from another component. If you have a clear path in mind of how I could overwrite that, any feedback appreciated.

Before I wrote this, I already tried to attach another listener to the DOM node and stop the propagation of the event there, but that didn't work.

wesleycho:
I don't think this is a bad feature to support - this component is probably one of the more complex ones in its functionality in UI Bootstrap, so it makes sense that many options are desired. The performance overhead of this looks minimal, so I am thinking that this is ok.

src/typeahead/test/typeahead.spec.js
```diff
@@ -579,6 +579,25 @@ describe('typeahead tests', function () {
       expect(element).toBeClosed();
     });
 
+    it('should show the popup after blur if show-after-blur is enabled', function () {
```
wesleycho: Remove space in `function (`

src/typeahead/test/typeahead.spec.js
```diff
@@ -579,6 +579,25 @@ describe('typeahead tests', function () {
...
+
+      $scope.items = function(viewValue) {
+        return $timeout(function(){
```
wesleycho: Add space after `)`

chrisirhc:
I think this feature can be allowed after the typeaheadController change. Users of the directive can then use programmatic triggers, to trigger opening of the popup whenever they want, or to prevent it from closing.

wesleycho:
Should we hold off on this until the refactor then?

chrisirhc:
I don't think we should consider this setting as it's very specific and is another potential for watcher bloat and complexity.

Users can add a directive that would require the directive controller to set up their own programmatic triggers, to write such features themselves.

wesleycho:
In this case, I don't see any watchers - is the concern potential `$watch`es being added? This could be avoided with `attrs.$observe` at worst if it is a concern.

chrisirhc:
It's yet another option in the long list of options. We're maintaining a long list of options that we're adding code complexity for. Before we know it, this becomes as complex or more complex than the datepicker.
It makes refactoring/changing the code harder as it adds more considerations.
Which leads to the final question, do we add features every time a single person asks for it? Do users download the library expecting these features?


This change breaks the contract and meaning of `hasFocus`. `hasFocus` can no longer be relied upon. A future developer now needs another way to tell if the element actually has focus.

wesleycho:
Closing due to prior comments

PR #3345, opened 2015-02-27
fix(datepicker): Check if getter.assign is function fixes #3155

rvanbaalen:
Closes #3155

PR #3719, opened 2015-05-27
fix(datepicker): check if getter.assign is function

### LINKED ISSUE #3155: Date picker TypeError: setAttribute is not a function setAttribute(scope.$parent, value);
I got error then datepicker is closed after date is selected
Version: 0.12.0 - 2014-11-16
```javascript
TypeError: setAttribute is not a function
setAttribute(scope.$parent, value);
```

rvanbaalen:
Closes #3155, #3345


rvanbaalen:
Travis seems to be busted. Some tests failed that are not related to this PR. @RobJacobs do you know anything about this? You're the last one who touched the master branch.

RobJacobs:
@rvanbaalen The package.json had the angular dependency as ^1.3.15 so it was pulling in the recently released 1.4 version.  I checked in a change with the dependency set to >=1.3.x. although there may be a better version syntax I should have used.

PR #3720, opened 2015-05-27
fix(datepicker): do not focus input element after selecting date

JohannesRudolph:
The focus causes the soft-keyboard to popup in mobile safari, which is rather unpleasant for the user.
We could also make this a configuration option but I fear proliferation of options is not a good idea.


rvanbaalen:
Please rebase.

wesleycho:
In general, I think selecting the input is the correct behavior - this is due to how forms work, and if the form is set up right, it allows the user to just hit `enter` and submit the form.

Maybe supporting an optional callback that gets fired if it is present instead of changing this behavior completely would be better.

This feature would require a documentation change.

rvanbaalen:
@wesleycho In my opinion, we should keep the focus on the input as-is. Implementing a feature like this one will probably be bad for accessibility.

JohannesRudolph:
Well, in its current state at least it has a very bad user experience on
mobile causing the keyboard to pop up intermittently. Not sure how other
datepickers solve this?

That we want to have focus for submit on enter and accessibility makes
sense.
Am 05.06.2015 05:02 schrieb "Robin van Baalen" <notifications@github.com>:

wesleycho:
How about using an input with type="date"?

wesleycho:
Considering that this seems to be against all recommendations on accessibility for datepickers, I am closing this PR.

Any solution to this issue must not break browser accessibility as a default - [here](http://www.w3.org/TR/wai-aria-practices/#datepicker) is the WAI-ARIA specifications for datepickers, and [here](http://oaa-accessibility.org/example/15/) is an example of a datepicker that conforms to the guidelines.

PR #3727, opened 2015-05-28
feat(tabs): prevent tab switch

### LINKED ISSUE #2715: A way to intercept tab change
Hello,

I need a way to intercept a tab change. Here is a situation where such functionality would be useful:

I have two tabs. Each has a form inside. The requirement is that if form in the current tab becomes `dirty`, I need to show a popup message and depending on which option user chooses ((1) "save changes", (2) "discard changes", (3) "cancel") I should either (1) save and switch tab, (2) don't save and switch tab, or (3) don't save and stay on current tab.

Please let me know if that functionality already exists. If it does not, what approach do you think would be most suitable?

larvanitis:
Closes #2715.

### DIFF
--- a/template/tabs/tab.html
+++ b/template/tabs/tab.html
@@ -1,3 +1,3 @@
 <li ng-class="{active: active, disabled: disabled}">
-  <a href ng-click="select()" tab-heading-transclude>{{heading}}</a>
+  <a href ng-click="select($event)" tab-heading-transclude>{{heading}}</a>
 </li>
--- a/src/tabs/docs/readme.md
+++ b/src/tabs/docs/readme.md
@@ -38,3 +38,7 @@ AngularJS version of the tabs directive.
  * `deselect()`
  	_(Defaults: null)_ :
  	An optional expression called when tab is deactivated.
+
+ * `before-select()`
+ 	_(Defaults: null)_ :
+ 	An optional expression called before the tab is activated. The `ng-click` [event object is available as `$event`](https://docs.angularjs.org/guide/expression#-event-) so that you can use `$event.preventDefault()` to prevent the tab switch.
--- a/src/tabs/tabs.js
+++ b/src/tabs/tabs.js
@@ -192,6 +192,7 @@ angular.module('ui.bootstrap.tabs', [])
     scope: {
       active: '=?',
       heading: '@',
+      beforeSelect: '&', //Called before tab switch and can prevent it
       onSelect: '&select', //This callback is called in contentHeadingTransclude
                           //once it inserts the tab's content into the dom
       onDeselect: '&deselect'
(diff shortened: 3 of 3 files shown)


larvanitis:
Fixed missing $event in scope.select

rvanbaalen:
Seems like some good work to me. Can you also add some tests for this new feature?

template/tabs/tab.html
```diff
@@ -1,3 +1,3 @@
 <li ng-class="{active: active, disabled: disabled}">
-  <a href ng-click="select()" tab-heading-transclude>{{heading}}</a>
+  <a href ng-click="select($event)" tab-heading-transclude>{{heading}}</a>
```
wesleycho: With this change, documentation should be added for the `select` option as well.

larvanitis: This `select` is only used internally, and the event is passed to `beforeSelect` expression which is documented.

Do you mean document the `scope.select` function in [src/tabs/tabs.js](https://github.com/larvanitis/bootstrap/blob/prevent-tab-switch/src/tabs/tabs.js#L229) using JSDoc?

wesleycho: Ah, ignore my comment, I was mistaken.

wesleycho:
This just needs tests.

I still want to think whether this is the best approach, but otherwise this is generally ok.

wesleycho:
Any luck with tests @larvanitis?

larvanitis:
Sorry mate. I really don't have any free time yet. Could you have a take at tests?

wesleycho:
Shouldn't be too hard to write I think - in the `basics` describe section, one just needs something like
```javascript
it('should cancel selection of the tab', function() {
  scope.beforeSelect = function($event) {
    $event.preventDefault();
  });
  spyOn(scope, 'beforeSelect').and.callThrough();
  elm = $compile([
    '<tabset class="hello" data-pizza="pepperoni">',
    '  <tab heading="First Tab {{first}}" active="actives.one" select="selectFirst()" deselect="deselectFirst()">',
    '    first content is {{first}}',
    '  </tab>',
    '  <tab active="actives.two" select="selectSecond()" before-select="beforeSelect($event)" deselect="deselectSecond()">',
    '    <tab-heading><b>Second</b> Tab {{second}}</tab-heading>',
    '    second content is {{second}}',
    '  </tab>',
    '</tabset>'
  ].join('\n'))(scope);
  expect(scope.actives.one).toBe(true);
  expect(scope.actives.two).toBeFalsy();

  titles().eq(1).find('a').click();
  expect(scope.beforeEach).toHaveBeenCalled();
  expect(scope.actives.one).toBe(true);
  expect(scope.actives.two).toBeFalsy();
  expect(scope.deselectFirst).not.toHaveBeenCalled();
  expect(scope.selectSecond).not.toHaveBeenCalled();
});
```

I think this would be a sufficient test.

wesleycho:
The PR also needs to be rebased due to code style fixes in `master`.

wesleycho:
Closing this as outdated - feel free to open a new PR reimplementing this based on `master` if you feel it is a sensible feature

PR #3737, opened 2015-06-01
fix(dateParser): added date validity check

pedroclayman:
added a date validity check
added zeros as default values for hours, minutes and seconds in addition to milliseconds

### DIFF
--- a/src/dateparser/dateparser.js
+++ b/src/dateparser/dateparser.js
@@ -125,7 +125,7 @@ angular.module('ui.bootstrap.dateparser', [])
 
     if ( results && results.length ) {
       var fields, dt;
-      if (baseDate) {
+      if (baseDate && !isNaN(baseDate.getTime())) {
         fields = {
           year: baseDate.getFullYear(),
           month: baseDate.getMonth(),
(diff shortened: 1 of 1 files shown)


rvanbaalen:
Can you add some tests for this?

pedroclayman:
angular.isNaN does not exist as far as I know. Did you mean angular.isNumber ?

pedroclayman:
angular.isNumber does not behave the way I need it to. neither does angular.isDate
```javascript
var date = new Date('foo'); // invalid date
console.log(date); // Date { invalid Date }
console.log(date.getTime()); // NaN
console.log(angular.isNumber(date.getTime())); // true
console.log(angular.isDate(date)); // true
console.log(isNaN(date.getTime())); // true - this is how I identify an invalid date
```
this solution is based on this: http://stackoverflow.com/questions/1353684/detecting-an-invalid-date-date-instance-in-javascript

Do you see a more suitable approach?

rvanbaalen:
Ah well you're right. My bad. It was early in the morning :wink:

menssen:
This is a duplicate of both #3731 AND #3708, both of which have tests.

rvanbaalen:
Closing in favour of #3708. Thanks for reporting duplicate @menssen

Issue #3729, opened 2015-05-28
Just click quickly among the indicators to break the carousel


cho-is:
So, you just click quickly among the indicators to break the carousel.

Any clue?

rvanbaalen:
Bug confirmed. Clicking aggressively on the carousel indicators temporarily breaks the functionality.

PR #3757, opened 2015-06-04
fix(carousel): disable transition until animation completes

wesleycho:
- Force the carousel indicator to not select the slide if the animation has not completed
- Factor out `goNext` function to not redefine on each execution of `select`

Plunker of fix in action [here](http://plnkr.co/edit/tMmhMqs6mfRWUzcOh80D?p=preview).

This addresses #3729


wesleycho:
I'm willing to make that change, but it should be noted that both the `next` and `prev` methods check that property as well, and if it is falsey, then `select` is called. Theoretically, I would prefer to have it be inside the function instead of being parsed by Angular, but then it requires thought about those methods. I could move the check into `select` instead and delete those lines from the `next` and `prev` functions.

template/carousel/carousel.html
```diff
@@ -1,6 +1,6 @@
 <div ng-mouseenter="pause()" ng-mouseleave="play()" class="carousel" ng-swipe-right="prev()" ng-swipe-left="next()">
     <ol class="carousel-indicators" ng-show="slides.length > 1">
-        <li ng-repeat="slide in slides | orderBy:'index' track by $index" ng-class="{active: isActive(slide)}" ng-click="select(slide)"></li>
+        <li ng-repeat="slide in slides | orderBy:'index' track by $index" ng-class="{active: isActive(slide)}" ng-click="!$currentTransition && select(slide)"></li>
```
rvanbaalen: Shouldn't this !$currentTransition be done inside the select() instead of explicitly having to set it within the ng-click?

wesleycho: I'm willing to make that change, but it should be noted that both the `next` and `prev` methods check that property as well, and if it is falsey, then `select` is called. Theoretically, I would prefer to have it be inside the function instead of being parsed by Angular, but then it requires thought about those methods. I could move the check into `select` instead and delete those lines from the `next` and `prev` functions.

rvanbaalen: I would go for that indeed. Just keep the `currentTransition` variable as an internal flag like suggested also keeps the template cleaner. 

We promote overwriting templates to implement custom designs/functionality. By keeping `$currentTransition` out of the template, there is one less variable to keep in mind that could potentially break a custom template.

Issue #3701, opened 2015-05-22
Datepicker clicking clear button on popup gives "ng-model" value must be a Date object


gjmoyer:
If you have a Datepicker directive as a popup, and choose a valid date, then popup again and click the clear button, you get the following exception:
Datepicker directive: "ng-model" value must be a Date object, a number of milliseconds since 01.01.1970 or a string representing an RFC2822 or ISO 8601 date.

This then leaves the form.$invalid property as true, which is not correct, if this is not a required field.

Open the console window on the demo page, and follow the above 2 steps, and you will see the error emitted to console window.

rvanbaalen:
Plunkr please.

dkavun:
Looks like the problem is either in $viewChangeListener or in the dateParser: 'clear' passes an empty string as a selected value and the dateParser parses it to 'Invalid date'. It seems to me, that null should be returned in this case.

gjmoyer:
In datepicker.js line 652, I changed the code to this, and it appears to fix the issue.
      ngModel.$viewChangeListeners.push(function () {
          if (ngModel.$viewValue) scope.date = dateParser.parse(ngModel.$viewValue, dateFormat, scope.date) || new Date(ngModel.$viewValue);
          else scope.date = null;
      });

slavafomin:
This is a very annoying bug indeed. Is there a temporary workaround for this that we can use externally?

Thanks!

ruionwriting:
Reverted back to 0.12.0 for now.

DeWet-vanRooyen:
I still get this error when I use the format of 'dd/MM/yyyy'

when manually typing in the date and starting with a number greater than 12 it wont work.
Any ideas ? Thanks !

kenthamilton:
I am guessing this is happening because there is no date present on the page when it loads. What I ended up doing was setting the value to undefined instead of null if there is no value for that field. If there is a value then I convert it to a new Date object. I hope this code helps. I believe that the value cannot be null but it can be undefined. In my case I was setting the value as it was returned from the JSON object in my service like this:
_.each(_estimate.EstimateLineItems, function(li){
                        if(li.IssueDate == null) {
                            li.IssueDate = undefined;
                        } else {
                            li.IssueDate = new Date(li.IssueDate);
                        }
                    });

Issue #3713, opened 2015-05-26
DatepickerPopup error when changing input field


csvn:
When changing an input field with the datepicker-popup directive (in 0.13.0), an error is thrown when the starting value is a time string ('yyyy-MM-dd').

**TypeError: baseDate.getFullYear is not a function**
http://plnkr.co/edit/oKNbNp07cJompsK6NBVr?p=preview

This does not occur in 0.12.1:
http://plnkr.co/edit/GFHEvBbmsgfHKK7NrYeN?p=preview

Is this a bug or am I missing something?

gmarziou:
I experience same problem when I manually enter a date in the input text not using the popup.

wesleycho:
I have a proposed fix, if you take a look at the referenced PR above

gmarziou:
Great, thanks

bob-lee:
After upgrading from 0.14.3 to 1.1.2, I am getting "TypeError: this.activeDate.getFullYear is not a function". It seems like previous version handles both UTC string and Date object ok but 1.1.2 seems to have issue with UTC string?

thowar2:
Same, I had to pass it a new Date object, the date string was throwing an error.

bob-lee:
It turns out that ngmodel should be JavaScript date object. Fair enough. I was maintaining dates as utc strings for sorting.

icfantv:
Yes.  The API was changed to require the model to be a Javascript `Date` object due to issues that arose with how different browsers handle date strings - even ISO 8601 formatted ones.  The issues forum is literally littered with issues that discuss the reason for this so if you're looking for something more in-depth, please feel free to search that.

Also, @bob-lee, it's never a bad idea to go over the CHANGELOG for each version to look for breaking changes.  This was one such change, however it was made in 0.13.x and a regression was introduced that allowed strings to be used again - which was rectified in the 1.1.0 release (I believe).

ilaipi:
So, After the change, what's the right way to use this?
I have an element bind to foo.date, and I response
```javascript
[
    {
        name: 'bar', 
        date:"2016-08-28T16:00:00.000Z"
    },
    //......
]
```
So, I need to for-each my response and transform `date`?

bob-lee:
Between server and client, date is normally transfered as string whereas datepicker requires to feed date object, so string to date (on response / page load) and date to string (on post) transforms needed.

PR #3759, opened 2015-06-04
fix(dateparser): add type and validity check

wesleycho:
- Add type and validity check to ensure proper date object is created
- Make code style uniform in scripts

Plunker with fix in action [here](http://plnkr.co/edit/k54L20Ux3VkND7yAK2Kh?p=preview)

This addresses #3713


wesleycho:
I believe this also fixes #3701

rvanbaalen:
Leaving a comment to keep track on this one. Will review this saturday.

chrisirhc:
@wesleycho , there's a lot of whitespace changes in this PR. Any chance of editing that away? Just want to avoid adding noise to blame.

chrisirhc:
Code changes looks good to me.
I was only wondering if this should log a warning when the base date argument is provided but invalid (`isNaN` check returns true).

wesleycho:
Sure, I can do that sometime later today or tomorrow.

I can add a $log.debug or $log.warn statement about a string where isNaN returns true and falling back to date constructor.

wesleycho:
I reverted all code style changes

src/dateparser/dateparser.js
```diff
@@ -136,6 +136,9 @@ angular.module('ui.bootstrap.dateparser', [])
           milliseconds: baseDate.getMilliseconds()
         };
       } else {
+        if (baseDate) {
+          $log.warn('dateparser:', 'baseDate is a string with no date representation');
```
chrisirhc: How about `baseDate is not a valid date` ?

wesleycho: Much better than what I had - changing shortly

chrisirhc:
Hm, I was just testing this, I think this introduces another bug, now there's time added to the string, possibly due to a toISOString conversion that shouldn't be done:

See http://plnkr.co/edit/XCvfhGTYUUraB6mTSVSK?p=preview (there are steps to reproduce on the Plunker)

I'm thinking the change in https://github.com/angular-ui/bootstrap/pull/3294 is needed to actually fix this at the root.

wesleycho:
I tried some experimenting, and it looks like stripping out the `|| new Date(...)` portions from the $viewChangeListener and $parser does not fix this issue at all. In fact, this looks to be a preexisting issue, due to the fact that the view value gets converted to a date as the model value.

wesleycho:
#3643 may also be related.

cbosco:
+1

IRus:
Hi, don't understand this check:

```
if (angular.isDate(baseDate) && !isNaN(baseDate.getTime())) {
```
if baseDate is date getTime always will return an number, right?

wesleycho:
That is not true - try doing `new Date('foo').getTime()`

IRus:
Yep, you right, Thank you for example.

wesleycho:
@chrisirhc what are your thoughts on this? I do not think that the model aspect is fixable, since the model value will always become a date object or `null` currently (even if the `new Date(...)` fallback is stripped).

This can potentially be alleviated if we are able to switch to allowing users to bring in a custom parser & formatter for converting between a user desired model and the view value, which would also open up moment.js support for users, but that is a large & separate task.

I think this PR should make it in in its current incarnation though - it fixes a badly breaking bug in the datepicker.

IRus:
@wesleycho please remove || new Date(...) "fallback" that actually introducing new bugs.

wesleycho:
That is a breaking change that has explicit tests written against it - it is outside the scope of this PR.

anilnatha:
Is this in any way related to Issue #3617?

Reason I ask is because one of the behaviors of the bug illustrated in 3617 is that when you clear a datepicker field, it sets a form's $invalid property to false, even though the field is blank and the field is marked as _not_ required. I created a plunker illustrating the issue and is posted in that thread.

llafuente:
Partially related.
I apply this patch, and the exception is gone, but form is still $invalid.

wesleycho:
@llafuente I am not seeing the form being invalid incorrectly - [here](http://plnkr.co/edit/7HUwgHvV5OfT8qLbpiHR?p=preview) is an updated Plunker.

llafuente:
This patch alone do not solve: #3617

I apply this patch to 0.13, and didn't work, form was $invalid.
I can confirm master + this patch works.

wesleycho:
`master` is the only version that matters. There are a lot of changes since 0.13.0.

wesleycho:
Merging this in shortly, as this check in itself is completely safe & necessary.

PR #3758, opened 2015-06-04
showSpinners directive to toggle timepicker spinners

armincifuentes:
Useful for proper form alignment

PR #3761, opened 2015-06-04
showSpinners directive to toggle timepicker spinners

armincifuentes:
Directive for togglings spinners in order to vertically align a timepicker to other form controls in multi column forms.

**Example:**
explanation

Incorporates **one-time binding** for this directive, as per suggested by @wesleycho (#3758)


wesleycho:
This is now in via 1f760eb35f5bad1aa76663f92259f546fe762d66

PR #3762, opened 2015-06-04
fix(typeahead): select match on tab for iOS webview

### LINKED ISSUE #3699: [typeahead] dropdown doesn't hide on blur (ios only)
Hi,
when using the typeahead component in a form with multiple fields, the dropdown doesn't hide when the field looses focus.
This only occurs on ios (safari and uiwebview).

Steps to reproduce:
- Browse to http://angular-ui.github.io/bootstrap/#/typeahead with iOS safari on any device.
- Start typing 'te' with the virtual keyboard, the dropdown appears
- Either use the ">" next field button or "ok" button of the keyboard, the field loses focus and the dropdown is still displayed
- You can do the same with the next field, to finally have the three typeahead fields dropdown open
- You can also see the bug when tapping on another field, but this randomly leaves the dropdown open or closes all dropdowns.

Expected behavior:
The dropdown should be hidden as soon as the field looses focus, even if no dropdown value has been selected (to allow custom values).

wesleycho:
- Select match on blur if only it has not been selected upon keydown

The fix can be seen in action in this [Plunker](http://plnkr.co/edit/MaHsszvZZu7Vqm4JjrmN?p=preview).

This fixes #3699


Foxandxss:
damn iOS and damn safari. LGTM.

Issue #3403, opened 2015-03-18
Datepicker - form is invalid with Long or String values


apandichi:
After setting a Long or String value to the ng-model of a datepicker, the form $invalid value is set to 'true'. Please see this Plunkr for a demo of this behaviour: http://plnkr.co/edit/L4MpJxYB0OYkn80FB9hw

I'm not sure if this is the intended behaviour, but it seems rather odd that the input text value is formatted correctly to a Date, but the form is set to invalid.
Any thoughts? Thanks!

P.S. It's funny that the only time when the form seems to be valid is when the 'dd.MM.yyyy' formatting is not applied, when setting the Date value.

RobJacobs:
I think this issue is related to [this](https://github.com/angular-ui/bootstrap/pull/3322/files) pull request.

wesleycho:
I believe this issue is currently fixed in master - see the forked version of that Plunker based on current `master` [here](http://plnkr.co/edit/U3nuipb1Z0aLhLCRS536?p=info).

Let us know if this is still an issue.

apandichi:
Hi @wesleycho, thanks for your input.

In your new plunkr, the date input is no longer formatted when String and Long values are set. Also, the  openStartDate($event) button on the right does not work anymore. When running the plunkr, the JS console is throwing an error: "TypeError: Cannot read property 'initDate' of undefined"

Is this expected behaviour? I see that ui-bootstrap.js version is 0.13.1-SNAPSHOT, so I'm guessing not all issues have been ironed out?

In conclusion, what I'd like to know is if I can expect the datepicker to accept the String and Long values as valid values, and if the validity will be reflected by the form $invalid value.

Thanks a lot!

wesleycho:
Hmm, so it looks like

wesleycho:
I will take a look at that today then.

wesleycho:
I updated the Plunker and removed the datepicker-options attribute there, since it was set to `undefined`. It appears that the formatting works, but the validation is not right when the number is set to a string representation. This would be buggy behavior here still and is a legitimate issue.

chrisirhc:
I'm thinking we should consider dropping support for any input type that can't be parsed using the date format available described in https://docs.angularjs.org/api/ng/filter/date .

[...]

PR #3768, opened 2015-06-05
fix(datepicker): add safety for unix timestamps in string form

wesleycho:
- Fixes unix timestamps not being considered valid date strings
- Adds coercion to number if model value is a string that has numeric value

Plunker in action [here](http://plnkr.co/edit/U3nuipb1Z0aLhLCRS536?p=preview).

This addresses #3403.


chrisirhc:
Hm I think this creates an asymmetric API, added a comment in #3403 . We should definitely stop using the new Date constructor for the same reason.

wesleycho:
I think I agree - the datepicker needs to have all of its date logic stripped and moved to the parser.

PR #3770, opened 2015-06-08
datepicker: reduce watchers

### LINKED ISSUE #3451: DatePicker - Performance issues with popup + other stuff
Using one of Chromes add-ons for performance evaluation, I have seen this:

image

I have only one datepicker (if I am not mistaken) in sight, so this is quite disturbing...
I cannot imagine what happens with more...
Do you think it would be possible to remove the popup from the DOM when it is not needed to be visible?
I think that would help a lot.
(I am referring mainly to `datepickerPopup`, but it may apply to other places as well)

---

Also, the popup doesn't work so good when placed inside some containers having CSS positioning and floating.
If you could make it work good in all cases (e.g. positioning it with JS) it would be nice.
(Then you might face problems with scrolling while the popup is open, but that's bearable)

---

Also, another issue:
I was using an old version of your library and I came across some issues with the week numbers.
I have found a correct version of the `getISO8601WeekNumber()` that you use:

```
    var d = new Date(+this);
    d.setHours(0, 0, 0);
};
```

I have noticed that in newer versions of you lib, you have improved this function, but I think it is still not the same to the one I use.
If you like, take a look at this implementation and compare, just to make sure you got it right.

chrisirhc:
Mode | No. of Watchers Before Change | No. of Watchers After Change
---- | -------------------------- | ----------------------------
daypicker | 152 | 28
monthpicker | 35 | 13
yearpicker | 29 | 11

Resolves #3451 #2613

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -105,7 +105,7 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
       date: date,
       label: dateFilter(date, format),
       selected: model && this.compare(date, model) === 0,
-      disabled: this.isDisabled(date),
+      disabled: this.isDisabled(date) || false,
       current: this.compare(date, new Date()) === 0,
       customClass: this.customClass(date) || null
     };
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -549,20 +549,21 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
       angular.forEach(['minDate', 'maxDate', 'datepickerMode', 'initDate', 'shortcutPropagation'], function( key ) {
         if ( attrs[key] ) {
           var getAttribute = $parse(attrs[key]);
-          scope.$parent.$watch(getAttribute, function(value){
-            scope.watchData[key] = value;
-          });
+          var propConfig = {
+            get: function () {
+              return getAttribute(scope.$parent);
+            }
+          };
           datepickerEl.attr(cameltoDash(key), 'watchData.' + key);
 
           // Propagate changes from datepicker to outside
           if ( key === 'datepickerMode' ) {
             var setAttribute = getAttribute.assign;
-            scope.$watch('watchData.' + key, function(value, oldvalue) {
-              if ( angular.isFunction(setAttribute) && value !== oldvalue ) {
-                setAttribute(scope.$parent, value);
-              }
-            });
+            propConfig.set = function (v) {
+              setAttribute(scope.$parent, v);
+            };
           }
+          Object.defineProperty(scope.watchData, key, propConfig);
         }
       });
       if (attrs.dateDisabled) {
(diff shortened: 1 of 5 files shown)


src/datepicker/datepicker.js
```diff
@@ -727,6 +731,105 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
...
+          });
+        }
+        else {
+          expToData = {};
+        }
+      }
+    }
+  };
+}])
+
```
RobJacobs: Seems like this would be useful in other areas, could me move it to a service?

chrisirhc: I was thinking of using it here first then move it elsewhere only we have more uses, and when more people have looked at it, heh. Can you think of other places that might need this?
It's useful for classes that apply to one element/child rather than all, in an ng-repeat, to reduce watchers.

RobJacobs: Carousel, pagination, and typeahead are using ng-class to match an active state.  I have a few directives outside of angular-ui-bootstrap that I thnk could benefit from this approach.

wesleycho: This would be pretty nifty to release as a standalone library, although that is probably outside the scope of this PR.

wesleycho:
This needs tests, but otherwise looks generally good to me.

chrisirhc:
Indeed, the new directive needs tests. I'll work on that. We can port it to other libraries later, once I'm done with adding tests. This has most gain on the default use of datepicker so it makes sense for it to have it first.

src/datepicker/datepicker.js
```diff
@@ -727,6 +731,105 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
...
+                newActivated = instance;
+                return true;
+              }
+            });
+            if (data.lastActivated !== newActivated) {
+              if (data.lastActivated) {
+                $animate.removeClass(data.lastActivated.element, clazz);
+              }
+              if (newActivated) {
+                newActivated.element.attr('has-been-activated', 'true');
```
chrisirhc: I'll be removing this. It was a debug statement.

jaskaye17:
+1 Let's get this badboy live!

jaskaye17:
This PR looks good in terms of reducing the number of watchers. The issue that I see for many users is that the popup is always rendered even when not visible. The display is simply set to none. This leaves all of the watchers. When using multiple datepickers on a single page (or in a grid as some users have done) the number of watchers still becomes unwieldy even with the above reduced number of watchers. Are there any plans to conditionally render (ng-if) the popup when clicked. 

I see that this ticket does in fact accomplish the task of reducing the watcher count on the datepicker and is not exactly tailored to the use case of multiple watchers. I ask here, however, because all datepicker watcher related tickets seem to lead me here

RobJacobs:
@jaskaye17 ng-if was added to the datepicker popup under this commit: https://github.com/angular-ui/bootstrap/commit/b72efed56f68e9e5dbed934df7f47e0b5c94242c

PR #3771, opened 2015-06-08
Code element wrapped around rendered promise

sime:


wesleycho:
This is an autogenerated page - can you file your change against `master` editing https://github.com/angular-ui/bootstrap/blob/master/src/modal/docs/readme.md ?

Issue #2421, opened 2014-07-04
Nested dropdowns


Yankovsky:
Any plans to support nested dropdowns? I made example [jsfiddle](http://jsfiddle.net/5KsLW/1/). Currently, only one dropdown can be open at a time due to [dropdownService's open method](https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L10).
I tried to make a use of ```require: '^dropdown'``` in child dropdown directive, but it is returning child dropdown controller, not his parent because of [bug](https://github.com/angular/angular.js/issues/4518).
So, any ideas how to handle this situation?
Thank you.

lazercore:
Anyone?

theaqua:
Anyone?

dmitriz:
Here is a pure CSS implementation if that helps:
http://bootsnipp.com/snippets/featured/multi-level-dropdown-menu-bs3

RobJacobs:
@dmitriz Nice find, thanks for sharing.

sebastianhenneberg:
I added nesting capabilities to the dropdowns and created a PR. Please check out the referenced PR #3776 or grab my [monkey-patch](https://gist.github.com/sebastianhenneberg/acb0a5e7b0e63006c699). Any feedback is highly appreciated!

wesleycho:
I put some more thought about this, and I am making the decision to reject this feature request - this adds complexity for what is a bad UX, which is a net negative for the project.

shyamal890:
@wesleycho WHY?!!! This is very much needed. There are use cases where you would need a dropdown inside a dropdown.

wesleycho:
@shyamal890 the amount of code required to support this, which will have lots of side effects, is not justified with why this is a valid UI/UX for the user.

azahreba:
Actually, I fixed it in 3 lines in JS, without any tricks with CSS or HTML.
https://github.com/azahreba/bootstrap/commit/b86cbda99275670e27aa94f293423ee7cef303d7 (a simple factory pattern is used there).
Now there are 2 failed unit-tests and that's the question - how to fix them.
Because I've tried to reproduce those unit tests on the real app, and everything is working properly with ```class="open"```.

Mike-Loffland:
@sebastianhenneberg Thanks for the monkey patch... works like a charm.

notbrain:
[...]

PR #3776, opened 2015-06-08
feat(dropdown): add nesting capability

sebastianhenneberg:
Adds nesting support for dropdowns as discussed in #2421. Allows to reflect deep hierarchies using dropdowns. The implementation uses an array as a stack to keep track of opened dropdown scopes.


RobJacobs:
@sebastianhenneberg There is a css solution for implementing this feature [here](http://bootsnipp.com/snippets/featured/multi-level-dropdown-menu-bs3)

sebastianhenneberg:
@RobJacobs I have seen the CSS solution in the issue already. You get this with pure bootstrap (without angular-ui/bootstrap). I required a solution that utilizes the dropdown directive and allows me to customize the behaviour.

wesleycho:
I am not sure we want to support something like this due to poor UX, although if I remember correctly, Bootstrap also had some controversy related to this as well.

wesleycho:
Closing as per https://github.com/angular-ui/bootstrap/issues/2421#issuecomment-140869946

PR #3777, opened 2015-06-08
feat(datepicker): optional force-position attribute

hxtpoe:
Specify where the datepicker element should be displayed


hxtpoe:
sth wrong with branch

PR #3782, opened 2015-06-09
feat(datepicker): add optional force-position attr

hxtpoe:
Specify where the datepicker element should be displayed

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -471,7 +471,9 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
       clearText: '@',
       closeText: '@',
       dateDisabled: '&',
-      customClass: '&'
+      customClass: '&',
+      position: '@'
+
     },
     link: function(scope, element, attrs, ngModel) {
       var dateFormat,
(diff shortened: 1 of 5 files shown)


hxtpoe:
@rvanbaalen could you look at the PR in free-time? That is the next time when I need this feature, so I've decided to implement. Let me know if you've got some suggestions.

src/datepicker/datepicker.js
```diff
@@ -471,7 +471,8 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
       clearText: '@',
       closeText: '@',
       dateDisabled: '&',
-      customClass: '&'
+      customClass: '&',
+      forcePosition: '@'
```
wesleycho: I think just calling it `position` would be fine.

src/datepicker/datepicker.js
```diff
@@ -682,7 +683,28 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
         if (value) {
           scope.$broadcast('datepicker.focus');
           scope.position = appendToBody ? $position.offset(element) : $position.position(element);
-          scope.position.top = scope.position.top + element.prop('offsetHeight');
+
+          var clearPosition = function() {
+            scope.position.top = scope.position.bottom = scope.position.left = scope.position.right = 'initial';
+          };
+
+          if(!scope.forcePosition || scope.forcePosition === 'left-bottom') {
```
wesleycho: Add space before `(`

wesleycho: Also this is more semantics, but `bottom-left`, `bottom-right`, `top-right`, and `top-left` are better strings to use.

src/datepicker/datepicker.js
```diff
@@ -682,7 +683,28 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
...
+            clearPosition();
+            scope.position.left = 0;
+            scope.position.top = $position.position(element).height  + 'px';
+          } else if(scope.forcePosition === 'right-bottom') {
```
wesleycho: Add space before `(`

src/datepicker/datepicker.js
```diff
@@ -682,7 +683,28 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
...
+            clearPosition();
+            scope.position.right = 0;
+            scope.position.top = $position.position(element).height  + 'px';
+          } else if(scope.forcePosition === 'right-top') {
```
wesleycho: Add space before `(`

src/datepicker/datepicker.js
```diff
@@ -682,7 +683,28 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
...
+            clearPosition();
+            scope.position.right = 0;
+            scope.position.bottom = $position.position(element).height + 'px';
+          } else if(scope.forcePosition === 'left-top') {
```
wesleycho: Add space before `(`

wesleycho:
History needs to be fixed, there is pollution from what appears to be other repositories. The base branch should branch directly from `master` on this repository, not any other.

hxtpoe:
-history fixed
-CS fixed
-attr name changed from force-position to position and options renamed

@wesleycho could you take a look?

wesleycho:
I'll take a harder look later today - my only main point of concern here is multiple PRs dealing with positioning but not implementing the feature in `$position` itself. I may look at tackling the positioning issue in `$position` so we do not run into these edge issues in each component & duplicate the logic.

wesleycho:
Thinking about this, I believe that the API should be tweaked to support single directions as well, i.e. `top`, `right`, `bottom`, and `left`. The attribute should take a string, i.e. `top left` to specify top and left, or `top` for just top (among other possibilities).

`$position.offset` should take a second parameter that takes a string of the form described above and parses it to determine how to set the offset. It should behave as it currently does otherwise. This portion should be in its own PR.

wesleycho:
Doing a little more investigation, `$position.offset` should not be modified - I am in the process of investigating an appropriate modification in `$position` to accommodate the proper approach in the datepicker for this feature.

I should have a PR ready by tomorrow with that modification.

chrisirhc:
Looking forward to $position changes. Thanks @wesleycho .

icfantv:
Closing in favor of #4762.

PR #3792, opened 2015-06-10
fix(dropdown): do not autoclose with outsideClick and append to body

### LINKED ISSUE #3645: Dropdown auto-close outsideClick not working when used with append-to-body
Simple, when `append-to-body` is set, and `auto-close` is set to `outsideClick`, closing behaves like if it has the default value (`always'), I believe it's somewhat related to catching the "onclick" event attached to the wrong element.

Plunkr: http://embed.plnkr.co/WpARFC/preview

wesleycho:
- Do not auto-close the dropdown menu when it is set to `outsideClick` and dropdown-append-to-body is used

Plunker of fix in action [here](http://plnkr.co/edit/yxHanZ?p=preview).

This fixes #3645.


Foxandxss:
Looks good to me.

PR #3795, opened 2015-06-10
fix(carousel): change to avoid references to debug info

### LINKED ISSUE #3794: Carousel depends on $compileProvider.debugInfoEnabled(true);
Been debugging this for some time. Found that our carousels are broken because we disable debugInfo for performance within our live environment. It seems the carousel depends on debugInfoEnabled to be true. `element.parent().scope()` is unavailable when debguInfo is disabled.

```
TypeError: Cannot read property 'noTransition' of undefined
    at Object.beforeRemoveClass [as fn] (carousel.js:324)
    at angular-animate.js:741
    at forEach (angular.js:339)
    at run (angular-animate.js:726)
    at Object.angular.module.directive.factory.config.$provide.decorator.before (angular-animate.js:767)
    at performAnimation (angular-animate.js:1410)
    at angular-animate.js:1215
    at angular-animate.js:538
    at Scope.$get.Scope.$digest (angular.js:14092)
    at Scope.$get.Scope.$apply (angular.js:14304)
```

wesleycho:
- Change usage of `.scope()` and `.isolateScope()` to grab metadata from elements

Plunker with fix in action [here](http://plnkr.co/edit/e1jjlHwxFne6JipgqFZN?p=preview).

This fixes #3794.


chrisirhc:
Nice. LGTM

PR #3797, opened 2015-06-11
docs(modal): add note about `$modalProvider`

### LINKED ISSUE #1759: docs(modals): $modalProvider.options is undocumented
For a while I thought it was impossible to set global defaults. Even a google search for modalProvider.options is lacking helpful info. I think it would be helpful for others to know how this works. A simple snippet like this would suffice:
```
angular.module('myApp')
  .config(['$modalProvider', function($modalProvider) {
    $modalProvider.options = {
      backdrop: 'static'
    };
  }])
```
I can try to find time to submit a PR for this sometime before the end of the week

wesleycho:
- Add note about setting global defaults via `$modalProvider.options`

Closes #1759


Foxandxss:
The more doc, the better.

PR #3805, opened 2015-06-11
chore(typeahead): change to one-way binding for position

wesleycho:
- Change position to one-way binding to reduce an unnecessary $watch


wesleycho:
Here is a [Plunker](http://plnkr.co/edit/nrG2bPHNm5ehEwViJD2c?p=preview) with this in action.

These changes speed up the typeahead component

chrisirhc:
Nit-picking but this should've been kept as two separate commits.
Anyways, LGTM

wesleycho:
Landed in 49e73a89f36b1c44d7547b8910bcb17e96135294

PR #3806, opened 2015-06-11
fix(modal): defer promise resolution until animation starts

### LINKED ISSUE #3787: Modal dialog does not disappear with 0.13 and Angular 1.3 when destroying a passed scope
Angular version: 1.3.16
ui-bootstrap version: 0.13.0

This is related to, but not the same as #3633 and #3620. If you create a new scope to pass to $modal.open, it needs to be destroyed when the dialog closes to prevent memory leaks. The only event we currently get to perform this action is the result promise. However, this resolves before the closing animation is executed. If you destroy the new scope at that point, the animation cannot execute, because of its dependency on ngClass to remove the 'in' CSS class.

Here is a Plunkr that shows the behavior: http://plnkr.co/edit/yaFvaOGu1NUxkzXwfFuJ?p=preview

This same Plunkr works when using version 0.12.1, so it is a breaking change.

wesleycho:
- Defer resolution of modalInstance promise until the animation cycle starts

It turns out that the promise resolution completes before the animation cycle starts, so it is possible for DOM elements to become stranded if it is desired for the $scope passed in to be destroyed.

Plunker of fix in action [here](http://plnkr.co/edit/NsEDlDDX772CIhTmlrDg?p=preview).

Fixes #3787


chrisirhc:
Hm.. testing this might be quite challenging. Perhaps add a test that the promises are resolved after the animation has ended?

wesleycho:
The promises do not resolve before the animation has ended - they just resolve after the animation starts.

wesleycho:
I added tests for this case and rebased

PR #3807, opened 2015-06-11
chore(pagination): improve performance

wesleycho:
- Use bind once for non-watched text blocks

According to the [docs](http://angular-ui.github.io/bootstrap/#/pagination), we are not currently supporting watching of text changes for the boundary links and direction links. This change enforces bind once on those, so it is technically a breaking change, but helps improve the performance of this component.


Foxandxss:
LGTM

PR #3808, opened 2015-06-12
chore(popover): add bind once to title

wesleycho:
- Add bind once to title bindings


wesleycho:
Closing this as I think losing the dynamic title and content is the wrong approach.

PR #3809, opened 2015-06-12
chore(timepicker): add bind once to template

wesleycho:
- Add bind once to appropriate values not being watched in template


Foxandxss:
LGTM

PR #3812, opened 2015-06-12
chore(tooltip): add bind once to content

wesleycho:
- Add bind once to content bindings


chrisirhc:
I think this might be a breaking change. Some users might be expecting the content to be modifiable.

wesleycho:
I was thinking the same - I'm fine with this not being merged if it's a big enough concern.

wesleycho:
Rejecting this, as I think we don't actually want to one-way bind this - users can override the template if they are interested in a perf gain.

PR #3830, opened 2015-06-17
fix(progressbar): use max value on stacked progress bar

### LINKED ISSUE #3618: Progress + Bar not respecting the Max value set on Progress
I've got a max value set on 'progress' element (with many 'bar' elements beneath it) and the value is never making it's way into the actual controller for the entire progress bar. This means that the max will always be the default value of 100 for a 'progress' element and incorrectly calculate the width.

I believe this is due to a recent change in ProgressController where it is attempting to default $scope.max if not set by doing $scope.max = $scope.max || default. In the previous version of the code it would look in $attrs for max instead ($scope.max = $attrs.max || default). The reason the new version does not work is because for the 'progress' directive (but not the ProgressBar directive), there is no 'max' variable defined in the scope, thus $scope.max will never be set for this directive. It used to work before because it was looking in $attrs.

Example: http://plnkr.co/edit/K2Y3G39IvP8VJbgwl06D?p=preview
If you inspect the progress bar, you can see that it's taking up 100% width even though it's value is set to 186 out of 211 (211 being the maximum set). There is also another bar beneath it that is invisible with the value of 25.

I believe to fix this while maintaining the current implementation,  'max: "=?"' should be added to the scope object of the 'progress' directive. Although I don't know the full ramifications of this and there certainly are other ways (such as reverting to $attrs again).

EDIT: trying to fix some terrible english.

groupsky:
Fixes #3618

### DIFF
--- a/src/progressbar/progressbar.js
+++ b/src/progressbar/progressbar.js
@@ -39,14 +39,12 @@ angular.module('ui.bootstrap.progressbar', [])
         this.bars.splice(this.bars.indexOf(bar), 1);
     };
 
-    $scope.$watch('max', (function(self){
-        return function(max) {
-            self.bars.forEach(function (bar) {
-                bar.max = $scope.max;
-                bar.recalculatePercentage();
-            });
-        };
-    })(this));
+    $scope.$watch('max', function(max) {
+        self.bars.forEach(function (bar) {
+            bar.max = $scope.max;
+            bar.recalculatePercentage();
+        });
+    });
 }])
 
 .directive('progress', function() {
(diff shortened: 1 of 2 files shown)


src/progressbar/progressbar.js
```diff
@@ -32,6 +38,15 @@ angular.module('ui.bootstrap.progressbar', [])
     this.removeBar = function(bar) {
         this.bars.splice(this.bars.indexOf(bar), 1);
     };
+
+    $scope.$watch('max', (function(self){
```
wesleycho: This block is a bit awkward - why not use the `self` variable defined at the top of the controller?

wesleycho:
Other than that minor issue, this LGTM.

src/progressbar/progressbar.js
groupsky: didn't notice the reference. updated the PR

PR #3833, opened 2015-06-17
feat: timepicker - fire 'blur' on parent element when fired on inputs

hryanjones:
Having the ability to add an ng-blur directive directly on the `<timepicker>` element is very useful for validating without annoying the user. This allows this to happen by firing a blur event on the parent when either the hour or minute input is blurred.

I wasn't able to run the test. I sent an email to angular-ui@googlegroups.com , but I got a bounce notification.

Here's the email:

> Hi,
> 
> I'm trying to make a small change to the timepicker to allow blur events on the hour and minute inputs to be bubbled up to the top-level event, but I'm having trouble running the tests on Ubuntu 14.04 with both chromium-browser and google-chrome installed.
> 
> Grunt complains about not being able to open Chrome. [1]
> It appears to be looking for the right binaries according to [2], I put in a console.log at this point, so it didn't appear to be [related to this issue][3] I found.
>
> Any pointers?
>
> Thanks,
Ryan

[1]
```
$ grunt --force
Running "enforce" task

Running "ddescribe-iit:files" (ddescribe-iit) task

Running "jshint:files" (jshint) task
>> 66 files lint free.

Running "html2js:dist" (html2js) task
Successfully converted 29 html templates to js.

Running "test" task

Running "karma:continuous" (karma) task
INFO [karma]: Karma v0.12.36 server started at http://localhost:9876/
INFO [launcher]: Starting browser Chrome
ERROR [launcher]: Cannot start Chrome
	
INFO [launcher]: Trying to start Chrome again (1/2).
ERROR [launcher]: Cannot start Chrome
	
INFO [launcher]: Trying to start Chrome again (2/2).
ERROR [launcher]: Cannot start Chrome
	
ERROR [launcher]: Chrome failed 2 times (cannot start). Giving up.
Warning: Task "karma:continuous" failed. Used --force, continuing.

Running "build" task
File dist/ui-bootstrap-0.13.1-SNAPSHOT-csp.css created

Running "concat:dist" (concat) task
File dist/ui-bootstrap-0.13.1-SNAPSHOT.js created.

Running "concat:dist_tpls" (concat) task
File dist/ui-bootstrap-tpls-0.13.1-SNAPSHOT.js created.

Running "uglify:dist" (uglify) task
>> 1 file created.

Running "uglify:dist_tpls" (uglify) task
>> 1 file created.

Running "makeModuleMappingFile" task
File dist/assets/module-mapping.json created.

Running "makeRawFilesJs" task
File dist/assets/raw-files.json created.

Running "copy:demohtml" (copy) task
Copied 1 file

Running "copy:demoassets" (copy) task
Created 2 directories, copied 15 files

Done, but with warnings.
```

[2]
```
$ cat node_modules/karma-chrome-launcher/index.js | grep google-chrome
    linux: getBin(['chromium-browser', 'chromium', 'google-chrome', 'google-chrome-stable']),
    linux: 'google-chrome-canary',
```

[3]: https://github.com/yeoman/yeoman/issues/1031#issuecomment-18867693

### DIFF
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -189,7 +189,7 @@ angular.module('ui.bootstrap.timepicker', [])
           $scope.hours = pad( $scope.hours );
         });
       }
-      parentEl.triggerEvent('blur', e);
+      parentEl.triggerHandler('blur', e);
     });
 
     $scope.updateMinutes = function() {
@@ -209,7 +209,7 @@ angular.module('ui.bootstrap.timepicker', [])
           $scope.minutes = pad( $scope.minutes );
         });
       }
-      parentEl.triggerEvent('blur', e);
+      parentEl.triggerHandler('blur', e);
     });
 
   };
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -16,7 +16,7 @@ angular.module('ui.bootstrap.timepicker', [])
       ngModelCtrl = { $setViewValue: angular.noop }, // nullModelCtrl
       meridians = angular.isDefined($attrs.meridians) ? $scope.$parent.$eval($attrs.meridians) : timepickerConfig.meridians || $locale.DATETIME_FORMATS.AMPMS;
 
-  this.init = function( ngModelCtrl_, inputs ) {
+  this.init = function( ngModelCtrl_, inputs, parentEl ) {
     ngModelCtrl = ngModelCtrl_;
     ngModelCtrl.$render = this.render;
 
@@ -38,7 +38,7 @@ angular.module('ui.bootstrap.timepicker', [])
     }
 
     $scope.readonlyInput = angular.isDefined($attrs.readonlyInput) ? $scope.$parent.$eval($attrs.readonlyInput) : timepickerConfig.readonlyInput;
-    this.setupInputEvents( hoursInputEl, minutesInputEl );
+    this.setupInputEvents( hoursInputEl, minutesInputEl, parentEl );
   };
 
   var hourStep = timepickerConfig.hourStep;
@@ -154,7 +154,7 @@ angular.module('ui.bootstrap.timepicker', [])
     });
   };
 
-  this.setupInputEvents = function( hoursInputEl, minutesInputEl ) {
+  this.setupInputEvents = function( hoursInputEl, minutesInputEl, parentEl ) {
     if ( $scope.readonlyInput ) {
       $scope.updateHours = angular.noop;
       $scope.updateMinutes = angular.noop;
@@ -189,6 +189,7 @@ angular.module('ui.bootstrap.timepicker', [])
           $scope.hours = pad( $scope.hours );
         });
       }
+      parentEl.triggerEvent('blur', e);
     });
 
     $scope.updateMinutes = function() {
@@ -208,6 +209,7 @@ angular.module('ui.bootstrap.timepicker', [])
           $scope.minutes = pad( $scope.minutes );
         });
       }
+      parentEl.triggerEvent('blur', e);
     });
 
   };
@@ -259,10 +261,10 @@ angular.module('ui.bootstrap.timepicker', [])
     selected.setHours( dt.getHours(), dt.getMinutes() );
     refresh();
   }
-  
+
   $scope.showSpinners = angular.isDefined($attrs.showSpinners) ?
     $scope.$parent.$eval($attrs.showSpinners) : timepickerConfig.showSpinners;
-  
+
   $scope.incrementHours = function() {
     addMinutes( hourStep * 60 );
   };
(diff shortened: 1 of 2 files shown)


hryanjones:
Also, I figured out that tests weren't running because I was in a remote terminal session, however, when I run them now 430 out of 785 tests fail. That doesn't seem right...

hryanjones:
The errors are only of a few types:

```fish
$ag "Ubuntu 0.0.0" /tmp/bootstrap_test.log -A 1 | grep -v "Ubuntu 0.0.0"| sed -e 's/^[0-9]*- *\t*//' | sort | uniq -c | sort -n
      1 Warning: Task "karma:continuous" failed. Use --force to continue.
     12 Error: [$injector:unpr] Unknown provider: $templateRequestProvider <- $templateRequest <- tooltipTemplateTranscludeDirective
     26 TypeError: Cannot read property 'then' of undefined
     40 Error: [$injector:unpr] Unknown provider: $templateRequestProvider <- $templateRequest
     48 Error: [$injector:unpr] Unknown provider: $templateRequestProvider <- $templateRequest <- $modal
     54 Error: [$parse:syntax] Syntax Error: Token ':' not a primary expression at column 1 of the expression [::match.id] starting at [::match.id].
     55 Error: [$parse:syntax] Syntax Error: Token ':' not a primary expression at column 1 of the expression [::showSpinners] starting at [::showSpinners].
     67 TypeError: Cannot set property 'date' of undefined
    128 Error: [$parse:syntax] Syntax Error: Token ':' not a primary expression at column 1 of the expression [::uniqueId] starting at [::uniqueId].
```

src/timepicker/timepicker.js
```diff
@@ -189,6 +189,7 @@ angular.module('ui.bootstrap.timepicker', [])
           $scope.hours = pad( $scope.hours );
         });
       }
+      parentEl.triggerHandler('blur', e);
```
wesleycho: You can just do `parentEl.triggerHandler(e)`.

src/timepicker/timepicker.js
```diff
@@ -208,6 +209,7 @@ angular.module('ui.bootstrap.timepicker', [])
           $scope.minutes = pad( $scope.minutes );
         });
       }
+      parentEl.triggerHandler('blur', e);
```
wesleycho: Ditto

wesleycho:
Thinking about this, this is not the correct way to implement such a feature. `blur` has a specific meaning when talking about DOM events, and this abuses the event since the element still has focus.

If such an end goal is desired, another way to implement it must be found.

I am closing this PR as a result - feel free to open a new PR with a better implementation.

hryanjones:
Don't suppose you have any ideas of a better implementation?

> this abuses the event since the element still has focus.

I'm not sure I understand this, do you mean if someone changes focus from the Hour to the Minute input that the event would fire? In that case it doesn't seem like it would be hard to put in a check for when the "other" input has focus and not fire on the parent...

wesleycho:
It still would be incorrect to do so - it breaks the native DOM api developers expect. Any implementation should not break expected browser API.

hryanjones:
Okay, that blanket statement without any explanation (or at least link to more info) was not very helpful, and you didn't seem to attempt to answer either of my questions.

Your whole response here is leaving me feeling pretty rebuffed and not super willing to contribute or really even use this library in the future. If you're fine with that, then please ignore me; I just feel this is not a great way to treat potential contributors...

wesleycho:
My point is your implementation breaks a fundamental contract a library should have with developers in not breaking how the browser operates. If blur is fired on a child element, it does not bubble. It only fires on the element if the element itself has lost focus, so it should not fire on the element of the directive unless that element itself has lost focus. This implementation breaks that expectation.

I appears that your expectation is for me to give a thorough explanation on how the browser works & a tantrum is thrown when it is not given - there are plenty of more appropriate channels for that sort of learning & detail than a pull request with no issue attached to it, including almost any JS related channel on FreeNode on IRC or #angularjs on FreeNode. Frankly, this demanding attitude without doing the proper research makes a developer (in this case myself) wanting to quit working on OS since it is disrespectful & irresponsible, especially trying to frame it as a developer's fault when the potential contributor has made a modification involving a browser API without understanding how said browser API works.

The only thing more I can do for you is link you to [the MDN documentation on blur](https://developer.mozilla.org/en-US/docs/Web/Events/blur). This is not the place for such a discussion though. If you want such a potential feature to be discussed, please open an issue ticket about your desired features and more guidance can be given about implementation there, as your base request does not seem unreasonable. I would be happy to review & accept a pull request that is refined from collaboration on how the API should look & be implemented.

pkozlowski-opensource:
> Frankly, this demanding attitude without doing the proper research makes a developer (in this case myself) wanting to quit working on OS

@wesleycho please don't! You are doing awesome job! 

Human interactions are hard and even more so on the net, where we can't see each other. But I agree with you that there is no room for demands in OS - for me GitHub is a place where _developers_ meet to work _together_ on pieces of software they judge useful / interesting. This requires time, dedication and skill. There is no room for any demand here as many OS developers contribute not only code but also they free time that could be spent with family etc.

hryanjones:
I apologize, I was just upset as I was excited to try and contribute to this library which I've been using, and felt summarily shot down -- I should've waited a day to cool down before responding.

I [created an issue](https://github.com/angular-ui/bootstrap/issues/new) to discuss. Thank you for your responses.

PR #3834, opened 2015-06-18
feat(datepicker): custom display text of null date values

overule:
A small feature that allows custom text to be displayed when the date value is null. e.g. Displaying "Not Specified" rather than an empty string in the underlying text input.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -455,6 +455,7 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   currentText: 'Today',
   clearText: 'Clear',
   closeText: 'Done',
+  nullDateText: '',
   closeOnDateSelection: true,
   appendToBody: false,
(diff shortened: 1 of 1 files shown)


wesleycho:
I don't think this is a good idea - this mutates the value of the input, which increases complexity when supporting inputs of type text & date, as well as managing the interplay between the view value and the model value.

tystol:
@wesleycho  Thanks Wesley, fair enough.

Any ideas how to achieve our requirements using the datepicker as-is? Note in our actual case the datepicker represents a 'Cease Date' and we need to handle a 'no cease date' case. We've changed the Clear button's text to 'Ongoing', and we wanted to be able to display 'Ongoing' or something to that effect within the input. (which in the model is represented by a null ceaseDate).

wesleycho:
Is there anything wrong with simply using an appropriate placeholder text via the input placeholder attribute?

wesleycho:
Closing this as per prior comments - unfortunately this would require a lot of complexity to maintain this feature.

Issue #2618, opened 2014-08-21
min-mode doesn't work with datepicker popup


Zacharias3690:
fix was pushed for datepicker-mode on popup in #2180, but min-mode still doesn't work in 0.12.0.

because 0.12.0 isn't available on plunker, here's my markup for reference

    <input type="text" class="form-control" 
                                datepicker-popup="MMMM yyyy" 
                                datepicker-mode="'month'" 
                                min-mode="'month'" 
                                ng-model="vm.endDate" 
                                is-open="endOpen" 
                                ng-click="endOpen=true" />

kannajune:
thanks suksit 
its working well

davidhantson:
I'm experiencing the same issue as johncronan,
any news on this? I really need to change the minMode on the fly, I have some graphs depending on it.
Thanks for the nice work on the component though, it really saves my day.

prettycode:
+1 `min-mode` isn't respected, whether it's specified as attribute or in `datepicker-options` attribute value object.

Slimmmo:
The min-mode problem is very odd. Would love a fix.
http://plnkr.co/edit/CoqkFvOM7GGNPHzb3iGq?p=preview
Thanks.

rvanbaalen:
Feel free to submit a PR, @Slimmmo

Slimmmo:
I think I've fixed the issues in this thread with my [PR](https://github.com/angular-ui/bootstrap/pull/3843). It fixed my problem at least. Enjoy.

urvashi-bhatt:
It is Possible to set [min-mode="week"](url) ?

PR #3843, opened 2015-06-22
datepicker dynamic min-mode and max-mode

Slimmmo:
Simply enabling the ability to change modes dynamically without having to use two different datepickers and hiding one based on the current mode.

Also fixes min-mode for popups. #2618


wesleycho:
Can you add a test for this? I will merge this afterwards - otherwise this looks good.

Slimmmo:
Hi Wesley,

I made a little change since you posted to make it slightly more fool-proof. If you update min-mode to something larger than datepicker-mode, it will now automatically update datepicker-mode to match min-mode. Likewise for max-mode but for smaller values.

Please let me know if those tests aren't adequate. There isn't really that much to test though.

Thanks.

wesleycho:
I need to review this deeper - the split off of the `minMode` and `maxMode` from the original iterator into a custom one has me wary.

wesleycho:
Thanks for this work - in the future though, please squash the commits. It ended up being quite a pain resolving the conflict multiple times because the commits were never squashed.

PR #3853, opened 2015-06-24
fix(dateparser): do not try to process invalid baseDate

cbosco:
This is causing a lot of issues for me when I manually enter date strings. This could also be related to issues reported with leading zeroes.

[Plunker](http://plnkr.co/Bx7RtEwbEc6smFQDsZho)

Enter the date "2001-01-01" into the input field and see that in the end the model becomes an invalid date.

The root of the issue is the parsing of `baseDate` inside dateparser.

The string '2001-01-0' is an invalid date, passed as baseDate to the next input, '2001-01-01' which *is* valid but gets corrupted by the previous baseDate because it is assumed to be valid


cbosco:
Fixed by 4f1e03f1147b9a50566df064db72522a130f8135

PR #3864, opened 2015-06-27
fix(alert): adjust check for close attribute

### LINKED ISSUE #3848: Alert always shows close button
The alert directive always shows the close button ('x'), even if no close attribute is present.
This behaviour only occurs when using Angular.js v. 1.4.1. With version 1.4.0 everything is Ok.

To reproduce:
Use Angular.js v. 1.4.1 and use alert directive without 'close' attribute.

wesleycho:
- Adjust check due to change in Angular 1.4 where the presence of attribute gives key value of `undefined`, causing variable to contain incorrect truthiness

This fixes #3848 in a way that does not conflict with Angular 1.3.

A Plunker with the 1.4 fix can be seen in action [here](http://plnkr.co/edit/d7xnZbFTKdtHBo0czIiK?p=preview).

A Plunker with the fix applied to 1.3 can be seen in action [here](http://plnkr.co/edit/GEM6UYZouFZfl4yOXCQX?p=preview).


src/alert/alert.js
```diff
@@ -1,7 +1,7 @@
 angular.module('ui.bootstrap.alert', [])
 
 .controller('AlertController', ['$scope', '$attrs', function ($scope, $attrs) {
-  $scope.closeable = 'close' in $attrs;
+  $scope.closeable = $attrs.close;
```
chrisirhc: `!!$attrs.close` might be clearer since we intend closeable to be a boolean.

chrisirhc:
Did a review. This looks good. This change is needed due to: https://github.com/angular/angular.js/blob/d193c3a25caa0d2c6dd149941c23163dbd062e4d/src/ng/compile.js#L2570-L2574

wesleycho:
Alright, changed

chrisirhc:
LGTM

PR #3865, opened 2015-06-27
fix(typeahead): add support for ngModelOptions getterSetter

### LINKED ISSUE #3823: Typeahead does not support ng-model-options="{getterSetter: true}"
AngularJS: v1.3.
ui-bootstrap: v 0.13.

Looks like when I set `ng-model-options="{getterSetter: true}" ` typeahead for some reasons replaces getter/setter method with selected value.

Here the plankr: http://plnkr.co/edit/a3Ld9eVmysrt8RsyzbhX?p=preview
Steps:
0) Note from the sources that `suggestions.value` is a getter/setter. 
1) Type "one" or "two"
2) Note that `suggestions.value` is not a function any more. It is a choosen value from typeahead suggestion.

wesleycho:
- Add support for getterSetter with ngModelOptions

Fix can be seen in action in [this Plunker](http://plnkr.co/edit/FLxL41HQQCMbGygfKcBm?p=preview).

This fixes #3823.


pkarikh:
Works for me! Thank you. :)

lehresman:
This fixes the problem for me as well.  Please merge this for 0.13.1.  Thanks!

chrisirhc:
Argh, this feels like we're copying internal ngModel APIs to be able to replicate its functionality. Is there no other alternative?

wesleycho:
I don't like it either, but upon inspecting how ng-model implements it, this seems like how we have to do it - ideally the typeahead probably should be refactored since the whole $setModelValue is a bit of a hack.

wesleycho:
After discussing this a bit with some others on this, the opinion is that this is an ok change - IMO this is a weakness with the `getterSetter` API with ngModel in Angular itself.

Issue #3852, opened 2015-06-24
New popover-template not working with Angular 1.3.8


jadarnel27:
Here is a plunkr for the issue: http://plnkr.co/edit/spe6hpCHzGhjxUFVJIHl?p=preview

Clicking the button (to produce the popover) does nothing (no errors, but also no popover).  Compare to [this plunkr](http://plnkr.co/edit/nbbfpG2TjB3uvSORsSAJ?p=preview) where the popover-template is working.  The only difference I can see is that it's using Angular 1.3.15.

The GitHub site for UI Bootstrap says "*requires AngularJS 1.3.x, tested with 1.3.13*," so I suppose there was some kind of breaking change between Angular 1.3.13 and 1.3.8.

dovy:
So I found a workaround.

If you declare the exact template in your view, be it popOverTemplate.html, and declare it within the HTML, it fails. However, if you declare a scope variable inside your controller, and call that as your template, the popover works...

jadarnel27:
I definitely don't understand everything that's going on here, but maybe this will help point things in the right direction.  The reason the popover isn't displaying is here (in the `show()` method):

    // Don't show empty tooltips.
    if ( !(options.useContentExp ? ttScope.contentExp() : ttScope.content) ) {
      return angular.noop;
    }

In the case of using a popoverTemplate, `useContentExp` is true; so the result of `contentExp()` is used to determine if the popover (tooltip) is "empty."  Here's that function:

    ttScope.contentExp = function () {
      return scope.$eval(attrs[type]);
    };

The value of `type` is "popoverTemplate" and the value of the `popoverTemplate` property of the `attrs` object is the name of / path to the template ("*app/template/my-popover-template.html*" in my example).  

Calling `$eval` on that returns `NaN`, which is falsey - thus the popup doesn't display.

jadarnel27:
Actually, I think this fixes the problem:

    ttScope.contentExp = function () {
      return scope.$eval(attrs[type]) ? scope.$eval(attrs[type]) : attrs[type];
    };

If popover-template was set to a valid expression, then evaluate that expression (maybe a scope function that generates the template link) and return it.  Otherwise, it *is* the template link, so just return that value.

jadarnel27:
I will go ahead and submit a pull request to that effect with some unit tests that expose the issue.

jadarnel27:
[...]

PR #3866, opened 2015-06-29
fix(tooltip): allow string literal templates

jadarnel27:
Can now pass a string literal with the templateUrl to tooltip-template and popover-template,
rather than just an expression / scope variable. This is a potential solution to my issue here: https://github.com/angular-ui/bootstrap/issues/3852

### DIFF
--- a/src/tooltip/test/tooltip-template.spec.js
+++ b/src/tooltip/test/tooltip-template.spec.js
@@ -75,5 +75,22 @@ describe('tooltip template', function() {
     expect( elmBody.children().length ).toBe( 1 );
   }));
 
+  it('should open when templateUrl is specified as a string literal rather than an expression - issue 3852', inject(function($compile) {
+    elmBody = angular.element(
+      '<div><span tooltip-template="myUrl">Selector Text</span></div>'
+    );
+    $compile(elmBody)(scope);
+    scope.$digest();
+    elm = elmBody.find('span');
+    elmScope = elm.scope();
+    tooltipScope = elmScope.$$childTail;
+
+    
+    elm.trigger( 'mouseenter' );
+
+    expect( tooltipScope.isOpen ).toBe( true );
+    expect( elmBody.children().length ).toBe( 2 );
+  }));
+
 });
 
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -65,7 +65,7 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
    * Returns the actual instance of the $tooltip service.
    * TODO support multiple triggers
    */
-  this.$get = [ '$window', '$compile', '$timeout', '$document', '$position', '$interpolate', function ( $window, $compile, $timeout, $document, $position, $interpolate ) {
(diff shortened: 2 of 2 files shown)


jadarnel27:
This is my first submission to the project.  I made sure to add a test that exposed the problem, and then verified all tests passed after making my changes.  I am totally open to feedback on my solution.  I wasn't sure if adding the $templateCache dependency to the directive was the best way to go about this, but I couldn't think any other good way to make sure the tooltip / popover wouldn't come up "blank" (because of an invalid template).

src/tooltip/test/tooltip-template.spec.js
```diff
@@ -75,5 +75,22 @@ describe('tooltip template', function() {
     expect( elmBody.children().length ).toBe( 1 );
   }));
 
+  it('should open when templateUrl is specified as a string literal rather than an expression - issue 3852', inject(function($compile) {
+    elmBody = angular.element(
+      '<div><span tooltip-template="myUrl">Selector Text</span></div>'
```
wesleycho: I think if this change is to be made to support an expression, there should be a breaking change where the attribute value is not assumed to be a string - otherwise this creates a confusing api.

RobJacobs:
If you wrap the template string in single quotes it works.  I recall this being the suggested approach when working with string literals for other directive attributes.  We should check with Chris though and see what his intent was.

jadarnel27:
@RobJacobs Thanks again for sharing that workaround - that makes a lot of sense.

I guess I just found it confusing that this works:

    $scope.myUrl = 'myTemplate.html';
    ...
    <div><span tooltip-template="myUrl">Selector Text</span></div>

But this doesn't:

    <div><span tooltip-template="myTemplate.html">Selector Text</span></div>

Which is, intuitively, the same thing.

src/tooltip/test/tooltip-template.spec.js
jadarnel27: Maybe I need more coffee, but I'm having a little trouble understanding your comment.  The change I suggested is to support string literals - expressions are currently the only thing that is supported.  So, currently, something like `<span tooltip-template="myUrl">` doesn't work unless you wrap `myUrl` in single quotes as well.

pkozlowski-opensource:
> Which is, intuitively, the same thing.

Nope, it is not. Expressions are evaluated against variables on the scope, so doing `myTemplate.html` would imply: find the `myTemplate` variable on a scope.

Anyway, I don't think this PR is valid, as it would introduce confusing API. There are already directives in AngularJS (ex.: `ngInclude`) that behave exactly like this one so I don't think we should deviate from the well-established practices in the community. 

This would be -1 from me when it comes to merging this PR.

jadarnel27:
"*Nope, it is not.*" - Perhaps I should have said that it was intuitive to me, rather than generalizing.  Sorry about that, @pkozlowski-opensource. 

I can definitely see how this change would cause confusion, especially after you pointing out the ngInclude example.  I completely understand why this change wouldn't be merged.

At the very least, I can at least now better understand why my code was not working =)  Thanks for taking the time to look into this.

wesleycho:
I too used `ng-include` for reference before responding, since for developer intuitive use we should match how Angular handles situations & agree with the others on this.

Closing as per discussion.

PR #3874, opened 2015-06-30
Typeahead popup position

davidrus:
## Append-to-body dropdown position 
Add support for typeahead-dropdown position after scroll and resize events. 

### Scenario 1
* Type some text into input (show dropdown appended to body)
* Scroll down
* We need dropdown in right place after input

### Scenario 2
* Type some text into input (show dropdown appended to body)
* Resize after breakpoint in media query (change position of input)
* We need dropdown in right place after input

Thanks


src/typeahead/typeahead.js
```diff
@@ -170,6 +169,28 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
         });
       };
 
+      // bind events only if appendToBody params exist - performance feature
+      if ( appendToBody ) {
+        angular.element( $window ).bind( 'resize', fireRecalculating );
+        angular.element( document.body ).bind( 'scroll', fireRecalculating );
```
wesleycho: These bindings should be throttled.

Also you can use $document.find('body') instead.

src/typeahead/typeahead.js
```diff
@@ -170,6 +169,28 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+      }
+
+
+      function fireRecalculating() {
+        // if popup is visible
+        if ( scope.matches.length > 0 ) {
```
wesleycho: It is enough to do `if (scope.matches.length) {`

src/typeahead/typeahead.js
```diff
@@ -170,6 +169,28 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+          recalculatePosition();
+          scope.$apply();
```
wesleycho: In this case, I believe you can get away with just `scope.$digest()`

src/typeahead/test/typeahead.spec.js
```diff
@@ -752,6 +752,29 @@ describe('typeahead tests', function () {
...
       expect(findDropDown($document.find('body')).length).toEqual(0);
     });
+
+    it('should have right position after scroll', function() {
+      var element = prepareInputEl('<div><input ng-model="result" typeahead="item for item in source | filter:$viewValue" typeahead-append-to-body="true"></div>');
+      var dropdown = findDropDown($document.find('body'));
+      var body = angular.element(document.body);
+
+      // Set body height to allow scrolling
+      body.css({height:"10000px"});
```
wesleycho: Use single quotes

davidrus:
Thanks for comments, all done. Travis OK

davidrus:
Any news for this enhancement?

src/typeahead/typeahead.js
```diff
@@ -170,6 +169,28 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
         });
       };
 
+      // bind events only if appendToBody params exist - performance feature
+      if ( appendToBody ) {
+        angular.element( $window ).bind( 'resize', fireRecalculating );
+        $document.find( 'body' ).bind( 'scroll', fireRecalculating );
```
wesleycho: Both these event listeners need to be throttled - the resize and scroll events both fire often.

wesleycho: Actually, a debounce on each of the event listeners would be better, since we only care about the trailing trigger of the event.

davidrus:
I added debounced execution of event logic and support for hiding popup if scroll & resize is "in progress".

src/typeahead/typeahead.js
```diff
@@ -170,6 +171,52 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
         });
       };
 
+      // bind events only if appendToBody params exist - performance feature
+      if ( appendToBody ) {
```
wesleycho: Remove spaces in between `(` and `)`

src/typeahead/typeahead.js
```diff
@@ -170,6 +171,52 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+        angular.element( $window ).bind( 'resize', fireRecalculating );
```
wesleycho: Remove spaces just after `(` and just before `)`

src/typeahead/typeahead.js
wesleycho: Remove spaces just after `(` and just before `)`

src/typeahead/typeahead.js
```diff
@@ -170,6 +171,52 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+
+      // Declare the timeout promise var outside the function scope so that stacked calls can be cancelled later
+      var timeoutEventPromise;
+
+      // Default progress type
+      scope.moveInProgress = false;
+
+      function fireRecalculating() {
+
+        if( !scope.moveInProgress ){
```
wesleycho: Change to `if (!scope.moveInProgress) {`

src/typeahead/typeahead.js
```diff
@@ -170,6 +171,52 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+          scope.moveInProgress = true;
+          scope.$digest();
+        }
+
+        // Cancel previous timeout
+        if ( timeoutEventPromise ) {
```
wesleycho: Remove spaces just after `(` and just before `)`

src/typeahead/typeahead.js
```diff
@@ -170,6 +171,52 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+          $timeout.cancel(timeoutEventPromise);
+        }
+
+        // Debounced executing recalculate after events fired
+        timeoutEventPromise = $timeout(function () {
+
+          // if popup is visible
+          if ( scope.matches.length ) {
```

src/typeahead/typeahead.js
```diff
@@ -170,6 +171,52 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+        }
+
+        // Cancel previous timeout
+        if ( timeoutEventPromise ) {
+          $timeout.cancel(timeoutEventPromise);
+        }
+
+        // Debounced executing recalculate after events fired
+        timeoutEventPromise = $timeout(function () {
+
```
wesleycho: Remove this line

src/typeahead/typeahead.js
```diff
@@ -170,6 +171,52 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+          // if popup is visible
+          if ( scope.matches.length ) {
+            recalculatePosition();
+          }
+
+          scope.moveInProgress = false;
+          scope.$digest();
+
```
wesleycho: Remove this line

src/typeahead/typeahead.js
```diff
@@ -170,6 +171,52 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
...
+        }, eventDebounceTime);
+
+      }
+
+      // recalculate actual position and set new values to scope
+      // after digest loop is popup in right position
+      function recalculatePosition() {
+        scope.position = appendToBody ? $position.offset(element) : $position.position(element);
+        scope.position.top = scope.position.top + element.prop('offsetHeight');
```
wesleycho: Why is there a separate top calculation?

Also this should be `scope.position.top += element.prop('offsetHeight')`

davidrus: This is function copied from original code and top, must be calculate with offsetHeight of input, because dropdown must be showed after bottom edge of input.

davidrus:
Done

src/typeahead/typeahead.js
wesleycho: Would it make more sense to move the functionality into the `$position` service perhaps?

davidrus: Do you think something like ```$position.bottomEdge(element)```?

wesleycho:
Adding a comment here for better visibility - there are multiple PRs involving positioning, so I am going to look into seeing if there is a better way of handling it inside $position instead.

wesleycho:
This PR looks safe to merge, although I think the debounce should be split off into a separate service - I'm inclined to hold off until this is done.

I could merge this in now, or wait until the debounce is split off into a service - the benefit of holding off would be that debounce would be available as a utility injectable for anyone who needs it, but it does put some more burden on you.

What are your thoughts?

davidrus:
OK, the idea of separate debounce in service is good. But I think this should be in separate PR. I can write this after merge this PR.

davidrus:
Only for my information: when are you planning to merge it into production?

wesleycho:
I think the debounce should be split off here - can you update your PR with that change?

wesleycho:
Actually, scratch my comment - I will merge this in today, and work can be done to split the debounce to a separate component after.

Issue #3875, opened 2015-06-30
Using alternative formatting breaks validation


robbaman:
When using an alternative formatting (in my case dd-MM-yyyy) the validation breaks when typing in the textbox.

http://plnkr.co/edit/ZdyqcUpFW6dxcqZsS8CE?p=preview

This appears to be due to an issue where the baseDate passed to the parse function of the parse may be an invalid date. If this is the case, the parser will always keep spitting out invalid dates.

I'll try to issue a PM with a fix for the issue.

wesleycho:
This appears to be a duplicate that should be fixed by #3759.

PR #3876, opened 2015-06-30
Fix for issue #3875 

robbaman:
Issue: https://github.com/angular-ui/bootstrap/issues/3875

Detect invalid date using the method proposed in this SO topic:

http://stackoverflow.com/questions/1353684/detecting-an-invalid-date-date-instance-in-javascript

### DIFF
--- a/src/dateparser/dateparser.js
+++ b/src/dateparser/dateparser.js
@@ -125,7 +125,7 @@ angular.module('ui.bootstrap.dateparser', [])
 
     if ( results && results.length ) {
       var fields, dt;
-      if (baseDate) {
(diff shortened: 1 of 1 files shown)


src/dateparser/dateparser.js
```diff
@@ -125,7 +125,7 @@ angular.module('ui.bootstrap.dateparser', [])
 
     if ( results && results.length ) {
       var fields, dt;
-      if (baseDate) {
+      if (baseDate && Object.prototype.toString.call(baseDate) === "[object Date]" && !isNaN(baseDate.getTime())) {
```
wesleycho: Why not use `angular.isDate` to check if it is a date instead of the `toString` call?

wesleycho:
This is a duplicate of #3759.

PR #3888, opened 2015-07-02
fix(modal): Modal Dialog doesn't fade out

### LINKED ISSUE #3633: Modal Dialog doesn't fade out
Under certain circumstances, the modal dialog will simply disappear instead of fading out (even though it successfully fades in).

The example in the documentation is what it should do (fade in and out).
https://angular-ui.github.io/bootstrap/#/modal

Ironically, the plunker link it generates shows the modal not properly fading out. I've saved my own to serve as an example here:
http://plnkr.co/edit/lmFTpQtfgV6qB4AuFrXj?p=preview
[*Edit from @chrisirhc to reduce confusion*: Please take note that the above Plunker does **not** include **angular-animate**. As such, no animations should be seen. This is a new requirement for animations, [see README](https://github.com/angular-ui/bootstrap#installation).]

I have the same issue in a separate project of mine where I am handling the modal differently but have the same bug. I can provide the 2nd example if it would help but I imagine the plunker should be enough.

The bug appears to be browser independent as it occurs both when I use Chrome v42.0.2311.135 m and in the developer edition of Firefox (39.0a2).

UPDATE: Note that the plnkr is using the latest version of angular-ui (0.13.0) and the latest supported version of angular (1.3.13). This is not about angular 1.4.

ihenshaw:
This fixed the issue for me. I have tested it with Angular 1.4 and 1.3 and it seems to work for both.

It seems in Angular 1.4, the animation events are no longer triggered on elements, instead there are new functions on $animate for registering / deregistering listeners. It also seems that the "close" event doesn't exist anymore, at least I found no reference to it.

The approach to fixing the issue was to listen for a "removeClass" event. When we get one of those, we check the event phase, it needs to be "close". This seems to be about the point where the old $animate:close event would have been triggered.

Unfortunately, the $animate service does not provide functions for registering one off listeners like element.one, so I've had to manually remove the event listener in the handler.

Finally, when all that was working, I re-introduced the old mechanism for Angular 1.3 support (I assume we'll want to support both).

This fixes #3633 

Hopefully that'll sort it!

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -234,10 +234,21 @@ angular.module('ui.bootstrap.modal', [])
         scope.animate = false;
 
         if (domEl.attr('modal-animation') && $animate.enabled()) {
-          // transition out
-          domEl.one('$animate:close', function closeFn() {
-            $rootScope.$evalAsync(afterAnimating);
-          });
+          // transition out, angular 1.4 or <1.4?
+          if ($animate.on) {
+              // angular 1.4. use $animiate.on to capture a removeClass event in the "close" phase to detect the
+              // completion of the animation.
+            $animate.on('removeClass', domEl, function closeFn(element, phase) {
+              if (phase === 'close') {
+                $rootScope.$evalAsync(afterAnimating);
+              }
+            });
+          } else {
+              // angular <1.4 pickup the "close" event from the element to detect the completion of the animation.
+            domEl.one('$animate:close', function closeFn() {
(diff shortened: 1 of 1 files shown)


chrisirhc:
This will indeed fix it. However, this might silently break users of the custom template because of the dependency on the `removeClass` animation event. I'll refactor the template to pass in the `in` class to the controller so the controller can make use of the `$animate` service (like `$animate.addClass(inClass).then(removeModal)` and avoid the need for listening to event of the end of the animation.

PR #3895, opened 2015-07-04
fix(modal): closing breaks on missing scope, 1.4

### LINKED ISSUE #3787: Modal dialog does not disappear with 0.13 and Angular 1.3 when destroying a passed scope

### LINKED ISSUE #3873: Version 0.13 - modal does not close correctly
Backdrop still remains and modal itself is still there, just with opacity of 0.  Don't think this is a duplicate of https://github.com/angular-ui/bootstrap/issues/3633 .    Does not occur with 0.12.

chrisirhc:
- supports Angular 1.4

Fixes #3787
Fixes #3806
Fixes #3873
Closes #3888

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -78,12 +78,13 @@ angular.module('ui.bootstrap.modal', [])
     }
   }])
 
-  .directive('modalWindow', ['$modalStack', '$q', function ($modalStack, $q) {
+  .directive('modalWindow', [
+           '$modalStack', '$q', '$animate',
+  function ($modalStack ,  $q ,  $animate) {
     return {
       restrict: 'EA',
       scope: {
-        index: '@',
-        animate: '='
+        index: '@'
       },
       replace: true,
       transclude: true,
@@ -119,8 +120,14 @@ angular.module('ui.bootstrap.modal', [])
         });
 
         modalRenderDeferObj.promise.then(function () {
-          // trigger CSS transitions
-          scope.animate = true;
+          if (attrs.modalInClass) {
+            $animate.addClass(element, attrs.modalInClass);
+
+            scope.$on($modalStack.WINDOW_CLOSING_EVENT, function (e, setIsAsync) {
+              var done = setIsAsync();
+              $animate.removeClass(element, attrs.modalInClass).then(done);
+            });
+          }
 
           var inputsWithAutofocus = element[0].querySelectorAll('[autofocus]');
           /**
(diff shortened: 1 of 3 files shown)


wesleycho:
This is meant to supersede my PR, correct?

chrisirhc:
Yes, I think it fixes the root cause of a lot of issues in the repo.

wesleycho:
LGTM

esion:
Thank you @chrisirhc, unfortunatly there is still a bug in your plnkr you gave http://plnkr.co/edit/R8322USvwVICHy8wjM7Y?p=preview : 

* the backdrop element is still in the DOM and prevent the modal to be open again.

Other bug in the animation : 

* the modal just shows up without animation (top to middle)
* the modal just disapears without animation (middle to top)

It requires to wait the animation is finished before removing element from DOM : I've started this hack that uses the event from $animate : https://github.com/esion/bootstrap/pull/1/files#diff-15d409302580c93c654a382e2da68396R245  
[demo](http://plnkr.co/edit/EOGddlgF36e8ZusbtJUt?p=preview)

chrisirhc:
@esion you're right, I haven't fixed the backdrop animation. I'll make another PR for that in a bit.

chrisirhc:
@esion , I've just submitted a PR #3899 to fix the issue you're having.
Updated Plunker: http://plnkr.co/edit/WlA6T6I63Og4Hkmk2FIh?p=preview

vamsibassetty08:
I've tried animation off globally and inclusion ngAnimate but still the backdrop doesn't close the modal with angular 1.3.2 and angular-ui-bootstrap 0.12.0. Even I tried with angular-ui-bootstrap 0.13.0 & 0.13.1 etc but still modal doesn't close as backdrop is still present..

Any other alternatives?

PR #3899, opened 2015-07-06
fix(modal): backdrop animation on AngularJS 1.4

### LINKED ISSUE #3896: Modal Backdrop not being Removed from Page Upon Modal Close
I'm opening a new issue based off of the following two threads <a href="https://github.com/angular-ui/bootstrap/issues/3620">3620</a> and <a href="https://github.com/angular-ui/bootstrap/issues/3633">3633</a>. Both threads discuss this issue in depth but apparently it needs its own dedicated space. The issue is with ngAnimate and the updates that have changed the way its functions interact with Angular Bootstrap's. The opening functions are still wired up and fire but the close events do not cause ngAnimate to fire. There are several work arounds provided in <a href="https://github.com/angular-ui/bootstrap/issues/3633">3633</a>.

chrisirhc:
Fixes #3896

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -57,7 +57,9 @@ angular.module('ui.bootstrap.modal', [])
 /**
  * A helper directive for the $modal service. It creates a backdrop element.
  */
-  .directive('modalBackdrop', ['$timeout', function ($timeout) {
+  .directive('modalBackdrop', [
+           '$animate', '$modalStack',
+  function ($animate ,  $modalStack) {
     return {
       restrict: 'EA',
       replace: true,
@@ -69,12 +71,14 @@ angular.module('ui.bootstrap.modal', [])
     };
 
     function linkFn(scope, element, attrs) {
-      scope.animate = false;
+      if (attrs.modalInClass) {
+        $animate.addClass(element, attrs.modalInClass);
 
-      //trigger CSS transitions
-      $timeout(function () {
-        scope.animate = true;
-      });
+        scope.$on($modalStack.NOW_CLOSING_EVENT, function (e, setIsAsync) {
+          var done = setIsAsync();
+          $animate.removeClass(element, attrs.modalInClass).then(done);
+        });
+      }
     }
   }])
 
@@ -123,7 +127,7 @@ angular.module('ui.bootstrap.modal', [])
           if (attrs.modalInClass) {
             $animate.addClass(element, attrs.modalInClass);
 
-            scope.$on($modalStack.WINDOW_CLOSING_EVENT, function (e, setIsAsync) {
+            scope.$on($modalStack.NOW_CLOSING_EVENT, function (e, setIsAsync) {
               var done = setIsAsync();
               $animate.removeClass(element, attrs.modalInClass).then(done);
             });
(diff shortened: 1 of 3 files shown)


src/modal/test/modal.spec.js
```diff
@@ -142,6 +142,7 @@ describe('$modal', function () {
   function open(modalOptions) {
     var modal = $modal.open(modalOptions);
     $rootScope.$digest();
+    $timeout.flush(0);
```
wesleycho: Does the 0 need to be passed in?

chrisirhc: There's [a test that has a $timeout resolve](https://github.com/chrisirhc/angular-ui-bootstrap/blob/feature/fix-backdrop-animate/src/modal/test/modal.spec.js#L469-L480) , this prevents flushing that timeout so that that test is successful.

This really is supposed to just run the next tick, it's only for resolving the newly added $q promises, since there weren't any $timeout timers added.

src/modal/modal.js
```diff
@@ -261,18 +249,21 @@ angular.module('ui.bootstrap.modal', [])
...
+            closingPromise = closingDeferred.promise;
+          }
+
+          return function () {
+            closingDeferred.resolve();
+          };
+        };
+        scope.$broadcast($modalStack.NOW_CLOSING_EVENT, setIsAsync);
+
+        return $q.when(closingPromise).then(afterAnimating);
```
wesleycho: I'm a bit confused by this - is there a possibility that `closingPromise` could be `undefined` at this point of time? This seems potentially dangerous.

chrisirhc: Yes, there is, if `setIsAsync` is never called after the `NOW_CLOSING_EVENT` has broadcasted, `closingPromise` will be `undefined`. This is intentional. In such a case, the promise should resolve immediately (asynchronously), as it has no need to wait for anything (animations etc.)

src/modal/modal.js
```diff
@@ -261,18 +249,21 @@ angular.module('ui.bootstrap.modal', [])
...
-        if (domEl.attr('modal-animation') && $animate.enabled()) {
-          // transition out
-          domEl.one('$animate:close', function closeFn() {
-            $rootScope.$evalAsync(afterAnimating);
-          });
-        } else {
-          // Ensure this call is async
-          $timeout(afterAnimating);
-        }
+        var closingDeferred;
```
wesleycho: Would it make sense to do
```
var resolveFn;
var closingPromise = $q(function(resolve) {
  resolveFn = resolve;
});
```
And change setIsAsync to
```
function setIsAsync() {
  return function() {
    resolveFn();
  };
}
```
Seems like this would be potentially less prone to error with ensuring `closingPromise` exists.

chrisirhc: See below, it's actually intended that it be allowed to be undefined.

wesleycho:
Hm, I think a comment explaining the behavior would be good - otherwise, this LGTM.

chrisirhc:
Added a comment and updated the variable names to hopefully clarify things.

wesleycho:
Looks good here

diwu1989:
This is really important for us, can you make a 0.13.1 release soon with this included?

LunarDevelopment:
:+1: 
this needs released asap, to bower preferably!

sagar-ranglani:
Any updates about the release?

Xal3ph:
When I built 13.1 based on 158d267 it fixed the problems I was experiencing.  Thank you!  Looking forward to the official release :beers:

PR #3900, opened 2015-07-06
feat(position): implement directional positioning

wesleycho:
- When positioning based off `position` or `offset`, allow specification of direction

I will work on tests for this, but I am opening this up for review so that this code can be cleaned up some (I'm not a fan of all of these if conditionals here).

This will assist with #3782 with a cleaner implementation. Implementing this part is as simple as changing `template/datepicker/popup.html` to
```
<ul class="dropdown-menu"
  ng-style="{
    display: (isOpen && 'block') || 'none',
    top: position.top,
    right: position.right,
    bottom: position.bottom,
    left: position.left
  }"
  ng-keydown="keydown($event)">
```
And the relevant lines in the `datepickerPopup` directive to
```
        if (value) {
          scope.$broadcast('datepicker.focus');
          scope.position = $position.placement(element, attrs.datepickerPlacement, appendToBody);
          if (angular.isNumber(scope.position.bottom)) {
            if (appendToBody) scope.position.bottom -= element.prop('offsetHeight');
            scope.position.bottom += 'px';
            scope.position.top = 'initial';
          } else {
            scope.position.bottom = 'initial';
            if (!appendToBody) scope.position.top += element.prop('offsetHeight');
            scope.position.top += 'px';
          }

          if (angular.isNumber(scope.position.right)) {
            scope.position.left = 'initial';
            scope.position.right += 'px';
          } else {
            scope.position.left += 'px';
            scope.position.right = 'initial';
          }

          $document.bind('click', documentClickBind);
        } else {
          $document.unbind('click', documentClickBind);
        }
      });
```
[Here](http://plnkr.co/edit/ugEPcBOggaWO2j4HNhr9?p=preview) is a Plunker with those changes implemented.


wesleycho:
It might make sense to move the conditional adding and subtracting of `top` and `bottom` values in that code block example to the `placement` method.

It also might make sense to just move all of the CSS logic right into the method as well.

src/position/position.js
```diff
@@ -39,6 +41,66 @@ angular.module('ui.bootstrap.position', [])
...
+              newPosition.left = 0;
+            }
+          }
+        } else {
+          newPosition.top = offset === 'offset' ? 0 : position.top;
+          newPosition.left = position.left;
+        }
+      } else {
+        newPosition.top = offset === 'offset' ? 0 : position.top;
+        newPosition.left = position.left;
```
chrisirhc: I'm quite confused about this line and 92-94. What are these cases? Are they fallbacks?

wesleycho: 92-94 is the default case if it is a string but invalid value.

96-98 is the default case if it has no value set.

src/position/position.js
```diff
@@ -7,6 +7,8 @@ angular.module('ui.bootstrap.position', [])
  * typeahead suggestions etc.).
  */
   .factory('$position', ['$document', '$window', function ($document, $window) {
+    var POSITION_REGEX = /((top|bottom)(\s+(left|right))?)|(left|right)/,
```
chrisirhc: There are non-capturing groups, which could be useful in cleaning up your logic below. For example: `/(?:(top|bottom)(?:\s+(left|right))?)|(left|right)/` means that `match[1]` will be top/bottom and `match[2]` will be right/left (only if there's top/bottom), and `match[3]` will be right/left without top/bottom.

src/position/position.js
chrisirhc: If it's the default case, perhaps it be handled via a `isInvalid` flag or something, and a check below, since the code is identical, might make sense to put it together.

wesleycho: Makes sense - I am not happy with how I wrote this, there are too many if-else conditionals here.

src/position/position.js
wesleycho: Good catch - I never thought about doing `?:(foo)` for some reason

icfantv:
Closing in favor of #4762.

Issue #3008, opened 2014-11-24
Change regex for dateparser format 'd' and 'M' to allow for leading zeros and fix regex for 'M' format


dkuerner:
Allow leading zeros for day and month for format d and M
Use case:
Format d.M.y should allow for entering dates like 02.05.12 and like 2.5.12
This format is very common.

In addition, the regex for 'M' is incorrect. The order of the alternation should be changed.

cleftheris:
This is true I had the same issue. @dkuerner please make a pull request with your commit so that the someone from the team can review/merge. (hope sometime soon :+1: )

wesleycho:
We are not interested in supporting this at this point in time - the dateparser currently aims to support the behavior found in Angular's dateFilter.

cleftheris:
Are you serious. Filters are meant primarily for presenting data not user input. Try putting a date picker on an input type text and explain to the user why is 1/2/2015 is not a valid date. IMHO this is a bug

wesleycho:
Nothing is stopping you from overwriting the `dateparser` factory to implement your own custom parser.

That type of attitude though is unwarranted and unprofessional.

cleftheris:
My sincere apologies for my outburst.

wesleycho:
You'll be happy to know @cleftheris that I figured out a good implementation for this feature, and it will make it shortly to `master` - see #4809 :) .

cleftheris:
@wesleycho thanks much appreciated. Although currently I have overridden the `DateParser` as you suggested and works quite well I will switch back to default in order to get updates as soon as it gets merged.

cheers

PR #3901, opened 2015-07-06
allow for leading zeros #3008

dkuerner:
Allow leading zeros for day and month for format d and M
Use case:
Format d.M.y should allow for entering dates like 02.05.12 and like 2.5.12
This format is very common.

In addition, the regex for 'M' is incorrect. The order of the alternation should be changed.


wesleycho:
The regex change for `M` should be split off to another PR.

Leading zeroes should not be supported - this is as per adherence to how the [date filter](https://docs.angularjs.org/api/ng/filter/date) works in Angular.

wesleycho:
Closing as per comments - feel free to open a new PR with comments addressed.

PR #3905, opened 2015-07-07
Replaced templateUrl param in typeaheadPopup Directive for external templates

GTTo:
Replaced templateUrl param in typeaheadPopup Directive for support external templates in addition to the self contained template.

If you try to use the templateUrl attrib before this modification, you get an error if you don't have access to the original template on url 'template/typeahead/typeahead-popup.html'.

Now, you don't need the original template in the route if you provide an additional template.

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -361,10 +361,10 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
         select:'&'
       },
       replace:true,
-      templateUrl:'template/typeahead/typeahead-popup.html',
+		templateUrl: function (element, attrs) {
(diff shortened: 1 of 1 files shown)


wesleycho:
The tests need to be fixed, but this otherwise looks like a reasonable change.

GTTo:
When I run the tests over the master branch I get a tons of errors.
I revised the node versions on my machine and it's up to date.
How can I run the tests without errors??

GTTo:
Ok, sorry....
I update the node version from 0.10.X to 0.12.7 but I don't clean the dependencies on the project.
After delete the ``node_modules`` folder and re-run ``npm install`` and ``grunt test`` everything works like a charm.

image

I'll check the tests over my fork.

Thank you!

wesleycho:
I believe this might be superceded on `master` currently.

Closing as already addressed - if it is still an issue, feel free to file an issue.

Issue #3904, opened 2015-07-07
Accordion: href="javascript:void(0);" Content Security Policy (CSP) Error


roomond:
PR https://github.com/angular-ui/bootstrap/pull/2869 introduced an empty href attribute to enable keyboard accessibility for toggling accordion groups. This introduced a page refresh issue which was fixed by https://github.com/angular-ui/bootstrap/issues/3299. The fix for this was introducing a noop via javascript.void(0) to the href attribute. The result of this change is that it breaks the CSP rules such as our use case where we've disabled all inline javascript execution. This needs an alternative workaround that doesn't rely on having an inline script for the empty href attribute.

Please see https://developer.mozilla.org/en-US/docs/Web/Security/CSP for a quick introduction to CSP.

realityking:
CSP compatability would be a great goal for UI Bootstrap. It's certainly a lot of work, but it's great if libraries don't stand in the way of hardening web sites.

wesleycho:
Is there any other problems as far as CSP is concerned inside UI Bootstrap other than the improper href usage?

realityking:
The other - and probably more difficult one - would be inline styles.

wesleycho:
Is mg-style still considered ok?

realityking:
TBH, I don't know. Without `unsafe-inline`, the style attribute is disallowed but CSSOM is allowed. My hunch is that it would work, but it should be tested. That said, the JS issue is much more interesting, as the security benefit is larger.

wesleycho:
I have opened a new ticket for the broader CSP question at #3911.

PR #3909, opened 2015-07-07
Make accordion work with inline js CSP directive

roomond:
This PR fixes issue https://github.com/angular-ui/bootstrap/issues/3904 where CSP errors are triggered as the code href="javascript:void(0)" violates the inline js content security policy. This also preserves the original intent of adding an empty href attribute to enable accessibility while preventing a page refresh.

### DIFF
--- a/template/accordion/accordion-group.html
+++ b/template/accordion/accordion-group.html
@@ -1,7 +1,7 @@
 <div class="panel panel-default">
   <div class="panel-heading">
     <h4 class="panel-title">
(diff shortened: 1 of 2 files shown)


wesleycho:
There is similar usage [here](https://github.com/angular-ui/bootstrap/blob/master/template/typeahead/typeahead-match.html) - mind filing a PR for that one too?

roomond:
@wesleycho Will do

Issue #3820, opened 2015-06-15
dropdown-append-to-body doesn't work with .dropdown-menu-right


bytasv:
Hello,

When adding `dropdown-append-to-body` to `dropdown` component and inner menu has `.dropdown-menu-right` class (for right side alignment) the dropdown menu position is set incorrectly. I've checked source code and noticed that when calculating position for `dropdown-append-to-body` option only `left` property is used, hence not positioning correctly the menu... It should be using `right` instead of left..

cpitt:
Should be able to fix it by checking if element has class dropdown-menu-right and subtracting el width from left offset on line 126ish of src/dropdown/dropdown.js . I'm unfortunately not in a position to try it right now if someone wants to take a crack at it in the mean time.

rvanbaalen:
Please demonstrate your issue in a plunkr.

rysilva:
Does this really need a plunkr?  It's really straightforward.

I wrote the append-to-body feature of dropdown, and I actually just came across this issue myself.

Here's the plunkr so we can get that out of the way: http://plnkr.co/edit/21xjDthnZsbC2StDy2VY?p=preview

schmuli:
I tried out a fix, by looking for the dropdown-menu-right class, and changing the calculation for the CSS positions, so in [dropdown.js](https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L165) instead of:

```
  if ( appendToBody && self.dropdownMenu ) {
      var pos = $position.positionElements(self.$element, self.dropdownMenu, 'bottom-left', true);
      self.dropdownMenu.css({
        top: pos.top + 'px',
        left: pos.left + 'px',
        display: isOpen ? 'block' : 'none'
      });
    }
```

I used:

```
        if (appendToBody && self.dropdownMenu) {
            var rightalign = self.dropdownMenu.hasClass('dropdown-menu-right');
            var pos = $position.positionElements(self.$element, self.dropdownMenu, 'bottom-left', true);
            var css = {
                top: pos.top + 'px',
                display: isOpen ? 'block' : 'none'
            };
            if (!rightalign) {
                css.left = pos.left + 'px';
                css.right = 'auto';
            } else {
                css.left = 'auto';
                css.right = (window.innerWidth - (pos.left + self.$element.prop('offsetWidth'))) + 'px';
            }
            
            self.dropdownMenu.css(css);
        }
```

[...]

PR #3913, opened 2015-07-08
Added support for dropdown-menu being right aligned, see angular-ui/bootstrap#3820

schmuli:
Bootstrap has a 'dropdown-menu-right' class, which is used to display a dropdown menu aligned to the right edge of the toggle, rather than the left.

This PR adds a check in the case of append-to-body, and calculates the correct right-aligned position of the dropdown menu element.

This would then be able to close angular-ui/bootstrap#3820.


wesleycho:
This looks good - [here](http://plnkr.co/edit/SgTp6f4BRlZkuY7CCOya?p=preview) is a Plunker with your changes for reference.

PR #3916, opened 2015-07-08
Great job!

richardaanzee:
Hi, you did a great job. But bower states that packages should be pre-build. 
When deploying to heroku it is hard to configure a postinstall script that builds the plugin. 
So please always provide a build version.


pkozlowski-opensource:
Nope, we don't want to keep build results in this repo, as it creates many hard problems for the day-to-day maintenance. If you are looking for the build results for bower consumption you can find those in https://github.com/angular-ui/bootstrap-bower

wesleycho:
As noted, it is not desired to keep the result of the builds in the repository - it also bloats Git, which can cause problems as well.

If one needs to consume a built version from `master` or a particular commit, one can write a build script around creating the build from that version.

Issue #3526, opened 2015-04-10
Discussion - Adding Aria compatibility & improving accessibility


rvanbaalen:
Opening this tread to discuss Aria compatibility / accessibility in general with @angular-ui/bootstrap and it's community. I'll also reference all Aria related PR's and issues in here.

Related issues:  #494, #2249, #2303, #2688, #3095, #3199, #3253, #3264, #3474 

Related PR's: #1714, #2174, #3329, #2316, #2515, #3329

### Resources (curated list)

- http://www.w3.org/TR/wai-aria-practices
- https://developer.mozilla.org/en-US/docs/Accessibility/ARIA
- https://docs.angularjs.org/guide/accessibility
- http://a11yproject.com/checklist.html
- http://oaa-accessibility.org
- https://medium.com/@cironunesdev/angularjs-cares-about-accessibility-607785648c58
- [Fully accessible datepicker](http://oaa-accessibility.org/example/15/)
- https://www.marcozehe.de/2015/02/05/advanced-aria-tip-2-accessible-modal-dialogs/
- https://egghead.io/lessons/angularjs-using-ng-aria-to-automatically-improve-your-angularjs-accessibility

- http://www.webdesignerdepot.com/2015/04/8-reasons-to-embrace-website-accessibility/

### Tools

- [Automated Accessibility testing](https://github.com/nature/pa11y)

### Milestone 1.1

Everything accessibility and ARIA related will be covered in milestone `v1.1.0`.

rvanbaalen:
Related: 

#494
#2249
#2249
#2303
#2688
#3095
#3199
#3253
#3264
#3474

Tobino:
Is it possible to get a label "accessibility"?
Like [twbs/bootstrap](https://github.com/twbs/bootstrap/labels/accessibility)

rvanbaalen:
@Tobino your prayers are heard :wink: `type: accessibility` tag is created

icfantv:
Closing in favor of #4772.

icfantv:
Closing in favor of #4772.

PR #3920, opened 2015-07-08
Improved accessibility of collapse

RicardoRFaria:
- Added aria attributes to collapse
- Updated documentation with accessibility example
- Created tests to new attributes

Referencing the discussion #3526


src/collapse/collapse.js
```diff
@@ -26,6 +29,8 @@ angular.module('ui.bootstrap.collapse', [])
             // prevents the animation from jumping to collapsed state
             .removeClass('collapse')
             .addClass('collapsing');
+          element.attr('aria-expanded', false);
```
wesleycho: You can chain on the other calls to `element` right after `.addClass(...)` instead of calling it again

src/collapse/collapse.js
```diff
@@ -6,6 +6,9 @@ angular.module('ui.bootstrap.collapse', [])
       link: function (scope, element, attrs) {
         function expand() {
           element.removeClass('collapse').addClass('collapsing');
+          element.attr('aria-expanded', true);
```
wesleycho: Just chain on the `element` actions instead of typing `element` again.

wesleycho:
Other than those two minor issues, this looks good.

RicardoRFaria:
Fixed. Thanks.

PR #3927, opened 2015-07-09
Only call timeout if the time is a valid integer

steven10172:
Allows the use of `ng-attr-dismiss-on-timeout="{{alert.dismissOnTimeout}}"` when `alert.dismissOnTimeout` isn't defined. This fix will keep the alert on the screen instead of running `$timeout(function(){ alertCtrl.close(); }, NaN);` which runs instantly.

### DIFF
--- a/src/alert/alert.js
+++ b/src/alert/alert.js
@@ -23,9 +23,12 @@ angular.module('ui.bootstrap.alert', [])
   return {
     require: 'alert',
     link: function(scope, element, attrs, alertCtrl) {
-      $timeout(function(){
-        alertCtrl.close();
-      }, parseInt(attrs.dismissOnTimeout, 10));
+      var dismissOnTimeout = parseInt(attrs.dismissOnTimeout, 10);
+      if(dismissOnTimeout) {
+        $timeout(function(){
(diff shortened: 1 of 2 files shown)


wesleycho:
I think doing
```
$timeout(..., parseInt(attrs.dismissOnTimeout, 10) || 0)
```
would be a sufficient fix for this.

wesleycho:
Ahh, I was mistaken about the intent here - I'm not sure this is desired - if the user intent is to block closing, one should create the blocking behavior in the close method itself, instead of hacking the timeout.

steven10172:
@wesleycho Your fix is definitely a step in the right direction, but I feel that there should be a way to keep the alert on the screen by passing a blank (invalid) number or -1.

Using -1 would be the best option as IE 8 and lower don't fire the callback setTimeout function if the value is negative.

wesleycho:
As we are mainly supporting 1.3+, we are not supporting IE8

wesleycho:
As per my comments, I believe this approach should not be handled by the library itself, as this is too much of an edge case that has a legitimate workaround on the developer's side.

PR #3933, opened 2015-07-10
Make sure baseDate is actually a valid date.

rwlogel:
If the baseDate is an invalid date object it will not default the fields structure correctly.  This is can easily be reproduced by just trying to type in a date on the http://angular-ui.github.io/bootstrap/ datepicker popup example.


wesleycho:
This is a duplicate of #3759.

PR #3940, opened 2015-07-13
feat(datepicker): allow to only display days of selected month

christian-bromann:
#### Problem

Days that don't belong to the selected month could confuse people. There is no option to handle this issue.

#### Solution

Add another config attribute that allows us to handle this. Per default it is set to `false` so basically nothing changes until someone explicitly set it to `true`

**daysOfMonthOnly: `false` (default)**

screen shot 2015-07-13 at 23 06 31

**daysOfMonthOnly: `true`**

screen shot 2015-07-13 at 23 08 00

#### Comments

- Not sure if `daysOfMonthOnly` is really a good name for that property. Any other suggestions?
- Instead of using `ng-if` we could just apply a certain css class like `dayInMonth` to the button so people can use CSS to handle this use case.

#### Possible improvements

If show weeks is set to `true` it can happen that all buttons in a row get disabled and the row is a bit short in height. There is definitely a better way to handle this:

screen shot 2015-07-13 at 19 35 35

But I first wanted to know if you guys think that this patch is worth adding or maybe there is already a different approach of achieving this.

### DIFF
--- a/src/datepicker/docs/readme.md
+++ b/src/datepicker/docs/readme.md
@@ -85,6 +85,9 @@ All settings can be provided as attributes in the `datepicker` or globally confi
     _(Default: false)_ :
     An option to disable or enable shortcut's event propagation.
 
+ * `days-of-month-only`
+    _(Default: false)_ :
+    If true datepicker will only show days of selected month
 
 ### Popup Settings ###
 
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -15,7 +15,8 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   yearRange: 20,
   minDate: null,
   maxDate: null,
-  shortcutPropagation: false
+  shortcutPropagation: false,
+  daysOfMonthOnly: false
 })
 
 .controller('DatepickerController', ['$scope', '$attrs', '$parse', '$interpolate', '$timeout', '$log', 'dateFilter', 'datepickerConfig', function($scope, $attrs, $parse, $interpolate, $timeout, $log, dateFilter, datepickerConfig) {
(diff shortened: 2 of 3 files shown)


src/datepicker/docs/readme.md
```diff
@@ -134,4 +137,4 @@ Depending on datepicker's current mode, the date may reffer either to day, month
  * `Enter`/`Space`: Select date.
  * `Ctrl`+`Up`: Move to an upper mode.
  * `Ctrl`+`Down`: Move to a lower mode.
- * `Esc`: Will close popup, and move focus to the input.
\ No newline at end of file
+ * `Esc`: Will close popup, and move focus to the input.
```
wesleycho: What modification was made here?

wesleycho:
I'm not sure how I feel about this feature - this is not standard behavior for datepickers, and so I'm not sure how confusing this actually is for users. This also adds performance overhead on the `ng-if`.

src/datepicker/docs/readme.md
christian-bromann: My IDE added a new line character. I haven't read anything about new line at the end of the file in your contribution guidelines, shall I remove it?

christian-bromann:
@wesleycho thanks for the quick response. Don't know if there is something like a standard behaviour for datepicker, however I found couple examples where it is exactly like this:

jQueryUI: https://jqueryui.com/datepicker/
datepickr: https://code.google.com/p/datepickr/
Mootools: http://aryweb.nl/projects/mootools-datepicker/Test/ (kinda)
Dojo Bootstrap: http://xsokev.github.io/Dojo-Bootstrap/datepicker.html (kinda too)

All these examples have at least the possibility to disable these days using CSS:

screen shot 2015-07-14 at 21 51 16

How about we are going the CSS solution I was mentioning above: adding a certain css class to the button tag for days that don't belong to the selected month, e.g.:

screen shot 2015-07-14 at 21 51 16

wesleycho:
Sorry for taking a long time to get back to you.

CSS solution may be possible, but if you do attempt it, performance is something you would have to be careful about.

I'm going to close this PR for now since `master` changed quite a lot in the past half year, especially the datepicker. Feel free to reopen a PR implementing this again with tests.

Issue #3515, opened 2015-04-08
Popover to remain open when hovering over it


boillodmanuel:
Popover triggered by mouseenter event hide himself as soon as the mouse leave the target element. And it's impossible to click on a link which was in the popover.

I've found some solution in SO which can help for the PR :
- link to SO: http://stackoverflow.com/questions/20267967/create-hoverable-popover-using-angular-ui-bootstrap
- direct link to jsfiddle: https://jsfiddle.net/6awhrcec/

And another link with a vanilla js solution: http://jsfiddle.net/WojtekKruszewski/Zf3m7/22/

Could you add this feature?

RobJacobs:
The popover-trigger attribute allows you to define which triggers show and hide the popover using the following format:
 
    popover-trigger="showtrigger hidetrigger"

The directive is simply doing the following for you:

    element.bind(showtrigger, showTooltipBind)
    element.bind(hidetrigger, hideTooltipBind)

The [plunker](http://plnkr.co/edit/ScAI7a5lliaCpxWPVaA6?p=preview) demonstrates using  'mouseenter click' trigger that will leave the popup open during interaction.  While this get's us most of the way there, it's not as stable as I would like it to be.  

I have been working on a blurNotChildren trigger that will not call the close method until the trigger element or the popup are no longer the document.activeElement.  Here is a [plunker](http://plnkr.co/edit/C6peZ8ZEWgSlcfw3Q65E?p=preview) that demonstrates that approach.

The other ways I've seen of accomplishing this require listeners on several elements and document body click which get's messy to clean up and doesn't necessarily account for a blur event.

chrisirhc:
I'm marking this as Purgatory as I think this sort of custom handling should be left to the user/developer. Instead, we should expose the triggers through a directive controller to allow users to show/hide the tooltip/popovers with their custom handling logic.

If this becomes a high frequency problem, we can consider it further. However, right now, this adds a new feature that isn't there on Bootstrap and I'm worried about bloating this directive with features that most users wouldn't use.

tmcgee123:
I had a similar issue and addressed it by creating a close delay that I could set through the popover directive. The problem is that if you attempt to override the fade class that bootstrap provides or attempt to hijack the listeners, things get a little wonky. I have created this feature via PR #3576 https://github.com/angular-ui/bootstrap/pull/3576

erquhart:
[...]

PR #3942, opened 2015-07-14
patch to allow hoverable tooltips/popovers

plong0:
adding `tooltip-hoverable=true` or `popover-hoverable=true` will enable the feature when hide trigger is mouseleave

### DIFF
--- a/template/accordion/accordion-group.html
+++ b/template/accordion/accordion-group.html
@@ -1,7 +1,7 @@
 <div class="panel panel-default">
   <div class="panel-heading">
     <h4 class="panel-title">
-      <a href="javascript:void(0)" tabindex="0" class="accordion-toggle" ng-click="toggleOpen()" accordion-transclude="heading"><span ng-class="{'text-muted': isDisabled}">{{heading}}</span></a>
+      <a href="#" tabindex="0" class="accordion-toggle" ng-click="$event.preventDefault(); toggleOpen()" accordion-transclude="heading"><span ng-class="{'text-muted': isDisabled}">{{heading}}</span></a>
     </h4>
   </div>
   <div class="panel-collapse collapse" collapse="!isOpen">
--- a/template/typeahead/typeahead-match.html
+++ b/template/typeahead/typeahead-match.html
@@ -1,1 +1,1 @@
-<a href="javascript:void(0)" tabindex="-1" bind-html-unsafe="match.label | typeaheadHighlight:query"></a>
+<a href="#" ng-click="$event.preventDefault()" tabindex="-1" bind-html-unsafe="match.label | typeaheadHighlight:query"></a>


plong0:
adding `tooltip-hoverable=true` or `popover-hoverable=true` will enable the feature when triggers.hide is mouseleave

RobJacobs:
There is already a mechanism in place to handle this, see the triggers section [here](http://angular-ui.github.io/bootstrap/#/tooltip).

plong0:
Thanks, I've read over the tooltip triggers section, but could not see a trigger I could use that would allow the popover to stay open while I hover on the popover itself.
I did not see a specific answer in Issue #3515 either.

RobJacobs:
I think you can accomplish this by creating your own trigger with the $tooltipProvider like so:

    $tooltipProvider.setTriggers({'mouseenter manual': 'manual' });

Then:

    <div tooltip-trigger="'mouseenter manual'" />

wesleycho:
@plong0 can you try @RobJacobs solution and verify whether this is still an issue or not?

RobJacobs:
@plong0 @wesleycho [Here](http://plnkr.co/edit/IlbdVm1aT70D4UB0NFA7?p=info) is a plunk that demonstrates the approach I suggested.  It would be nice if the [tooltip-prepTriggers](https://github.com/angular-ui/bootstrap/blob/master/src/tooltip/tooltip.js#L327) method could be modified to not bind triggers set as 'none' or 'manual'

plong0:
@RobJacobs thanks for the plunk... I understand better where you are going with using the existing tooltip-enable attribute.  In that plunk using the 'none' trigger, the tooltip is staying open even if I enter something in the input and move my mouse away.

The root of the problem is in detecting when the tooltip itself is hovered/unhovered, then triggering the tooltip to open/close.

larserikfinholt:
What was the actual sln to this problem? @plong0 - did you manage to make a hoverable popover with the existing tooltip-enable attribute - in case how did you do that? Or was your PR accepted? (I could figure that out)

plong0:
No I couldn't figure it out with the existing tooltip-enable attribute.  PR never got accepted either... I am just using my forked version for now.

kentor:
great. another monkey patched solution

RossRogers:
@RobJacobs,

You referenced a Plunk at this address http://plnkr.co/edit/IlbdVm1aT70D4UB0NFA7?p=info .  Unfortunately, that URL now redirects to a new, empty Plunkr.  Can you restore that Plunkr or recreate it, per chance?

paulcamantigue-kaiser:
This saved me!  Had to adjust it for a very old version of angular-ui but it did the job :D

asitpanda:
@RobJacobs , Given plunk url is redirecting to empty plunk.Could you give some link?

I am looking for this feature. Can any one here help me.

icfantv:
@asitpanda, not sure it's fair to ask someone to recreate a plunk from 6 months ago and lord knows how many versions.  If you have a support-related question, please read our project documentation for how to receive support.

asitpanda:
i requested a SO http://stackoverflow.com/questions/37108913/uib-tooltip-should-appear-while-mouse-over-on-it, i am using angular-bootstrap Version: 0.14.3, Can any one suggest something.

odedkeren1:
I have solved it in a very cleaned way and thought to share it: .popover is being created not as a child of the uib-popover so the idea is to wrap uib-popover with a parent and to control show&hide on hovering the parent. .popover and uib-popover are children of this parent so just left to set popover-trigger=none and you have what you are wishing for.

I created a plunk example: https://plnkr.co/edit/6vzrrH1KUyQeLHDz1hWp?p=preview

bmrinal:
@odedkeren1 this doesn't work with popover-append-to-body="true" option set !!

Issue #3811, opened 2015-06-12
Carousel does not work with Angular 1.4.0


smajl:
EDIT: I have just noticed, that you only support Angular 1.3.x. You can close this if desired.

Carousel does only one cycle, then stops. Sliding manually does not work. There is no such problem with Angular 1.3.16 - just change version number in Plunker and see for yourself.

Angular: 1.4.0
UI-Bootstrap: 0.13.0
Plunker: http://plnkr.co/edit/IlauLdEJoar76cD14BLc?p=preview

wesleycho:
If I had to guess, this is due to the breaking changes with ngAnimate in 1.4.

Soon the team is going to have a discussion on what needs to be done to get 0.13.1 out the door, then we will be looking into Bootstrap 3.3 support & likely Angular 1.4

dkegle:
Is there any timeline for support Angular 1.4?

chrisirhc:
I'm looking forward to supporting 1.4 . Unfortunately, the current carousel animations are a bunch of hacks (my fault) made to run on $animate before $animateCss was introduced. It currently relies on JS and CSS animations running in parallel. However, this is no longer allowed as of 1.4 .

mchlbrnd:
A bit of experimental stuff on 13.0 because I need it for a project. Perhaps a nice PR

What do you think? Also, how would you go about supporting @chrisirhc on > 1.3; using angular.version to inject differently?

Cheers!

https://github.com/mchlbrnd/bootstrap/tree/carousel-animatecss.

wesleycho:
$animateCss does not exist in 1.3, and so we cannot move to changeover to use it until we start work on a release to support 1.4 unless we consume $injector directly and execute code depending on the angular version loaded. We are currently working towards a 0.13.1 release of bugfixes, and will work on 1.4 support after then.

Using $injector may not be a bad idea necessarily - any solution that would currently be accepted in master would have to support angular 1.3 though.

wesleycho:
In order to support that though, we would need some build tooling adjustment to run the tests for angular 1.3 and 1.4

mchlbrnd:
Indeed $injector + angular.version would work perfectly for now.

I haven't checked all specs for mocked $animate other than carousel, but the latter does not feature it. On the road now..

mchlbrnd:
Check https://github.com/mchlbrnd/bootstrap/commit/0490507f4652160961bb1ab58998f1cc4fc5a88e!

[...]

PR #3946, opened 2015-07-14
fix(carousel): Animation Callback in Angular 1.4

idealeric:
$animate no longer triggers events on the element itself, but rather has its own interface for managing these callbacks.

I wasn't sure what the protocal for supporting angular 1.3 vs 1.4 was, so I just put the version check in. 
Grunt passed locally and also worked in my manual testing.

Let me know. :)


wesleycho:
See the discussion in #3811 for guidance on the preferred implementation.

src/carousel/carousel.js
```diff
@@ -40,9 +40,18 @@ angular.module('ui.bootstrap.carousel', [])
...
       slide.$element.data(SLIDE_DIRECTION, slide.direction);
       $scope.$currentTransition = true;
-      slide.$element.one('$animate:close', function closeFn() {
-        $scope.$currentTransition = null;
-      });
+      if ( angular.version.minor >= 4 ) {
+        $animate.on('addClass', slide.$element, function closeFn14(aniElem, phase) {
+          if ( phase === 'close' ) {
+            $scope.$currentTransition = null;
+            $animate.off('addClass', aniElem);
```
chrisirhc: Shouldn't this be `closeFn14` ?

chrisirhc:
Did a review. This looks good. I thought at first that we need to check the major version of angular but I realized that if Angular 2 is used, it would probably be a breaking change anyways.

wesleycho:
Ideally, shouldn't we migrate over to $animateCss conditionally through usage of $injector similar to the suggestion that was made in the aforementioned issue? It contains explicit support for adding and removing classes, which I would be concerned about code fragility with ngAnimate support in abusing $animate to get 1.4 support given that it has been stated that $animateCss should be used for anything related to CSS animations (and that CSS animations otherwise would not be supported).

If this patch is to make it in, I would want to refactor it asap to change over to use $animateCss in the 1.4+ situation, since it could potentially be dangerous.

chrisirhc:
You're right, we still need to migrate over to `$animateCss` to do this animation the correct way.
However, I think this change is still needed even if we switch over to using `$animateCss` and this could be the bare minimum to get things working in 1.4 (I haven't tested to verify this). 

I'm not at all opposed to getting the $animateCss refactor in. It's just that if this is the bare minimum, we shouldn't block merging this since we can always do the $animateCss refactor later and things would still be working for a while.

src/carousel/carousel.js
```diff
@@ -40,9 +40,18 @@ angular.module('ui.bootstrap.carousel', [])
       slide.$element) {
       slide.$element.data(SLIDE_DIRECTION, slide.direction);
       $scope.$currentTransition = true;
-      slide.$element.one('$animate:close', function closeFn() {
-        $scope.$currentTransition = null;
-      });
+      if ( angular.version.minor >= 4 ) {
```
chrisirhc: Alternatively, test for existence of `$animate.on`.

idealeric:
I made these minor changes. :)

Also i'm with @chrisirhc. From my testing, the use of $animateCss is actually more or less equivalent to the use of .addClass currently. I would not be surprised if .addClass uses animateCss under the sheets anyway (the docs said as much, but I did not bother reading and angular source code).

wesleycho:
My comment about $animateCss is mainly due to the official stance on the ngAnimate api - in the future, it is conceivable that $animate will make a breaking change with regards to how it handles animation involving classes with $animate.

That said, I am not against getting the change in without $animateCss support, but it makes changing to use $animateCss for 1.4+ more urgent whenever possible to avoid breakage in the future.

wesleycho:
I did a build based off of this PR, and it seems like the animation is a little borked - check out the Plunker [here](http://plnkr.co/edit/XBChPC6SJPvkUVgNdiwk?p=preview) to see the UI glitch.

chrisirhc:
Hm.. indeed, in the Plunker, it looks like the classes from `$animate.addClass` isn't being applied. Looks like we'll need $animateCss after all.

Issue #3947, opened 2015-07-15
Possibility to show/hide indicators


demiro:
component: carousel
For some reason, there is no way to turn off the indicators, they are added automatically
Would be nice to have an option to hide them or better, not include them at all!

RicardoRFaria:
I think this would be a nice feature.

wesleycho:
This can be accomplished via CSS with ng-class on the carousel element.

I do not believe this is something UI Bootstrap should support, as the ng-class solution would be more performant than adding watchers for each potential toggle.

PR #3949, opened 2015-07-15
Hide carousel navigation arrows

RicardoRFaria:
As requested in issue #3947 now you can hide arrows in carousel using option hide-arrows
```html
<div style="height: 305px">
            <carousel interval="myInterval" hide-arrows="true">
                <slide ng-repeat="slide in slides" active="slide.active">
                    <img ng-src="{{slide.image}}" style="margin:auto;">
                    <div class="carousel-caption">
                        <h4>Slide {{$index}}</h4>
                        <p>{{slide.text}}</p>
                    </div>
                </slide>
            </carousel>
        </div>
```


demiro:
I didn't request arrows to be hidden, but indicators ;) But both options would be very good to have!

RicardoRFaria:
Sorry @demiro, I got confused. Bad day.
As requested now you can hide Indicators and with bonus, hide arrows.
```html
<div style="height: 305px">
            <carousel interval="myInterval" hide-arrows="true" hide-indicators="true">
                <slide ng-repeat="slide in slides" active="slide.active">
                    <img ng-src="{{slide.image}}" style="margin:auto;">
                    <div class="carousel-caption">
                        <h4>Slide {{$index}}</h4>
                        <p>{{slide.text}}</p>
                    </div>
                </slide>
            </carousel>
        </div>
```

demiro:
Thanks! You are "da man"

src/carousel/test/carousel.spec.js
```diff
@@ -138,6 +138,37 @@ describe('carousel', function() {
       expect(navPrev.length).toBe(0);
     });
 
+	it('should hide arrows when hideArrows is true', function () {
```
wesleycho: Fix indentation.

wesleycho:
If one wants to hide the arrows or indicators, I think it would make sense to use ng-class on the carousel element and use CSS - it would also be more performant.

Thoughts?

demiro:
It's how I did it...but its not a good solution since there are still unused elements in the DOM that have event listeners that are just there for no reason...


Proper way is to not includ them at all



Sent by Outlook for Android





On Fri, Jul 17, 2015 at 8:13 AM -0700, "Wesley Cho" <notifications@github.com> wrote:
If one wants to hide the arrows or indicators, I think it would make sense to use ng-class on the carousel element and use CSS - it would also be more performant.

Thoughts?

---

wesleycho:
But modifying ng-show for this doesn't change this situation - this implementation is wholly inferior in approach to my suggested solution, and does not buy anything except minor convenience.

I am inclined to reject this unless it adds a performance benefit, as this is something that can be done already without a hack on the developer side.

RicardoRFaria:
@wesleycho You right about performance, but I think that we can provide some option to remove this controls.
How about detach this elements from DOM?

demiro:
If it uses ngShow...than it is good only to hide stuff on the fly (dependi g on the controller) otherwise it is not what I was actually looking for


Sent by Outlook for Android





On Fri, Jul 17, 2015 at 8:54 AM -0700, "Wesley Cho" <notifications@github.com> wrote:
But modifying ng-show for this doesn't change this situation - this implementation is wholly inferior in approach to my suggested solution, and does not buy anything except minor convenience.

I am inclined to reject this unless it adds a performance benefit, as this is something that can be done already without a hack on the developer side.

---

wesleycho:
After reading the original issue, I am closing this, as this has a simple solution currently & only would serve to mislead developers as to the best course of action for their applications.

PR #3954, opened 2015-07-16
Fix Collapse flickering

### LINKED ISSUE #3801: Collapse flickers at page load in 0.13.0
In 0.12.1 'collapse' works well, but in 0.13.0 the element 'flickers' at page load when I set an initial $scope.collapse = true; as the first line in my controller. 

The 'collapsed' content just pops into view in a fast moment just after page load. It is very obvious on pages, were a lot of elements are collapsed at page load.
It happens the most in Chrome and Firefox, but IE seems to handle it better !?!

I have made a <a href="http://plnkr.co/edit/VSbgaHR4vlxOyAz69iVz?p=preview" target="_blank">Plunker</a> with the issue.

Try to Run / Stop the Plunker a few times, because the issue is not as obvious in all cases.

mseancole:
Only add `collapsing` class if `collapse` or `in` are already attached to element.

Fixes #3801
[Plunker](http://plnkr.co/edit/GYRzp1ZnPmnb080Nr25K?p=preview)

### DIFF
--- a/src/collapse/collapse.js
+++ b/src/collapse/collapse.js
@@ -17,6 +17,10 @@ angular.module('ui.bootstrap.collapse', [])
         }
 
         function collapse() {
+          if(! element.hasClass('collapse') && ! element.hasClass('in')) {
+            return collapseDone();
+          }
+
           element
             // IMPORTANT: The height must be set before adding "collapsing" class.
             // Otherwise, the browser attempts to animate from height 0 (in


wesleycho:
Question - does this make #3804 obsolete, or is it still necessary?

If indeed this solves the problem, then I have no problem merging this, as this looks like a better fix.

mseancole:
@wesleycho This would make it obsolete

wesleycho:
For posterity, [here](http://plnkr.co/edit/kJv0Mr0idGeMosneNNOQ?p=preview) is a working Plunker with your fix.

wesleycho:
This has now landed as ede9ea460629a044241c713a62e4de74c62d9d9a - thanks!

PR #3956, opened 2015-07-17
Add option to disable pagination from outside.

DarkAvanger:
I want to use this plugin in my project to paginate through tables, but because table content is ajax-based, I dont want to allow user clicking two times at next button at once.
The problem is not only mine, as can be seen in [this question](http://stackoverflow.com/questions/21609545/how-to-disable-ui-bootstrap-pagination/26902100#comment50910741_26902100).

To solve this issue I have added attribute paginationDisabled to pagination widget. When it is set to true - the widget does not responds to clicking and all buttons are marked as disabled.
Is should be used like this:

    <pagination 
        total-items="totalItems" 
        ng-model="currentPage" 
        ng-change="pageChanged()" 
        pagination-disabled="blocked"
    ></pagination>

### DIFF
--- a/src/pagination/pagination.js
+++ b/src/pagination/pagination.js
@@ -1,5 +1,3 @@
-angular.module('ui.bootstrap.pagination', [])
-
 .controller('PaginationController', ['$scope', '$attrs', '$parse', function ($scope, $attrs, $parse) {
   var self = this,
       ngModelCtrl = { $setViewValue: angular.noop }, // nullModelCtrl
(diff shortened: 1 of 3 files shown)


wesleycho:
I am not sure this is the best approach - I think hooking into ng-disabled and observing the value would be more performant, as well as less complicated to grasp.

src/pagination/pagination.js
```diff
@@ -43,11 +42,12 @@ angular.module('ui.bootstrap.pagination', [])
   };
 
   this.render = function() {
-    $scope.page = parseInt(ngModelCtrl.$viewValue, 10) || 1;
+      $scope.page = parseInt(ngModelCtrl.$viewValue, 10) || 1;
```
wesleycho: Undo this change.

src/pagination/pagination.js
```diff
@@ -43,11 +42,12 @@ angular.module('ui.bootstrap.pagination', [])
...
   };
 
   $scope.selectPage = function(page, evt) {
-    if ( $scope.page !== page && page > 0 && page <= $scope.totalPages) {
+    var disabled_click = $scope.paginationDisabled && evt !== undefined;
```
wesleycho: Use a camelcased variable - `disabled_click` is bad JS code style.

In addition, `$scope.paginationDisabled && evt` is a simpler to read check since checking for definedness is all the statement seems to be doing.

wesleycho:
LGTM, I will add some documentation and then merge it after getting some sleep.

PR #3961, opened 2015-07-18
href on tab changes location when deleting tab

JeffPlummer:
When I delete a tab in my angular app it changes the location, when it shouldn't.

Here's some context related to the issue
http://stackoverflow.com/questions/18080715/href-causes-location-address-to-change-can-we-avoid-it
and
https://github.com/angular/angular.js/issues/1225


wesleycho:
There was a problem with the `javascript:void(0);` method - it causes an error for CSP.

wesleycho:
This is superceded by 2b27dcbf997c3ddac37bfef75c0b256958fdd232

Issue #3404, opened 2015-03-18
Modal - Add bindToController


eranimo:
It would be nice to have a bindToController option for the $modal service. This would allow resolves to be bound to the controller instead of the scope much like it does in directives in Angular 1.3.

See the Angular documentation for bindToController:
> When an isolate scope is used for a component (see above), and controllerAs is used, bindToController: true will allow a component to have its properties bound to the controller, rather than to scope. When the controller is instantiated, the initial values of the isolate scope bindings are already available.

https://docs.angularjs.org/api/ng/service/$compile

Foxandxss:
First we need to see if we want to break compatibility with 1.2.x

wesleycho:
I think we've already broken compatibility with 1.2.x with the ngAnimate changes.

iyel:
+1

awerlang:
```
scope - a scope instance to be used for the modal's content 
(actually the $modal service is going to create a child scope of a provided scope). 
Defaults to $rootScope
```

I receive them on controller constructor function. As I use TypeScript, I'm able to prefix them with either `public` or `private` in order to have automatic properties.

Isn't your use case covered by means of constructor function parameters? Or perhaps I'm missing something?

ouchxp:
+1 for this. 
The inconsistency of controller sytay is really important. 
In the mean time. This issue really bothers our new angularjs developers. We have to add exceptions to the coding guideline that explain to our developers that when and why they are allow to break this consistency. That's really painful experience.

Quirksmode:
+1

kamarouski:
+1 

As we have Angular 2.0 coming it's generally recommended not to use $scope anymore and prefer bindToController for directives. 

The absence of bindToController is frustrating and painful.

wesleycho:
I'll take a look at implementing this tomorrow.

PR #3965, opened 2015-07-19
feat(modal): add support for bindToController

wesleycho:
- Add `bindToController` support in modal options

This implements #3404.


src/modal/docs/readme.md
```diff
@@ -8,6 +8,7 @@ The `$modal` service has only one method: `open(options)` where available option
 * `scope` - a scope instance to be used for the modal's content (actually the `$modal` service is going to create a child scope of a provided scope). Defaults to `$rootScope`
 * `controller` - a controller for a modal instance - it can initialize scope used by modal. Accepts the "controller-as" syntax in the form 'SomeCtrl as myctrl'; can be injected with `$modalInstance`
 * `controllerAs` - an alternative to the controller-as syntax, matching the API of directive definitions. Requires the `controller` option to be provided as well
+* `bindToController` - when used with `controllerAs` & set to `true`, it will bind the controller properties onto the `$scope` directly
```
Foxandxss: onto the `$scope` or onto `this.` ?

wesleycho: As it says, `$scope`. If you see the implementation, it becomes instantly clear. This is similar to how `bindToController` works in Angular for directives with isolate scope.

Foxandxss: Oh, you're right.

Foxandxss:
@wesleycho LGTM

src/modal/docs/readme.md
drpicox: I think that you understood `bindToController` in the other way around:

https://docs.angularjs.org/api/ng/service/$compile
> *bindToController*: When an isolate scope is used for a component (see above), and controllerAs is used, bindToController: true will allow a component to have its properties bound to the controller, rather than to scope. 

Bind to controller enforces the use of *controllerAs*, and it publishes the controller into the scope as it happens since 1.2. The new point is that all bindings to the directive are made towards the controller instead of towards the scope, but the scope does not dissappear, and you always use in template: ctrl.property.

The problem in angular 1.X is that you need to use the dot notation because you have a high risk (specilly with you work with non expert programmers) to mess it up with double bindings and nested scopes, for instance:

```javascript
Value: {{vm.value}}
<div ng-if="vm.isInputActive()">
  <input ng-model="vm.value">
</div>
```

vs.

```javascript
Value: {{value}}
<div ng-if="vm.isInputActive()">
  <input ng-model="value">
</div>
```

Which will not work as expected according the current implementation proposed in this thread: it will save the value in the scope created by ngIf but not in the directive context.

So, one of two:

1) bindToController does not make sense at all here, there are no bindings to the modal, so better to remove it (my recommendation)

2) ok, we want to bind controller properties to the scope, although it is dangerous, but we should name another name like: controllerPropertiesInScope, to avoid confusions with angular meaning

And according the presented implementation, it does not uses the controller instance at all, it is just a temporary object which instance reference is lost in favor to scope:

https://github.com/angular-ui/bootstrap/commit/8adfc833f710fd1aff5d91950e78362515946767
```javascript
angular.extend(modalScope, ctrlInstance);
```

Foxandxss: Check #4054 :)

drpicox: Nice!

It happened to me looking for ways to make "mastermodals" that I can accomodate for many uses. Ex:

- A generic PurchaseModal which is extended by MacaroonsPurchaseModal in which you for example need to request remote macaroons colors and prices and pass it to the directive or whatever to choose macaroon, select a price and go ahead with purchase.

The main problem is how to extend specific existing controller with some nice locals. In such case something like but different to bindToController would be useful. For example binding resolve results to controller.

Foxandxss: Please continue this conversation on #4054.

PR #3974, opened 2015-07-20
chore(bindHtml): add deprecation notice

wesleycho:
- Add deprecation notice for bindHtmlUnsafe

Closes #3574


Foxandxss:
LGTM, we should warn about deprecations, *cough*, *cough* ;( :P

PR #3975, opened 2015-07-20
fix(typeahead): don't close popup on right click

### LINKED ISSUE #3973: Typeahead inconsistent behavior on contextmenu in different browsers (Firefox)
Hi!

I've faced with issue that when typeahead dropdown is shown and on contextmenu (right click) triggered at any page area, it forces to close dropdown in FF, in Chrome dropdown still shown.

wesleycho:
- Ensure typeahead popup does not dismiss on right click in Firefox

This fixes #3973.


Foxandxss:
Easy enough.

Issue #3978, opened 2015-07-20
Tooltip placement does not update dynamically


cvn:
When using an interpolated value for `tooltip-placement` or `popover-placement` attributes, the tooltip does not update it's apperance until it is hidden and shows again.

http://plnkr.co/edit/nDerP4FeiaMIZTOrClxb?p=preview

Issue #3964, opened 2015-07-19
Tooltip shifts 1px


cvn:
Hello,

I'm seeing an issue where tooltips shift 1px horizontally at the end of their fade in animation.

Here's a plunkr: http://plnkr.co/edit/nv8iDI6dyGxvByomZaNj?p=preview

Using Chrome or Safari, hover over each of the buttons a few times. As you move between them, you'll see the problem on the left and rightmost buttons.

Here's [a gif](https://cloud.githubusercontent.com/assets/774864/8764414/b2f48eb0-2d8c-11e5-9fbe-55d2c60c6db5.gif) showing it. In this gif, I log the `offsetWidth` of the tooltip in the `positionElements` function, because I think it's related. It returns different values for the same tooltip.

Here's [another example](https://cloud.githubusercontent.com/assets/774864/8764424/2221f2d2-2d8d-11e5-9364-eedd64b3ee1c.gif), where the tooltip never stopped moving.

I think there's some kind of Webkit/Blink bug happening. The problem doesn't affect Firefox. I searched a little, but didn't find much about it. The closest thing was this [Stack Overflow post](http://stackoverflow.com/questions/26124678/chrome-bug-changing-left-css-changes-offsetwidth) about a similar offsetWidth problem.

This problem is only in ui.bootstrap 0.13.0. It is a regression from 0.12.1.

I'm using Angular 1.4.2 and Bootstrap 3.3.5 CSS. Testing with Chrome 43 and Safari 8 on a retina screen.

cvn:
I can prevent the shift from happening by removing this `$watch` on [line 249](https://github.com/angular-ui/bootstrap/blob/6af627a8edcbce0370ca2d90f9a575cd6770f6c1/src/tooltip/tooltip.js#L249-L251):

    tooltipLinkedScope.$watch(function () {
      $timeout(positionTooltip, 0, false);
    });

but I'm sure it's there for a good reason.

wesleycho:
Thanks for this clear investigation - if I had to guess, this looks like a potential Webkit/Blink bug. I am not able to reproduce this on Firefox or on IE11.

The watch is there to reposition in the event something changed in the UI - a bit of a hack, but I'd be hesitant to touch it.

cvn:
Thanks @wesleycho. I took a shot at fixing this in the above PR. It avoids the shift by being more deliberate about when to update the position.

Issue #288, opened 2013-03-31
"Edit in Plunker" Button Busted!


joshdmiller:
A new version of Plunker got pushed sometime in the last day or so, and now all our "Edit in Plunker" buttons error out: "500 TypeError: Cannot set property 'bootstrap' of undefined". I'm not sure yet if the Plunker API changed or if this is a regression, but since I don't have time to look into it right now, I'll just throw open the issue here.

PR #3980, opened 2015-07-20
Tooltip positioning & placement improvements

cvn:
This PR contains two changes:

1. Only update tooltip position when content changes, instead of every $digest. This fixes a bug where the tooltip would shift a pixel #3964, and should be a little more performant.
2. Adds the ability for changes to `tooltip-placement`/`popover-placement` to be reflected by the tooltip immediately #3978. The test for this is arguably incomplete, it only checks that the tooltip scope is updated, and doesn't check that the tooltip appearance has changed. I don't really know how to test the latter.

@chrisirhc It looks like you've done the most work on the tooltips recently, I would love to get your opinion on these changes.

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -307,6 +307,15 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
               positionTooltipAsync();
             });
 
+            attrs.$observe( prefix + 'Placement', function () {
+              if (ttScope.isOpen) {
+                $timeout(function () {
+                  prepPlacement();
+                  show()();
+                }, 0, false);
+              }
+            });
+
             function prepPopupClass() {
               ttScope.popupClass = attrs[prefix + 'Class'];
             }
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -136,6 +136,10 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
               tooltip.css( ttPosition );
             };
 
+            var positionTooltipAsync = function () {
+              $timeout(positionTooltip, 0, false);
+            };
+
             // Set up the correct scope to allow transclusion later
             ttScope.origScope = scope;
 
@@ -246,12 +250,9 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
                 }
               });
 
-              tooltipLinkedScope.$watch(function () {
-                $timeout(positionTooltip, 0, false);
-              });
-
               if (options.useContentExp) {
                 tooltipLinkedScope.$watch('contentExp()', function (val) {
+                  positionTooltipAsync();
                   if (!val && ttScope.isOpen ) {
                     hide();
                   }
@@ -287,6 +288,7 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
             if (!options.useContentExp) {
               attrs.$observe( type, function ( val ) {
                 ttScope.content = val;
+                positionTooltipAsync();
 
                 if (!val && ttScope.isOpen ) {
                   hide();
@@ -302,6 +304,7 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
 
             attrs.$observe( prefix+'Title', function ( val ) {
               ttScope.title = val;
+              positionTooltipAsync();
             });
 
             function prepPopupClass() {
--- a/src/tooltip/test/tooltip.spec.js
+++ b/src/tooltip/test/tooltip.spec.js
@@ -68,6 +68,24 @@ describe('tooltip', function() {
     expect( tooltipScope.placement ).toBe( 'bottom' );
   }));
 
+  it('should update placement dynamically', inject( function( $compile, $timeout ) {
+    scope.place = 'bottom';
+    elm = $compile( angular.element(
+      '<span tooltip="tooltip text" tooltip-placement="{{place}}">Selector Text</span>'
+    ) )( scope );
+    scope.$apply();
+    elmScope = elm.scope();
+    tooltipScope = elmScope.$$childTail;
+
+    elm.trigger( 'mouseenter' );
+    expect( tooltipScope.placement ).toBe( 'bottom' );
+
+    scope.place = 'right';
+    scope.$digest();
+    $timeout.flush();
+    expect(tooltipScope.placement).toBe( 'right' );
+  }));
+
   it('should work inside an ngRepeat', inject( function( $compile ) {
 
     elm = $compile( angular.element(
--- a/src/tooltip/test/tooltip.spec.js
+++ b/src/tooltip/test/tooltip.spec.js
@@ -474,24 +474,26 @@ describe( 'tooltip positioning', function() {
     tooltipScope = elmScope.$$childTail;
   }));
 
-  it( 'should re-position on every digest', inject( function ($timeout) {
+  it( 'should re-position when value changes', inject( function ($timeout) {
     elm.trigger( 'mouseenter' );
 
     scope.$digest();
     $timeout.flush();
     var startingPositionCalls = $position.positionElements.calls.count();
 
+    scope.text = 'New Text';
     scope.$digest();
     $timeout.flush();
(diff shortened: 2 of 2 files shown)


wesleycho:
This PR looks pretty solid to me, although another set of eyes couldn't hurt.

RobJacobs:
I'm not sure if wrapping the call to positionTooltip in $timeout is necessary, have you tried without the $timeouts?

cvn:
Yeah, the $timeouts were necessary in my testing. It *might* be possible to set priorities on the directives to remove the need for the $timeouts (so that the position calculation always happened after the text was rendered), but I'm just speculating, I didn't attempt this.

src/tooltip/tooltip.js
```diff
@@ -302,6 +299,18 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
...
+              if (prefix === 'popover') {
+                $timeout(positionTooltip, 0, false);
+              }
+            });
+
+            attrs.$observe( prefix + 'Placement', function () {
+              if (ttScope.isOpen) {
+                $timeout(function () {
+                  prepPlacement();
+                  show()();
```
chrisirhc: Doesn't this mean that if the placement is changed while the tooltip is hidden, it'll show?

src/tooltip/tooltip.js
```diff
@@ -302,6 +299,18 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
 
             attrs.$observe( prefix+'Title', function ( val ) {
               ttScope.title = val;
+              if (prefix === 'popover') {
```
chrisirhc: Why does the `popover` prefix need special handling? If a user used the `$tooltip` service to create their own tooltip directive, they wouldn't be able to get the special behavior that `popover` has unless they use the same prefix.

src/tooltip/tooltip.js
```diff
@@ -302,6 +299,18 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
...
+                $timeout(positionTooltip, 0, false);
```
chrisirhc: Does [$applyAsync](https://docs.angularjs.org/api/ng/type/$rootScope.Scope#$applyAsync) work instead of $timeout ?

src/tooltip/tooltip.js
```diff
@@ -284,15 +280,16 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
             /**
              * Observe the relevant attributes.
              */
-            if (!options.useContentExp) {
-              attrs.$observe( type, function ( val ) {
+            attrs.$observe( type, function ( val ) {
+              $timeout(positionTooltip, 0, false);
```
chrisirhc: Does [$applyAsync](https://docs.angularjs.org/api/ng/type/$rootScope.Scope#$applyAsync) work instead of $timeout ?

src/tooltip/tooltip.js
```diff
@@ -246,10 +246,6 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
                 }
               });
 
-              tooltipLinkedScope.$watch(function () {
-                $timeout(positionTooltip, 0, false);
-              });
-
               if (options.useContentExp) {
                 tooltipLinkedScope.$watch('contentExp()', function (val) {
```
chrisirhc: You might also need to schedule a reposition over here since if `contentExp()` changes, the content changes as well.

src/tooltip/tooltip.js
```diff
@@ -284,15 +280,16 @@ angular.module( 'ui.bootstrap.tooltip', [ 'ui.bootstrap.position', 'ui.bootstrap
             /**
              * Observe the relevant attributes.
              */
-            if (!options.useContentExp) {
-              attrs.$observe( type, function ( val ) {
+            attrs.$observe( type, function ( val ) {
+              $timeout(positionTooltip, 0, false);
+              if (!options.useContentExp) {
```
chrisirhc: You shouldn't need to change/move the `useContentExp` check. If `options.useContent` is false, this attribute (`type`) is never expected change.

src/tooltip/tooltip.js
cvn: No, it only runs if `ttScope.isOpen` is true, so only when it's already open.

src/tooltip/tooltip.js
cvn: I did not test it. I'm using $timeout because that's what you had used with the watch and it worked. I can try $applyAsync if you think there's a benefit.

src/tooltip/tooltip.js
chrisirhc: Ah right, I missed that check, but there's also when the value is a blank string, see above at line #288: `if (!val && ttScope.isOpen ) {`

src/tooltip/tooltip.js
chrisirhc: Ah, okay, makes sense. In that case, it might make sense to make a function called `repositionTooltip` that does the `$timeout` call, as you're trying to make the same reposition call in multiple places, it helps reading. I missed the fact that all the $timeout calls have the same arguments.

src/tooltip/tooltip.js
cvn: I moved the check inside the $observe because now it's also running positionTooltip inside the $observe independently of the check. This saves us from making two $observes on the same attr.

chrisirhc: `options.useContentExp` is used when the attribute *never changes* because it's actually an expression such as: `tooltipTemplate="'someUrl.html'"`. In such cases, the `tooltipTemplate` attribute never changes (it's always `'someUrl.html'`). So I'm not sure why you would need to run positionTooltip if `options.useContentExp` is true.

src/tooltip/tooltip.js
cvn: I did this because only the popover needs to reposition when its title changes. But I can remove the check to allow extensibility.

src/tooltip/tooltip.js
cvn: I'll do that.

src/tooltip/tooltip.js
chrisirhc: It's probably better to remove the check since any user of the `*Title` attribute would expect a re-render if the attribute's value changes. Users can make use of the `title` variable in the ttScope in their custom tooltips.

src/tooltip/tooltip.js
cvn: I think I explained myself poorly. I'm not messing with the contentExp logic. I'm just moving the conditional inside the $observe because I also want to use the same $observe for watching the tooltip value and repositioning. I'm using one $observe for two different purposes instead of adding another $observe.

chrisirhc: I understand what you're saying. I'm explaining to you that if `options.useContextExp` is true, the tooltip's value will never change. That's why the check was there, this observer isn't needed if that check returns true. It also means you don't need to observe the tooltip's value if the the `options.useContextExp`.

I'm not saying to separate the repositionTooltip logic out of this observer, I'm saying that the check should still be outside. You shouldn't need to observe the tooltip's value at all if `options.useContextExp` is true because it'll never change..

cvn: I see now. I'll make the change. In light of that I think I also need to add a reposition to the contentExp watch, as you suggested in an earlier comment.

src/tooltip/tooltip.js
cvn: I think this is a slightly different case that warrants a different conditional. On 288, there's new content to deal with and if the content is blank, it should hide the tooltip. Here there's no new content, just a new placement.

src/tooltip/tooltip.js
RobJacobs: From what I understand, $timeout(fn, 0, false) is sometimes preferred over $applyAsync as it won't trigger a digest cycle.  There is also $evalAsync with a good explanation of the difference between $applyAsync [here](http://www.bennadel.com/blog/2751-scope-applyasync-vs-scope-evalasync-in-angularjs-1-3.htm)

cvn:
Thanks @chrisirhc and @RobJacobs for the feedback. I pushed new commits incorporating the suggestions.

wesleycho:
The changes look good - if @chrisirhc signs off, this should be good to go.

wesleycho:
I built and tested this PR against the existing reproduction cases - this is some good work. I am going to merge it in right now.

PR #3982, opened 2015-07-21
make alert protractor testable

vlajos:
Basicall using the `$interval` module. More details: https://github.com/angular/protractor/issues/169#issuecomment-49078228

### DIFF
--- a/src/alert/alert.js
+++ b/src/alert/alert.js
@@ -19,13 +19,11 @@ angular.module('ui.bootstrap.alert', [])
   };
 })
 
-.directive('dismissOnTimeout', ['$timeout', function($timeout) {
+.directive('dismissOnTimeout', ['$interval', function($interval) {
   return {
     require: 'alert',
     link: function(scope, element, attrs, alertCtrl) {
-      $timeout(function(){
-        alertCtrl.close();
-      }, parseInt(attrs.dismissOnTimeout, 10));
+      $interval(alertCtrl.close, parseInt(attrs.dismissOnTimeout, 10), 1);
     }
   };
 }]);
--- a/src/alert/test/dismissOnTimeout.spec.js
+++ b/src/alert/test/dismissOnTimeout.spec.js
@@ -1,13 +1,13 @@
 describe('dismissOnTimeout', function () {
 
-  var scope, $compile, $timeout;
+  var scope, $compile, $interval;
 
   beforeEach(module('ui.bootstrap.alert'));
   beforeEach(module('template/alert/alert.html'));
-  beforeEach(inject(function ($rootScope, _$compile_, _$timeout_) {
+  beforeEach(inject(function ($rootScope, _$compile_, _$interval_) {
     scope = $rootScope;
     $compile = _$compile_;
-    $timeout = _$timeout_;
+    $interval = _$interval_;
   }));
 
   it('should close automatically if auto-dismiss is defined on the element', function () {
@@ -15,7 +15,7 @@ describe('dismissOnTimeout', function () {
     $compile('<alert close="removeAlert()" dismiss-on-timeout="500">Default alert!</alert>')(scope);
     scope.$digest();
 
-    $timeout.flush();
+    $interval.flush(500);
     expect(scope.removeAlert).toHaveBeenCalled();
   });
 });


vlajos:
(Sorry for the couple of rebases... this $interval is a bit new for me.)

wesleycho:
I'm hesitant on this - this smells like a hack, and one I don't believe that should be supported.

I would try resolving the issue using the other suggested approach in that issue of `ptor.ignoreSynchronization`

vlajos:
I think both solutions are a bit hack like.
But hopefully the $interval based solution makes all the library's users life easier.
They won't see the obscure error message, they won't have to figure out that the underlying $timeout is the problem and they won't have to implement a hack.
I think this is the most important difference.

src/alert/alert.js
```diff
@@ -19,12 +19,13 @@ angular.module('ui.bootstrap.alert', [])
...
 
-.directive('dismissOnTimeout', ['$timeout', function($timeout) {
+.directive('dismissOnTimeout', ['$interval', function($interval) {
   return {
     require: 'alert',
     link: function(scope, element, attrs, alertCtrl) {
-      $timeout(function(){
+      var intervalCanceller = $interval(function() {
         alertCtrl.close();
+        $interval.cancel(intervalCanceller);
```
petrjasek: not sure how this works with protractor, but it should be enough to set count to 1:
```js
$interval(function() {
     alertCtrl.close();
}, parseInt(attrs.dismissOnTimeout, 10), 1);
```

vlajos: @petrjasek Thanks! You are right. I have just updated this change.

wesleycho:
Problem is, this is a hack that is a completely incorrect use of $interval for what is very clearly a $timeout situation. This should be fixed in Protractor itself if this is a problem.

Closing as invalid.

vlajos:
OK. No problem.
However currently testing this project with protractor is a bit painful.
Could you please maybe contact them and explain your arguments?
It would be good to find an acceptable solution for both projects making the shared users life a bit easier. :-)

Foxandxss:
Even if I agree with @wesleycho in here, I had to make those changes to my toastr as well. They are not that problematic to merge and won't change the app code that much.

adiherzog:
I would have appreciated if you had merged this small change. Now I had to implement my own closeAlert function so that the $interval service is used. Protractor suggests to use the `ignoreSynchronization` setting, but I don't think that would be a superior solution. Protractor does not show any signs of fixing it on their side either (https://github.com/angular/protractor/issues/169#issuecomment-49078228).

Foxandxss:
I will probably merge it for the next version.

pkozlowski-opensource:
@Foxandxss how this would be a valid fix? $interval will keep firing, after initial close...

I'm worried here that we start to hack around bugs in other places and I agree with @wesleycho that we should fix the bug where it originates.

Foxandxss:
@pkozlowski-opensource You can make `$interval` to fire once. Basically make it work as a single `$timeout`. It won't make any difference at all in how it would work. It won't fire after initial close.

I agree that it is not the proper solution, better to have it fixed on Protractor than hacked on a library. My point is that sometimes we shouldn't simply say: Not my problem, wontfix.

This bug on protractor is well known and it haven't been fixed for a year (or even more). So we can simply ignore that it exists and make the end users angry because they will have to make even a bigger hack to be able to test, or we can merge a really **simple** hack, that won't do us any harm at all. It is not like we are adding an extra `$timeout`, that is an horrible hack, we are just swapping `$timeout` for a single time `$injector`. That would make the exact same behavior and all the ui-bootstrap users that does protractor testing would be really happy. If the Protractor teams fix it, we can simply revert it back.

I am not a member of the protractor library, so I can only speak for my users, and they need a solution for a bug that for some reason or another, is not fixed yet.

I had to fix this in a library of mine as well Foxandxss/angular-toastr#28 almost a year ago and everything works perfect.

----

Said that, I agree with both of you, in a good day, we won't need hacks, but the proper solution is out of my reach so I only worry for ui-bootstrap users.

vlajos:
(In this given patch $interval fires once.)

pkozlowski-opensource:
> My point is that sometimes we shouldn't simply say: Not my problem, wontfix.

It is not my point either. I'm just proposing a different action here: instead of hacking on multiple libraries (yours, here and probably hundreds of others) let's fix it at the source?

> the proper solution is out of my reach 

Why are you saying so? IMO focusing on a fix at the source would be more "economical" as compared to work-around in many libraries facing the same issue.

Foxandxss:
Out of my reach because Protractor is not a library of my competence.

Anyhow, looking at the protractor issue, seems like Julie recommended to use the `$interval` hack to solve this issue, you can see it here: https://github.com/angular/protractor/issues/169#issuecomment-26442090

Then she closed the issue.

If you think that it is still something that should be fixed in Protractor, let's talk with Julie again and close this issue. Best solution is always recommended, but sometimes that is not that clear which one is.

wesleycho:
My preference is that this should be fixed in Protractor - otherwise, what is the point in ever using $timeout? It has a strong smell to me, since this workaround would force a core piece of browser functionality to be avoided in favor of one that can be used in the same way, but not intended to. In addition, the browser probably has setTimeout optimized over setInterval for this use case, and we would lose any optimizations present by making this switch.

Foxandxss:
That is fine for me then :+1:

rcollette:
After reading lots of long threads in lots of places about this issue, it turns out $interval was specifically intended to provide a mechanism for creating timers/intervals that wouldn't affect Protractor synchronization.  It's not just some random alternate function whose usage doesn't match the intent of long running timers.  The "fix" was to add it to angular core and not protractor.  The intent is right there in the commit comment.
https://github.com/angular/angular.js/commit/2b5ce84fca7b41fca24707e163ec6af84bc12e83

PR #3996, opened 2015-07-22
chore(karma): Be able to test against 1.3 and 1.4

Foxandxss:
So by default there is no change (grunt will still test against 1.3 and same for travis), but if we bump this to support 1.4 (with all tests passing), we can uncomment the line to make travis and continuous test for 1.4 working as well.

By now you can simply:

```plain
$ grunt karma:continuous14
```

To test angular 1.4.x

I accept suggestions to make this better.


wesleycho:
So one thing to note, this means we have to manually maintain the versions of angular in the repository.

Foxandxss:
I don't have a better idea.

wesleycho:
I am not necessarily against it, I thought it was worth noting though.

Foxandxss:
Yeah, it is a bit of duplicating stuff, but grunt-karma doesn't seem to be that flexible.

wesleycho:
I'm fine with merging this, but I'd like a second opinion just to be sure.

Foxandxss:
Me too. Let's ping @kentcdodds for being helpful today.

kentcdodds:
I personally don't like including libs in my repos... I would recommend against doing this. I'm not sure what you're trying to accomplish though, why do you feel like you have to do this?

Foxandxss:
We want to be sure that all our tests run well with angular 1.3.x and 1.4.x

wesleycho:
I'm against it too in general, but I don't feel too strongly either way - in order to test against 1.4, one could just clear out the bower components, and bump up the version on bower.json temporarily & revert all changes when ready to file a PR.

kentcdodds:
With angular-formly, I say that angular-formly works with angular 1.2 and up. But all my tests run against 1.4. Occasionally, I'll pull down one of the others and test them, but most of the time I try to be aware of what the differences are and make sure I don't break anything. If I'm concerned, I'll make sure to pull it down locally and run the tests.

kentcdodds:
One thing that you may consider looking into is [`dont-break`](http://npm.im/dont-break)

Foxandxss:
Yeah, that is a good way, but the problem with this concrete library is that it does a lot of things and all of them different. It can potentially break with any change.

Foxandxss:
Well, I guess we can bump it to 1.4 and test it to 1.3 time to time. Less complications. @wesleycho ?

wesleycho:
Sounds good to me if we have this last 1.4 issue fixed.

Foxandxss:
Alrighty, we will simply test against 1.4 and check once a while for 1.3.

Issue #3941, opened 2015-07-13
Datepicker: Popout day of month width squishes day view in firefox


atentis:
If you have a limited column size your day view for your popout date picker will be squished in firefox. This is not an issue with chrome as it keeps its static size. Resize the screen to see the difference if necessary. 

http://plnkr.co/edit/kV8ICxL8Jiwmttzyt13G?p=preview

![capture](https://cloud.githubusercontent.com/assets/8903400/8662007/8fbeee22-2981-11e5-9062-a674fdf4bd47.JPG)

wesleycho:
Perhaps modifying the templates for the `day`, `month`, and `year` components with an appropriate min-width would work to fix this - feel free to file a PR for this.

atentis:
@wesleycho That worked. I will try to file a pr soon

PR #4000, opened 2015-07-22
Min-width fix on date picker

atentis:
Added min-width to fix firefox css issue. See
https://github.com/angular-ui/bootstrap/issues/3941#issuecomment-122363776


wesleycho:
This looks fixed according to [this Plunker](http://plnkr.co/edit/8jDz4mw4nZtYYKfPd7fS?p=preview), but this does add a non-ideal scrollbar...perhaps this is the best solution, given that a datepicker is not ideal for small screens.

Issue #3862, opened 2015-06-26
Datepicker - Date is Always Required?


hancock808:
See my Plunkr > http://plnkr.co/edit/1iGDEnnpEOtBPymPPiLS?p=preview

Dates are not required on my form, however when a date is entered then cleared out the form is marked invalid and the user cannot submit.

wesleycho:
This looks like a legitimate bug - this needs to be rewritten to use `ngModel.$validators`

onenive:
@wesleycho , I have a fix to this bug in my forked repo: https://github.com/blendlabs/bootstrap. Should I add tests and submit a PR for this issue?

wesleycho:
I am assuming that [this](https://github.com/blendlabs/bootstrap/commit/55334fb22ba263374508ca08ae5162158d7e2f4b) is the commit?

Feel free to open a PR - I think this is not quite the correct way to fix this though. The proper fix would change the setting of validity to be in `ngModelCtrl.$validators.date` and not manually checking `$attrs.ngRequired`.

onenive:
Would that be to:
1) log the error: "Datepicker directive: "ng-model" value must be a Date object..." regardless of what ng-required is set as
2) remove ngModelCtrl.$setValidity('date', ...) from this.render() and move all the validation logic to ngModelCtrl.$validators.date

pavlovt:
the bug still exists in version 1.3.2 (date is always required for uib-datepicker-popup component)

brunosantosnasajon:
Bug still exists in version 2.0.1. Is this project still maintained?

MatheusArleson:
Bug still happens on version 2.5.6.

I have an optional datepicker field initialized with null value.
When the datepicker is initialized, angular has a $error.date

An null date is an invalid date, but a null value for the model maybe not.

PR #4002, opened 2015-07-22
fix(datepicker): validate using ng-required attribute

onenive:
This is a solution for the the issue (https://github.com/angular-ui/bootstrap/issues/3862), where date is always required for datepicker, even when ng-required is set to false.

(todo: still need to add tests)

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -78,14 +78,14 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   this.render = function() {
     if ( ngModelCtrl.$viewValue ) {
       var date = new Date( ngModelCtrl.$viewValue ),
-          isValid = !isNaN(date) || !$attrs.ngRequired;
+          isValid = !isNaN(date);
 
       if ( isValid ) {
         this.activeDate = date;
-      } else {
+      } else if (!!$attrs.ngRequired){
         $log.error('Datepicker directive: "ng-model" value must be a Date object, a number of milliseconds since 01.01.1970 or a string representing an RFC2822 or ISO 8601 date.');
       }
-      ngModelCtrl.$setValidity('date', isValid);
+      ngModelCtrl.$setValidity('date', isValid || !$attrs.ngRequired);
     }
     this.refreshView();
   };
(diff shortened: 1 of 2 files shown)


onenive:
@wesleycho, moving the discussion to this PR: 

In response to your comments, I had some questions. Would the proper fix be to:
 
1) log the error: "Datepicker directive: "ng-model" value must be a Date object..." regardless of what ng-required is set as
2) remove ngModelCtrl.$setValidity('date', ...) from this.render() and move all the validation logic to ngModelCtrl.$validators.date

wesleycho:
Correct on both counts

onenive:
does this look good?

src/datepicker/datepicker.js
```diff
@@ -616,6 +615,11 @@ function ($compile, $parse, $document, $position, dateFilter, dateParser, datepi
 
       function validator(modelValue, viewValue) {
         var value = modelValue || viewValue;
+
+        if (!attrs.ngRequired && !value) {
```
wesleycho: Is this even necessary? Does this run if ngRequired is falsy?

onenive: Yes, it works even if ngRequired is falsy. If ngRequired is falsy (i.e. anything other than ng-required=true), then the datepicker is not required. So, it considers an empty value valid in that case.

wesleycho: I wasn't saying just working - I was wondering whether the validator even runs if `!attrs.ngRequired`. If it doesn't run, it wouldn't be necessary, but if it does, then this line is fine.

onenive: The validator does run if !attrs.ngRequired

wesleycho:
This LGTM, I will do a full check on Sunday and hopefully will merge it in shortly after.

onenive:
hey, is there any update on this?

wesleycho:
Sorry, haven't had the chance to check this out yet, as it requires a little more setup/evaluation than the other stuff I have looked at so far.

I do intend on reviewing this thoroughly this week though.

wesleycho:
Looks good [here](http://plnkr.co/edit/gIp1ixngsjx8pEp6jU7N?p=preview).

PR #4004, opened 2015-07-23
Focus is now trapped when modal is opened. It works in both directions with tab or shift+tab.

paulorbpacheco:
Fix Modal: Focus is now trapped inside when modal is opened. It works in both directions with tab or shift+tab. No more focus tab on background.
Refs: #738, #3689


wesleycho:
Does this work with multiple modals?

paulorbpacheco:
I didn't use with multiple modals, I tested using the main components page. I will create a plunk today and post here.

paulorbpacheco:
http://plnkr.co/izQCgs71XzDt3anp7ZH7

e-cloud:
nice, :+1:

src/modal/docs/demo.html
```diff
@@ -6,7 +6,7 @@ <h3 class="modal-title">I'm a modal!</h3>
         <div class="modal-body">
             <ul>
                 <li ng-repeat="item in items">
-                    <a ng-click="selected.item = item">{{ item }}</a>
+                    <a href="#" ng-click="selected.item = item">{{ item }}</a>
```
wesleycho: Any reason for this change? I think you should revert this, as I think it would have an adverse effect on the docs page.

wesleycho: Hmm, I think I get it - maybe change ng-click to
```
$event.preventDefault(); selected.item = item
```

wesleycho:
Work here is pretty solid, just needs some code style clean up - I will merge this shortly with all of the needed minor/cosmetic fixes.

e-cloud:
@wesleycho , @paulorbpacheco
**thanks for your effort** on the **traping tab feature on the modal module**. but i have a trouble with the new feature.

i use a tabset in the modal. when i update to **v0.13.2**, my modals without tabs work as expected, but it turn to tabs and modal aggregation, it can only trap the `tab` key on the second tab of the tabset.

here is a plukr that reproduce the problem.
http://plnkr.co/edit/wwgZhXu0fru4L703z5by?p=preview

i'm not sure if i use it on the wrong place or the modal source code has some bugs not concerned yet.
can you guys offset some help?

wesleycho:
Can you file a new issue? This isn't the place to discuss it.

e-cloud:
ok, i have filed it just now

paulorbpacheco:
Hello @e-cloud,
Thank you so much for find this bug. I would never think to use tabs inside modals :). Did you create the issue?

wesleycho:
#4149

PR #4005, opened 2015-07-23
Accordion accessibility enhancements

RicardoRFaria:
- Added roles to title for better reading of screen readers
- Make content of accordion tabbable when he is open (for keyboard navigation)
- Added aria attributes to describe accordion state

Using aria attributes I described title like a button, aria doesn't define accordion, only buttons and tabs. There is no consensus in this approach, but I think that for blind, hear button is conceptually closer than tab. We can discuss this.

### DIFF
--- a/template/accordion/accordion-group.html
+++ b/template/accordion/accordion-group.html
@@ -1,10 +1,11 @@
 <div class="panel panel-default">
   <div class="panel-heading">
-    <h4 class="panel-title">
-      <a href="#" tabindex="0" class="accordion-toggle" ng-click="$event.preventDefault(); toggleOpen()" accordion-transclude="heading"><span ng-class="{'text-muted': isDisabled}">{{heading}}</span></a>
+    <h4 class="panel-title" role="presentation">
+      <a href="#" tabindex="0" class="accordion-toggle" ng-click="$event.preventDefault(); toggleOpen()"
+         role="button" aria-label="{{heading}}" aria-expanded="{{isOpen}}" accordion-transclude="heading"><span ng-class="{'text-muted': isDisabled}">{{heading}}</span></a>
     </h4>
   </div>
   <div class="panel-collapse collapse" collapse="!isOpen">
-	  <div class="panel-body" ng-transclude></div>
+	  <div class="panel-body" tabindex="{{isOpen ? '0' : '-1'}}" role="document" aria-hidden="{{!isOpen}}" ng-transclude></div>
   </div>
 </div>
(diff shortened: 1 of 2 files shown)


template/accordion/accordion-group.html
```diff
@@ -1,10 +1,11 @@
...
-    <h4 class="panel-title">
-      <a href="#" tabindex="0" class="accordion-toggle" ng-click="$event.preventDefault(); toggleOpen()" accordion-transclude="heading"><span ng-class="{'text-muted': isDisabled}">{{heading}}</span></a>
+    <h4 class="panel-title" role="presentation">
+      <a href="#" tabindex="0" class="accordion-toggle" ng-click="$event.preventDefault(); toggleOpen()"
+         role="button" aria-label="{{heading}}" aria-expanded="{{isOpen}}" accordion-transclude="heading"><span ng-class="{'text-muted': isDisabled}">{{heading}}</span></a>
     </h4>
   </div>
   <div class="panel-collapse collapse" collapse="!isOpen">
-	  <div class="panel-body" ng-transclude></div>
+	  <div class="panel-body" tabindex="{{isOpen ? '0' : '-1'}}" role="document" aria-hidden="{{!isOpen}}" ng-transclude></div>
```
wesleycho: The attribute shouldn't be in the DOM when the element is shown though - wouldn't it be better to use `ng-attr-aria-hidden="!isOpen"`?

RicardoRFaria: I think that the result is the same. But I dont have certain.
Should I change?

wesleycho: The way it currently is, it is present always - similarly with aria-expanded. The attributes should only be present in the appropriate situation.

RicardoRFaria: Oh, you right, I will change.

RicardoRFaria: I read the documentation of [ng-Aria](https://docs.angularjs.org/api/ngAria) and he will mark as hidden only if is properly configured.
So I will remove the attribute and leave in hands of angular.

wesleycho:
There should be tests to test the proper presence of `aria-hidden` and `aria-expanded` when appropriate.

RicardoRFaria:
Can I include ngAria lib as test dependency?

wesleycho:
Sure, but why is ngAria needed here if we are using `ng-attr-`?

wesleycho:
Oh, I see now - sure.

RicardoRFaria:
Ok

RicardoRFaria:
@wesleycho I returned the tests for aria-hidden but now checking the previos div, aparently angular is adding the aria-hidden attribute automatically, the problem is that he don't remove this aria attribute never.

icfantv:
Closing in favor of #4772.

PR #3344, opened 2015-02-27
fix(datepicker): date format in EcmaScript for firefox and others

FernandoVelcic:

PR #4014, opened 2015-07-24
update datepicker.js

saichaitanya88:
On line 656:

Angular validation is triggered when date is an empty string because an empty string is an invalid date object. Setting date to null mitigates this issue.

This happens when the input type is set to 'date'


wesleycho:
I like this change - good catch.

RobJacobs:
See related PRs #2513, #3344

PR #4019, opened 2015-07-27
feat(timepicker): Added min/max attributes to timepicker.

mikeknoll:
Adds attributes for timepicker's min and max time.

### DIFF
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -56,20 +56,16 @@ angular.module('ui.bootstrap.timepicker', [])
   }
 
   var min;
-  if ($attrs.min) {
-    $scope.$parent.$watch($parse($attrs.min), function(value) {
-      var dt = new Date(value);
-      min = isNaN(dt) ? undefined : dt;
-    });
-  }
+  $scope.$parent.$watch($parse($attrs.min), function(value) {
+    var dt = new Date(value);
+    min = isNaN(dt) ? undefined : dt;
+  });
 
   var max;
-  if ($attrs.max) {
-    $scope.$parent.$watch($parse($attrs.max), function(value) {
-      var dt = new Date(value);
-      max = isNaN(dt) ? undefined : dt;
-    });
-  }
+  $scope.$parent.$watch($parse($attrs.max), function(value) {
+    var dt = new Date(value);
+    max = isNaN(dt) ? undefined : dt;
+  });
 
   $scope.noIncrementHours = function() {
     var incrementedSelected = addMinutes( selected, hourStep * 60 );
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -310,8 +310,9 @@ angular.module('ui.bootstrap.timepicker', [])
 
   function addMinutes( date,  minutes ) {
     var dt = new Date( date.getTime() + minutes * 60000 );
-    var newDate = new Date( date.getTime() ).setHours( dt.getHours(), dt.getMinutes() );
-    return new Date( newDate );
+    var newDate = new Date( date );
+    newDate.setHours( dt.getHours(), dt.getMinutes() );
+    return newDate;
   }
 
   function addMinutesToSelected( minutes ) {
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -55,6 +55,50 @@ angular.module('ui.bootstrap.timepicker', [])
     });
   }
 
+  var min;
+  if ($attrs.min) {
+    $scope.$parent.$watch($parse($attrs.min), function(value) {
+      var dt = new Date(value);
+      min = isNaN(dt) ? undefined : dt;
+    });
+  }
+
+  var max;
+  if ($attrs.max) {
+    $scope.$parent.$watch($parse($attrs.max), function(value) {
+      var dt = new Date(value);
+      max = isNaN(dt) ? undefined : dt;
+    });
+  }
+
+  $scope.noIncrementHours = function() {
+    var incrementedSelected = addMinutes( selected, hourStep * 60 );
+    return incrementedSelected > max || (min && incrementedSelected < selected);
+  };
+
+  $scope.noDecrementHours = function() {
+    var decrementedSelected = addMinutes( selected, - hourStep * 60 );
+    return decrementedSelected < min || (max && decrementedSelected > selected);
+  };
+
+  $scope.noIncrementMinutes = function() {
+    var incrementedSelected = addMinutes( selected, minuteStep );
+    return incrementedSelected > max || (min && incrementedSelected < selected);
+  };
+
+  $scope.noDecrementMinutes = function() {
+    var decrementedSelected = addMinutes( selected, - minuteStep );
+    return decrementedSelected < min || (max && decrementedSelected > selected);
+  };
+
+  $scope.noToggleMeridian = function() {
+    if(selected.getHours() < 13) {
+      return addMinutes( selected, 12 * 60 ) > max;
+    } else {
+      return addMinutes( selected, - 12 * 60 ) < min;
+    }
+  };
+
   // 12H / 24H mode
   $scope.showMeridian = timepickerConfig.showMeridian;
   if ($attrs.showMeridian) {
@@ -177,7 +221,11 @@ angular.module('ui.bootstrap.timepicker', [])
 
       if ( angular.isDefined(hours) ) {
         selected.setHours( hours );
-        refresh( 'h' );
+        if( selected < min || selected > max) {
+          invalidate(true);
+        } else {
+          refresh( 'h' );
+        }
       } else {
         invalidate(true);
       }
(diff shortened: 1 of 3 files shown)


wesleycho:
That is what `min-date` and `max-date` are for.

wesleycho:
I apologize, this is for a different component than I thought - I will review this.

src/timepicker/timepicker.js
```diff
@@ -254,9 +312,14 @@ angular.module('ui.bootstrap.timepicker', [])
     $scope.meridian = selected.getHours() < 12 ? meridians[0] : meridians[1];
   }
 
-  function addMinutes( minutes ) {
-    var dt = new Date( selected.getTime() + minutes * 60000 );
-    selected.setHours( dt.getHours(), dt.getMinutes() );
+  function addMinutes( date,  minutes ) {
+    var dt = new Date( date.getTime() + minutes * 60000 );
+    var newDate = new Date( date.getTime() ).setHours( dt.getHours(), dt.getMinutes() );
+    return new Date( newDate );
```
wesleycho: Is there any reason why `newDate` can't just get returned? It should already be an independent reference.

src/timepicker/timepicker.js
```diff
@@ -55,6 +55,50 @@ angular.module('ui.bootstrap.timepicker', [])
...
+  if ($attrs.max) {
+    $scope.$parent.$watch($parse($attrs.max), function(value) {
+      var dt = new Date(value);
+      max = isNaN(dt) ? undefined : dt;
+    });
+  }
+
+  $scope.noIncrementHours = function() {
+    var incrementedSelected = addMinutes( selected, hourStep * 60 );
+    return incrementedSelected > max || (min && incrementedSelected < selected);
```
wesleycho: If !$attrs.max or !$attrs.min, this does not seem to be the behavior we want.

wesleycho:
Thanks for this, this is mostly solid work - I do have two comments that should be addressed before this is merged.

src/timepicker/timepicker.js
mikeknoll: Date.setHours() returns an int, not a Date for some reason.  I returned new Date( newDate ) to convert the int back into a Date.

src/timepicker/timepicker.js
mikeknoll: I'm not sure I understand.  I was following the pattern of the other attributes, such as $attrs.hourStep and $attrs.minuteStep.  If the attribute is present, add a watch for the expression in it.

wesleycho: These functions will run regardless of whether they exist

src/timepicker/timepicker.js
wesleycho: Oh, I see now.

It would likely be more efficient to do
```
var newDate = new Date( date.getTime() + minutes * 60000 );
newDate.setHours( dt.getHours(), dt.getMinutes() );
return newDate;
```
The `dt` date existed as a temporary structure to figure out how to modify the existing `selected` date, but now we can just return the date of interest

mikeknoll:
I believe I addressed your concerns.  I was able to remove an instance of Date in addMinutes(), slightly different than your suggest because it needs to "rollover" from 11:59PM to 12:00AM on the same day, not the next.

I removed the unnecessary conditionals on $attrs.min/max.

src/timepicker/timepicker.js
```diff
@@ -55,6 +55,46 @@ angular.module('ui.bootstrap.timepicker', [])
...
+
+  var max;
+  $scope.$parent.$watch($parse($attrs.max), function(value) {
+    var dt = new Date(value);
+    max = isNaN(dt) ? undefined : dt;
+  });
+
+  $scope.noIncrementHours = function() {
+    var incrementedSelected = addMinutes( selected, hourStep * 60 );
+    return incrementedSelected > max || (min && incrementedSelected < selected);
```
wesleycho: So what I meant was more `return (max && incrementedSelected > max) || incrementedSelected < selected;`, and so on.

src/timepicker/timepicker.js
```diff
@@ -55,6 +55,46 @@ angular.module('ui.bootstrap.timepicker', [])
...
+  };
+
+  $scope.noDecrementMinutes = function() {
+    var decrementedSelected = addMinutes( selected, - minuteStep );
+    return decrementedSelected < min || (max && decrementedSelected > selected);
+  };
+
+  $scope.noToggleMeridian = function() {
+    if(selected.getHours() < 13) {
```
wesleycho: Add space here, i.e. `if (`

src/timepicker/timepicker.js
```diff
@@ -177,7 +217,11 @@ angular.module('ui.bootstrap.timepicker', [])
 
       if ( angular.isDefined(hours) ) {
         selected.setHours( hours );
-        refresh( 'h' );
+        if( selected < min || selected > max) {
```
wesleycho: Change to `if (selected...`

src/timepicker/timepicker.js
```diff
@@ -196,7 +240,11 @@ angular.module('ui.bootstrap.timepicker', [])
 
       if ( angular.isDefined(minutes) ) {
         selected.setMinutes( minutes );
-        refresh( 'm' );
+        if( selected < min || selected > max) {
```
wesleycho: Change to `if (selected...`

src/timepicker/timepicker.js
```diff
@@ -222,7 +270,13 @@ angular.module('ui.bootstrap.timepicker', [])
       if ( date ) {
         selected = date;
       }
-      makeValid();
+      if( selected < min || selected > max ) {
```
wesleycho: Ditto

src/timepicker/timepicker.js
```diff
@@ -264,19 +324,29 @@ angular.module('ui.bootstrap.timepicker', [])
...
   
   $scope.incrementHours = function() {
-    addMinutes( hourStep * 60 );
+    if(!$scope.noIncrementHours()) {
+      addMinutesToSelected( hourStep * 60 );
+    }
   };
   $scope.decrementHours = function() {
-    addMinutes( - hourStep * 60 );
+    if(!$scope.noDecrementHours()) {
```
wesleycho: Change to `if (`

src/timepicker/timepicker.js
```diff
@@ -264,19 +324,29 @@ angular.module('ui.bootstrap.timepicker', [])
...
+      addMinutesToSelected( - hourStep * 60 );
+    }
   };
   $scope.incrementMinutes = function() {
-    addMinutes( minuteStep );
+    if(!$scope.noIncrementMinutes()) {
```
wesleycho: Ditto

src/timepicker/timepicker.js
```diff
@@ -264,19 +324,29 @@ angular.module('ui.bootstrap.timepicker', [])
...
+      addMinutesToSelected( minuteStep );
+    }
   };
   $scope.decrementMinutes = function() {
-    addMinutes( - minuteStep );
+    if(!$scope.noDecrementMinutes()) {
```
wesleycho: Ditto

src/timepicker/timepicker.js
```diff
@@ -264,19 +324,29 @@ angular.module('ui.bootstrap.timepicker', [])
...
+      addMinutesToSelected( - minuteStep );
+    }
   };
   $scope.toggleMeridian = function() {
-    addMinutes( 12 * 60 * (( selected.getHours() < 12 ) ? 1 : -1) );
+    if(!$scope.noToggleMeridian()) {
```
wesleycho: Ditto

wesleycho:
Once these changes are in, this is good to go.

src/timepicker/timepicker.js
mikeknoll: The logic is a bit different than that.  `incrementedSelected > max` is catching the common case, increasing the time would exceed the max.  If max is undefined, `incrementedSelected > max` will be false, which is what we want(to allow the increase).

The other part, `(min && incrementedSelected < selected)` was intended to disable the increment/decrement if the time would rollover(at midnight) if the opposite direction's wouldn't be allowed.  For example, if it is 11:30pm, you cannot increase to 12:30am if the minimum is 1:00am.

This code isn't quite there though.  This just prevents a rollover if the minimum/maximum is defined. For example, if it is 11:30pm, you cannot increase to 12:30am if the minimum is 12:01am.  This should be allowed, as the updated time is still later than the minimum.

I believe the correct code will be `return incrementedSelected > max || (incrementedSelected < selected && incrementedSelected < min);`.  I've manually tested it.  I'll write up some unit tests and submit it.

mikeknoll:
I updated the style, and changed the noIncremenentHours() etc... to disable the time step in a direction if when time rolls over, the result exceeds the limit for movement in the opposite direction.

wesleycho:
LGTM - merging shortly with documentation & code style fixes.

PR #4023, opened 2015-07-27
Correction in the popup position

marlonmleite:
Upon receiving position in the directive, we need to use '=' to place the correct bind the scope position.


wesleycho:
This is a breaking change and needs a lot more justification than given.

wesleycho:
Closing as invalid - feel free to open an issue with a reproduction of the problem & as much detail as possible so we can properly fix the issue at hand.

PR #4026, opened 2015-07-28
fix(buttons): respect disabled attribute

### LINKED ISSUE #4013: disabled and ng-disabled ineffective on radio and checkbox buttons with Bootstrap CSS 3.3.5
With Bootstrap CSS 3.3.4, `ng-disabled` and `disabled` applied to `label` associated with the `btn-radio` or `btn-checkbox` directives work has expected (See this [plunk](http://plnkr.co/edit/m5sHzwsIPPQTQpAgJsNJ?p=preview))

After upgrading to Bootstrap CSS 3.3.5, it does not work anymore (See this [plunk](http://plnkr.co/edit/O2kghWBeYszJWwXpdC8X?p=preview)).

To be more precise, styling is good (even better than with 3.3.4) but behaviour is not what it should be.

I guest solving issue [#16092](https://github.com/twbs/bootstrap/pull/16092) on Bootstrap CSS had a harmful effect on `radio` and `checkbox` `buttons` behaviour.

wesleycho:
- Ensure disabled attribute is respected due to a change in Bootstrap 3.3.5 CSS

Plunker in action [here](http://plnkr.co/edit/Xbjb2oAzfx2yt95VA0C6?p=preview).

This fixes #4013


src/buttons/buttons.js
```diff
@@ -24,6 +24,10 @@ angular.module('ui.bootstrap.buttons', [])
 
       //ui->model
       element.bind(buttonsCtrl.toggleEvent, function () {
+        if ('disabled' in attrs) {
+          return;
```
RobJacobs: Is this the preferred syntax over if(attrs.disabled) which dropwdownToggle and tab directives use?

wesleycho: The issue is Angular 1.4 - it turns out that `attrs.disabled` is `undefined` there.

wesleycho: In the Tabs component, this should be fine - the only way there would be a problem is if the `disabled` attribute is used instead of `disable`, which is deprecated use.

For the dropdown toggle, I opened #4033

wesleycho: Update: Looks like the dropdown works perfectly fine in 1.4

wesleycho: Tabs look [fine too](http://plnkr.co/edit/NU5aUt2rK5NpYXQ2N2eW?p=preview).

Foxandxss:
LGTM

PR #4033, opened 2015-07-28
fix(dropdown): change check for angular 1.4

wesleycho:
- Fix attribute check due to breaking change in Angular 1.4


wesleycho:
I need to investigate this more properly.

wesleycho:
This appears to be invalid - there is something else going on with the dropdown for 1.4 though, as exhibited [here](http://plnkr.co/edit/LNXHR0uWL3CkTmJPnmuw?p=preview)

Issue #3937, opened 2015-07-13
datetimepicker - wrong validatio


agrudz:
I use datepicker and i'm trying to type date using keyboard. The specified format for datetimepicker is yyyy-MM-dd. When i type whole data it works correct. when i tries to change for examle month it starts to fail.
The path to the bug looks like this:
1. I have correct date in my datetimepicker for example 2015-07-14
2. I whant to change it to 2015-08-14 so i start editing it and im removing 7 - i have 2015-0-14 writen in my datetimepicker. The validation shows me that it is not correct date ant this is fine for me
3.I type 8 and i have 2015-08-14 in my datetimepicker. The validation message still shows up...
There is a screen from my debuger - 
![debuger](https://cloud.githubusercontent.com/assets/13309807/8644345/9de19cfe-293e-11e5-9686-801f46bf8860.png)
You can notice that houres, minutes and seconds have NaN values. The bug is easy to correct.
I edited this piece of code
i replace this :
     if ( isValid(fields.year, fields.month, fields.date) ) {
        dt = new Date(fields.year, fields.month, fields.date, fields.hours, fields.minutes, fields.seconds,
          fields.milliseconds || 0);
      }
 with:

       dt = new Date(fields.year, fields.month, fields.date, fields.hours || 0, fields.minutes || 0, fields.seconds || 0,
          fields.milliseconds || 0);

goliney:
I have problems with direct input editing too. Here is plunkr with bug description:
http://plnkr.co/edit/SJpireJtJfTmeK80QmGi?p=preview

inaccessible:
+1

zlailari:
+1

wesleycho:
This is a duplicate of #3713 . I believe that #3759 will address this.

PR #4035, opened 2015-07-29
Fixed bug "datetimepicker - wrong validatio #3937"

darkcrou:
Fixed by changing "baseDate" parser argument if-statement condition (absence and valid Date type instance)


wesleycho:
Closing as duplicate of #3759

PR #4041, opened 2015-07-30
fix(carousel): fix animations with 1.4

### LINKED ISSUE #3811: Carousel does not work with Angular 1.4.0

wesleycho:
- Change to use `$animateCss` in angular 1.4 and fallback to existing behavior for angular 1.3

The behavior can be observed in [this Plunker](http://plnkr.co/edit/MNmmdR7ychBC1ADJAg0Y?p=preview) - there is a minor UI glitch where the first slide slides out to the right with Angular 1.4 currently that I am investigating. This otherwise works fine with 1.4 and 1.3.

Many thanks to @mchlbrnd for the fix linked in the issue.

This fixes #3811 and supercedes #3946

Edit: Fixed the 1.4 specific transition issue, updated Plunker.


mchlbrnd:
My bad; I was not able to pull request this due to other business, namely holidays. ;)

Many thanks @wesleycho for getting it done!

Foxandxss:
That plunker is working weird for me. Sometimes the slides doesn't appear and if you change it (in the application, not source) to 500 delay, it works really bad.

wesleycho:
Hmm, good catch - I'll have to investigate that.

wesleycho:
@Foxandxss looks like this issue is because the carousel has a transition for each slide set to 0.6s. I am not sure whether we should try to necessarily support the old behavior again for 1.4, given that a transition that is slower than the slide interval will cause slides to buffer - there likely is a change in ngAnimate that interplays with this poorly due to how rAF (if a change is buffered right before another is complete, it will cause the browser to not render anything).

At the very least, this is still superior to the current behavior & think this is fine to merge in as is - I am willing to investigate this further though, as I do think we need to figure out a strategy for handling this situation.

Foxandxss:
I am OK with that. It is better from what we had so I it is good to go. We can make it better on a future.

PR #4043, opened 2015-07-30
chore(build): update dependencies

wesleycho:
- Change the angular version to 1.4.3 for docs
- Change the Bootstrap version to 3.3.5 for docs


Foxandxss:
We sure we want to bump the bootstrap version?

wesleycho:
I checked every component when I was checking for angular 1.4 compatibility, didn't see any issues - it would also allow us to fix them quicker if they do happen to come up.

Foxandxss:
Oh, that is good to hear. The LGTM

wesleycho:
Hmm looking at the 0.14 milestone, perhaps it would be wise to reset that - there are a few minor issues around.

PR #4054, opened 2015-07-31
fix(modal): fix `bindToController` implementation

### LINKED ISSUE #4051: modal bindToController inconsistent behavior
In the modal options, the bindToController's behavior is inconsistent with how angular controller's bindToController behaves.

In the example, the controllerAs name is "ctrl" and also bindToController set to true.

The direction of "binding" appears to be reversed.  In angular, the bindToController takes the variables bound in $scope and put them INTO the controller.  As a result, references to "$scope.ctrl.field" will find the fields.  Referencing these fields in the html should be "ctrl.field"  If the field is a simple value (like a string), binding it there will update ctrl.field (or $scope.ctrl.field).

From what i'm seeing in the code, bootstrap is doing the opposite.  When bindToController is set to true, It is taking a "copy" of the fields in the controller and then pu those inside the the $scope variable.  As a result, the fields actually now exists in $scope and not $scope.ctrl.  The code did an angular.extend(scope, controller).  Once the copy is done, the linkage will be broken on the simple fields.  Also, no $scope.ctrl object is linked/created (since that is on the "else" path).  This doesn't look right.  Can someone clearify/verify.

Thanks.

Tony

wesleycho:
- Correct the implementation of `bindToController` to match Angular with properties on `$scope` being bound on the controller

Fixes #4051


drpicox:
Nice!

(as following which I started in https://github.com/angular-ui/bootstrap/pull/3965 )

> It happened to me looking for ways to make "mastermodals" that I can accomodate for many uses. Ex:

In angular 1.4, although it is not well documented, you can use bindToController as follows:

```javascript
scope: {}, // <-- personal recommendation to avoid problems
bindToController: {
   name: '=',
},
```

What it is doing is assigning binding values to controller. It may be useful for assigning values to modal controllers (without creating custom $scopes):

```javascript
bindToController: {
   name: function(nameService) { return nameService.get(); },
}
```

or may be:

```javascript
resolve: {
   name: function(nameService) { return nameService.get(); },
},
bindToController: {
   name: '=',
},
```

But I still do not like any of both (probably just not the right naming).  

The idea is to allow do something like the following example but without doing a sort of hacking (something too complex to write and understand because the lacking of desidered/useful feature):

```javascript
scope: customScope,
resolve: {
   __unused_by_controller__name: function(nameService) {
     return nameService.get().then(function(name) {
        customScope.name = name;
     };
  }
},
```

It is required because if you do an modal.open you are expected to return a modal, not a promise to a modal. The modal itself has the promise inside.

If anyone has a good idea how to implement it, it would be nice to heard of.

wesleycho:
I think that sort of support should wait, if it should even be implemented - that would require some further thought on what is the meaning of an isolate $scope inside a modal.

wesleycho:
@Foxandxss said LGTM, so merging now.

PR #4061, opened 2015-08-01
chore(tab): change to linking function

wesleycho:
- Change to use linking function as the use of `transclude` on the compile has been deprecated by Angular

The deprecation notice in the documentation is [here](https://docs.angularjs.org/api/ng/service/$compile).


chrisirhc:
LGTM

PR #4063, opened 2015-08-01
fix(typeahead): reset matches if enter is hit

### LINKED ISSUE #3545: typeahead close select on form submit
I would like to get the select list to close when the form is submitted with a value which is not on the list (using `typeahead-focus-first` set to `false`). 
I used the [typeahead](http://angular-ui.github.io/bootstrap/#/typeahead) plunker example as a starting point.

When the user hits enter, the `storeItem` function fires, but the list does not close.
```
    <form ng-submit="storeItem()">
      <input type="text"
        typeahead="state for state in states | filter:$viewValue | limitTo:8" 
        typeahead-focus-first="false"/>
    </form>
```
The full function:

```
    $scope.storeItem = function() {
       $scope.selected = Date.now();
    };
```

See full plunker [here](http://plnkr.co/edit/nsQXVAyjuB5WpIgE2rWZ?p=preview)

If this is not an issue, and you have an idea for a workaround, I'd appreciate if you could answer this [SO question](http://stackoverflow.com/questions/29626129/angular-bootstrap-typeahead-close-select)

wesleycho:
- If no match is selected and enter is hit, reset matches

Plunker with fix is [here](http://plnkr.co/edit/4Iodg8wQrd7IVnucme3D?p=preview)

This fixes #3545.


wesleycho:
Vetted by @Foxandxss

Issue #3457, opened 2015-03-28
Consider dropping ngDocs


chrisirhc:
Looks like this isn't used, and it's sort of broken. Doing `grunt ngdocs` creates a documentation page that doesn't seem to work properly. Not sure whether to remove it entirely or fix it later.

wesleycho:
We should probably remove it completely if it is broken - the demo page serves as our documentation page currently, as the library generally is pretty simple to use.

I know the Angular team created dgeni, we can probably try migrating to that as a docs solution.

Foxandxss:
Yes, deleting it completely, then in a future consider other kind of doc.

PR #4064, opened 2015-08-01
chore(build): remove ngdocs

wesleycho:
- Remove ngdocs since it is not functional

This addresses #3457.


Foxandxss:
We added this long time ago but we never used it actually. so fire with it.

PR #4069, opened 2015-08-02
fix(buttons): handle expressions in `btn-radio` directive

### LINKED ISSUE #3220: Adding Tooltips to Radio Buttons Allows Multiple Buttons in Group to Appear Active
If you put the tooltip directive on each button acting as a radio button in a group, clicking the button does not cause the other buttons in the radio button group to be deselected, and therefor it appears that multiple radio buttons in the group are selected.  It appears that the "active" css class is being added to the button that was clicked but not removed from the other buttons in the group.  When the tooltip directive is removed, the "active" css class is added and removed as expected.  My tooltip directive options are:

tooltip="string"
tooltip-add-to-body="true"
tooltip-placement="right"
tooltip-trigger="mouseenter"

image

image

icfantv:
- modified btnRadio directive to correctly handle angular expression values
- added unit test to verify code change
- added additional unit test to test for string values
- added JetBrains project config directory to git ignored

Closes #3220


icfantv:
closing to fix quote marks

PR #4078, opened 2015-08-02
fix(typeahead): return `null` if empty

### LINKED ISSUE #3176: Setting an empty string instead of null or empty object when editable is false
When the user clears the input field, the parser returns an empty string instead of null. This may cause problems when the controller expects the ng-model to be an object.

I fixed this changing the line 3719, inside the parser function:
 ```javascript
if (!inputValue) {
            // Reset in case user had typed something previously.
            // An empty string is not always the best value to return
            modelCtrl.$setValidity('editable', true);
            return null;
```

wesleycho:
- Return `null` instead to unify falsy values

Plunker in action [here](http://plnkr.co/edit/TlzYtI?p=preview).

Fixes #3176.


Foxandxss:
LGTM

Issue #2773, opened 2014-10-02
 datepicker $setValidity $error does not follow convention naming


emouawad:
In datepicker.js
ngModelCtrl.$setValidity('date-disabled', !date || (this.element && !this.isDisabled(date)));

$setValidity(validationErrorKey, isValid)
validationErrorKey: does not follow convention 

The validationErrorKey should be in camelCase and will get converted into dash-case for class name. Example: myError will result in ng-valid-my-error and ng-invalid-my-error class and can be bound to as {{someForm.someControl.$error.myError}} .

rvanbaalen:
You're right that should be edited. 

https://docs.angularjs.org/api/ng/type/ngModel.NgModelController#$setValidity

PR #4080, opened 2015-08-02
fix(datepicker): change to `dateDisabled`

wesleycho:
- Change validation key to `dateDisabled` to align better with Angular's convention

This addresses #2773.


Foxandxss:
LGTM after you change the title and you put the breaking change in the message.

PR #4081, opened 2015-08-02
fix(progressbar): fix `min-width` for Bootstrap 3.2

### LINKED ISSUE #2511: stacked progress bars fail with bootstrap 3.2.0
Bootstrap 3.2.0 has introduced a new CSS rule, which sets the min-width of progress bars with an aria-valuenow attribute of "0", "1" or "2" to 30px. 
This causes stacked progress bars generated by angular-ui-bootstrap to fail when one of the not-last bar has one of these values (because the last progress bar goes down).

Here's a plunkr demonstrating it: http://plnkr.co/edit/kqrKNDvH8icXUNWycL9I?p=preview

I'm not sure what the best solution is. Maybe this should be mentioned in the documentation. Or the aria-valuenow attribute could be omitted for stacked bars, or it could be omitted using an additional attribute of the directive.

wesleycho:
- Fix styling of progressbar for Bootstrap 3.2

Plunker with fix in action [here](http://plnkr.co/edit/ymHigGyZWh8wUuJBFap8?p=preview).

This fixes #2511.


Foxandxss:
LGTM

PR #4084, opened 2015-08-03
feat(accordion): add `templateUrl` support

wesleycho:
- Add ability to override the template of the `accordion` and `accordion-group` directives on an instance by instance basis


Foxandxss:
I am not any sure we want this kind of upgrades. The whole point of having the `tpls` is to include the templates. You can always use the other version if you don't want the templates.

Still, you can override the templates in the `tpls` version with $templateCache easily.

wesleycho:
This would let you override it on a case-by-case basis - say you want one template for one instance and another for a different one

Foxandxss:
Uhmm, uhmmm. I can see that.

I would put in the doc that it only accepts a string, no interpolation.

wesleycho:
Alright, I'll make the change - my test has a bit of a hack in it btw. Is it fine or do you have a better recommendation there?

PR #4092, opened 2015-08-03
fix(carousel): fix animation direction

### LINKED ISSUE #4087: carousel contents collide when going left for the first time
When clicking the right arrow on the carousel, the current content exists to the left and the next content enters from the right, just as expected.

When clicking the left arrow on a carousel, the next content correctly enters from the left. However, the current content also exists to the left, instead of to the right. For all subsequent left-arrow clicks, it works as expected.

The same wrong behavior is also observed when clicking the right arrow *after* having clicked the left arrow at least once.

wesleycho:
- Properly fix the animation direction of the slides

Plunker in action [here](http://plnkr.co/edit/YRcO7j5c1H5jzqNPm29M?p=preview).

This fixes #4087.


Foxandxss:
Now it is a :+1:

PR #4096, opened 2015-08-04
chore(pagination): preventDefault on click

wesleycho:
- Add `event.preventDefault` call in order to avoid potential router bugs


Foxandxss:
:+1:

PR #4097, opened 2015-08-04
fix(modal): dismiss modal on unschedule destruction

### LINKED ISSUE #3694: $modal cant be dismissed if state [ui-router] is changed
Hi,

Using ui-router and $modal brings confustion when $modal inherits scope from current state controller's scope. If state is changed, modal's methods are gone, and its impossible to close it (I tried using $stateChangeStart and $destroy events, but in both - its already too late and dismiss and close have no effect).

One solution was not using inherited scopes but in case they are needed - modal stays open forever.

Br

wesleycho:
- Dismiss modal via `modalInstance.dismiss` if the `$scope` undergoes an unscheduled destruction

Plunker in action [here](http://plnkr.co/edit/pzISwjbVSOdCCZ3v7G67?p=preview)

This fixes #3694.


chrisirhc:
Smart solution. Perhaps this needs tests?

wesleycho:
I'll write up a test tomorrow - should be easy.

wesleycho:
What are your thoughts on what the dismissal should reject the promise with? I just chose an arbitrary string to have something that can be hooked to, but I'm not partial to any particular form of the api there.

Foxandxss:
Yup, need tests.

wesleycho:
Added test

Foxandxss:
I don't mind the string chosen. It just need one unique.

chrisirhc:
@wesleycho sorry, didn't see your reply until now, "$uibUnscheduledDestruction` could be clearer?

wesleycho:
Sure, I can change that right now.

Issue #3611, opened 2015-05-04
Issue #3167 still present if using `tooltip-popup-delay`


ntrifunovic:
Issue #3167, fixed in 87a3607632cae03ead57b377bb91ca27610e7730, is still present if using  `tooltip-popup-delay` with the tooltip.

**Explanation:** The tooltip is opened after the check is made to see if the tooltip is opened, when the button becomes disabled. This will lead to the same issue as described in #3167 .

rvanbaalen:
Updated plunkr: http://plnkr.co/edit/AUKz0kVzwMwlzZJDiHsr

rvanbaalen:
@ntrifunovic Can you confirm your issue is fixed in my plunkr?

http://plnkr.co/edit/AUKz0kVzwMwlzZJDiHsr?p=preview

I'll submit a PR for this if you confirm it's fixed.

ntrifunovic:
Yes.

              if (popupTimeout && val) {
                  $timeout.cancel(popupTimeout);
              }

That seams to do the trick :smiley:

rvanbaalen:
Allright ill squeeze it in soon. Ill close this automatically when the PR lands.

PR #4098, opened 2015-08-04
fix(tooltip): prevent opening when `tooltipPopupDelay` is present

wesleycho:
- Cancel the timeout when `tooltipPopupDelay` is present and the element with the tooltip directive becomes disabled

This implements @rvanbaalen's fix for #3611.


Foxandxss:
:+1:

Issue #3555, opened 2015-04-16
Popover-tempate not working with jqLite


hrajchert:
If I have simple popover it works great, but when I change to popover-template I get 

Error: [jqLite:nosel] Looking up elements via selectors is not supported by jqLite!

rvanbaalen:
Plunker please. Popover-template is not available as a feature yet. Will be released withversion 0.13.0

RobJacobs:
@hrajchert  Here is a [plunk](http://plnkr.co/edit/nbbfpG2TjB3uvSORsSAJ?p=preview) with the latest build from the master branch with a popover-template demo.  I don't see any jqLite errors.  If you can reproduce the problem using that plunk I'll reopen the issue, thanks.

hrajchert:
I used this [plunkr](http://plnkr.co/edit/mYiDJpdCqKGXvOzcOOFR?p=preview), In which I grabbed the latest commit and made a custom build using this command
```
grunt dist:../bootstrap-bower/ version build:buttons:datepicker:dropdown:modal:tooltip:popover:tabs
```

If I add jquery as a library I don't get the error but it still doesn't work. I will now see @RobJacobs plunkr to see whats different

@rvanbaalen I do know that this is supposed to be in 0.13.0 but I wanted to test it out.

chrisirhc:
This is because your template only contains text. It doesn't contain at least an HTML tag. Looking at the jqLite implementation, there's no easy way to fix this.
As such, we may need to improve the documentation to specify that if a template is used, it needs to be wrapped in a tag.

PR #4101, opened 2015-08-04
chore(tooltip): add note on wrapping in a tag

wesleycho:
- Add documentation to tooltip and popover about usage of `*-template` and the template needing to be wrapped in a tag

I just want a check on whether the wording is good here.

This addresses #3555.


Foxandxss:
:+1:

PR #4107, opened 2015-08-04
fix(dropdown): handle `keyboard-nav` correctly

### LINKED ISSUE #4091: Dropdown Key Navigation doesn't work when "dropdownAppendToBody" is enabled
When pressing up/down arrow buttons it causes an exception in the "keyboardNav" directive in this line:
```javascript
var elems = angular.element(element).find('a');
```

I think this line fixes this issue (accessing the dropdownMenu jQuery element from a reference):
```javascript
var elems = dropdownCtrl.dropdownMenu.find('a');
```

icfantv:
- properly handle arrow key navigation in dropdown button menus

fixes #4091


icfantv:
closing as there's another bug in the dropdown directive that needs to be fixed.  this was missed by the functional tests because of this second bug.

Issue #3748, opened 2015-06-03
[date-picker] Is it possible to get hook for current selected date when using the keyboard navigation


sushantdhiman:
Date-picker has good aria support, it allow navigation through keyboard. But date is changed only when user hits enter / space.

Are there any possible hooks which give the `current selected date` when navigating via keyboard.

rvanbaalen:
I think you mean current focussed date, based on the fact that you're talking about keyboard navigation?

sushantdhiman:
I have noticed that there is already something like that [activeDate]( https://github.com/angular-ui/bootstrap/blob/master/src/datepicker/datepicker.js#L59) available. But its not exposed to public API.

May be if we can expose `activeDate`. Then code should be able to `watch` it safely without any performance issues.

sushantdhiman:
I have created an addon which can work independently from Datepicker API.

[Here](https://gist.github.com/sushantdhiman/0b1957bfb096c1addccf) is the GIST if anyone need this feature without waiting for "Purgatory" updates.

It works by watching the [activeDate](https://github.com/angular-ui/bootstrap/blob/master/src/datepicker/datepicker.js#L59).

wesleycho:
This sounds like a reasonable feature, although this should be opt-in if it were implemented.

icfantv:
Closing in favor of #4772.

PR #4109, opened 2015-08-04
feature(datepicker): fix the issue #3748

sushantdhiman:
This update will add new directive which expose the `activeDate` for date picker. Thus able to watch current focused date.

Also fixes the issue [#3748](https://github.com/angular-ui/bootstrap/issues/3748)


wesleycho:
The implementation looks incorrect here - the datepicker & datepicker popup directives should support an optional two way binding, where if present, it would update the data being bound to.

sushantdhiman:
This directive only suppose to read the current focused date so user can take some action based on that date. Like fetching the data from api etc.

wesleycho:
As built, there are problems with this implementation - this sort of wrapper is better provided as an external library, and not part of UI Bootstrap.

In addition, it contains hacks such as $timeout usage instead of instantiating it on proper timing, whereas a built in optional binding to the active date would allow the user to more easily consume it without mangling the HTML.

From an api standpoint, this is a poor implementation, and contains inefficient & fragile initialization logic.

sushantdhiman:
Yeah I understand your point. I will update the implementation to expose the `activeDate` in `datepicker` rather than wrapping a component around it.

PR #4110, opened 2015-08-04
fix(dropdown): handle `keyboard-nav` correctly

### LINKED ISSUE #4091: Dropdown Key Navigation doesn't work when "dropdownAppendToBody" is enabled

icfantv:
- fix bug in directive whereby list item selector is not being initiated on
  correct DOM element.
- fix bug in directive whereby invalid `keyCode` property of jQuery Event
  object is being referenced instead of `which`.
- update functional test to trigger keydown event on correct DOM element
  because otherwise the keydown handler in the directive is not getting
  executed.
- remove invalid functional test because it is not possible to trigger a
  keydown event on a hidden DOM element through normal UI interaction.

fixes #4091


wesleycho:
LGTM

Issue #3836, opened 2015-06-18
Datepicker - disable validation option


Blackbaud-PatrickOFriel:
I have a situation where in some cases where I am using the datepicker popup and I don't want the manually entered input to go through the datepicker's validation (I want to do custom validation instead). Are there any plans to have either an option to disable validation on the datepicker or an option to provide custom validation for the datepicker (or maybe there's already a way that I'm too thick to discover today)? Thanks for all your work!

adouglas:
+1

wesleycho:
I am currently in the process of working on a fix to the validation of the datepicker component [here](https://github.com/wesleycho/bootstrap/tree/fix/datepicker-validators) - once this is implemented, one can write a custom directive with a `$timeout` hack to overwrite `ngModelCtrl.$validators`, allowing for custom validators.

Blackbaud-PatrickOFriel:
I found a way to hack around the current validation (albeit in a very strange way), my main pain point at the moment is this guy popping up angry and red in the console when I'm doing my own validation: `$log.error('Datepicker directive: "ng-model" value must be a Date object, a number of milliseconds since 01.01.1970 or a string representing an RFC2822 or ISO 8601 date.');`. Any way this can be a warning instead of an error?

Blackbaud-PatrickOFriel:
Any more info on whether we can have that $log.error be a warning, or better yet, not exist? I've managed to hack my own validation, but that console error keeps coming up.

wesleycho:
One can disable `$log` via

```javascript
$provide.factory('$log', function() {
  return {
    debug: angular.noop,
    error: angular.noop,
    log: angular.noop,
    warn: angular.noop
  };
});

Blackbaud-PatrickOFriel:
I thought about doing that, but I really only wanted to disable the validation for this case, since I know that I'm overriding the typical validation here. I dont want to suppress all log messages in my application. I'll try to think of another way to work around it.

wesleycho:
One way is to file a PR that creates an injectable `$datepickerSuppressError` and `$dateparserSuppressWarning` and the $log usage in those components can then check for the truth condition. similar to how it is done in `$transition`.

Blackbaud-PatrickOFriel:
Okay I'll take a look at that, thanks!

distinctdan:
[...]

PR #4115, opened 2015-08-05
feat(datepicker): add suppression of validation for #3836

Blackbaud-PatrickOFriel:
This adds the `$datepickerSuppressError` value provider which has an `isSuppressed` property to allow the option to suppress the `$log` error in `DatepickerController`. This was made to address #3836


src/datepicker/datepicker.js
```diff
@@ -1,5 +1,9 @@
 angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootstrap.position'])
 
+.value('$datepickerSuppressError', {
```
wesleycho: Just make it a boolean - no need for an object here.

wesleycho:
Just make all changes related to my comment and this PR should be good to get in.

Blackbaud-PatrickOFriel:
The only reason I made it an object is because when it's a literal, changes you make upon injection don't take effect (setting `$datepickerSuppressError` to `true`). You could only change the value in the config phase using `$provide`. If that's the route I should take, then I'll have to make the test a bit different since you can't use `$provide` after you've used `inject`.

wesleycho:
I think that is fine - there really isn't a great use case for toggling suppression on or off once the app is running.

Blackbaud-PatrickOFriel:
Okay I'll go ahead and make the change then

Blackbaud-PatrickOFriel:
So in order to test properly I had to wrap the other tests in their own describe block so that I could override the `$provide` for my test cases before the `inject` function was ever called. And now the diff tool makes it look like I deleted everything and added it back...

wesleycho:
That's perfectly fine

wesleycho:
Can you squash your commits into one?

Blackbaud-PatrickOFriel:
Squash complete

wesleycho:
LGTM - merging shortly. Thanks for the work!

Blackbaud-PatrickOFriel:
No problem! Thanks for this entire library, it's great!

PR #4117, opened 2015-08-05
[kaidjohnson] Fix up dateparser support for 12 hour datetime format.

kaidjohnson:
Using a date format that includes a 12-hour time format (eg. dd MMMM yyyy hh:mma) can result in a permanent invalid parsing of the value.

I've created a plunk to demonstrate the issue (try manually deleting the 'M' from 'PM' then putting the 'M' back. It becomes invalid without the M but never is valid again when the M is restored.):
http://plnkr.co/edit/CvJgnoCE08XGlDbVQka8?p=preview

Here's the same plunkr, but with the fix found in this PR:
http://plnkr.co/edit/uTKEjRFcY3CsWDHyZ0bZ?p=preview


icfantv:
@kaidjohnson please add the relevant functional tests for this and update/rebase your PR.  thanks.

src/dateparser/dateparser.js
```diff
@@ -53,12 +53,16 @@ angular.module('ui.bootstrap.dateparser', [])
...
+      regex: '0[0-9]{1}|1[0-2]{1}',
+      apply: function(value) { this.hours = +value; }
+    },
     'H': {
       regex: '1?[0-9]|2[0-3]',
       apply: function(value) { this.hours = +value; }
     },
     'mm': {
-      regex: '[0-5][0-9]',
+      regex: '[0-5]{1}[0-9]{1}',
```
wesleycho: Revert this change - this is no different than the previous version, as tested [here](http://regexpal.com/).

src/dateparser/dateparser.js
```diff
@@ -53,12 +53,16 @@ angular.module('ui.bootstrap.dateparser', [])
       regex: '(?:0|1)[0-9]|2[0-3]',
       apply: function(value) { this.hours = +value; }
     },
+    'hh': {
+      regex: '0[0-9]{1}|1[0-2]{1}',
```
wesleycho: Remove `{1}` in both instances in this line, they are unnecessary.

kaidjohnson:
Thanks @icfantv and @wesleycho -- I've added tests, removed redundant {1}, and corrected AM/PM parsing issue discovered during tests. Also updated plunkr

wesleycho:
LGTM - good work!

Issue #4119, opened 2015-08-05
Carousel broken with Angular 1.4.3 + UI-Bootstrap 0.13.2 + ng-animate + modal dialog


alpeshgaglani:
I'm aware of the issue #3811 , and have the latest fixes from master. The carousel works just fine outside of a dialog box. However, when I use ngAnimate inside a $modal dialog, I see just the first slide. The navigation anchors are hidden, and I cant move to the second slide. Any ideas?

wesleycho:
Looks like you are correct to a degree, I am able to reproduce [here](http://plnkr.co/edit/b3ddvoTPVFjBx5H92pFx?p=preview) incorrect behavior of a different nature.

wesleycho:
Ah, now I see the behavior you describe (Plunker was behaving weird for me there).

wesleycho:
Investigating this, this appears to be a problem with the modal service, and it not being converted to handle $animateCss support. There seems to be some interplay problems between the two due to the new restrictions in ngAnimate.

wesleycho:
This should be fixed in `master` now - thanks a lot for this find, I almost cut a new release today before encountering this issue (& the tooltip/popover one).

I highly recommend checking out the comments in #4121, it is highly relevant if you're living on the edge :) .

PR #4121, opened 2015-08-06
fix(modal): add `$animateCss` support

wesleycho:
- Add support for `$animateCss`

The carousel change is to safely detect whether $animateCss is truly present (if the user is not using `ngAnimate`, then this would throw an exception before this patch.

The modal change is to correctly use `$animateCss` - this can be observed in action [here](http://plnkr.co/edit/75rLxCoDUlh1r6ecehVX?p=preview).

I am currently still investigating #4119, and hopefully will be able to patch that up on top of this or as a part of this PR. These changes are not enough to fix the broken behavior, as can be seen [here](http://plnkr.co/edit/Oi7YsUHB1QTIXSTb2GqC?p=preview).


wesleycho:
I updated the Plunker, with usage from the snapshot version of Angular, due to a major ngAnimate fix that is currently not released. This is still currently broken in 1.4.3, but only with the indicators and controls being missing.

icfantv:
LGTM

mattlewis92:
FYI you can use `$injector.has('$animateCss')` instead of using the try catch to detect if $animateCss is available

wesleycho:
I should have checked the docs...I never knew that existed - thanks for that tidbit, I will refactor that now.

PR #4124, opened 2015-08-06
feat(modal): allow users to resolve with strings

### LINKED ISSUE #2676: 'string' type resolves in $modals are not supported
As per https://github.com/angular-ui/bootstrap/blob/master/src/modal/modal.js#L322 it seems that 'string' type resolves are not handled properly in $modal.

wesleycho:
- Allows users to specify a string reference to a service for injection

Closes #2676


Foxandxss:
:+1:

PR #4125, opened 2015-08-06
feat(docs): explain the difference between `-tpls` and not

### LINKED ISSUE #420: docs(): ambiguity in standard vs. tpls
Correct me if I'm wrong.

ui-bootstrap-0.3.0.js contains all the main code, but requires external template files to be available, which will be pulled down to the client at runtime

ui-bootstrap-tpls-0.3.0.js is a copy of the main code, but also inlines all the templates as string literals so that they are copied down with the code rather than separately.

To be honest, from reading the docs, I initially got this backwards, until I compared the two.

wesleycho:
- Add documentation on the difference between `-tpls` and the regular
  built scripts

Closes #420


pkozlowski-opensource:
Maybe we should link to https://github.com/angular-ui/bootstrap/wiki/FAQ#full-explanation or https://github.com/angular-ui/bootstrap/tree/gh-pages#build-files as well?

wesleycho:
That is a good idea - I added in those links to the commit.

PR #4127, opened 2015-08-06
feat: added build task for separated components

m99coder:
Maybe this small enhancement is useful for you in case somebody (as me) needs separated compiled components.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ pids
 logs
 results
 dist
+dist-separate
 # test coverage files
 coverage/
 
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -292,6 +292,50 @@ module.exports = function(grunt) {
     if (dir) { grunt.config('dist', dir); }
   });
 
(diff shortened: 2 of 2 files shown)


wesleycho:
I'm confused - how is this significantly different from the current task that compiles to `dist`?

m99coder:
Current `grunt dist` compiles together all components in one file and creates a second one for the templates beside. We need artifacts which are separated into their components to only include `modal` for instance. Some of your directives create a naming conflict with our own library. It’s a little bit more flexible if you ask me.

wesleycho:
Maybe a subfolder inside `dist` such as `dist/components` might be more desirable then?

m99coder:
That would be nice. Feel free to decide where it should reside best. Finally we created a corresponding bower-compatible repository too from where these artifacts can be included in our applications. So maybe you could deploy the whole `dist` folder contents to `https://github.com/angular-ui/bootstrap-bower`?

wesleycho:
Closing as we decided instead to move to more robust npm support so that one can load/require the scripts one needs and use Browserify/Webpack plugins to load the templates into JS.

PR #4128, opened 2015-08-06
Feature/add typeahead popup height

miyukiw:
Add optional `typeahead-popup-height` attribute to set maxHeight and overflow:hidden on typeahead popup.
And in that case, adjust scroll position when arrow keys are used to make highlighted element always visible.

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -71,6 +71,9 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position'])
         //If input matches an item of the list exactly, select it automatically
         var selectOnExact = attrs.typeaheadSelectOnExact ? originalScope.$eval(attrs.typeaheadSelectOnExact) : false;
 
+        // if PopupHeight is specified, popup will be set maxHeight and overflow:hidden
+        var popupMaxHeight = parseInt(attrs.typeaheadPopupHeight, 10) || 0;
+
         //INTERNAL VARIABLES
 
         //model setter executed upon match selection
@@ -110,7 +113,8 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position'])
           select: 'select(activeIdx)',
           'move-in-progress': 'moveInProgress',
           query: 'query',
-          position: 'position'
+          position: 'position',
+          'max-height': popupMaxHeight
         });
         //custom item template
         if (angular.isDefined(attrs.typeaheadTemplateUrl)) {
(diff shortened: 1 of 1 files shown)


wesleycho:
This fix would be better to be added on the dropdown component itself, and then plugged into the typeahead.

miyukiw:
Thank you for your reply.
I checked dropdown component behavior with `keyboard-nav` and maxHeight. Dropdown component itself seems OK, it automatically scrolls when I use arrow keys.
sample: https://jsfiddle.net/miyukiw/wx1a903p/1/

So I think only typeahead component needs this fix. Or typeahead should change the template structure to use `keyboard-nav` instead of this fix.
How do you think? @wesleycho

wesleycho:
Is there no way to plug into the `keyboard-nav` feature in the regular dropdown in the typeahead then?

miyukiw:
You mean this way?
>  Or typeahead should change the template structure to use keyboard-nav instead of this fix.

Since I didn't notice `keyboard-nav` feature when I made this PR, I haven't tried that way yet, though.

wesleycho:
Whenever you get the chance, please rebase the PR on top of `master` (a lot of code style changes has went on in `master` to make it more consistent) and squash the commits.

miyukiw:
rebased & squashed.

wesleycho:
Have you investigated an approach using the existing `keyboard-nav` directive for the typeahead?

miyukiw:
To see `dropdown.js` and  `typeahead.js` , these seem to handle keyboard event respectively. Then introducing ` keyboard-nav` approach into typeahed is gonna be big changes. For now, I wanna simply fix the problem by this PR rather than change whole structure of typeahead, since I'm not sure how should it be done...

wesleycho:
I'm hesitant to merge in this approach due to the CSS modifications via JS, especially since it involves reading DOM positional values, which has a large performance hit - a bigger change is likely preferred here to use what is already built.

miyukiw:
I see...
Then you want to introduce `<a>` element and focus() approach into typeahead just as `keyboad-nav`? `keyboad-nav` directive is supposed to use with `dropdown` directive, meanwhile typeahead only uses `dropdown-menu`. no?

BTW my approach referred here: 
https://github.com/angular-ui/ui-select/blob/master/src/uiSelectController.js#L492
(JFYI

miyukiw:
Moving focus to matched item let the input field blur, then I think it wouldn't be nice for typeahead UI.

wesleycho:
The scrolling problem is now solved via a1355e76874a68e65d9b481fa42de335d7349b98 - otherwise, one can use CSS to alter the height to a particular height.

PR #4132, opened 2015-08-07
feat(modal): add ability to change class on body

### LINKED ISSUE #2633: Abilility to change OPENED_MODAL_CLASS
I'm trying to use Nifty Modals with this, so far so good except it would be nice to have the ability to change the opened modal class. It's currently hard coded and doesn't apply to the modal window.

wesleycho:
- Add support for `openedClass` to allow overriding the default modal class added & removed on the `body` element (`modal-open`)

Closes #2633


icfantv:
:thumbsup:

Issue #4130, opened 2015-08-06
Pagination 'direction-links' to support glyphicons


Frondor:
just by providing a string like `glyphicon-chevron-right` the directive should render a `<span class="glyphicon glyphicon-chevron-right"></span>` instead of a simple character or string.

wesleycho:
Any reason why you wouldn't just use `&gt;` or `&lt;` for that?

Frondor:
In a small pagination they are really small, and not good looking at all.
In bootstrap you can place whatever you want inside the list items

wesleycho:
The problem is, your suggestion is tied too heavily to one particular implementation. What if someone wants to use Font Awesome icons for example?

The best solution is to allow `templateUrl` overriding at the directive level IMO.

RobJacobs:
We could potentially add unique classes to the pagination first, last, next, previous elements that would allow people to use the css content property to set the icon, so the pagination button would look like so:

    <li ng-if="::boundaryLinks" ng-class="{disabled: noPrevious()||ngDisabled}"><a href class="pagination-first" ng-click="selectPage(1, $event)">{{::getText('first')}}</a></li>

and the class would look something like:

    .pagination-first: after {
         font-family: 'Glyphicons Halflings';
         content: '\fe079'; //glyphicon-chevron-left
    }

wesleycho:
That is also a good idea - I was thinking about that too, it would make it more flexible to style.

wesleycho:
I opened a PR with classes added to each of the `li` elements.

PR #4137, opened 2015-08-07
feat(pagination): add `templateUrl` support

wesleycho:
- Add support for overriding the template url on an instance by instance
  basis
- Expose the controller to the view via `controllerAs`

This addresses #4130.


icfantv:
:thumbsup:

PR #4138, opened 2015-08-07
feat(accordion): add `controllerAs` support

wesleycho:
- Expose the controller to the view via `controllerAs`


icfantv:
@wesleycho Is this something we want to document?  I think it would be helpful for someone who uses the custom template option to know they have access to the controller via `controllerAs`.

wesleycho:
I'm not sure - if someone is overriding the template, the person is highly likely to be looking at the existing template and directive.

icfantv:
That's fair.  LGTM.

PR #4139, opened 2015-08-07
feat(alert): add `templateUrl` support

wesleycho:
- Adds support for `templateUrl` to override template on an instance by
  instance basis
- Exposes controller to the view via `controllerAs`


icfantv:
:+1:

PR #4140, opened 2015-08-07
feat(buttons): add `controllerAs` support

wesleycho:
- Expose controller to the view for `btnCheckbox` and `btnRadio`
  directives

Not quite sure this would be useful to the user, but it doesn't hurt IMO.


icfantv:
@wesleycho Is this something we want to document? I think it would be helpful for someone who uses the custom template option to know they have access to the controller via `controllerAs`.

icfantv:
Per #4138 comments re documentation.  LGTM.

PR #4141, opened 2015-08-07
feat(carousel): add `templateUrl` support

wesleycho:
- Add support for `templateUrl` attribute for `carousel` and `slide`
  directives for instance by instance overriding of default templates


icfantv:
:+1:

Issue #4090, opened 2015-08-03
Popover expanding height


jameshyde:
Hi,
I updated to 0.13.2 earlier and it appears to have broken the script to calculate where the top of a popover should be based on the height.

For example, I have a popover open set to "popover-placement="top" and add some content in which will increase the height. Incorrectly, the bottom of the popover moves down, and the top stays static. Prior to 0.13.2, the top of the popover would have moved up as the bottom stayed static.

See images for an example - note the button is covered in the 0.13.2 example:

![0 13 1](https://cloud.githubusercontent.com/assets/7480500/9039857/2de491cc-39f6-11e5-9f70-bfc23d8c0cd6.png)
![0 13 2](https://cloud.githubusercontent.com/assets/7480500/9039906/7e01edee-39f6-11e5-9a75-74d52c85cdfe.png)

wesleycho:
Hm, I wonder if this is related to the changes made in the $tooltip service due to a bug in tooltip positioning found in tooltips.

wesleycho:
Can you create a reproduction in Plunker?

RobJacobs:
Here is a [plunk](http://plnkr.co/edit/e2sS7vFj5p5mVl8hcG6s?p=preview) reproducing the problem.  Open the popover, then start typing in the input until the content wraps.

wesleycho:
Hm, probably related, although that change fixes a positional bug with the tooltip that results from calling it too often.

wesleycho:
Do we need to just reposition on every digest? I'm not sure how we would be able to reliably position the tooltip otherwise.

RobJacobs:
@wesleycho I'll play with this for a bit and will have some suggestions tomorrow

RobJacobs:
Here's what I'm thinking, rather than watching content (type, title, etc..) create a watch for the tooltip element size to trigger a call to position, something like:

    tooltipLinkedScope.$watch(function() { 
      return tooltip[0].clientHeight + tooltip[0].clientWidth; }, 
      function() {
         positionTooltip(); 
    });

We could potentially remove the other calls to the positionTooltip and just have that one watch responsible for handling when to position the tooltip.  We may also want to consider implementing a timeout on the actual positionTooltip function like so:

    var positionTimeout;
    var positionTooltip = function () {
      if (!tooltip) { return; }
    
      if (positionTimeout) {
        $timeout.cancel(positionTimeout);
      }
      positionTimeout = $timeout(function() {
[...]

PR #4144, opened 2015-08-07
fix(tooltip): tooltip w/ template position

RobJacobs:
When using a template with the tooltip/popover
the position funciton needs to be called at the
end of the $digest cycle to ensure correct
positioning in case the template content size
has changed

Fixes: #4090


wesleycho:
LGTM - good optimizations here here too.

wesleycho:
Ah crap, the semicolon threw off the automatic close :( .

PR #4145, opened 2015-08-08
issue(accordion): add open` class and `aria-expanded` attribute

icfantv:
- add `open` class to `.panel-collapse` element when opened.
- add `aria-expanded` to `.panel-collapse` element
- fix missing closing brace in another test.


icfantv:
One should really not attempt to do a PR when tired.

Issue #2546, opened 2014-08-06
when datepicker opening,it's focused,please do not this,ok?


bondli:
var focusElement = function() {
    $timeout(function() {
      //self.element[0].focus();
    }, 0 , false);
  };

stephanie-dm:
+1

stephanie-dm:
Will this issue ever get fixed? The input field losing focus is a major problem for us, users always have to click twice if they want to type a date in the input field. It would be great if this was configurable...

wesleycho:
Nothing is stopping anyone from opening a PR from being opened fixing this if it is desired - there are other much higher priority issues in need of tackling though, which is likely why there hasn't been any movement on this.

stephanie-dm:
Thanks for the fix!

PR #4146, opened 2015-08-08
feat(datepicker): add `onOpenFocus` support

### LINKED ISSUE #2303: Request "open-on-focus" attribute on datepicker-popup.
After upgrading to 0.11 our datepicker popups no longer work as expected. I've asked around on IRC and came to the conclusion that there are very good reasons for this (w.r.t. accessibility). 

Nevertheless, the guys from business prefer the old behavior. As they are the end-user, there is little I can do to convince them that the current behavior is fine.

As a compromise I propose an "open-on-focus" attribute, which could be put on datepicker-popups to re-enable the old behavior. A note should be made in the documentation to explain the accessibility issues involved. 

Thank you.

wesleycho:
- Add `onOpenFocus` attribute for optional disabling of focus of popup
  after datepicker popup is opened

Closes #2303 & #2546.


Foxandxss:
Why are we relying on events that much? The PR is :+1: for me, but that is something to check better.

wesleycho:
It probably should call a method on the controller to do focus, but that can be fixed at any time.

PR #4152, opened 2015-08-08
fix(accordion): add `open` class when expanded

### LINKED ISSUE #3419: Accordion: add a class at the panel level to indicate open state
Would you add a class to indicate the panel is open at the panel level?

I'm using a caret indicator in the `panel-title` like so:
```sass
.panel-title > a:before { 
  content: $fa-var-angle-right;
  font: normal normal normal 14px/1 FontAwesome; }

.panel-open .panel-title > a:before {
  content: $fa-var-angle-down; }
```

Adding a class like `panel-open` on the panel div would allow me to just go like
```html
<accordion-group is-open="thisDrawer.open">
```
instead of
```html
<accordion-group is-open="thisDrawer.open" ng-class="{'panel-open': thisDrawer.open}">
```

icfantv:
- add `open` class to `.panel-collapse` when open.
- fix missing closing `<` on test.

fixes #3419


wesleycho:
LGTM, assuming `open` is changed to `panel-open` - I'll make that change when merging now.

PR #4153, opened 2015-08-08
chore(accordion): tighten up code style

wesleycho:
- Make the code style uniform for consistency


wesleycho:
Should note that I'd like multiple people to review this as to whether my changes in these code style tweak PRs are good. Code style has a lot of opinions, and I want to make sure that it is something people find agreeable.

Foxandxss:
Good luck.

icfantv:
:+1:   I caught the extra spaces between `function` and the `()` a little while back, but wasn't going to change it since my mantra is usually if I'm contributing to a codebase that's not mine, I just match the existing code style.  Even if it drives me nuts.  That said, I'm cool with the changes as I'm used to gslint.

We may want to look at using some sort of linter.  I know we do some sort of linting because I had a build fail due to using `"` instead of `'` in a string.  But I have not investigated what form of linting that is.

wesleycho:
I think we should consider adding JSCS as soon as we tighten up the code style of the whole codebase. The inconsistent code style is a little out of hand, and confuses people.

PR #4157, opened 2015-08-08
feat(datepicker): allow custom templates

wesleycho:
- Allow custom templates for datepicker and popup
- Expose controller to template


Foxandxss:
Good for me, but when we allow custom templates to be used on directives, we need to double check that pulling custom templates from remote (a different file in a plunker would work) doesn't break it. Angular won't wait for a directive (and its children) to be fully there in the DOM so animations (if any) could break.

src/datepicker/datepicker.js
```diff
@@ -225,21 +225,22 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
...
+      return attrs.templateUrl || 'template/datepicker/datepicker.html';
+    },
     scope: {
       datepickerMode: '=?',
       dateDisabled: '&',
       customClass: '&',
       shortcutPropagation: '&?'
     },
-    require: ['datepicker', '?^ngModel'],
+    require: ['datepicker', '^ngModel'],
```
wesleycho: Should note, this change would cause #1913 to be resolved.

wesleycho:
I think that is fine - they should be bundling it in $templateCache.

PR #4160, opened 2015-08-10
fix(timepicker): leave view alone if either input is invalid

### LINKED ISSUE #3825: Timepicker: Invalid value in the field is removed when focus lost
Timepicker shows behaviour that is inconsistent with datepicker (and other input fields as well) when handling invalid input.

Steps to reproduce:
* Enter 25 in first field
* Tab to second field
* Enter 6 => first field is changed to 2 => should stay 25 and marked as invalid

Expected behaviour: Invalid fields are marked but invalid input is still displayed in the field.
Same problem in the other direction as well.

wesleycho:
- Change to not auto-update inputs if either input is invalid

Fix can be seen in action [here](http://plnkr.co/edit/t4RWbtfTlPJQyzgZc68v?p=preview).

This fixes #3825.


Foxandxss:
:+1:

PR #4162, opened 2015-08-10
feat(pagination): add `templateUrl` support

wesleycho:
- Add support for overriding the `templateUrl` on an instance by instance basis
- Expose controller via `controllerAs`


icfantv:
:+1:

PR #4166, opened 2015-08-10
feat(tabs): add support for `x-tab-heading`

### LINKED ISSUE #1893: Support for x-tag-names
I haven't looked through the entire code base, but in the tabs module I found that the checks for `tab-heading` look for an attribute called `tab-heading` or `data-tab-heading` or a tag also called `tab-heading` or `data-tab-heading`. The correct prefixed version of the *tag* name should probably be `x-tab-heading` (which is what web components will use), not `data-tab-heading` (which seems to have no basis as a tag name in the current HTML spec).

AngularJS is less strict about the distinction and simply strips `x-` and `data-` prefixes from attribute names and tag names alike. AngularUI should probably adopt this behaviour and simply accept `x-` prefixed attribute names and tag names in addition to the `data-` prefixed ones it already handles.

Maybe this logic could be factored out into a utility function.

wesleycho:
- Add support for `x-tab-heading` attribute and element

This fixes #1893


Foxandxss:
:+1:

Issue #590, opened 2013-06-28
[popover/tooltip] Support Programmatic Triggers


joshdmiller:
The upstream JavaScript version supports `trigger="manual"` to allow for programmatic triggering.

christopherthielen:
@asafdav @marian2js Thanks for your code.  I had a problem where the toggle functionality was becoming confused whether or not the tooltip should be shown, when the tooltips had a popup delay.  

This was due to `toggleTooltipBind` being used.  `ttScope.isOpen` is not set to true until `show()` function executes, which happens after the delay timeout.  Thus:

- the expression becomes truthy
- it invokes `toggleTooltipBind`
- delays (for 1 second, let's say)
- the expression becomes falsey before `show()` is called
- `toggleTooltipBind` is invoked again
- ttScope.isOpen is still false, so it invokes `showTooltipBind()` again

I fixed this by registering a custom event *pair* and invoking the proper one based on truthy/falsey.

See this gist: 
https://gist.github.com/christopherthielen/4980c9117f3185906627

sdasun:
@asafdav  @marian2js  @christopherthielen Why these examples only work with old angular-js none of these examples working with angular-js 1.3.9

christopherthielen:
@sdasun FYI I'm using (a modified version of) my gist on angular 1.3.4 and ui-bootstrap 0.12.0

christopherthielen:
@sdasun yeah it works fine for me:
http://plnkr.co/edit/yaENfy8Tbmxz6hcRNsY1?p=preview

sdasun:
@christopherthielen  Thanks for Plnk r:+1: , I think something in my app interfering the event trigger.

sdasun:
@christopherthielen worked after updating my ui-bootstrap from 0.11.0v to 0.12.0 :+1:

ingshtrom:
1. I tried @luthur's fix by loading 0.12.1 of ui.bootstrap, but couldn't get it to work. I followed the examples given in this thread but nothing worked.  I am using angular 1.2.28 and it looked like it was compatible with that version. What exactly do I need to do to get it working?
2. Alternatively, I tried the plunkr given by @christopherthielen, but had to modify it to get it to work. The `closeTrigger` wasn't working for me.  Every time I sent the `openTrigger` event, it toggled the tooltip, so I modified his code to work for me. Any ideas why I would have such a different experience?

``` javascript
'use strict';

angular.module('ngen.shared')
[...]

PR #4179, opened 2015-08-10
feat(tooltip): expose isOpen property

RobJacobs:
Add support for toggling the isOpen
property of the tooltip.

Closes: #2148
Fixes: #590


wesleycho:
LGTM - only question I have is whether it is worth having tests in the popover module for the same functionality except with the appropriate prefix.

wesleycho:
Otherwise, this just needs documentation update for popovers.

PR #4198, opened 2015-08-13
fix(accordion): add attr.openClass || 'panel-open' class when expanded

### LINKED ISSUE #4193: Cant use ng-class on accordion group since 1.33.3
You can no longer use `ng-class` on an accordion group since this commit.
https://github.com/angular-ui/bootstrap/commit/ead15e37
---
https://github.com/angular-ui/bootstrap/pull/4152
https://github.com/angular-ui/bootstrap/issues/3419

davious:
uses toggleClass in isOpen watch
Fixes #4193

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/bootstrap/4198)


src/accordion/accordion.js
```diff
@@ -79,7 +79,9 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
     link: function(scope, element, attrs, accordionCtrl) {
       accordionCtrl.addGroup(scope);
 
+      scope.openClass = attrs.openClass || 'panel-open';
       scope.$watch('isOpen', function(value) {
+        element.toggleClass(scope.openClass, value);
```
wesleycho: This is incorrect - if the element has a different class for some reason, this will cause a bug.

You will want
```javascript
element[value ? 'addClass' : 'removeClass'](scope.openClass, value)
```
in this case to be safe.

davious: Hi @wesleycho,
Thanks for reviewing the PR. I'd like to understand your concern better. Please provide an example so I can get on the same page.

wesleycho: Any DOM manipulation can be done to change the class of the element - in that case, relying on toggle to correctly toggle the class would be completely incorrect, since it makes a poor assumption that the actual class reflects the state of `isOpen`. The coupling of the model value and the DOM value is not strong enough to make this assumption in a library.

davious: @wesleycho 
how about
`element.toggleClass(scope.openClass, !!value);`
is that OK?

wesleycho: Still has the same flaw of using toggleClass

davious: @wesleycho 
would you provide an example? If you are concerned with toggleClass mashing some other classes, I'm thinking toggleClass plays safe with them just like addClass and removeClass.

wesleycho: I apologize, I'm on not enough sleep it appears - this is completely fine due to the second parameter. Ignore my complaints here.

wesleycho:
This PR LGTM - will merge in sometime today.

src/accordion/accordion.js
davious: @wesleycho :grin: np. Thanks for staying up and accepting the PR

PR #4199, opened 2015-08-13
feat(datepicker): disable today button if invalid

### LINKED ISSUE #3988: Datepicker today option is enabled if today's date is outside of max-date/min-date range
The button to select Today in the datepicker is enabled and works even if today's date is outside of the valid max-date/min-date range.

I would imagine that if today's date falls outside the valid range of dates, then the today button should be hidden or disabled.

screen shot 2015-07-22 at 11 34 43

wesleycho:
- Disable today button if current date is before `min-date` or after
  `max-date`

This can be seen in action [here](http://plnkr.co/edit/ZKSIbj8mrDUOqJSg8MIH?p=preview).

This resolves #3988.


Foxandxss:
I see the code, but I don't see it working on the plunker. I toggle the minDate but "today" can be clicked.

In this type of plunker (where you have lot of buttons and widgets, could be good if you add some instructions at the top of the plunker to reproduce it. I am probably clicking the wrong buttons I guess.

wesleycho:
minDate is already toggled by default - toggling the button disables the minDate.

Foxandxss:
But if I click another date and I click today, it moves. Shouldn't it be disabled?

wesleycho:
I'm not seeing that behavior - the today button should only be enabled (in the popup mind you) if minDate is enabled in my example. I updated it to strip out any distractors.

Foxandxss:
Ah, I was checking the other today button with the inline calendar.

:+1: for me then.

Issue #3686, opened 2015-05-18
modal does not show up when on fullscreen mode on an element


nozer:
If I go full screen on a div element and then if I click on a button on this div to show a modal, that modal will not show up. 

Here is a plnkr demo (it is a modified version of what you have on the bootstrap modal template page). 
1 - Please click on "launch the preview in a separate window" icon of plnkr and then 
2 - Click on "Go Full Screen" button. 
3 - Then, when in full screen mode, click on any of the buttons to try to show a modal. 

http://plnkr.co/edit/oKZHZZebyNMwpG114Jyy?p=preview

I think that if I could specify in modal options what the parent element of the modal window is, then this could solve the issue.

rvanbaalen:
This has nothing to do with modal, it's an issue in your 'screenfull' implementation. Please ask questions on StackOverflow and report bugs on GitHub.

https://github.com/angular-ui/bootstrap#need-help

nozer:
Thank you. I will do as you suggested.

nozer:
Hi
Instead of using the `screenfull` library, I used `BigScreen` library to test this issue. I still have the same problem. This is not related to screenful implementation or BigScreen implementation. 

When browsers go on fullscreen on an **element**, for ui bootstrap's modal to appear in front, i think that, it must be a child of the element that is on fullscreen (I verified this by editing the html source of ui  modal from developer tools and pasting it inside the element that has gone full screen and the problem was fixed). 

You may test this again via this plnkr

To test it: 
1 - Click on this link  http://plnkr.co/edit/oKZHZZebyNMwpG114Jyy?p=preview
2 - Click on the  "launch the preview in separate window" icon (image shown below)
3 - Click on Go Full Screen Button (you will then be in full screen on the element with id `fullable`)
4 - Then click on any of the buttons to try to show a modal. 

Here is how to "launch  the preview in separate window" on plnkr: 

![screen shot 2015-05-19 at 09 54 23](https://cloud.githubusercontent.com/assets/6604091/7723127/f189860c-feee-11e4-9870-b4ea5c51c356.png)

nozer:
I solved this issue by editing the modal source. 

It involved modal window to be appended to the element of my choice (in this case, element that goes full screen).

Options object we pass to the `$modal.open()` function now accepts an `appendTo` property which is a css selector that will be used by `document.querySelector`.

In modal code, I changed from these:
```
var body = $document.find('body').eq(0),
[...]

PR #4200, opened 2015-08-13
added appendTo modal option

nozer:
appendTo modal option allows us to specify an element to append the modal to via a css selector that is used by document.querySelector. Related issue here -> https://github.com/angular-ui/bootstrap/issues/3686#issuecomment-130472014

### DIFF
--- a/src/modal/docs/readme.md
+++ b/src/modal/docs/readme.md
@@ -18,6 +18,7 @@ The `$modal` service has only one method: `open(options)` where available option
 * `windowTemplateUrl` - a path to a template overriding modal's window template
 * `size` - optional suffix of modal window class. The value used is appended to the `modal-` class, i.e. a value of `sm` gives `modal-sm`
 * `openedClass` - class added to the `body` element when the modal is opened. Defaults to `modal-open`
+* `appendTo` - a css selector that is supported by document.querySelector to specify the element to which the modal should be appended. Defaults to `body`
 
 Global defaults may be set for `$modal` via `$modalProvider.options`.
 
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -521,8 +521,7 @@ angular.module('ui.bootstrap.modal', [])
       options: {
         animation: true,
         backdrop: true, //can also be false or 'static'
-        keyboard: true,
-        appendTo: 'body'
+        keyboard: true
       },
       $get: ['$injector', '$rootScope', '$q', '$templateRequest', '$controller', '$modalStack',
         function ($injector, $rootScope, $q, $templateRequest, $controller, $modalStack) {
@@ -623,7 +622,7 @@ angular.module('ui.bootstrap.modal', [])
                 windowTemplateUrl: modalOptions.windowTemplateUrl,
                 size: modalOptions.size,
                 openedClass: modalOptions.openedClass,
-                appendTo: modalOptions.appendTo
+                appendTo: modalOptions.appendTo || 'body'
               });
 
             }, function resolveError(reason) {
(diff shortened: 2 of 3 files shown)


nozer:
Test is failing because in the tests because `body > ...` selectors fail. Any help to update tests would be appreciated

wesleycho:
There should not be a merge commit, and this needs tests.

src/modal/docs/readme.md
```diff
@@ -18,6 +18,7 @@ The `$modal` service has only one method: `open(options)` where available option
 * `windowTemplateUrl` - a path to a template overriding modal's window template
 * `size` - optional suffix of modal window class. The value used is appended to the `modal-` class, i.e. a value of `sm` gives `modal-sm`
 * `openedClass` - class added to the `body` element when the modal is opened. Defaults to `modal-open`
+* `appendTo` - a css selector that is supported by document.querySelector to specify the element to which the modal should be appended. Defaults to `body`
```
wesleycho: This needs to accept the DOM element itself that the user wants to append to, not a selector. The service itself should not do any selecting.

src/modal/modal.js
```diff
@@ -265,22 +265,22 @@ angular.module('ui.bootstrap.modal', [])
 
       function removeModalWindow(modalInstance, elementToReceiveFocus) {
 
-        var body = $document.find('body').eq(0);
         var modalWindow = openedWindows.get(modalInstance).value;
+        var modalContainer = angular.element($document[0].querySelector(modalWindow.appendTo)).eq(0);
```
wesleycho: This is no good, for reasons mentioned in prior comment.

src/modal/docs/readme.md
nozer: Will you please elaborate on why passing the css selector would be hazardous?

wesleycho: Testability, and poor API signature. Angular does not work off of selectors in general, especially for arbitrary element targeting - accepting the DOM element directly in the form of a jqLite wrapped element is more in line with how one works with Angular, as it can then be consumed by a custom directive with minimal code. This would also be a cause of issues being opened due to unexpected behavior.

Testing while dealing with `document.querySelector` also would be a major pita. It is much easier to work with the desired DOM node directly, which also does not create an extra coupling of the service to `$document`.

nozer: Thanks for the explanation. I understand and agree with your concerns. Regarding the  `$document` service, I saw it used liberally in the code and so I felt free to use it as well. 

I will update the code to use a jqLite wrapped element for appendTo as you suggested.

infnada:
What if you are launching the modal inside a directive and pass the angular `element` from the directive `link` function? or `$element` from the directive controller? Thats what i'm looking for right now :D

src/modal/modal.js
infnada: What if you are launching the modal inside a directive and pass the angular `element` from the directive `link` function? or `$element` from the directive controller? Thats what i'm looking for right now :D

like:

```javascript
.directive('someDirective', [function () {
	return {
		restrict: 'C',
		template: '<button ng-click="clickMe()">blabla</button',
		scope: true,
		replace: true,
		controller: ['$scope', '$modal', '$element', function ($scope, $modal, $element) {

			$scope.clickMe = function () {
				$modal.open({
					templateUrl: 'someUrl',
					appendTo: $element, //$element.parent() ....
					controller: ['$scope', '$modalInstance', function ($scope, $modalInstance) {

						$scope.close = function() {
							$modalInstance.dismiss('cancel');
						}

					}]
				})
			};

		}]
	};
}])
```

wesleycho: If it were just `modalContainer: modalWindow.appendTo`, it should work fine

wesleycho:
This feature landed recently as 16d854ca4c70aefb6b03faeb2a20105ce2bea8ec - thanks for the attempt!

PR #4202, opened 2015-08-13
feat(carousel): add model binding support to slide

### LINKED ISSUE #4201: Customization of the carousel
Cannot access $scope.slides object inside carousel controller to customize the carousel template. The answer in the stackoverflow explains the scenario and possible work around. 
Link: http://stackoverflow.com/a/31989387/1848109

> Keep up the good work. Your library is awesome.

wesleycho:
- Add support for binding slide model to slide

This resolves #4201.


Foxandxss:
Tests? The doc is a bit simple. I would like an example of it working and a test is a good candidate for the example.

PR #4203, opened 2015-08-13
Added ability to handle empty model in timepicker

### LINKED ISSUE #1114: TimePicker shows current time when model is empty.
TimePicker shows current time when the model is null or cleared. On the contrary, DatePicker is blank when model is empty, which is the correct approach. It is impossible to tell when the TimePicker model is not set by looking at the control. TimePicker should be blank when model is not set.

fmccown:
Model can be empty so user must type values into the hour and minute
text boxes. This closes #1114.

### DIFF
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -143,6 +143,9 @@ angular.module('ui.bootstrap.timepicker', [])
   }
 
   function pad(value) {
+    if (value === null) {
+      return '';
+    }
     return (angular.isDefined(value) && value.toString().length < 2) ? '0' + value : value.toString();
   }
 
@@ -218,12 +221,15 @@ angular.module('ui.bootstrap.timepicker', [])
       var hours = getHoursFromTemplate(),
         minutes = getMinutesFromTemplate();
 
-      if (angular.isDefined(hours) && angular.isDefined(minutes)) {
-        selected.setHours(hours);
-        if (selected < min || selected > max) {
-          invalidate(true);
-        } else {
-          refresh('h');
+      if (angular.isDefined(hours)) {
+        if (angular.isDefined(minutes)) {
+          selected.setHours(hours);
+          selected.setMinutes(minutes);
+          if (selected < min || selected > max) {
+            invalidate(true);
+          } else {
+            refresh('h');
+          }
         }
       } else {
         invalidate(true);
@@ -231,9 +237,12 @@ angular.module('ui.bootstrap.timepicker', [])
     };
 
     hoursInputEl.bind('blur', function(e) {
-      if (!$scope.invalidHours && $scope.hours < 10) {
+      ngModelCtrl.$setTouched();
+      if ($scope.hours === null || $scope.hours === "") {
+        invalidate(true);
+      } else if (!$scope.invalidHours && $scope.hours < 10) {
         $scope.$apply(function() {
-          $scope.hours = pad($scope.hours);
+        $scope.hours = pad($scope.hours);
         });
       }
     });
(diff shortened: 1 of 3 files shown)


src/timepicker/timepicker.js
```diff
@@ -143,6 +143,9 @@ angular.module('ui.bootstrap.timepicker', [])
   }
 
   function pad(value) {
+    if (value === null) {
```
wesleycho: `if (!value) {` may be a better check here.

src/timepicker/timepicker.js
```diff
@@ -231,9 +237,12 @@ angular.module('ui.bootstrap.timepicker', [])
     };
 
     hoursInputEl.bind('blur', function(e) {
-      if (!$scope.invalidHours && $scope.hours < 10) {
+      ngModelCtrl.$setTouched();
+      if ($scope.hours === null || $scope.hours === "") {
```
wesleycho: Change to `if (!$scope.hours) {`

src/timepicker/timepicker.js
```diff
@@ -255,7 +267,10 @@ angular.module('ui.bootstrap.timepicker', [])
     };
 
     minutesInputEl.bind('blur', function(e) {
-      if (!$scope.invalidMinutes && $scope.minutes < 10) {
+      ngModelCtrl.$setTouched();
+      if ($scope.minutes === null) {
```
wesleycho: Change to `if(!$scope.minutes) {`

src/timepicker/timepicker.js
```diff
@@ -300,17 +315,23 @@ angular.module('ui.bootstrap.timepicker', [])
...
   function updateTemplate(keyboardChange) {
-    var hours = selected.getHours(), minutes = selected.getMinutes();
-
-    if ($scope.showMeridian) {
-      hours = (hours === 0 || hours === 12) ? 12 : hours % 12; // Convert 24 to 12 hour system
-    }
+    if (ngModelCtrl.$modelValue == null) {
+      $scope.hours = null;
+      $scope.minutes = null;
+	  $scope.meridian = meridians[0];
```
wesleycho: Indentation is off here

wesleycho: Actually, seems like the whole block's indentation is off.

src/timepicker/timepicker.js
```diff
@@ -353,8 +374,15 @@ angular.module('ui.bootstrap.timepicker', [])
...
   $scope.toggleMeridian = function() {
+    var minutes = getMinutesFromTemplate(),
+        hours = getHoursFromTemplate();
+    
     if (!$scope.noToggleMeridian()) {
-      addMinutesToSelected(12 * 60 * (selected.getHours() < 12 ? 1 : -1));
+      if (angular.isDefined(minutes) && angular.isDefined(hours)) {
+        addMinutesToSelected(12 * 60 * (selected.getHours() < 12 ? 1 : -1));
+      } else {
+        $scope.meridian = $scope.meridian == meridians[0] ? meridians[1] : meridians[0];
```
wesleycho: use `===` here.

template/timepicker/timepicker.html
```diff
@@ -8,11 +8,11 @@
     </tr>
     <tr>
       <td class="form-group" ng-class="{'has-error': invalidHours}">
-        <input style="width:50px;" type="text" ng-model="hours" ng-change="updateHours()" class="form-control text-center" ng-readonly="::readonlyInput" maxlength="2">
+        <input style="width:50px;" type="text" placeholder="HH" ng-model="hours" ng-change="updateHours()" class="form-control text-center" ng-readonly="::readonlyInput" maxlength="2">
```
wesleycho: Remove the `placeholder` change

template/timepicker/timepicker.html
```diff
@@ -8,11 +8,11 @@
...
       </td>
       <td>:</td>
       <td class="form-group" ng-class="{'has-error': invalidMinutes}">
-        <input style="width:50px;" type="text" ng-model="minutes" ng-change="updateMinutes()" class="form-control text-center" ng-readonly="::readonlyInput" maxlength="2">
+        <input style="width:50px;" type="text" placeholder="MM" ng-model="minutes" ng-change="updateMinutes()" class="form-control text-center" ng-readonly="::readonlyInput" maxlength="2">
```
wesleycho: Ditto

anthonmanix:
I don't think he should change value === null to !value, because that won't work for 0 hours and 0 minutes.

wesleycho:
Good point - however, the other aspects are still issues.

wesleycho:
Even with code style cleaned up, this PR fails tests.

It needs to be worked on before this can be accepted.

ftorghele:
the PR is green now

wesleycho:
History needs to be cleaned up - there should not be any merge commits.

PR #4212, opened 2015-08-16
feat(typeahead): add `typeaheadFocusOnSelect`

Foxandxss:
This is the continuation of #4211. The failure test is here.

Since we now attach stuff directly to the DOM, it is better to clean it after each test .


wesleycho:
LGTM

Issue #2952, opened 2014-11-11
UTC mode for DatePicker or default timezone option


slavafomin:
Hello!

I have a problem with UTC dates in DatePicker component.

When I provide a date to it in UTC format it appears to work normally, however, when I set initial value to `null` and then user selects some date, it returns date in local timezone, but I want it to always return dates in UTC format.

Maybe you can create some configuration option like `utcMode` or provide an option to specify default timezone?

Thanks!

Iku-turso:
+1

karianna:
@chrisirhc Can you add a comment for the Purgatory reasoning?

jasontk19:
+1 for UTC

piupiupiu:
+1

MJGrant:
+1

chrisirhc:
@karianna I did that because we lacked resources to look into this and also, datepicker is pending quite a bit of refactoring. I'm putting it into Backlog. This is clearly a very desirable feature.

bguezzie:
Using 0.13.0 library:
I was able to work around this by adding a directive to my element (the datepicker popup trigger) that basically did the following:

```javascript
ngModelCtrl.$formatters.push(function (value) {
    var date = new Date(value);

    if (isNaN(date)) {
        return value;  // needed for null/empty/undefined values
    }

    date.setMinutes(date.getTimezoneOffset());  // forces the "local" version of the date to match the intended UTC date

    return date;  // yes, this is a Date object being applied to the $viewValue...don't judge, it works!
});
```

Once I did that, the datepicker used the $viewValue (a Date object in this case) as its model, and since a decent offset had been applied the correct date was shown in my input as well as in the graphical day picker.

Caveat:  Some further conversion might be needed when consuming the timestamp provided by the datepicker, but that hasn't proven problematic in our project so far.

ttschupp:
@bguezzie : I did something similar. 

    app.directive('datepickerTimezone', function () {
        return {
            restrict: 'A',
            priority: 1,
            require: 'ngModel',
            link: function (scope, element, attrs, ctrl) {
                ctrl.$formatters.push(function (value) {
                    var date = new Date(Date.parse(value));
                    date = new Date(date.getTime() + (60000 * date.getTimezoneOffset()));
                    //if (!dateFormat || !modelValue) return "";
                    //var retVal = moment(modelValue).format(dateFormat);
                    return date;
                });

                ctrl.$parsers.push(function (value) {
[...]

Issue #2906, opened 2014-10-29
Datepicker value is off by a day when date format is different from the datepicker's date format


Trish00:
I have my date picker format set to:  datepicker-popup="MM/dd/yyyy".

If a user types in a date in a different format the date ends as the day before.

i.e.  User types in 2013-10-07  date value is 2013/10/06.  But if the user types in 2013-10-7 the date is correct.
      User types in 2014-08-09 date value is 2014-08-08 but if the user types in 2014-9-09 the date is correct.

Here is an example using your the datepicker example on your website.  It would seem it doesn't like the leading zeros.

![datepicker - 2014-08-10](https://cloud.githubusercontent.com/assets/8975032/4832413/f70cbfc4-5f9b-11e4-94cd-626939a43e74.JPG)

cgwyllie:
+1 for exposing a configuration option to allow 'strict' parsing, we've just encountered this issue with 'relaxed' parsing too on a multi-locale app.

wesleycho:
This should now be fixed in `master`, if not in 0.14.3

Issue #1891, opened 2014-03-04
TZ localization in date picker should be optional


porschelau:
We have an application which operates solely on UTC time despite of the fact that we have works in PST as well as EST. We are running into issues where we really just want the date picker to show the date in UTC mode rather than localizing it by default.

I am proposing that we offer this as a bindable configuration at the date picker level and pump the configuration into the ng-date-filter. I have a separate ticket to represent the changes needed at the date filter.

ensAlejandro:
+1

przemoli:
Any update?

wesleycho:
Just noting, I have a current work in progress where the datepicker defaults to the timezone of the date bound to the model initially - I should be able to have this ready by the end of the week.

wesleycho:
Currently, one can set an initial model value of the desired timezone and that timezone will be used throughout the datepicker.

If one wants to dynamically change it, that still needs work, but is also captured in other issues.

PR #4218, opened 2015-08-17
feat(datepicker): timeless-json-mode

davious:
datepickerConfig option timelessJsonMode for json models when we don't care about the time.
* also available as attribute or via datepicker-options
* on init and parse: sets model to 'yyyy-MM-dd' format
* on format: converts model to date before normal formatting
* timelessJsonUtils constant with method jsonDateToTicks for use with min-date, max-date
	* alias jsonDateAddDays
	* set timelessJsonMode = 'withUtils' to add methods to parent scope
	* also: jsonToDate -- used internally

[plunker](http://plnkr.co/edit/WzoImdpZtQcQDfN7AN5e?p=preview)
[codepen](http://codepen.io/davious/pen/qOXNZY?editors=101)

Related: #1891, #2906, #2952

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/bootstrap/4218)

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -17,10 +17,12 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   yearRange: 20,
   minDate: null,
   maxDate: null,
-  shortcutPropagation: false
+  shortcutPropagation: false,
+  timelessJsonMode: false
 })
 
-.controller('UibDatepickerController', ['$scope', '$attrs', '$parse', '$interpolate', '$log', 'dateFilter', 'uibDatepickerConfig', '$datepickerSuppressError', function($scope, $attrs, $parse, $interpolate, $log, dateFilter, datepickerConfig, $datepickerSuppressError) {
+.controller('UibDatepickerController', ['$scope', '$attrs', '$parse', '$interpolate', '$log', 'dateFilter', 'uibDatepickerConfig', '$datepickerSuppressError', 'timelessJsonUtils',
+  function($scope, $attrs, $parse, $interpolate, $log, dateFilter, datepickerConfig, $datepickerSuppressError, timelessJsonUtils) {
   var self = this,
       ngModelCtrl = { $setViewValue: angular.noop }; // nullModelCtrl;
 
@@ -36,6 +38,13 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   angular.forEach(['showWeeks', 'startingDay', 'yearRange', 'shortcutPropagation'], function(key) {
     self[key] = angular.isDefined($attrs[key]) ? $scope.$parent.$eval($attrs[key]) : datepickerConfig[key];
   });
+
+  self.timelessJsonMode = angular.isDefined($attrs.timelessJsonMode) ?
+     $attrs.timelessJsonMode.toLowerCase() !== 'false' : datepickerConfig.timelessJsonMode;
+
+  if (($attrs.timelessJsonMode || datepickerConfig.timelessJsonMode || '').toString().toLowerCase() === 'withutils') {
+    angular.extend($scope.$parent, timelessJsonUtils);
+  }
 
   // Watchable date attributes
   angular.forEach(['minDate', 'maxDate'], function(key) {
@@ -106,8 +115,10 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
     if (ngModelCtrl.$viewValue) {
       var date = new Date(ngModelCtrl.$viewValue),
           isValid = !isNaN(date);
-
       if (isValid) {
+        if (self.timelessJsonMode) {
+          date = timelessJsonUtils.jsonToDate(date.toJSON(), ngModelCtrl, dateFilter);
+        }
         this.activeDate = date;
       } else if (!$datepickerSuppressError) {
         $log.error('Datepicker directive: "ng-model" value must be a Date object, a number of milliseconds since 01.01.1970 or a string representing an RFC2822 or ISO 8601 date.');
@@ -128,6 +139,9 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
 
   this.createDateObject = function(date, format) {
     var model = ngModelCtrl.$viewValue ? new Date(ngModelCtrl.$viewValue) : null;
+    if (model && self.timelessJsonMode) {
+      model = timelessJsonUtils.jsonToDate(model.toJSON());
+    }
     return {
       date: date,
       label: dateFilter(date, format),
@@ -162,6 +176,9 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
     if ($scope.datepickerMode === self.minMode) {
       var dt = ngModelCtrl.$viewValue ? new Date(ngModelCtrl.$viewValue) : new Date(0, 0, 0, 0, 0, 0, 0);
       dt.setFullYear(date.getFullYear(), date.getMonth(), date.getDate());
+      if (self.timelessJsonMode) {
+        dt = dateFilter(dt, 'yyyy-MM-dd');
+      }
       ngModelCtrl.$setViewValue(dt);
       ngModelCtrl.$render();
     } else {
@@ -543,14 +560,14 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   altInputFormats: []
 })
 
-.controller('UibDatepickerPopupController', ['$scope', '$element', '$attrs', '$compile', '$parse', '$document', '$rootScope', '$uibPosition', 'dateFilter', 'uibDateParser', 'uibDatepickerPopupConfig', '$timeout',
-function(scope, element, attrs, $compile, $parse, $document, $rootScope, $position, dateFilter, dateParser, datepickerPopupConfig, $timeout) {
+.controller('UibDatepickerPopupController', ['$scope', '$element', '$attrs', '$compile', '$parse', '$document', '$rootScope', '$uibPosition', 'dateFilter', 'uibDateParser', 'uibDatepickerPopupConfig', '$timeout', 'timelessJsonUtils', 'uibDatepickerConfig',
+function(scope, element, attrs, $compile, $parse, $document, $rootScope, $position, dateFilter, dateParser, datepickerPopupConfig, $timeout, timelessJsonUtils, datepickerConfig) {
   var self = this;
   var cache = {},
     isHtml5DateInput = false;
   var dateFormat, closeOnDateSelection, appendToBody, onOpenFocus,
     datepickerPopupTemplateUrl, datepickerTemplateUrl, popupEl, datepickerEl,
-    ngModel, $popup, altInputFormats;
+    ngModel, $popup, altInputFormats, timelessJsonMode, addTimelessJsonUtils;
 
   scope.watchData = {};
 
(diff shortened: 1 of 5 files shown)


src/datepicker/datepicker.js
```diff
@@ -649,7 +663,7 @@ function($compile, $parse, $document, $rootScope, $position, dateFilter, datePar
           if (isNaN(date)) {
             return undefined;
           } else {
-            return date;
+            return timelessJsonMode ? dateFilter(date, 'yyyy-MM-dd') : date;
```
wesleycho: This looks like it overrides the date format - shouldn't it be `dateFilter(date, dateFormat)`? What am I missing here?

src/datepicker/datepicker.js
```diff
@@ -535,6 +536,19 @@ function($compile, $parse, $document, $rootScope, $position, dateFilter, datePar
         return (new Date(date1.getFullYear(), date1.getMonth(), date1.getDate()) - new Date(date2.getFullYear(), date2.getMonth(), date2.getDate()));
       };
 
+      scope.jsonToDate = function(date) {
+        if (date && angular.isString(date) && date.indexOf('-') > -1) {
+          var update = date.indexOf('T') > -1;
+          date = date.substring(0, 10).split('-');
+          date = new Date(+date[0], +date[1] - 1, +date[2]);
+          if(update) {
```
wesleycho: Add space after `if`

src/datepicker/datepicker.js
```diff
@@ -684,11 +698,17 @@ function($compile, $parse, $document, $rootScope, $position, dateFilter, datePar
         ngModel.$validators.date = validator;
         ngModel.$parsers.unshift(parseDate);
         ngModel.$formatters.push(function(value) {
+          if(timelessJsonMode) {
```
wesleycho: Add space after `if`

src/datepicker/datepicker.js
```diff
@@ -684,11 +698,17 @@ function($compile, $parse, $document, $rootScope, $position, dateFilter, datePar
...
+            value = scope.jsonToDate(value);
+          }
           scope.date = value;
           return ngModel.$isEmpty(value) ? value : dateFilter(value, dateFormat);
         });
       } else {
         ngModel.$formatters.push(function(value) {
+          if(timelessJsonMode) {
```
wesleycho: Add space after `if`

wesleycho:
I need to probe this a little more, but this looks like mostly solid work, minus that question I have.

src/datepicker/datepicker.js
davious: @wesleycho this parseDate function is added to the ngModel.$parsers for non-Html5DatrInputs; see line 699; I am setting the model, usually a date object, to a date string. I want the model to be a json compatable string so I can post it over ajax without any fuss -- and I also want to have it play nicely as a date in angular-land. The 'yyyy-MM-dd' format allows this to happen.

src/datepicker/datepicker.js
```diff
@@ -535,6 +536,30 @@ function($compile, $parse, $document, $rootScope, $position, dateFilter, datePar
...
+            date = new Date(+date[0], +date[1] - 1, +date[2]);
+            if (update) {
+              ngModel.$setViewValue(dateFilter(date, dateFormat));
+              ngModel.$render(); 
+            }
+          }
+          return date;
+        };
+        
+        scope.$parent.jsonToTicks = function(date) {
```
wesleycho: This has me concerned - this would appear to pollute the parent $scope, which should only be read for config values.

src/datepicker/datepicker.js
```diff
@@ -535,6 +536,30 @@ function($compile, $parse, $document, $rootScope, $position, dateFilter, datePar
...
+          if (date && angular.isString(date) && date.indexOf('-') > -1) {
+            date = date.substring(0, 10).split('-');
+            date = new Date(+date[0], +date[1] - 1, +date[2]);
+          }
+          return date.getTime ? date.getTime() : null;
+        };
+        scope.$parent.oneDay = 1000 * 60 * 60 * 24;
```

src/datepicker/datepicker.js
davious: @wesleycho Yeah, that was nagging me too. I plan to fix this. When I go `scope.prop = function` it isn't available on the element scope. Do you happen to have any advice here: any special way of adding to this scope in a way that persists past the link function?

wesleycho: Maybe provide an optional binding on the isolate scope where the user can pull the value from the directive?

davious: @wesleycho OK. I understand what's happening now.

The `datepickerJsonUtils` directive that I was using in the [original commit](https://github.com/davious/angular-ui-bootstrap/commit/145578efd7a9e4ce8278f6b249b948e8a98a2fff#diff-6240fc17e068eaeef7095937a1d63eaeR842) didn't have an isolated scope, so it was actually doing the same thing as this [latest commit](https://github.com/davious/angular-ui-bootstrap/commit/e30d284aeab91ed21bfd537837cf247032b6d9e5). The only difference is the earlier way required additional markup.

Now that I realize this, I'm personally OK with the function and property living on the parent scope.

An incidental thought I had is I could move the `oneDay` property into an optional function parameter, for example: `jsonToTicks(someJsonDate, -1)`

davious:
@wesleycho I moved the `jsonDateToTicks` function into a `timelessDateUtils` service with a `jsonDateAddDays` alias. This allows the client to bring in the function as they see fit.

I also added a `timelessJsonUtils` convenience option to allow the client to intentionally add the service methods to the parent scope.

davious:
changed `timelessJsonUtils` option to `timelessJsonMode='withUtils'`

BobbieBarker:
LGTM, Docs, Demo, and readme need updated?

davious:
added `timelessJsonMode` support for `uib-datepicker`.
plunker demo: http://plnkr.co/edit/WzoImdpZtQcQDfN7AN5e?p=preview
codepin: http://codepen.io/davious/pen/qOXNZY?editors=101

had trouble with datepicker initial formatting to 'yyyy-MM-dd', but it's working in codepin even after removing the conjoined datepicker-popup

src/datepicker/datepicker.js
```diff
@@ -33,6 +35,13 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
     self[key] = angular.isDefined($attrs[key]) ? (index < 6 ? $interpolate($attrs[key])($scope.$parent) : $scope.$parent.$eval($attrs[key])) : datepickerConfig[key];
   });
 
+  self.timelessJsonMode = angular.isDefined($attrs.timelessJsonMode) ?
+     $attrs.timelessJsonMode.toLowerCase() != 'false' && ($attrs.timelessJsonMode || true)
```
wesleycho: `($attrs.timelessJsonMode || true)` is redundant - omit this condition completely.

This whole line can be reduced to `$attrs.timelessJsonMode !== 'false'`.

src/datepicker/datepicker.js
```diff
@@ -859,4 +902,45 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
       return attrs.templateUrl || 'template/datepicker/popup.html';
     }
   };
-});
+})
+.constant('timelessJsonUtils', (function() {
```
wesleycho: This should not be a constant, but a service.

src/datepicker/datepicker.js
davious: @wesleycho `&& ($attrs.timelessJsonMode || true)` allows the `withUtils` option to be passed in.

wesleycho: Ah, I see - in any case, this can still be simplified to `self-timelessJsonMode = angular.isDefined($attrs.timelessJsonMode) ? $attrs.timelessJsonMode !== 'false' && $attrs.timelessJsonMode`

I don't think it should default to true, this should be opt-in.

davious: It is opt-in. The `true` is there if the attribute is added without a value. It allows for `<div><input ng-model="date" uib-datepicker-popup="MM/dd/yyyy" timeless-json-mode></div>` to trigger the mode.

paulflo150:
Out of curiosity, what would be the purpose of having a "timeless" option? I would think this just adds to the complexity of the datepicker without a gain - most people want a way to set the date/ time pickers to UTC without any hacks - and that would yield a reliable date only part, as well as time.

davious:
this dateInterceptProvider is working for me on 0.13.4 and 0.14.3
```
(function(window, angular) {
  'use strict';
  
  angular.module('app')
    .provider('dateInterceptor', dateInterceptorProvider);
    
  function dateInterceptorProvider() {
    this.$get = function() {
      return {
        response: response,
        request: request
      };
    };
    
    function response(response) {
      convertJsonToDates(response.data);
      return response;
    }
    
    function convertJsonToDates(obj) {
      for (var key in obj) {
        var value = obj[key];
        var type = typeof value;
        if (type === 'string') {
          if(/^\d{4}-\d{2}-\d{2}T/.test(value)) {
            var dateArray = value.split('T')[0].split('-');
            var date = new Date(+dateArray[0], +dateArray[1] - 1, +dateArray[2], 0, 0, 0, 0);
            date.setFullYear(+dateArray[0]);
            obj[key] = date;
          }
        } else if (type === 'object') {
          convertJsonToDates(value);
        }
      }
    }
  
    function request(config) {
      convertDatesToJson(config.data);
      return config;
    }
    
    function pad(num) {
      return (num < 10 ? '0' : '') + num;
    }
    
    function convertDatesToJson(obj) {
      if (typeof obj !== "object") {
        return;
      }

      for (var key in obj) {
        if (!obj.hasOwnProperty(key)) {
          continue;
        }
        var value = obj[key];
        if (value && angular.isDate(value) && !isNaN(value)) {
          obj[key] = value.getFullYear() + '-' + pad(value.getMonth() + 1) + - + pad(value.getDate());
        } else if (typeof value === "object") {
          convertDatesToJson(value);
        }
      }
    }
  }
})(window, window.angular);
```

davious:
One nit I have with the above approach (https://github.com/angular-ui/bootstrap/pull/4218#issuecomment-158249540) is it produces a `flash-of-invalid` as the date model is transformed to a string on the request.

wesleycho:
Question - is this PR relevant anymore after the timezone work done now in the library?

davious:
For the envisioned use-case, it is possible to use `ng-model-options = {timezone: utc}`; but, I still need to convert my date strings and all my client-side dates to utc dates.

From my point of view, this does not match the intent of the pr, which is to provide a no-brainer date picker. A typical datepicker, say [jQuery UI datepicker](https://jqueryui.com/datepicker/), leaves the model in a date format. I leave it up to you, the owners, to decide whether you'd like to allow for a date string mode.

Without such a mode, I still would like to be able to have something out-of-the-box that supports want I consider the default datepicker case: one that converts date strings to dates and emits date strings. Unfortunately, because the javascript date object is natively formatted in utc, this a problem only a dateInterceptor can solve. Barring a date string mode, I'd like to salvage some ideas here: an out-of-the-box dateInterceptor and convenience date helpers for min-date/max-date comparisons (moving them to dateParser).

wesleycho:
UI Bootstrap isn't meant to necessarily be an out of the box library - it is meant to be more of an extensible library that does most of the difficult common operations to avoid a bloating of configuration options (which inevitably leads to configuration options that override the behavior of another configuration option, as well as library bloat). The datepicker has been the antithesis of that unfortunately, and we'd ideally like to keep it as thin as possible.

The description sounds like the opposite of our goals.

davious:
OK. Hmm. I sense tension. I'll presume a no on a date string mode.

What are your thoughts on an out-of-the-box dateInterceptor and convenience date helpers for min-date/max-date comparisons (moving them to dateParser)?

PR #4220, opened 2015-08-17
feat(dateparser): add support for the `h` format

JonathanGawrych:
Similar to #3417 and #4117, add the date format `h` from Angular's [dateFilter](https://docs.angularjs.org/api/ng/filter/date)


wesleycho:
Thanks for this (and the other PR)!

PR #4226, opened 2015-08-18
feat(modal): support multiple open classes

### LINKED ISSUE #4184: openedClass not removed when close/dismiss with 2 pop-ups
I have problem, one case like you have 2 opened pop-ups and i got 2 body classes ' class="modal-scrollable modal-open" ' when I close pop-up with modal-open class, class is not toggled 
```javascript
body.toggleClass(modalWindow.openedClass || OPENED_MODAL_CLASS, openedWindows.length() > 0).
```

wesleycho:
- Support multiple modal open classes when multiple modals are open

This fixes #4184.


Foxandxss:
The implementation and tests are nice. Simple issue, complex solution :)

wesleycho:
Yep, have to use a comb to filter each instance by the class type essentially to ensure all with that class are removed only when none of them are open.

PR #3060, opened 2014-12-06
fix(tabs):  #2883 select gets called twice

caseyg1204:
Fix for https://github.com/angular-ui/bootstrap/issues/2883

PR #4230, opened 2015-08-18
fix(tabs): ensure tab selection only occurs once

### LINKED ISSUE #2883: If initial active tab is not the first one, its select gets called twice
If the initial active tab is not the first tab, its select action gets called twice.

You can see this in the slightly modified sample code in this [Plunker](http://plnkr.co/edit/CErAYR11JWvjXgFaiKj5?p=preview).  All I did was add select="alertMe()" to additional tabs, and set active:true to the first Dynamic tab.

wesleycho:
- Fix issue where select callback is fired twice due to compilation order when using dynamic tabs mixed with static tabs

This incorporates the changes from #3060, and adds a unit test around it.

The fix can be seen in action [here](http://plnkr.co/edit/OVCJgHn2j6MyLMWhzZst?p=preview).

This closes #3060, and fixes #2883.


Foxandxss:
:+1:

PR #4231, opened 2015-08-18
Typeahead append element to div with id

trueinviso:
I'd like to add this feature so that the typeahead popup can be appended to another div.  This works on my site, however I can't get the test to pass.  I can see the popup get appended inside typeahead.js when I put in log statements.  However, the popup doesn't display when text is entered into the input field.  Suggestions?


src/typeahead/typeahead.js
```diff
@@ -419,6 +421,8 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position', 'ui.bootstrap
 
         if (appendToBody) {
           $document.find('body').append($popup);
+        } else if(appendToElementId !== false) {
+            $document.find('#' + appendToElementId).append($popup);
```
wesleycho: It should be noted that this library does not have a dependency on jQuery, so this would fail with purely jqLite.

wesleycho:
What is the use case for such a feature? This sounds like it would be bad for accessibility.

trueinviso:
I need the results to be appended under the Popular Searches heading in this screen shot.  My solution was to append the popup to the div using typeahead-append-to-element-id.  

screenshot

src/typeahead/typeahead.js
wesleycho: One workaround here is to do `$document[0].getElementById(appendToElementId)`

wesleycho:
The test needs to be fixed.

wesleycho:
Ahh, I know why it fails the test - the element doesn't exist on the DOM - in the test, just do `$document.find('body').append(element)` and remove it after the test with `element.remove(); element = null;` (for gc reasons).

PR #4238, opened 2015-08-20
prevent typeahead-template-url working wrong

qpitlove:
prevent that dropdown has typeahead directive, typeahead-template-url is working wrong.


wesleycho:
This needs more information - what is the problem, and why does this fix it? Reproduction via Plunker would be best.

Also the commit is from a different account - this needs to be verified that the person is ok with it being merged.

wesleycho:
Ah, I see the original issue now (saw the PR first) - this seems related to the datepicker issue that is similar.

This needs investigation as to what is going on exactly here - in addition, the test needs to be fixed if this is to be merged.

qpitlove:
Ok. I wrote issue before pr request.
https://github.com/angular-ui/bootstrap/issues/4239
Include reason and plunker example link

qpitlove:
And my suggestion pr sample,  i don't care whether merge or not.

qpitlove:
I'll close this pr.
Thanks.

PR #4242, opened 2015-08-20
feat(accordion): allow custom panel class

### LINKED ISSUE #3968: Allow different accordion-group classes
The class of panels (aka accordion groups) at https://github.com/angular-ui/bootstrap/blob/master/template/accordion/accordion-group.html are hard-coded to `panel-default`. I would like to be able to use a different one. Currently, even if I use `ng-class` to assign another class to the panel, the `panel-default` class is still there.

icfantv:
* add support for custom panel class

fixes #3968


template/accordion/accordion-group.html
```diff
@@ -1,4 +1,4 @@
-<div class="panel panel-default">
+<div class="panel panel-{{panelClass || 'default'}}">
```
wesleycho: I think I would prefer this be `{{panelClass || 'panel-default'}}` - gives the users of the library a little more flexibility without any cost.

icfantv: I really went back and forth here on this (as well as putting the default inside the directive ala `scope.panelClass = attrs.panelClass || 'panel-default';` but opted for what I did so as to keep the code as close to the element as possible) but ultimately decided to mimic [what we do in tabs](https://github.com/angular-ui/bootstrap/blob/master/template/tabs/tabset.html#L2).

Recall that we did talk about only supporting the built-in contextual panel classes from Bootstrap - which is also why I went this route.

But I don't have a strong feeling either way and if we want to be totally flexible that solution would do it.

wesleycho: There are some users of the library who use it with custom CSS or other CSS libraries - I think we should try to be less opinionated here for minimal pain.

icfantv: Sounds good.

wesleycho:
LGTM

Issue #2429, opened 2014-07-07
Progress directive overrides HTML 5 <progress> element.


blindworld:
The HTML 5 progress element (http://www.w3.org/html/wg/drafts/html/CR/forms.html#the-progress-element) gets overridden by the progress directive here (https://github.com/angular-ui/bootstrap/blob/master/src/progressbar/progressbar.js).

I realize we could create our own build and just not include the progress bar, but that would prevent us from using bower to keep up to date as well.

bzitzow:
+1

ryanhburbank:
I agree with @scottopherson, I would really like some better namespacing in the library.

LeDeep:
+1 please. Prefixing would be great

JobaDiniz:
+1

chrisirhc:
Will be fixed via #157 , moving milestone to align with that issue.

wesleycho:
This should now be fixed.

wesleycho:
Sorry, closed prematurely - this is still an issue for stacked bars.

grzegorzjudas:
+1, this is getting ridiculous, especially now, when HTML5 is being so widely used...

wesleycho:
It's not so simple - changing this is a breaking change, and guaranteed to break apps.

In addition there are a lot of feature requests/bug fixes that also need addressing, and short manpower.

If someone were to put together a high quality PR, it would be reviewed and merged in, but nobody has stepped up to the plate to contribute.

wesleycho:
I'm going to close this, as we have already marked the `progress` element directive as deprecated and it is scheduled to be removed in 1.0.0 along with all of the non-prefixed directives.

PR #4255, opened 2015-08-21
fix(progress): rename to avoid conflict

wesleycho:
- Rename to `ui-progress` and `ui-bar`

This is for fixing #2429, but does not fix it completely. The second step is removing the offending directives, which I have scheduled for 0.14.0 to give consumers of the library adequate time to migrate over.


icfantv:
What do you think about injecting the `$log` service in the offending directives and adding a warning statement about deprecation?

wesleycho:
Already done if you check it out :P

icfantv:
Wow.  That was really embarrassing.  I told you I was too impulsive...

icfantv:
:+1:

PR #4257, opened 2015-08-22
feat(collapse): convert to use `$animateCss`

wesleycho:
- Change to use `$animateCss` when available

I would like someone to pull this down and investigate the tests if possible - in particular, in the `should shrink accordingly when content size inside collapse decreases` test, if I do `$animate.triggerCallbacks()`, it causes the test to fail. Perhaps I am not understanding what is truly going on here with ngAnimate, since it seemed necessary in some of the other ones.


Foxandxss:
I would love to give you an answer, but it is still magic for me. Sometimes you need it, sometimes you don't. That is ngAnimate for you.

wesleycho:
Commenting that this should not be merged - ngAnimateMock changes in 1.4.5 means that this is broken.

wesleycho:
Going to need your assistance for this @Foxandxss - stuck on why these two tests don't pass :( .

PR #4267, opened 2015-08-25
docs(accordion): update demo and docs to match latest version

Foxandxss:
I added all the extra options we had without docs and I also mentioned the `accordion-heading`.


wesleycho:
LGTM

Issue #4277, opened 2015-08-26
Compatibility with Angular JS


vitaly-t:
Took me some time to figure out why upgrading from 0.13.0 to any of the later versions was breaking my UI everywhere...

Turns out, starting from 0.13.1, the library is no longer compatible with Angular JS 1.3 prior to 1.3.18, because 1.3.18 contains fixes in animation that directly effect Angular UI changes for 0.13.1, in particular related to the DateTime control.

I didn't expect it, while using Angular JS 1.3.17, and I couldn't find any notice here to state that compatibility issue. It becomes even stranger, if you consider that AngularJS 1.3.18 was released 8 days ago, while Angular UI 0.13.1 was released 1 month ago.

As a test, try to use Angular UI 0.13.1 - 0.13.3 with Angular JS 1.3.17. Just as it comes to using DateTime picker, it all stops working. Angular-animate 1.3.17 throws an error on line 1205:

```javascript
var parentNode = elementNode.parentNode;
```
saying that elementNode is `undefined`.

Only upgrading Angular JS to version 1.3.18 fixes the issue.

icfantv:
@vitaly-t in the future, please reserve the issues section for actual bugs rather than a place to vent.

Before upgrading any 3rd party library, the onus is on the user to read the appropriate changelogs to see if there are any breaking changes.  On that note, there were a lot of 3rd party libraries that were broken due to the animation error.

If you would like to suggest an update to the documentation to indicate library requirements, then please just suggest this, or better yet, submit a PR.

GunninDownCode:
I also ran into this issue, and upgrading to 1.3.18 fixed it for me as well. I did notice, however, that the animation on the datepicker is horrendously slow now. Any idea on why that is or how to fix it?

icfantv:
@GunninDownCode, are you able to test with the latest version of Angular 1.4 (1.4.7, I think) and Angular UI Bootstrap (0.14.3, or even `master`) to see if this is still an issue?

I ask because there were a few 1.4 releases (I want to say 1.4.4 - 1.4.6) where there were animation issues.

Without knowing your code base, I don't know that these will be drop in replacements (`master` on Angular UI Bootstrap won't be because we moved to prefixed directive names) but a basic plunker might suffice.

If it IS still an issue, please let us know and we'll probably have you open a new issue.  Thanks.

GunninDownCode:
[...]

PR #4278, opened 2015-08-26
docs(readme): add clarity on angular requirement

icfantv:
* add a note indicating that due to critical animation fixes, UI Bootstrap
  0.13.3 and higher requires Angular 1.3.18 or higher.

addresses #4277


wesleycho:
Maybe it's worth changing the comments to highly recommend using 1.3.18? I'm an advocate of letting people play with fire if they like, especially since that is sort of what we have seen to have enough of a presence in the community.

icfantv:
I'm cool with that.  @Foxandxss, I'll leave this as you're currently reworking the docs?

PR #4279, opened 2015-08-26
docs(alert): revamp to match the general style

Foxandxss:
I matched the doc style we use on the other components.

Also updated it with more options and a more complete example.


wesleycho:
LGTM

PR #4280, opened 2015-08-26
chore(alert): remove restrict

Foxandxss:
That is the default since 1.2.x or something like that so we are fine by removing it.


wesleycho:
My only concern is that this is a potential breaking change as far as CSS goes.

Otherwise this is probably fine.

Foxandxss:
I don't see what you mean. That line adds nothing at all to the code.

wesleycho:
Oh no, I thought replace.

Ok, I'm fine with this - the default is now EA right?

Issue #3264, opened 2015-02-06
Revisit Button Group HTML


kwinters:
In 772734177f2f0e82add275c1bfcfe73a1f32cce5 the button group code was changed from buttons to labels, Closes #1323. Closes #1482

But I think it's time to revisit that decision (2013).

http://getbootstrap.com/javascript/#buttons-checkbox-radio uses a label and an input, and none of the examples in http://getbootstrap.com/components/#btn-groups have labels at all.

The above commit's message is "Align with BS3 markup." but the code does not meet that goal, at least anymore.  From the other threads, it looks like the example started with the input + label format and then determined that the inputs were not required, ending up with just label.  However it appears that the typical usage is to use either actual buttons or divs, not orphaned labels.

And I think that the use of labels in this way should be discouraged in general.  Labels tend to have styling associated with them that are not appropriate for an element that acts like a button.

rvanbaalen:
@kwinters I think you have a totally valid point there. @chrisirhc Any idea why the buttons where changed to labels back then?

chrisirhc:
This refers to the demo site's code. We can improve on this by including the `input` checkboxes but state in that they're not required. As it's a lot more code but it's good for accessibility.

However, @rvanbaalen just found that the Bootstrap CSS only hides the inputs when it's nested inside a `data-toggle`: https://github.com/twbs/bootstrap/blob/master/less/button-groups.less#L221-L243
As such, if we want to do this, we need to add `data-toggle` in the parent of the demo site's button groups as well.

rvanbaalen:
After digging in and checking bootstrap's source I would say for semantical and accessibility reasons we should add an `<input>` inside the `<label>` tag.

But, after I checked out Bootstrap's CSS I noticed that they target the internal `<input>` element with a css selector prefix:

```
[data-toggle=buttons]>.btn input[type=checkbox] {
  position: absolute;
  clip: rect(0,0,0,0);
  pointer-events: none;
}
```

So either we have to add `data-toggle="buttons"` to the wrapping `div` or we have to handle the hiding of the checkbox / radio button ourselves within the directive.

rvanbaalen:
I would say adding `[data-toggle="buttons"]` adds no added value to our templates and could possibly result in unwanted behavior. My vote goes to hiding the `<input>` ourselves within the `btnCheckbox` directive.

chrisirhc:
I see, that sounds reasonable. If a btn-checkbox element contains any input elements within, hide them.

[...]

PR #4282, opened 2015-08-26
feat(buttons): hide nested inputs

wesleycho:
- Hide nested input elements

This is for #3264 - note that the demo site would also need a change to use this.


icfantv:
:+1:

Foxandxss:
I will probably revisit this library + docs + demo site on the next week so I can put that up to date.

Issue #4275, opened 2015-08-26
Timepicker - expose template url


waratah:
I have a need to override the formatting of the time picker and note the documentation does not include the ability to use a template url attribute which would be consistent with other components.

This would allow me to move the spinners around to best fit the screens.

wesleycho:
Currently that feature is not implemented - I believe the two components that do not have it fully available is the timepicker and typeahead. I have been bogged down on a major refactor of the datepicker to optionally support moment.js usage.

If you'd like to take a stab at it, feel free to file a PR implementing this - it should be pretty easy to follow the existing feature and tests on some of the other components for adding this to the timepicker.

icfantv:
@waratah please let us know if you want to do the PR for this.  Otherwise, I can knock this out in an hour or so.  Thanks.

waratah:
Sorry I am a git newbie.   @icfantv can you please.

icfantv:
I'll see if I can get to it tonight.  I have some work work to do.

PR #4284, opened 2015-08-27
feat(timepicker): add `templateUrl` and `controllerAs` support

icfantv:
* Add support for overriding the `templateUrl` on an instance by instance basis
* Expose controller via `controllerAs`

addresses #4275


Foxandxss:
That is good to go sir.

PR #4293, opened 2015-08-27
fix(position): improve placing around corners

ksheedlo:
This change improves corner positioning to make it more internally consistent and useful.

The current implementation of corner positioning aligns elements in a way that is less useful than it could be. For instance, passing `left` as the second position argument aligns the left edge of the target element with the left edge of the host, while passing `right` aligns the *left edge* of the target with the right edge of the host. This behavior makes it difficult to trigger popovers on a right-aligned control. With this PR, passing `right` as the second position argument will cause the right edge of the target to be aligned with the right edge of the host control; and likewise for top and bottom. Here is the [plunker](http://plnkr.co/edit/6m06t94Boy10Q3LMUYV7?p=preview) with the updated functionality.

This change is potentially breaking, but I feel it's worth it in terms of improving the usability of positioned elements in ui-bootstrap.

### DIFF
--- a/src/position/position.js
+++ b/src/position/position.js
@@ -26,6 +26,74 @@ angular.module('ui.bootstrap.position', [])
     }
 
     /**
+     * Computes the `left` position property of a target element to be
+     * positioned.
+     * @param place the placement of the target element relative to the
+     *    host. For instance, place `'bottom'` means the target element
+     *    will be placed below the host.
+     * @param align the alignment of the element relative to the host.
+     *    For instance, align `'left'` means align the element with the host's
+     *    left edge. `'center'` may be passed to center the element.
+     * @param hostElPos the host element position, having at least a `width`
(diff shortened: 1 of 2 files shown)


wesleycho:
Thanks for this PR - I will try to review this this week, but it may be difficult since I am heading to a week long vacation in a couple of days.

This generally looks good, but I just want to think about the API signature some and whether we could improve it.

wesleycho:
Thinking about this some more, can you demonstrate this in a Plunker? This may potentially cause issues if the user wants it aligned to the right but the host is not aligned to the right.

ksheedlo:
@wesleycho [Plunker](http://plnkr.co/edit/6m06t94Boy10Q3LMUYV7?p=preview)

icfantv:
Closing in favor of #4762.

PR #4299, opened 2015-08-28
fix issue 4298: Parent context keeps a reference on destroyed childre…

yoann-antoviaque:
…n through closures

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -230,9 +230,10 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
     }
   });
 
-  $scope.$on('$destroy', function() {
+  var offDestroy = $scope.$on('$destroy', function() {
     scope.$destroy();
   });
+  scope.$on("$destroy", offDestroy);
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
There are some style check failures, see: https://travis-ci.org/angular-ui/bootstrap/builds/77681435

icfantv:
@yoann-antoviaque is there a way to add a functional test for your code changes?  if so, you'll need to do that as well.

yoann-antoviaque:
Well the only way i spotted it was through heap dumps therefore i do not see any way to test it automatically...

Le 28 août 2015 5:31 PM, Adam Gordon <notifications@github.com> a écrit :

@yoann-antoviaque is there a way to add a functional test for your code changes? if so, you'll need to do that as well.

wesleycho:
This is a very good catch - we should try to get this in as soon as possible.

Is it possible for you to write a test for this in the dropdown and typeahead components @yoann-antoviaque ?

wesleycho:
Actually, not sure we can test for this - this is good as is.

PR #4300, opened 2015-08-28
fix(modal): Wait for animation before focus. 

### LINKED ISSUE #4274: Modal focus should wait after animation.
When a modal is opened focus to either the modal or the autofocus element is applied after waiting a $digest cycle using modalRender property. That is not enough as the focus happens before/during the animation in class. This could be avoided by using the promise returned from $animateCss or $animate.

I could submit a PR if you thing that is the right approach.

wawyed:
Previously modalWindow directive focused itself or the autofocus element immediately after triggering the modalInClass animation, which could result in undesired behaviour since browsers try to scroll focused elements into view.
Now modelWindow waits until the animation is finished (if an animation was required) before applying focus.

Closes #4274


wesleycho:
LGTM, but holding off until tests are fixed in `master` before fixing here and merging.

PR #4305, opened 2015-08-29
fix(dropdown): fix display when using with append-to-body

### LINKED ISSUE #4240: Template url dropdown doesn't work once apply dropdown-append-to-body.
[Here is the Plunker] (http://plnkr.co/edit/wBAbEIHpLR9P9hkudCz3?p=preview)

As you can see, once remove `dropdown-append-to-body`, this dropdown menu starts to work.

wesleycho:
- Change to add `dropdown` class and open class to the `<body>` element

BREAKING CHANGE: This differs from the existing behavior in that it no longer will toggle based on the existing `dropdown` directive element, but on the `body` element instead

Plunker with fix in action [here](http://plnkr.co/edit/dKmHtHQnAXCz53W3ExxF?p=preview).

This fixes #4240.


wesleycho:
Greenlit by @Foxandxss

icfantv:
@wesleycho, is that label sufficient to ensure the breaking change makes it into the change log?

wesleycho:
@icfantv: yup

icfantv:
coolness.

PR #4310, opened 2015-08-30
fix(modal): fix allowing promises to be resolved

### LINKED ISSUE #4309: Promise no longer injected in modals
When using the `resolve` options in modals, and the value to inject is a promise, it is not injected properly in the controller.

Basically, you are missing an `else` in the function `getResolvePromises`:

```js
function getResolvePromises(resolves) {
  var promisesArr = [];
  angular.forEach(resolves, function (value) {
    if (angular.isFunction(value) || angular.isArray(value)) {
      promisesArr.push($q.when($injector.invoke(value)));
    } else if (angular.isString(value)) {
      promisesArr.push($q.when($injector.get(value)));
    }
    // If value is a promise, then what ?
  });
  return promisesArr;
}
```
I would replace this with
```js
function getResolvePromises(resolves) {
  var promisesArr = [];
  angular.forEach(resolves, function (value) {
    if (angular.isFunction(value) || angular.isArray(value)) {
      promisesArr.push($q.when($injector.invoke(value)));
    } else if (angular.isString(value)) {
      promisesArr.push($q.when($injector.get(value)));
    } else if (angular.isObject(value) && angular.isFunction(value.then)) {
      promisesArr.push(value);
    } else {
      // We may need to push an empty value here since we rely on the order of the element
      // when doing:
      // ctrlLocals[key] = tplAndVars[resolveIter++];
    }
  });
  return promisesArr;
}
```

wesleycho:
- Wrap values otherwise and push to resolve handler

This fixes #4309


Foxandxss:
:+1:

Issue #4195, opened 2015-08-12
Tooltip bug (version 0.13.3)


AndreasHo:
Hello Guys,

i have a problem with showing the tooltip, wenn it get to much letters it crash.

Example:
"Data..." => the tooltip is good
"Advanced settings..." => the tooltip crash :(

it is over the Button or link and it shimmer.

I hope one of you can fix the problem :)

bye Andreas

wesleycho:
https://github.com/angular-ui/bootstrap/blob/master/CONTRIBUTING.md#you-think-youve-found-a-bug

Please provide a reproduction via Plunker, which would allow us to debug and fix this much quicker.

AndreasHo:
Ok i do it later, or tomorrow :)

AndreasHo:
http://plnkr.co/edit/QKnq5z?p=preview

this is the Plunker, i think the problem only is on 
         tooltip-placement="left"

skarllot:
I'm having trouble when Bootstrap version is 3.3.5. The tooltips are shown on wrong position.
If I use Bootstrap 3.3.5 and Angular-UI/Bootstrap 0.13.0 the tooltips are shown correctly.

http://embed.plnkr.co/SeGe0hanzddW3M9KcPAp/preview

crisbeto:
I also found this to be the case. The position was slightly off in 0.12.x, it got fixed in 0.13.0 and now in 0.13.3 its slightly off again.

dabos-GFI:
Hello,

I've analysed the problem:
- At first the tooltip is created empty and positioned,
- then the content is changed and repositioned.

When the content is changed, the tooltip box size is adapted to the content from the current position (few width because the content is push on the right side) - it can take multiples line because the word-wrap is normal.
When a new position is set, the box size can change because there is available space for the content and the new location becomes wrong.
In order to fix this problem, I've added a reset of the position (top: 0, left: 0) just before computing the offset. I will submit a fix very soon.

Best regards,

Damien

skarllot:
Still buggy. The tooltip appears on wrong position by a fraction of a second and then is shown correctly.

Move the mouse pointer between "R", "U" and "D" back and forth to see the bug.
http://embed.plnkr.co/4kCOfLf5cpTZWYvIq4Oe/preview

wesleycho:
There is an update in `master` that fixes that: http://plnkr.co/edit/YeIpLrqEy6Jlv2kNLnb6?p=info

[...]

PR #4311, opened 2015-08-31
fix(tooltip): Tooltip misplaced near right side

dabos-GFI:
When the tooltip is near the right side of the first parent with relative positionning, the box size used to compute the tooltip position can change after the new location is set.
Then, it is required to reset the position (top: 0, left: 0) before computing the position.

Closes: #4195


wesleycho:
This does not quite solve the problem, as can be seen [here](http://embed.plnkr.co/Lb1bm7WRd7R2bD5pvChh/preview) - note that in smaller viewports it does indeed work fine. It appears almost as if it only needs to be applied conditionally.

dabos-GFI:
I am on it :)

dabos-GFI:
Is there an issue associated with the problem?
My solution is to reset the width and height, and to fix the tooltip width and height width the one use to locate the tooltip.
It seems to work.

wesleycho:
See my example with your changes I linked - it demonstrates a positioning problem that results.

dabos-GFI:
Here is my new positionTooltip function version:

```js
var positionTooltip = function() {
              if (!tooltip) { return; }
              
              // Reset the positioning and box size for correct width and height values.
              tooltip.css({ top: 0, left: 0, width: 'auto', height: 'auto' });

              var ttBox = $position.position(tooltip);
              var ttCss = $position.positionElements(element, tooltip, ttScope.placement, appendToBody);
              ttCss.top += 'px';
              ttCss.left += 'px';
            
              ttCss.width = ttBox.width + 'px';
              ttCss.height = ttBox.height + 'px';

              // Now set the calculated positioning and size.
              tooltip.css(ttCss);
            };
```

For me it fixes your new problem too.

icfantv:
@dabos-GFI we will also need tests for this.

wesleycho:
I think for this sort of issue it's ok for there not to be tests.

wesleycho:
This appears to work, as can be seen [here](http://plnkr.co/edit/du7NBdkdg9Mg9v0suXSp?p=preview) and [here](http://embed.plnkr.co/du7NBdkdg9Mg9v0suXSp/preview) - good work!

dabos-GFI:
Edited example with last version of bootstrap-ui http://plnkr.co/edit/JU1jwv?p=preview

PR #4316, opened 2015-08-31
fix(datepicker): ensure the original target is not in popup

### LINKED ISSUE #4314: Datepicker Popups closing when selecting a month
I've got a strange scenario where clicking any of the buttons on the datepicker will close, which really isn't a problem unless you're trying to scroll through the months (or pick one explicitly). When this happens, the date picker doesn't actually trigger [the cleanup method](https://github.com/angular-ui/bootstrap/blob/master/src/datepicker/datepicker.js#L779-L782) 

I'm fairly confident it's something to do with my app / libraries, as I can't seem to replicate it on a clean project. However, I've been trying to debug _why_ it's happening and am totally up a creek without a paddle. 

One potential issue I came across while stepping through the events is that the [documentClickBind](https://github.com/angular-ui/bootstrap/blob/master/src/datepicker/datepicker.js#L717-L723) function is being triggered and toggling the state, even if the event is inside the popup. It looks like line 718 is checking to see if the clicked element is inside the datepicker target, but in a popup's case it should probably also be a searching for a child of `popupEl` instead, correct? 

Adjusting that line locally to the following fixes our issues, but I'm curious how this isn't also occurring in vanilla apps. Does the following make more sense as it's possible to click a custom element the popup that doesn't close the popup? 

``` 
var documentClickBind = function(event) {
        if (scope.isOpen && !(element[0].contains(event.target) || popupEl[0].contains(event.target))) {
          scope.$apply(function() {
            scope.isOpen = false;
          });
        }
      };
```

An example of one of our datepickers: 

```
<input type="text" class="form-control"
                               datepicker-popup="{{dp.format}}"
                               show-weeks="false"
                               show-button-bar="false"
                               is-open="__dp.reportStart"
                               ng-click="dpOpen($event,'reportStart')"
                               ng-model="report.startDate"
                               required/>
```

wesleycho:
- Ensure that the original event target is not a child of the popup element to avoid edge case where the datepicker closes on any click

This fixes #4314


Foxandxss:
:+1:

yjukaku:
This fix seems to have broken IE support. The `popupEl` seems to sometimes be a non-element node - IE doesn't support the `contains` method on  non-element nodes, so this crashes.

wesleycho:
@yjukaku can you try the change I just pushed up 65814f1 and let me know if that change fixes the issue?

yjukaku:
@wesleycho Nope, doesn't seem to work.

Take a look at this Plunkr: 
http://plnkr.co/edit/IlH0Y4CuhbAkL8mbthmT

This doesn't reference the master branch of the repo, but if you place a breakpoint on the `http://angular-ui.github.io/bootstrap/ui-bootstrap-tpls-0.13.4.js` at line 1851 you can see that the `popupEl` is actually pointing at a `ng-if` comment.  Replacing popupEl with `$popup` doesn't seem to make a difference (`$popup` just points to the `ng-if` comment but wrapped in a jQuery object).

yjukaku:
@wesleycho Any thoughts on this?

Do you want me to send up a pull request that fixes it?

wesleycho:
Sure - the issue to reference is #4411.

PR #4320, opened 2015-09-01
feat(typeahead): add custom popup template support

### LINKED ISSUE #3774: Customize typeahead popup template
I have three typeaheads on my html page. Each of these use AngularJs Bootstrap Typeahead. I can modify the template of the match by using the following: 
<script type="text/ng-template" id="template/typeahead/typeahead-match.html">
    <a class="abc" tabindex="-1" bind-html-unsafe="match.label | typeaheadHighlight:query"></a>
</script>

However I want to modify the outer template for the typeahead i.e. typeahead-popup.html and customize it differently for each of the three different typeaheads on the page. I searched for this but with no success. Any help would be very thankful.

wesleycho:
- Add support for custom popup templates

This resolves #3774.


Foxandxss:
Good. :+1:

PR #4322, opened 2015-09-01
fix(tooltip): switch to use raw DOM event bindings

### LINKED ISSUE #4060: Tooltip is not closed when exiting enclosing <div> containing disabled checkbox
Angular 1.4.3, Angular-UI-Bootstrap 0.13.1, Chrome 44.

* Put a tooltip on a `div`
* Within that `div`, put a disabled checkbox:
  * `<label><input type="checkbox" ng-disabled="true"><span>Label</span></label>`
* Run the app.
* Hover over the checkbox/label `div`... tooltip appears.
* Exit the `div` by moving the mouse to the *left*... tooltip doesn't close.

Note: this is using the standard Bootstrap styling.  You can make the problem less likely to appear by putting a margin (say, 5px) on the left side of the checkbox, but the problem can still occur if you move your mouse fast enough.

Here's a plnkr:  http://plnkr.co/edit/CZpwOw1NXcoI8LQAbnDV

wesleycho:
- Switch to use `addEventListener` and `removeEventListener` to prevent
  jqLite/jQuery bug where the events are swallowed on disabled elements

This fixes #4060.


Foxandxss:
Now that you fixed the tests, it is good to merge.

PR #4328, opened 2015-09-02
refactor(tooltip): optimize positioning routine

RobJacobs:
Wrapped position logic in a timeout to ensure the
digest cycle always completes before positioning
the popup.  Also optimized when the calls to
position the popup occur.


wesleycho:
LGTM

dumbNickname:
Hi,

It looks like after this change popovers/tooltips started to 'flicker' when appearing.

Cheers,
Bartek

wesleycho:
Already fixed in `master` - see https://github.com/angular-ui/bootstrap/pull/4455

PR #4330, opened 2015-09-02
fix(dropdown): fix up arrow nav support

### LINKED ISSUE #4327: keyboard-nav directive for dropdown menu throws error when using up arrow first
When you click on a drop down menu and hit the up arrow the following error is thrown. Uncaught TypeError: Cannot read property 'focus' of undefined

If you hit down first and then up it is not an issue

icfantv:
* add support for when user opens dropdown and then presses the up arrow key.

Fixes #4327


src/dropdown/test/dropdown.spec.js
```diff
@@ -587,6 +587,17 @@ describe('dropdownToggle', function() {
       expect(isFocused(focusEl)).toBe(false);
     });
 
+    it('should focus last list element when up arrow pressed after dropdown toggled', function() {
+      $document.find('body').append(element);
+      clickDropdownToggle();
+      expect(element.hasClass(dropdownConfig.openClass)).toBe(true);
+
+      triggerKeyDown($document, 38);
+      var elems = element.find('ul').eq(0).find('a');
```
wesleycho: This should be `var elems = element.find('ul li').eq(0).find('a');`

icfantv: the selector isn't the problem, rather the up trigger is not causing the anchor tag to become the `document.activeElement`. so the focus check will fail.  why this is the case, i don't know.  but essentially, the up arrow key is doing nothing here.

PR #4334, opened 2015-09-03
chore(tooltip): remove deprecated code

wesleycho:
BREAKING CHANGE:
Remove the unsafe html support from the tooltip


Foxandxss:
Deleting code is the best coding experience. :+1:

PR #4340, opened 2015-09-04
chore(collapse): clean some stale test headers

Foxandxss:
Not a deal breaker, but back in the day we used to "collapse" by hand with a "transition" service we rolled out. That is not true anymore so we can clean those headers.


icfantv:
:+1:

PR #4341, opened 2015-09-04
docs(accordion): fix accordion demo

icfantv:
* fix toggle last panel


wesleycho:
Can you port this change into the `index.html` in `gh-pages`?

icfantv:
sure.

PR #4342, opened 2015-09-04
update tooltip.js

DiHuang1986:
tooltip will be hidden before it is correct positioned. this change will prevent top left corner flicker...


wesleycho:
Is the flicker based on what is currently in `master`?

wesleycho:
Can you link a reproduction that demonstrates the flicker clearly on the most current version?

DiHuang1986:
http://plnkr.co/edit/jnTT6PT9IMaBER7FenZN?p=preview

plunr example. put mouse on button, notice above

wesleycho:
I like this solution - thanks a lot!

PR #4346, opened 2015-09-04
avoid the cannot read isOpen of null bug

btrauma8:
When removing an item that has a tooltip showing from an ng-repeat, you can get an "isOpen of null" bug. This prevents that.


leobudima:
Thanks a lot for this!

PR #4349, opened 2015-09-05
feat(rating): add `aria-valuetext` attribute

### LINKED ISSUE #4347: rating: needs to support aria-valuetext for accessibility
As pointed out in https://github.com/angular-ui/bootstrap/issues/4246#issuecomment-137495947, we should put `aria-valuetext` on the `ng-repeat` in the rating component for accessibility.

wesleycho:
- Add `aria-valuetext` attribute on each star element for accessibility

This resolves #4347.


icfantv:
:+1:

PR #4350, opened 2015-09-05
feat(progressbar): add `aria-labelledby` support

### LINKED ISSUE #4247: [a11y] aria label progressbar compliance
Each progressbar should be labelled and a reference to this label should be provided using aria-labelledby.

wesleycho:
- Add `aria-labelledby` support for accessibility

This resolves #4247.


src/progressbar/test/progressbar.spec.js
```diff
@@ -81,6 +81,7 @@ describe('progressbar directive', function() {
     expect(bar.attr('aria-valuemax')).toBe('100');
     expect(bar.attr('aria-valuenow')).toBe('22');
     expect(bar.attr('aria-valuetext')).toBe('22%');
+    expect(bar.attr('aria-labelledby')).toBe('foo');
```
icfantv: do we want a test case for when the title is omitted?  that the title is `progressbar`.

wesleycho:
Updated as requested

icfantv:
:+1:

PR #4351, opened 2015-09-05
feat(tooltip): hide tooltip when `esc` is hit

### LINKED ISSUE #4248: [a11y] keyboard interaction for tooltip
According to http://www.w3.org/TR/wai-aria-practices/#tabpanel
_Escape key_ shoud dismiss the tooltip.

wesleycho:
- Hide tooltip when `esc` is hit for accessibility

My only question with this PR is whether the listener should be on `$document`.

This resolves #4248.


wesleycho:
Closing, as it looks like this requires more work as per https://github.com/angular-ui/bootstrap/issues/4248#issuecomment-137941787

PR #3044, opened 2014-12-03
fix(modal): fix for conflicts with ngTouch module on mobile devices

jiniguez:
This change attempts to fix issue #2280 

ng-click directive has been removed from modal template and click event moved to the modal-window directive code, so the click event keeps firing and modal window closes when clicking on the backdrop. Not sure if there's a better place to put this event but the fix seems to be working.

PR #4357, opened 2015-09-07
fix(modal): fix for conflicts with ngTouch module on mobile devices

### LINKED ISSUE #2280: $modal + ngTouch input issues on iOS/Android
There is a weird issue with modal windows containing input fields on iOS and Android if ngTouch is used. For some reason, you can only focus on an input field with a long touch gesture. A regular touch gesture won't focus on the field (and thus won't open the virtual keyboard). On desktop browsers the issue does not exist but it also seems to work as it should on Mobile IE (WP8).

Here's a plunkr demonstrating the issue: http://plnkr.co/edit/X95ZBUd6FK2I2R14e3Lr?p=preview

chrisirhc:
Re-rolled #3044 

- Merge pull request #3044 from jiniguez/fix2280
Conflicts:
	src/modal/modal.js
	template/modal/window.html

Fixes #2280


icfantv:
:+1:

PR #4363, opened 2015-09-07
fix(tooltip): tooltip appears on wrong position by a fraction of a second

dabos-GFI:
- Avoid position computation when tooltip is empty
- Force visibility to none when resetting style

Fixes: #4195


busiczvonimir:
Doesnt work :(, http://makeagif.com/SnaCSS

PR #4370, opened 2015-09-08
feat(collapse): use uib- prefix

Foxandxss:
So this is the proposal for collapse prefixing.


Foxandxss:
@wesleycho I need you to check this out, I know you're in vacation, but if you agree, I can do a couple of this before you come back.

@icfantv you too.

wesleycho:
LGTM

Controllers need to be prefixed too for the others - this should be a 100% breaking change I think. Thoughts?

Foxandxss:
You're right @wesleycho, we should prefix controllers too. Will update the instructions.

This PR is not a breaking change tho, it still works as it did before. The breaking change will come when we decide to `rm` all the `.old` files.

We need somehow (for when we release 0.14 with all prefixes) to announce that we have prefixes now and that the users should update their projects ASAP.

wesleycho:
Right, my comment was just before you delved down into the other work :) .

As for the notifying users, perhaps we can mention it on the home page and the README - the deprecation notices will serve enough warning to current users I think.

Foxandxss:
@wesleycho Updated. Now we just need to remove from the comment to the bottom to clean the deprecated modules.

wesleycho:
Alright, this looks better to me - good to go!

PR #4379, opened 2015-09-08
fix(tooltip): add display block to style

RobJacobs:
Under PR #4363 the display block style was
removed but the popover implementation needs
that style as the bootstrap popover starts with
a display of none.


wesleycho:
LGTM

PR #4380, opened 2015-09-08
fix(tooltip): isOpen to work with expressions

RobJacobs:
The is-open attribute should work with expressions
as well as model values


src/tooltip/test/tooltip.spec.js
```diff
@@ -350,6 +350,28 @@ describe('tooltip', function() {
...
+      scope.isOpen = false;
+      elm = $compile(angular.element(
+        '<span tooltip="tooltip text" tooltip-is-open="isOpen === true" >Selector Text</span>'
+      ))(scope);
+      elmScope = elm.scope();
+      tooltipScope = elmScope.$$childTail;
+      scope.$digest();
+    }));
+
+    it( 'should show and hide with the expression', function() {
```
wesleycho: Remove space before `'`

src/tooltip/tooltip.js
```diff
@@ -221,7 +221,7 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position'])
 
               // And show the tooltip.
               ttScope.isOpen = true;
-              if (isOpenExp) {
+              if (isOpenExp && angular.isFunction(isOpenExp.assign)) {
```
wesleycho: I'm a bit confused - does this not work currently?

RobJacobs: It will throw an exception if using an expression instead of a model value, this change will allow for both implementations.

wesleycho: I should have commented, I read the related issue and understood.

PR #4383, opened 2015-09-09
chore(tabs): remove `disabled` support

wesleycho:
- Remove deprecated `disabled` support


icfantv:
Do we want to mark this as a breaking change?  I could go both ways.  Seems like removing deprecated features IS breaking, just that people have advance notice.

wesleycho:
You're right, I'll modify the commit when merging.

icfantv:
:+1:

Issue #4364, opened 2015-09-07
Custom trigger for tooltip doesn't seems to work on 0.13.4


cyrilgandon:
I have used this demo with great success for setting custom trigger on tooltip:

http://plnkr.co/edit/0wEqzz?p=preview

From that post : https://github.com/angular-ui/bootstrap/issues/625#issuecomment-41329842

It work until v0.13.3. Unfortunately, the release 0.13.4 of the 3 september broke the triggers:

http://plnkr.co/edit/KEGMhj2jbF7z6h2Qeu4i?p=preview

Ok never mind, I just read the doc:

"For any non-supported value, the trigger will be used to both show and hide the tooltip. Using the 'none' trigger will disable the internal trigger(s), one can then use the tooltip-is-open attribute exclusively to show and hide the tooltip."

So this becomes:
http://plnkr.co/edit/dm408dUoFKRVQKE9SOTb?p=preview

cyrilgandon:
even though their is a workaround, this should be log in the breaking changes of the releases notes.

https://github.com/angular-ui/bootstrap/blob/master/CHANGELOG.md

dbkup:
I came to the same realization after debugging my code for this problem for the last ~3 hours.

I read that differently, I believe that this is still a bug, i.e. it should work with the custom triggers. The note for 'none' explains how to disable internal triggers, but this should not affect setTriggers which sets custom triggers. If someone from the dev team would come and explain this a bit more.

readelobdill:
+1

readelobdill:
Hey fellers,

The reason this is happening is because of the change in #4060. The event listeners are being set up on the raw DOM elements so instead of ```elem.trigger('customEvent')``` we now need ```elem[0].dispatchEvent(new Event('customEvent))```

@cyrilgandon workaround does work although it may be easier the change the event triggering for some folks here rather than using the ```tooltip-is-open``` attribute in conjunction with ```tooltip-trigger="none"```

icfantv:
Hey guys, can any one of you please check to see if the fix for #4384 fixes this issue?  Currently it's a PR that has not yet been merged.

If you want to check that quickly, I can merge it into master.

dbkup:
I copied over the bootstrap-ui.js from plunk used in #4384 into a fork of @cyrilgandon's plunk http://plnkr.co/edit/aktrfXOUVtfwVdO06DTL?p=preview
[...]

PR #4384, opened 2015-09-09
fix(tooltip): fix binding to multiple triggers

### LINKED ISSUE #3699: [typeahead] dropdown doesn't hide on blur (ios only)

### LINKED ISSUE #4371: $tooltipProvider.setTriggers with multiple events broken with v. 0.13.4
I have the following code in my app:
```
$tooltipProvider.setTriggers({'mouseenter':'mouseleave click'});
```
to make my tooltip close either clicking OR leaving the elements.
Since the last update, none of these two events does trigger a tooltip hide action. If instead I keep just one of the two events ('mouseleave' or 'click'), the tooltip behaves has expected. So I guess that something is broken regarding multiple events triggers support.

Thank you in advance

wesleycho:
- Due to switch to raw `addEventListener`, changed to add and remove
  listeners to event triggers by iteration

[Here](http://plnkr.co/edit/t5Gl9HE6dkWLwz6UHumr?p=preview) is the fix in action.

This fixes #4371


icfantv:
:+1:

icfantv:
I think this may also fix #4364.

PR #4385, opened 2015-09-09
Accordion: don't toggle open class when is-open is undefined

inukshuk:
If `is-open` is not set the accordion group is closed, but the open class will be toggled, because jQuery expects a string or boolean (ignores undefined). See here: https://github.com/jquery/jquery/blob/c9cf250daafe806818da1dd207a88a8e94a4ad16/src/attributes/classes.js#L105

### DIFF
--- a/src/accordion/accordion.js
+++ b/src/accordion/accordion.js
@@ -82,7 +82,7 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
       scope.openClass = attrs.openClass || 'panel-open';
       scope.panelClass = attrs.panelClass;
       scope.$watch('isOpen', function(value) {
-        element.toggleClass(scope.openClass, value);
+        element.toggleClass(scope.openClass, Boolean(value));
         if (value) {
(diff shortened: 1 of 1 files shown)


src/accordion/accordion.js
```diff
@@ -82,7 +82,7 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
       scope.openClass = attrs.openClass || 'panel-open';
       scope.panelClass = attrs.panelClass;
       scope.$watch('isOpen', function(value) {
-        element.toggleClass(scope.openClass, value);
+        element.toggleClass(scope.openClass, Boolean(value));
```
wesleycho: I would change this to `!!value` - the reason is because the `Boolean` constructor can be overridden by the user, whereas using `!!` to force to a Boolean cannot be overridden.

wesleycho:
Scratch that, made the change - https://github.com/angular-ui/bootstrap/commit/c0647485362cb7a0692586c0e7b77fc662a52457

inukshuk:
Thanks!

PR #4389, opened 2015-09-09
feat(accordion): use uib- prefix

Foxandxss:
I had to tweak a bit `uib-accordion-transclude` to temporary support both the new prefix and the deprecated directives. The tweak shall be removed when we drop the deprecated directives.


wesleycho:
LGTM

PR #4394, opened 2015-09-10
chore(accordion): clean some stale code and comments

Foxandxss:
* Remove some `EA` and one `replace: false` (¿?)
* the convention is to name the attributes as `attrs`
* There is like hell of comments, I removed some that were a bit missleading.


wesleycho:
LGTM

PR #4395, opened 2015-09-10
chore(accordion): update docs to match the general style

Foxandxss:
This is my proposal for the final style (for now) of the docs. Things said, the actual overall CSS is a bit crap, but at least this is pretty much what we always had but a bit more normalized and some extra spacing.

image


wesleycho:
LGTM

PR #4399, opened 2015-09-10
chore(accordion): test that any other keyCode wont work

Foxandxss:
It is always good if apart from a working test, we test that something bad can't happen.


wesleycho:
LGTM

PR #4405, opened 2015-09-11
refactor(alert): move dismiss-on-timeout to an attribute

Foxandxss:
It never made sense as a separate directive, it only adds overhead to this little component and it is not needed at all. Also, it would need prefixing and that would make it uglier.

BREAKING CHANGE: Alert's dismiss-on-timeout is now an attribute inside `uib-alert`.


wesleycho:
LGTM - just a note, the dropdown has a similar problem with keyboard-nav being separate.

Foxandxss:
I am still in the A, will reach D soon.

PR #4409, opened 2015-09-12
fix(datepicker): focus input on keyboard selection in popup

stereocilia:
removed the `focusElement()` call from keydown handler so that the popup control is not focused after selecting a date with the keyboard. instead, the focus stays with the input, as called from `dateSelected()`


wesleycho:
Looking at an accessibility example [here](http://oaa-accessibility.org/example/15/), it looks like this agrees with the example, which is desired.

This LGTM.

Issue #4332, opened 2015-09-03
Custom CSS class in typeahead and dropdown when append-to-body is true


ehteshamkafeel:
Sometimes you want to style the drop down according to some particular styles matching to your look and feel or to achieve some very specific functionality. If the `typeahead-append-to-body="true"` you can't style the drop down by being more specific. e.g. you can't do `.my-component .dropdown {}` sure you can do `.dropdown {}` but that would globally apply style to every dropdown in your app. 

It would be really nice if we could specify another attribute like `dropdown-custom-class` to the directive and the class would be added to the main container i.e. `<div class="dropdown my-custom-class">`

wesleycho:
This makes a lot of sense - PRs welcome!

wesleycho:
As per discussion in fa1cdfc, this feature will not be supported. In order to accomplish custom classes currently, the solution espoused is to use a custom typeahead popup template.

datazen:
This actually makes more sense than you would think.   It would be NICE to simply have a class we can change in the ngb-typeahead-window div.   I mean how freaking hard would that be to do?   It seems a lot of users are having the same issues.

PR #4410, opened 2015-09-13
feat(typeahead): Custom CSS class for typeahead's dropdown menu

gorork:
PR for #4332 with the corresponding demo.

Typeahead directive checks if an optional attribute `typeahead-dropdown-custom-class` is specified, and adds it to the dropdown main container. Works with append-to-body attribute.

For example: attribute `typeahead-dropdown-custom-class="demo-class"` will add `demo-class` like this:
`<div class="dropdown-menu demo-class">`.

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -595,6 +595,19 @@ describe('typeahead tests', function() {
       findMatches(element).eq(1).trigger('mouseenter');
       expect(element).toBeOpenWithActive(2, 1);
     });
+
+    it('should add a custom class to popup', function() {
+      var element = prepareInputEl('<div><input ng-model="result" typeahead="item for item in source | filter:$viewValue" typeahead-dropdown-custom-class="demo-class"></div>');
+      changeInputValueTo(element, 'b');
+      expect(findDropDown(element).hasClass('demo-class')).toBeTruthy();
+    });
+
+    it('should add a custom class to popup with custom template', function() {
+      $templateCache.put('custom.html', '<div>foo</div>');
+      var element = prepareInputEl('<div><input ng-model="result" typeahead="item for item in source | filter:$viewValue" typeahead-dropdown-custom-class="demo-class" typeahead-template-url="custom.html"></div>');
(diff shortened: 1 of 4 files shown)


wesleycho:
This makes a lot of sense - can you add a test for this feature to verify that the class is properly being set?

gorork:
@wesleycho Added tests for: basic, with a custom template and with append-to-body.

src/typeahead/test/typeahead.spec.js
```diff
@@ -895,6 +908,12 @@ describe('typeahead tests', function() {
       $timeout.flush();
       expect(dropdown.css('top')).toEqual('500px');
     });
+
+    it('should add a custom class to append-to-body popup', function() {
```
wesleycho: I think this test begs the question, if the user uses `typeahead-append-to-body`, should the component remove this class from the body tag when the element is destroyed? I think it should here.

wesleycho:
Left a comment - I think we will need to garbage collect and remove the custom class from the `body` tag is using `append-to-body`. Can you add a listener via `$scope.$on('$destroy', removeCustomClass)` and add a simple test for this? This should be a quick change, and then we can merge this in.

gorork:
@wesleycho The custom class is being added to a popup element, not to `<body>` tag. So basically, we get `<ul class="dropdown-menu my-class">` element, which can be targeted in CSS like `.dropdown-menu.my-class { ... }`.

wesleycho:
Can you link me to a Plunker with the changes applied so I can see it in action?

gorork:
@wesleycho I've added a demo case for this PR with both "append-to-body" and "typeahead-dropdown-custom-class" attributes. Would it work instead of Plunkr?

wesleycho:
Just would be convenient - unfortunately my schedule is super packed for this week and a half, so it could be difficult for me to investigate in detail to verify, and any bit of help would be much appreciated.

gorork:
@wesleycho Np, here is Plunker: http://plnkr.co/edit/XasKUluu256ztava34oF?p=preview

wesleycho:
I see, thanks a lot, this looks good to me!

PR #4412, opened 2015-09-14
docs(typeahead): add example in demo

ehteshamkafeel:
the example for custom typeahead popup was missing. this PR adds a simple example to demonstrate how the custom popup template can be used specifying the `typeahead-popup-template-url` attribute in `typeahead ` directive.


Foxandxss:
I am going to revise all the docs and demos, but for the time being this can be added. Only one thing else. You need to scope those classes to the typeahead because that will be applied to the entire demo page.

ehteshamkafeel:
@Foxandxss done!

Issue #1055, opened 2013-09-22
How can I close a popover when clicking outside it?


tschofen:
I have multiple popovers on a page that are each opened by a link. I'd like a popover to disappear when I click anywhere outside it's space (just like the tooltip-trigger: focus). 

How would I go about this?

Radnyeechins:
hello everyone hope everyone doing fine 
my questions is ...
i want to close popover if it content input field....
here is demo
http://plnkr.co/edit/sIQevsrtTStDvqvG9ot4?p=preview 
add groups and click on the tabs then rename tab but if we click outside then it doesn't close, it closed iff we click on same tab

rvanbaalen:
@Radnyeechins ask questions on StackOverflow and dont hijack closed posts please.

pabacham:
@rvanbaalen no one knows the answer http://stackoverflow.com/questions/29611355/is-there-any-way-to-close-angular-ui-tooltip-by-clicking-outside-it . Please pay attention to the problem.

rvanbaalen:
Same goes for you @pabacham

Issue #618, opened 2013-07-04
close popover programmatically & only one popover open at the time


fxck:
is there any way to do these? I would like to have a "x" button inside the popover to close it and I would want all other popover to be closed when I open a new one..

jesselpalmer:
+1

smilelizy:
@vchatterji  
Thanks for the hacks there.

$('body').on('click', function (e) {
   $('*[popover]').each(function () {
                //Only do this for all popovers other than the current one that cause this event
                if (!($(this).is(e.target) || $(this).has(e.target).length > 0) 
                     && $(this).siblings('.popover').length !== 0 
                     && $(this).siblings('.popover').has(e.target).length === 0)                  
                {
                     //Remove the popover element from the DOM          
                     $(this).siblings('.popover').remove();
                     //Set the state of the popover in the scope to reflect this          
                     angular.element(this).scope().tt_isOpen = false;
                }
    });
});

In your solution here, I found everything works fine, except that : angular.element(this).scope().tt_isOpen = false; 
does not see to do anything.

when I tracing the scope(), there is no property called tt_isOpen, not sure if it because something has changed over the last year.

Now if your set trigger map as : click -> click,  you open the popover by clicking, and click somewhere else on the screen does close it (removing it from the DOM I believe) , however the state of of popover remain : isOpen = true. This results that next time you'd click twice to make the popover appear again. 

I'm not able to figure how to fix this.
for now I just can think of configuring the trigger as : click -> never 

Hope there is a better way of doing this.

thetaylor82:
Infuriating problem that should have been fixed long ago. On 0.12.1 I'm completely stuck as I have no access to tt_isOpen (or isOpen). Why can't I just trigger a close on the elements?

tmundt:
+1

2015-04-23 11:46 GMT+02:00 thetaylor82 <notifications@github.com>:

tompi:
tt_isOpen was removed, they use an "isOpen" attribute and I believe it's not in the same scope as before, anyway, lot's of trial and error leed us to this code which works unless you use the "append to body" attribute... (call hidePopovers  from your global click-handler))

var hidePopover = function(element) {
    //Set the state of the popover in the scope to reflect this
    var elementScope = angular.element($(element).siblings('.popover')).scope().$parent;
    elementScope.isOpen = false;
    elementScope.$apply();
    //Remove the popover element from the DOM
    $(element).siblings('.popover').remove();
};
[...]

PR #4419, opened 2015-09-15
hide tooltips and popovers when a user clicks away

SystemDisc:
> hide triggers when a click occurs that does not target the popover/tooltip link and does not target elements with the popover/tooltip classes

This solves issue #1055 and helps with similar issues (#618, #590, #749, #751)

Let me know if this is something that's desired and if an option needs to be added for toggling this behavior.


wesleycho:
This needs to be opt-in behind a feature - there are some cases where developers explicitly want to programmatically close it.

wesleycho:
In addition, this needs unit tests as well.

SystemDisc:
Is the feature to programmatically close Popovers and Tooltips already implemented? What are the relevant commits for that?

I can add an attribute for toggling this feature. Any preference as to what that attribute/directive should be called?

Also, what are you guys using for unit tests? I might not know how to do that, but I can look into adding them if you point me in the right direction.

SystemDisc:
Oh, I'm sorry, you're talking about the isOpen property, aren't you?

wesleycho:
Not just that - the programmatic triggers for hiding/showing

SystemDisc:
Using `$tooltipProvider.setTriggers( ... )`?

SystemDisc:
Should I just make this another trigger?

```diff
  var triggerMap = {
    'mouseenter': 'mouseleave',
    'click': 'click',
+   'clickaway': 'clickaway',
    'focus': 'blur',
    'none': ''
  };
```

wesleycho:
That could work, and logic should be added for the binding and unbinding of event listeners depending on if it matches the string `clickaway` - technically it would be a breaking change, but I think it's one that should be minimal in impact.

SystemDisc:
Something like that? b00a550

wesleycho:
Yep, that works - just needs unit tests now :) .

Also for simplicity, you can do `$document[0].body`

SystemDisc:
I just noticed that in `unregisterTriggers`, `toggleTooltipBind` is never unbound. Can I fix that in this same pull request?

wesleycho:
Sure

SystemDisc:
I made some updates, including unbinding toggleTooltipBind in unregisterTriggers, added a unit test, and updated the readme.

src/tooltip/tooltip.js
```diff
@@ -400,13 +401,34 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
               ttScope.popupDelay = !isNaN(delay) ? delay : options.popupDelay;
             }
 
+            // hide tooltips/popovers when click away
+            function bodyHideTooltipBind(e) {
+              var target_elem = angular.element(e.target);
+              if (e.target != element[0] && !target_elem.hasClass('popover') && !target_elem.hasClass('popover-inner') && !target_elem.hasClass('popover-title') && !target_elem.hasClass('popover-content') && !target_elem.hasClass('tooltip') && !target_elem.hasClass('tooltip-inner')) {
```
wesleycho: Can you explain the specific classes you're targeting here with the check? This seems a bit fragile to me.

Or better, can you explain what is being accomplished by the check.

SystemDisc:
The check makes sure you're not clicking on a popover or tooltip itself, allowing the user to select text in the popover or tooltip. These classes are part of bootstrap.

src/tooltip/tooltip.js
SystemDisc: The check makes sure you're not clicking on a popover or tooltip itself, allowing the user to select text in the popover or tooltip. Otherwise, if you clicked on anything that wasn't the popover or tooltip element/link, it would close the popover. These classes are part of bootstrap.

wesleycho: A better check out be to iterate over all of the elements in the stacked map and compare directly with each element - I am wary about tying it too heavily with a hardcoded class.

SystemDisc: Yeah, I agree with that. I hadn't thought much about it. I'll make the change when I have some time again - hopefully tomorrow.

RobJacobs:
I implemented this feature using a slightly different approach.  Rather than rely on the body click exclusively, I also watch for a focus event which will handle tabbing out of the popover.  The approach I use for adding the event handlers to the body are:

      $window.document.body.addEventListener('focus', eventHandler, true);
      $window.document.body.addEventListener('click', eventHandler, true);

With the useCapture enabled, I can simply use 'contains' to see if the triggering element is a child of the popover element like so:

      var eventHandler = function(evt) {
        if (element[0].contains(evt.target)) {
          // element is a child of the popover/tooltip
        }
      };

SystemDisc:
Clicks bubble up to body, so we're not just handling when the body is clicked - we're handling when anything is clicked. Listening for click on the body and ignoring it when the click is on the link or the tooltip/popover is the best way I can think of, because it allows users to click the popover, and select text.

Also, is .contains part of jqlite?

SystemDisc:
Also, this feature is specifically for "clicking away". I'm not sure it makes sense to hide the popover when a key is pressed, even tab to move focus.

wesleycho:
contains is native DOM api

SystemDisc:
Oh, ok. That's good to know. Let me think about this some more, and later today I'll make an update that I hope will satisfy you. Thanks for your continued feedback!

Hrafnkellos:
Can we merge this into the library?

gaui:
:+1:

SystemDisc:
Oops, I forgot about this. Let me tighten up the code a bit today, so everyone can be happy about it being merged in.

SystemDisc:
Everything has been so busy over here. I'll get to it as soon as I can.

Hrafnkellos:
ping

SystemDisc:
pong

I'm out of state, but getting home late tonight. I promise I haven't forgotten.

SystemDisc:
@wesleycho How's that?

SystemDisc:
I'm not sure why the travis-ci build is failing. It succeeds for me, locally, with both Chrome 45.0.2454, and Firefox 43.0.0

SystemDisc:
Yeah, I don't get it. All tests pass on my local machine.

src/tooltip/tooltip.js
```diff
@@ -422,13 +423,67 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
...
+                {
+                  element.unbind(trigger, showTooltipBind);
+                  element.unbind(trigger, toggleTooltipBind);
+                }
               });
               triggers.hide.forEach(function(trigger) {
                 trigger.split(' ').forEach(function(hideTrigger) {
-                  element[0].removeEventListener(hideTrigger, hideTooltipBind);
+                  if (trigger === 'clickaway') {
+                    angular.element($document[0].body).unbind('click', bodyHideTooltipBind);
```
wesleycho: Could do `$document.find('body').unbind(...)`

wesleycho: Hmm actually, may want to do `$document[0].body.removeEventListener(...)`

SystemDisc:
I guess that could save a millisecond of processing....

Updated.

src/tooltip/tooltip.js
```diff
@@ -422,13 +423,67 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
...
+                // hide by default
+                var shouldHide = true;
+
+                // loop through all ttScopes in the stackedMap (openedTooltips)
+                var ottScopes = openedTooltips.keys();
+                for (var index in ottScopes) {
+                  var ottScope = ottScopes[index];
+
+                  // see if the target element's scope is ottScope or is a descendant of ottScope (meaning it's part of the tooltip/popover element)
+                  var curScope = target_elem.scope();
```
RobJacobs: We can't access element scope like this anymore, see this [link](https://docs.angularjs.org/guide/production#!)

src/tooltip/tooltip.js
```diff
@@ -422,13 +423,67 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
...
+              var target_elem = angular.element(e.target);
+
+              // make sure the tooltip/popover link was not the element clicked (keep in mind, the click event on the link toggles show/hide already)
+              if (e.target != element[0]) {
+                // hide by default
+                var shouldHide = true;
+
+                // loop through all ttScopes in the stackedMap (openedTooltips)
+                var ottScopes = openedTooltips.keys();
+                for (var index in ottScopes) {
```
RobJacobs: Should consider using angular.forEach as that implements the hasOwnProperty check, [link](https://docs.angularjs.org/api/ng/function/angular.forEach#!)

src/tooltip/tooltip.js
SystemDisc: @wesleycho You recommended that I use the stacked map to figure out if the user is clicking inside a tooltip/popover, instead of using the bootstrap classes for popover/tooltip. Is there a different way you were thinking of doing this? It appears to me that the stacked map is just a list of scopes and the element's close method. Can I add the element itself to the stacked map?

src/tooltip/tooltip.js
SystemDisc: Thanks

SystemDisc: Can I just check hasOwnProperty? I won't be able to break the loop if I use angular.forEach.

SystemDisc:
@RobJacobs How's this?

SystemDisc:
Oops, I made a mistake. `element` is a reference to the element with the tooltip/popover directive, not the tooltip/popover element itself. I'm fixing that now.

SystemDisc:
Fixed

SystemDisc:
@RobJacobs Just so you know, for your code, you're checking the wrong element.

```diff
  var eventHandler = function(evt) {
-    if (element[0].contains(evt.target)) {
+    if (element[0].contains(evt.target) || tooltip[0].contains(evt.target)) {
      // element is a child of the popover/tooltip
    }
  };
```

`element` is a reference to the element that contains the directive. `tooltip` is a reference to the actual tooltip/popover element. `tooltip` may also be undefined for a time, so you may also want to check `tooltip == null`.

RobJacobs:
@SystemDisc That was a snippet from a different directive that was meant to demonstrate a different approach, but yes you are correct.

SystemDisc:
That makes sense. Do you approve of my updates? I think I've addressed your concerns.

RobJacobs:
Here is what I was thinking for the angular.forEach:

    angular.forEach(openedTooltips.keys(), function(ottScope) {
       var openedTooltip = openedTooltips.get(ottScope).value;
       ...
    });

But that's not a deal breaker for me.  Some linters complain when using a for...in loop without the hasOwnProperty check so I just use the angular.forEach.

Looks good, thanks for the PR...

SystemDisc:
Yeah, my only issue with angular.forEach is that I can't break out of it. Make sure to let me know if it needs any other updates before being merged in.

RobJacobs:
Ah yes, good point.

RobJacobs:
@wesleycho @SystemDisc Sorry for bringing this up so late in the game and I apologize if this has already been hashed out.

I noticed we have the same functionality in the popover directive with the auto-close attribute with the outsideClick option.  Should we call this trigger 'outsideClick' rather than 'clickaway' to keep the API's consistent?

The tooltip controller has a reference to the tooltip element (tooltip variable) so I'm wondering why that is not used in the bodyHideTooltipBind function?  Could we do:

    if (e.target != element[0] && e.target != tooltip[0] && !tooltip[0].contains(e.target)) {
      hideTooltipBind();
    }

and not add the element to the stackedMap?

We have run into problems with the if(ttScope,isOpen) check as ttScope can be set to null.  The hide() function has:

    if(!ttScope) {
      return;
    }

So I'm wondering if we can rely on that instead and change it to:

    if (!ttScope || !ttScope.isOpen) {
      return;
    }

SystemDisc:
`outsideClick` appears to be a directive for dropdowns, not popovers. Can you point me to the file/line you're referencing? Should we make an `outsideClick` directive for tooltips/popovers, instead of making it a trigger? I made it a trigger because people may not use click as a trigger at all, making `outsideClick` moot (if the popover closes on the blur trigger, the user can't click someone else while it's open).

We are using the stacked map because we want to check if the user is clicking in *any* popover, not just this current one. I figured if a user is clicking on a popover, that none of the popovers should be closed. Perhaps we should close a popover if anything other that the current popover is clicked, including other popovers. (?)

I agree with the ttScope thing.

SystemDisc:
To clarify, the way this is designed, a user would be able to click multiple popover elements (the element with the popover directive), which would open multiple popovers, and they would close if a user clicked anywhere that was not on any of the popover elements or the actual popovers.

RobJacobs:
@SystemDisc  My bad on the outsideClick, I meant the dropdown auto-close attribute where outsideClick is the option.  I just figured if we already have a behavior like that in the library, let's name it consistently.  So instead of naming the trigger clickaway, how about outsideClick instead?

BobbieBarker:
Need to rebase and squash your commit history too.

src/tooltip/tooltip.js
```diff
@@ -422,13 +426,56 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
...
+              }
+            }
+
             var unregisterTriggers = function() {
               triggers.show.forEach(function(trigger) {
-                element.unbind(trigger, showTooltipBind);
+                if (trigger === 'clickaway') {
+                  element[0].removeEventListener('click', toggleTooltipBind);
+                }
+                else
```
BobbieBarker: These could be rewritten with out the else

src/tooltip/tooltip.js
```diff
@@ -422,13 +426,56 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
...
+                  element[0].removeEventListener(trigger, toggleTooltipBind);
+                }
               });
               triggers.hide.forEach(function(trigger) {
                 trigger.split(' ').forEach(function(hideTrigger) {
-                  element[0].removeEventListener(hideTrigger, hideTooltipBind);
+                  if (trigger === 'clickaway') {
+                    $document[0].body.removeEventListener('click', bodyHideTooltipBind);
+                  }
+                  else
```
BobbieBarker: This could also be rewritten with out the else.

src/tooltip/tooltip.js
```diff
@@ -443,7 +490,14 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
                 triggers.show.forEach(function(trigger, idx) {
                   // Using raw addEventListener due to jqLite/jQuery bug - #4060
                   if (trigger === triggers.hide[idx]) {
-                    element[0].addEventListener(trigger, toggleTooltipBind);
+                    if (trigger === 'clickaway') {
+                      element[0].addEventListener('click', toggleTooltipBind);
+                      $document[0].body.addEventListener('click', bodyHideTooltipBind);
+                    }
+                    else
```
BobbieBarker: again.

src/tooltip/tooltip.js
```diff
@@ -443,7 +490,14 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
                 triggers.show.forEach(function(trigger, idx) {
                   // Using raw addEventListener due to jqLite/jQuery bug - #4060
                   if (trigger === triggers.hide[idx]) {
-                    element[0].addEventListener(trigger, toggleTooltipBind);
+                    if (trigger === 'clickaway') {
```
BobbieBarker: I don't think this should be nested, this would read clearer as an additional else if?

BobbieBarker:
I had some minor style comments, but other than that it lgtm.

src/tooltip/tooltip.js
```diff
@@ -422,13 +426,56 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
               }
             }
 
+            // hide tooltips/popovers for clickaway trigger
+            function bodyHideTooltipBind(e) {
+              if (ttScope == null || !ttScope.isOpen) {
+                return;
+              }
+              // make sure the tooltip/popover link was not the element clicked (keep in mind, the click event on the link toggles show/hide already)
+              if (e.target != element[0]) {
```
wesleycho: This should be a strong equality check `!==`

wesleycho:
This needs to have all of the commits squashed - we can handle the rest on merging.

src/tooltip/tooltip.js
SystemDisc: Can you elaborate?

src/tooltip/tooltip.js
SystemDisc: Please elaborate.

src/tooltip/tooltip.js
SystemDisc: ditto

src/tooltip/tooltip.js
SystemDisc: I disagree. Please elaborate.

src/tooltip/tooltip.js
SystemDisc: Agreed

SystemDisc:
Not to sound like a noob, but how would I go about squashing these commits, and what should the new commit message be? I would just Google how to squash commits, but I figured I may need some information directly related to this. I noticed there's a pre-commit hook for the structure of commit messages. Could you explain that to me?

wesleycho:
`git rebase -i HEAD~[number of commits without brackets]` - you can squash them all with the `fix` or `f` text when squashing commits into one.

SystemDisc:
Okay, what should the final commit message be? What do I need to include?

Also, should I change `clickaway` to `outsideClick` or `outsideclick` or keep it `clickaway`?

wesleycho:
Don't worry about the message - we can fix it up.

SystemDisc:
Let me know if any other changes need to be made. I'm interested to hear more from @BobbieBarker about how and why I should remove a couple of the `else`s in the code.

abelmokadem:
Question, how will this work with multiple angular apps on 1 page? Will one app close all other popovers from the other app?

SystemDisc:
No, it should not affect popovers in other apps on the page.

alanirudh:
popoveraction

@SystemDisc Question: Does this also work for the above workflow, like when you open a popover will it close the already opened popover? Or will it require more logic to get this to work?

wesleycho:
Should be fine - there is a `contains` check to check whether the event target is coming from a child of the popover in question.

SystemDisc:
It would keep popovers open when you click on another popover. It closes
them when a click happens on something that isn't a popover. See the gif
above for an example of the current functionality.

It would be easy to change the functionality so that only one
"outsideClick" popover can be open at a time. Is this more desirable?

alanirudh:
Thanks for the quick response. The reason I have asked about it is because if we have multiple rows with popover menu opening on left or the right. Then previously opened popover will be covered by the newly opened popover. It's nice that it doesn't close the popover when we click with in the popover though.

We have something called autoclose in dropdown module which closes any opened dropdown when opening new dropdown. So, to be consistent with outsideClick action of dropdown we can have this behavior which closes when we click outside.

SystemDisc:
Good point. I'll make the behavior of popover's outsideClick consistent
with the behavior of dropdown's outsideClick.

SystemDisc:
I'll try to get to it this weekend.

SystemDisc:
tooltip/popover `outsideClick` should now behave more closely to dropdown `outsideClick`

SystemDisc:
@wesleycho This PR no longer needs to be squashed, and I think it works the way everyone wanted it to.

src/tooltip/tooltip.js
```diff
@@ -174,7 +175,9 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
             // TODO add ability to start tooltip opened
             ttScope.isOpen = false;
             openedTooltips.add(ttScope, {
-              close: hide
+              close: hide,
+              triggerElement: element,
+              element: null
```
RobJacobs: I don't think we need the 'triggerElement' and 'element' to be in the modal stack since you are no longer ignoring close events from other tooltip elements?

src/tooltip/tooltip.js
```diff
@@ -287,6 +290,7 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
                   element.after(tooltip);
                 }
               });
+              openedTooltips.get(ttScope).value.element = tooltip;
```
RobJacobs: Same observation as noted above.

src/tooltip/tooltip.js
```diff
@@ -442,7 +470,11 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
               if (triggers.show !== 'none') {
                 triggers.show.forEach(function(trigger, idx) {
                   // Using raw addEventListener due to jqLite/jQuery bug - #4060
-                  if (trigger === triggers.hide[idx]) {
+                  if (trigger === 'outsideClick') {
+                    element[0].addEventListener('click', toggleTooltipBind);
+                    $document[0].body.addEventListener('click', bodyHideTooltipBind);
```
alanirudh: Shouldn't this be added on $document instead of body? Also, don't we need to close the tooltip on contextmenu event?

src/tooltip/tooltip.js
SystemDisc: No, we don't, but I figured it might be useful in the future. I can remove it.

src/tooltip/tooltip.js
SystemDisc: I add the event listener on body because click events bubble up, and the topmost element that's clickable (or that we care about) is the body.

src/tooltip/tooltip.js
SystemDisc: @RobJacobs

RobJacobs: My philosophy is don't add it unless it's needed.  It just adds unnecessary overhead to the openedTooltips collection.

SystemDisc: @RobJacobs Pushing now

SystemDisc:
@wesleycho I've addressed some concerns @RobJacobs brought up. Let me know if you guys need anything else before merging.

wesleycho:
Can you clean up the eslint errors?

src/tooltip/tooltip.js
wesleycho: I think it's better to have the listener on $document, for consistency with the rest of the library.

thegreenpizza:
@wesleycho I submitted a new PR with those errors fixed #4871 . Normally I wouldn't do this, but I'm interested in getting this change in asap. Travis build is running now.

wesleycho:
Actually, did the cleanup - thanks for sticking with this PR & putting in the good work @SystemDisc!

rand0m86:
@wesleycho, please, can you make a release for bower version that includes this update? It will be awesome.

icfantv:
The release will be available when we release 1.0.  We don't have a date for this release yet but want to get it out before the end of the year.  We still have some things we need to finish.

In the meantime, please feel free to pull down the latest code and do an `npm install` followed by `grunt` to get the latest SNAPSHOT build.  The files will be in the `dist` directory.

h6784rfg6:
Hello, I also am interested in this feature. I installed the npm version: npm install angular-ui-bootstrap and linked the ui-bootstrap-tpls.js (which was not in the disc folder). I could't get the popover open providing the popover-trigger="outsideClick" option. Could you please provide some plunker with the example usage of this feature?

wesleycho:
@h6784rfg6 currently this is not released - if one wants to use this feature, one has to consume `master` directly. Otherwise, keep an eye out for #4838 - once that issue is complete, we will release 1.0 shortly after

PR #4425, opened 2015-09-16
reset locale format when locale changes

tobigun:
When angular-dynamic-locale is used, the locale can change over time.
The datetime, month and day formats should be reset whenever the locale changes. Otherwise the parser will fail parse dates after locale switches.


wesleycho:
This LGTM - it sounds like a sensible and performant way to update the parsers.

This needs a test though.

PR #4426, opened 2015-09-16
add support for value literals

tobigun:
Add support for angular date value literals ('..'), e.g. used in the spanish longdate:
"longDate": "d 'de' MMMM 'de' y"


Foxandxss:
We would like to see some tests please.

wesleycho:
What are your thoughts on this @Foxandxss ? Aside from code style in some areas and the need to replace with `uibDateParser` in the tests, this LGTM, but I'd like to hear your thoughts on the API signature here.

PR #4418, opened 2015-09-15
Fix tooltip continuation

bretkikehara:
Deleted branch instead of doing a rebase #4408 branch, so I can't reopen that PR. Sorry! >_<

I have rebase the code.

PR #4428, opened 2015-09-17
fix(popover): avoid flash of unstyled content

### LINKED ISSUE #4427: Popover in 0.13.4 shows flash of unstyled content
Regression in 0.13.4 from 0.13.3 generates a FOUC on the popover directive.  It's very apparent on any page that has a slow digest cycle.  Making sure to hide element before the `$timeout` then promptly show works around the issue.

See plunkr to reproduce problem: http://plnkr.co/edit/XKtxNa67aaT2TsS7YcNV?p=preview

angular version: 1.4.5
angular-ui bootstrap version: 0.13.4
browser: latest stable Chrome

Workaround at https://github.com/bikegriffith/bootstrap/commit/af9d18df666a66a7c40c5409f89e7bdab073ef19 (I can pull request it if desired)

bikegriffith:
Fixes #4427

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -149,8 +149,10 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
               if (!tooltip || !tooltip.html()) { return; }
 
               if (!positionTimeout) {
+                tooltip.hide(); // Hide to avoid FOUC until $timeout completes
                 positionTimeout = $timeout(function() {
                   // Reset the positioning and box size for correct width and height values.
+                  tooltip.show();
                   tooltip.css({ top: 0, left: 0, width: 'auto', height: 'auto', visibility: 'hidden' });
 
                   var ttBox = $position.position(tooltip);


RobJacobs:
This has been fixed under PR #4363 and PR #4379.  This updated [plunk](http://plnkr.co/edit/JERhdfUrRrOrzhHtkAYh?p=preview) shows the behavior with the latest code from master.

src/tooltip/tooltip.js
```diff
@@ -149,8 +149,10 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
               if (!tooltip || !tooltip.html()) { return; }
 
               if (!positionTimeout) {
+                tooltip.hide(); // Hide to avoid FOUC until $timeout completes
```
bikegriffith: `positionTooltip` is called frequently in some cases for an already open tooltip, so this could cause flickering.

bikegriffith:
Thanks @RobJacobs. Just realizing this while auditing the code again and testing against a hacked version of 0.13.4 in my app.  Will shut this down.

RobJacobs:
@bikegriffith We're doing some more work to tighten up when the positioning method gets invoked and how often.  Next significant change is PR #4418.

PR #4429, opened 2015-09-17
Infinite digest loop when using filters inside tooltip expression properties (like popover-html)

forceuser:
There is an error when using filters inside expression (as example ```popover-html="'some html' | to_trusted"```)  it caused infinite digest loop (```Error message: [$rootScope:infdig] 10 $digest() iterations reached. Aborting!```). 
The reason is:
scope updates > triggers tooltipLinkedScope 'contentExp()' watcher > scope.$eval(attrs[type]) can again trigger update of the scope

I changed contentExp to evaluate using ```$parse``` instead of ```scope.$eval``` and moved it to ```scope``` watcher

Now contentExp() returns already evaluated value and can't cause infinite loop


wesleycho:
@RobJacobs what are your thoughts on this? This sounds like a good change IMO, since evaluating the attribute expression each time is inefficient.

RobJacobs:
@wesleycho @forceuser I like where this is going but have a suggestion.  Lets see if we can move the watch to the createTooltip function and replace the watch on line 229 for contentExp(), so that it would become:

    if (options.useContentExp) {
      tooltipLinkedScope.$watch(contentExpGetter, function(val) {
        contentExpVal = val;
        if (!val && ttScope.isOpen) {
          hide();
         }
        });
    ....
    }

That way, when the tooltipLinkedScope gets destroyed, so does the watch and that watch can serve both purposes (hide tooltip when template is empty).

forceuser:
but the parent of tooltipLinkedScope is ttScope and its isolated, so contentExpGetter may not return any value in tooltipLinkedScope.$watch

PR #4431, opened 2015-09-17
fix(tooltip): check if ttScope exists

### LINKED ISSUE #3347: Tooltip fails when closing on an element that's removed
I'm using the tooltip on a delete button, which delete a row in a table. The button has a mouse enter/leave effect. When I click the button, which is then deleted with an animation, the following error is logged:
Uncaught TypeError: Cannot read property 'isOpen' of null

The following seems to fix the problem:
            function toggleTooltipBind () {
              if (!ttScope) return; // Check that the element still exists. 
              if ( ! ttScope.isOpen ) {
                showTooltipBind();
              } else {
                hideTooltipBind();
              }
            }

wesleycho:
- Check for ttScope existence to avoid situations when null

This fixes #3347.


pkozlowski-opensource:
Is there any way of testing this? Do we know what is the condition under which this error is triggered?

wesleycho:
The condition is when the tooltip scope is destroyed (when not being displayed), and then a change is made to each of the attribute values on the directive - theoretically this should be testable in testing to see if an exception is thrown or not.

PR #4435, opened 2015-09-17
Corrected the meridian threshold when determining if toggling should be allowed in the timepicker.

kmccaw:
Edited the noToggleMeridian method in the timepicker directive to correctly compare against 12 (instead of 13) when determining if the resulting time should be compared against the minimum or the maximum. If the selected time was after noon, the resulting time after toggling would be in the morning of the same day, and should be compared against the minimum and not the maximum.


wesleycho:
I took a harder look at this, and this LGTM. Thanks for this catch!

PR #4440, opened 2015-09-19
Added angular object in define statement.

wsalazar:
When I added this code to my library I received an error in my console warning me that angular was not defined.


Foxandxss:
The problem is in your application. I guess you're using require.js or something like that. To be honest I am not that familiar with require.js, but that change you made is not something that should be made here.

I am sorry for not be able to help you more with your setup, but that is something you need to solve within your app / library and not with ui-bootstrap.

PR #4446, opened 2015-09-21
chore(demo): update collapse to use prefix

Foxandxss:
There was a deprecation warning all day long and I thought that it was for a directive I haven't update yet. Ended being the demo file :P


wesleycho:
LGTM

PR #4449, opened 2015-09-23
feat(tabs): use `uib-` prefix

icfantv:


Foxandxss:
Few things. No backquotes on title.

The demo html needs to be updated to use the prefix and the readme if mentions (and it does in this case) need to be updated with prefixes as well.

The main demo app has tabs so that needs to be fixed (I can do that in a separate commit, no problem).

My biggest problem here is the `<tab-heading>`. We have two directives with this type of feature (accordion and this) and they both behave differently.

In Accordion we have our own directive for that and for this one, we have sort of a hack. My point in here is that in here we don't have `<tab-heading>` prefixed.

So I think that we need to prefix that as well but also supporting it without prefix for the deprecated one. Thoughts on this @wesleycho ?

icfantv:
@Foxandxss, ok.  the title can be fixed on the merge.  demo has been updated.  `tab-heading` was updated to contain the `uib-prefix` and all relevant tests were updated as well.  per our discussion, it **NOT** made into a directive at this time.

src/tabs/tabs.js
```diff
@@ -274,12 +274,154 @@ angular.module('ui.bootstrap.tabs', [])
...
+        });
+
+        //We need to transclude later, once the content container is ready.
+        //when this link happens, we're inside a tab heading.
+        scope.$transcludeFn = transclude;
+      }
+    };
+  }])
+
+        .directive('tabHeadingTransclude', ['$log', '$tabSuppressWarning', function($log, $tabSuppressWarning) {
```
wesleycho: Fix indentation

Foxandxss: Yup, forgot to mention that too.

icfantv: done.  there were a couple of lines that needed indent fixes.

icfantv:
I think my comment got lost.  Fixed indentation in a couple of spots.

wesleycho:
This LGTM - @Foxandxss ?

icfantv:
Hang on, there's some duped lines in the second test.  I'll remove and repush.

icfantv:
The template was getting compiled and digested twice which was resulting in double the number of `$log` warning messages than expected.  This is fixed.

Foxandxss:
Alrighty merged. Changed $tabSuppress to $tabsSuppress and there was a missing prefix on the demo.

PR #4464, opened 2015-09-24
style(buttons): change bind to on

Foxandxss:


Foxandxss:
Update it to match `.on` which is what is used in other places.

wesleycho:
This bothered me in a couple of areas in the code - LGTM

Issue #4452, opened 2015-09-23
ProgressController recalculatePercentage bug


JDeuker:
AngularJS v1.4.5
angular-ui-bootstrap Version: 0.13.4 

What to do:
See plnkr demo: http://plnkr.co/edit/zf2jb4?p=preview
Go inside the first textbox (100) and use Backspace to reduce it to 10.

The Problem:
The Bars are recalculated wrong.

The Origin:
First the watcher for "max" is proceeded for both bars.
Next the watcher for "value" is proceeded for only the first bar.
While calculating the totalPercantage the second bar has still the wrong percentage. See:
         var totalPercentage = self.bars.reduce(function (total, bar) {
            return total + bar.percent;
         }, 0);

The Solution:
Recalculate the bar percentage:
         var totalPercentage = self.bars.reduce(function (total, bar) {
            bar.percent = +(100 * bar.value / bar.max).toFixed(2); //Bugfix
            return total + bar.percent;
         }, 0);

wesleycho:
This sounds reasonable, although I'm a bit worried as to why the unit test does not capture this situation - can you file a PR with this fix?

JDeuker:
There is also a problem with removing an item, because recalculatePercentage is still calculating with the removed one including.

I found a solution with recalculating after removing:
   this.removeBar = function (bar) {
      this.bars.splice(this.bars.indexOf(bar), 1);
      this.bars.forEach(function (bar) {   //Bugfix
         bar.recalculatePercentage();
      });
   };

I updated the Plunker: http://plnkr.co/edit/zf2jb4?p=preview

PR #4471, opened 2015-09-25
fix(progress): re-calculate percentage when a bar is removed or altered

JDeuker:
See: https://github.com/angular-ui/bootstrap/issues/4452


pkozlowski-opensource:
@JDeuker thnx for the PR. It will need a test(s) before merging, though. Also, could you please change the commit message to follow our convention. Ex.: `fix(progress): re-calculate percentage when a bar is removed` or something along those lines.

PR #4475, opened 2015-09-25
Added typeaheadFocusLast attribute

Squazz:
Made it possible to focus the last element instead of just the first or nothing


wesleycho:
This seems like a poor UX - can you justify why one might want this behavior over selecting the first?

pkozlowski-opensource:
Well, it is not only about poor UX - we simply can't keep adding flags to the typeahead like this... What if someone else is coming and saying that they need to focus a middle element? Or a second one? It just doesn't scale.

If we believe that we need to be able to focus an arbitrary match (which I don't think we should enable as part of the standard lib) we should do it in a generic manner.

I'm going to close this PR as I don't think we want this functionality as part of the core (not to mention the fact that this doesn't have tests, documentation, doesn't follow our commit guidelines etc.)

@Squazz thnx for your work, but we wan't take this patch as-is due to the explained reasons. I would be still curious to hear more about your use-case though.

Squazz:
Our usecase is the following:
We use the typeahead for a general searchfield on our site.
When people search for something with more than x (lets say 5) we add a item to the list of suggested items from the search. This extra item is a "see more results" entry.
We want this last entry to have the focus, so that if the user presses enter or his/her keyboard, this they will select the "See more results" entry in the list.

I hope this gives some kind of clarification for what it is that I'm trying to do.

As we found us selves doing this multiple places we figured out that we could as well just add it to the library which we have done internally. 

I apologize for the missing tests, documentation etc.,

Squazz:
@wesleycho  & @pkozlowski-opensource
I guess that you don't seem to agree that our case is a realistic case for productions?

icfantv:
@Squazz, apologies for not getting back to you.  What you describe is a pretty specialized use case.  As I'm sure you're aware, it is nearly impossible to capture _every_ user request and turn it into a library feature.  We need to have a core set of API options and then see what additional features, if any, users want most.

As @pkozlowski-opensource pointed out above, we have to be very careful about continually adding options to the directives or it just gets crazy - see the datepicker if you want an example.

If this is a feature you need to have, you're going to need to roll your own or use ours as a launchpad.

PR #4479, opened 2015-09-25
fix(carousel): Better a11y with sr-only text and markup structure

### LINKED ISSUE #4478: Default carousel template is not accessible
* Angular - 1.2.28
* ui-bootstrap - 0.11.2

Here is a plunker of a failing scenario: http://plnkr.co/edit/pouPdsf3ugooUiyHMPTV?p=preview

The default template is not accessible via keyboard navigation nor with screen reader assistance. While tabbing through the document you should be able to get keyboard focus on the carousel controls. While using assistive technology, you should also be able to tab-stop on the carousel controls. Neither of these are working as-is.

I understand that newer versions allow for custom templates but there are very minimal changes to the default template that would allow all versions to be much more accessible.

eduncan26:
fixes #4478

### DIFF
--- a/template/carousel/carousel.html
+++ b/template/carousel/carousel.html
@@ -1,8 +1,16 @@
 <div ng-mouseenter="pause()" ng-mouseleave="play()" class="carousel" ng-swipe-right="prev()" ng-swipe-left="next()">
-    <ol class="carousel-indicators" ng-show="slides.length > 1">
-        <li ng-repeat="slide in slides | orderBy:indexOfSlide track by $index" ng-class="{active: isActive(slide)}" ng-click="select(slide)"></li>
-    </ol>
-    <div class="carousel-inner" ng-transclude></div>
-    <a class="left carousel-control" ng-click="prev()" ng-show="slides.length > 1"><span class="glyphicon glyphicon-chevron-left"></span></a>
-    <a class="right carousel-control" ng-click="next()" ng-show="slides.length > 1"><span class="glyphicon glyphicon-chevron-right"></span></a>
+  <div class="carousel-inner" ng-transclude></div>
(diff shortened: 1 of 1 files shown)


template/carousel/carousel.html
```diff
@@ -1,8 +1,16 @@
 <div ng-mouseenter="pause()" ng-mouseleave="play()" class="carousel" ng-swipe-right="prev()" ng-swipe-left="next()">
-    <ol class="carousel-indicators" ng-show="slides.length > 1">
-        <li ng-repeat="slide in slides | orderBy:indexOfSlide track by $index" ng-class="{active: isActive(slide)}" ng-click="select(slide)"></li>
-    </ol>
-    <div class="carousel-inner" ng-transclude></div>
-    <a class="left carousel-control" ng-click="prev()" ng-show="slides.length > 1"><span class="glyphicon glyphicon-chevron-left"></span></a>
-    <a class="right carousel-control" ng-click="next()" ng-show="slides.length > 1"><span class="glyphicon glyphicon-chevron-right"></span></a>
-</div>
+  <div class="carousel-inner" ng-transclude></div>
+  <a role="button" href="#" class="left carousel-control" ng-click="prev()" ng-show="slides.length > 1">
```
wesleycho: This needs to be an empty `href` due to some edge cases

template/carousel/carousel.html
```diff
@@ -1,8 +1,16 @@
...
+    <span aria-hidden="true" class="glyphicon glyphicon-chevron-left"></span>
+    <span class="sr-only">previous</span>
+  </a>
+  <a role="button" href="#" class="right carousel-control" ng-click="next()" ng-show="slides.length > 1">
```
wesleycho: Ditto

wesleycho:
Other than those two small issues that we can fix on merge, I'm fine with this.

eduncan26:
Sweet!

PR #4487, opened 2015-09-28
fix(datepicker): prevent default click behavior on pop-up

nwhittaker:
In the case where the datepicker pop-up is in embedded within a "label" tag, a click on the pop-up (despite propagation being stopped) causes the label's default behavior of triggering a click event on its associated "input" field. This is problematic if clicks on the "input" field are configured to toggle the visibility of the datepicker pop-up -- specifically, clicking anywhere in the datepicker pop-up can cause it to close prematurely.


wesleycho:
Do you know exactly why the event is still propagating to the label? This sounds like contrary behavior to what is intended.

wesleycho:
Closing due to lack of information, and that this does not look like a valid fix for the described issue.

If this is still an issue, please open an issue first with valid reproduction on Plunker.

billccn:
Hi, I have encountered the same problem and make a test case for this:

http://plnkr.co/edit/Bsnh1WUfmx4nuDZQdvTf?p=preview

The html looks like this at runtime:

```html
<label>
    <input ... >
    <div uib-datepicker-popup-wrap ... >
</label>
```

So any `click` event in the `div` will naturally propagate to the `label`.

The `<ul>` element inside the `div` has the attribute `ng-click="$event.stopPropagation()"`, but this is insufficient to stop the event because `ng-click` does not seem to handle events not originating from the element it is marked on.

PR #4488, opened 2015-09-28
Add Dropdown AppendTo functionality

### LINKED ISSUE #4467: Add "appendTo" functionality to dropdown
I noticed there were a couple of PRs for the modal component that try to do something similar, this would be for the dropdown component.

I have an interesting use case where append-to-body isn't quite what we need, but not having it suffers from the issues that append-to-body was made for.

This was actually mentioned in #4240 (second comment) which is demonstrated by this plunkr: http://plnkr.co/edit/1EuLWcCSv47tCi7i5H6N?p=preview

Like the plunkr, I have a static body element that contains a scrollable div, which is where the dropdown toggles are.  If the dropdown is opened (and appended to body), and the user scrolls, the dropdown body stays in place (since it's appended to body, and the body isn't the element that's scrolling).

Appending inline (default behavior without append-to-body) doesn't work in this case, as the element that the dropdowns will appear in may have hiding overflow, so the dropdowns will appear cut off.

I'd like to look into adding an "appendTo" (dropdown-append-to) feature to dropdowns so that I can pass in an element that the dropdown body would be appended to.

Thoughts on this?

jaydiablo:
This fixes #4467.

In a UI that we're building we ran into an issue where the dropdowns had to be attached to body (because their containing element had some sort of hidden overflow, like the bootstrap responsive tables) but the body in this app is a static "window", as there's a div element that the main content is scrolled in (this is because there's a snap.js drawer used).  In this situation, if the dropdowns are appended to the body element, when the user scrolls after opening a dropdown, the dropdown menu would stay in a fixed position above the scrolling content beneath.

I've added a "dropdown-append-to" attribute that accepts an "angular.element" object and will use that element as the container for the dropdown menu when it is opened.  This allows us to attach the dropdown to the scrolling container which makes them scroll with the content, but don't get cutoff by hidden overflow in their immediate container.

Here's a little gif I put together of the three types of "attachment" and how they look in this sort of situation:

uprcvrq1v7

This was actually taken from the angular-ui bootstrap docs that has the changes from this PR in them.

I added tests for this, and modified the docs/demo from dropdown as well.

### DIFF
--- a/src/dropdown/test/dropdown.spec.js
+++ b/src/dropdown/test/dropdown.spec.js
@@ -220,6 +220,10 @@ describe('dropdownToggle', function() {
       element = dropdown();
     });
 
+    afterEach(function() {
+      $document.find('body').removeClass('dropdown');
+    });
+
     it('adds the menu to the body', function() {
       expect($document.find('#dropdown-menu').parent()[0]).toBe($document.find('body')[0]);
     });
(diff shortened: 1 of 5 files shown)


wesleycho:
Can you rebase?

src/dropdown/test/dropdown.spec.js
```diff
@@ -220,6 +220,10 @@ describe('dropdownToggle', function() {
       element = dropdown();
     });
 
+    afterEach(function() {
```
wesleycho: This `afterEach` is unnecessary once this is rebased off of current `master`.

wesleycho:
I took a look - this PR looks good, just needs to have the commits squashed and then rebased on top of current `master`.

wesleycho:
I made a couple of tweaks to get everything to work on `master` - I will have this merged shortly.

Thanks for the work!

jaydiablo:
Thanks for getting that up to date and merged!

PR #4497, opened 2015-09-30
feat(typeahead): add `appendElementToId`

wesleycho:
- Add appending popup to specific id

This supercedes #4231


wesleycho:
Will have doc update shortly

src/typeahead/typeahead.js
```diff
@@ -435,6 +437,8 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.position'])
 
         if (appendToBody) {
           $document.find('body').append($popup);
+        } else if (appendToElementId !== false) {
```
Foxandxss: !== false is needed in here? Any case where it could be falsy but acceptable?

wesleycho: Given the above setting of the variable to `false` by default, I think this is a fine check.

wesleycho:
Greenlit by @Foxandxss

PR #4500, opened 2015-10-01
Prevent `contains` error in IE 11

acbabis:
In IE11, void elements have the `contains` function but throw an operation not supported error when called. This change makes it check to see if the node has children before calling `contains`.


wesleycho:
I'm not sure that loosening this check is a good thing here - I found [this](https://connect.microsoft.com/IE/feedback/details/780874/node-contains-is-incorrect), and it looks like MS is not willing to reconsider this behavior, so potentially one can run into a situation where the element isn't attached to a node, and thus not have the contains method.

wesleycho:
Oh, now I understand - misread the statement at first. Can you post about more information about this existence? I can't seem to be able to reproduce this scenario

acbabis:
It looks like I was using an older version of angular-bootstrap. The browser error "Object doesn't support property or method 'contains'" made me believe the function was there but not supported. I didn't notice that the code was outdated because I was tweaking minified code in my development environment. My apologies.

wesleycho:
Not a problem, thanks for trying though :) .

PR #4501, opened 2015-10-01
feat(carousel): use uib- prefix

BobbieBarker:
Added uib prefixing to the carousel directive.


wesleycho:
Clean up the history (there should not be a merge commit) - also we need tests for the deprecation warning.

BobbieBarker:
i knew I was forgetting something!

Foxandxss:
There is a list of things that needs to be fixed:

Slide needs prefixing
You are creating the carousel module twice
Remove the controller from the deprecated part
Remove the animation from the deprecated part
Make the deprecated carousel use the prefixed controller
Fix the deprecation comment to use carousel
Change $alertSupressWarning to collapse (in all different places)
Add tests for deprecation warning
Fix all the tests to use the new prefixed versions

Perhaps I am forgetting something, but that is a good start.

Foxandxss:
Nice werk. Remaining.

Remove all the extra docs on the deprecated part (I want to remove them all but that is a different topic).
The deprecated slide needs deprecation messages too
You still need to change the $alertSuppress thingy to carousel.

Foxandxss:
Nice.

So, the demo is still using non prefixed slide
For the deprecation tests, no need of two describes. Check how the others are done. One describe with one tests with logs suppressed and one with all errors.

src/carousel/carousel.js
```diff
@@ -413,3 +405,65 @@ function ($injector, $animate) {
     }
   };
 }]);
+
```
wesleycho: Remove 3 of the lines

src/carousel/carousel.js
```diff
@@ -413,3 +405,65 @@ function ($injector, $animate) {
...
+
+
+
+/* Deprecated alert below */ 
+
+/**
```
wesleycho: No need for this ngdoc, especially of a deprecated component.

src/carousel/test/carousel.spec.js
```diff
@@ -67,12 +67,12 @@ describe('carousel', function() {
...
-        '<carousel interval="interval" no-transition="true" no-pause="nopause">' +
-          '<slide template-url="foo/bar.html"></slide>' + 
-        '</carousel>'
+        '<uib-carousel interval="interval" no-transition="true" no-pause="nopause">' +
+          '<uib-slide template-url="foo/bar.html"></uib-slide>' + 
+        '</uib-carousel>'
       )(scope);
       $rootScope.$digest();
-
+
```
wesleycho: Remove the whitespace

PR #4502, opened 2015-10-01
feat(rating): add `uib` prefix

wesleycho:
- Adds `uib` prefix to all injectibles and directives
- Adds deprecation warning


Foxandxss:
When merging, please change the commit title to match the others.

On the accordion, I decided to leave the local variable for the settings without prefix. Leads to a less verbose code and I think we should maintain one style. Which one? :)

Foxandxss:
:+1: (it is use and not user :P)

PR #4503, opened 2015-10-01
feat(progressbar): change to fully add `uib` prefix

wesleycho:
- Fully prefix with `uib`


Foxandxss:
Prefix the config and and dots at the end of each deprecation warning.

Foxandxss:
The tests are missing the dots on the deprecation.

PR #4504, opened 2015-10-01
feat(dateparser): use uib- prefix

wesleycho:


Foxandxss:
:+1:

llafuente:
Another breaking change.

I decorate `dateParser` and now is renamed.

icfantv:
@llafuente, the `dateParser` service was previously internal-only and thus, was not supported.  Moving forward, however, it is both documented and supported.  That said.  I'll mark this as a breaking change simply for those of you who like to live on the edge.

PR #4511, opened 2015-10-02
feat(modal): add uib- prefix

Foxandxss:


src/modal/docs/readme.md
```diff
@@ -20,7 +20,7 @@ The `$modal` service has only one method: `open(options)` where available option
 * `size` - optional suffix of modal window class. The value used is appended to the `modal-` class, i.e. a value of `sm` gives `modal-sm`
 * `openedClass` - class added to the `body` element when the modal is opened. Defaults to `modal-open`
 
-Global defaults may be set for `$modal` via `$modalProvider.options`.
+Global defaults may be set for `$uibModal` via `$modalProvider.options`.
```
wesleycho: Should be `$uibModalProvider.options`

src/modal/modal.js
```diff
@@ -689,3 +689,670 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
+    }])
+
+  .provider('$modal', function() {
+    var $modalProvider = {
+      options: {
+        animation: true,
+        backdrop: true, //can also be false or 'static'
+        keyboard: true
+      },
+      $get: ['$injector', '$rootScope', '$q', '$templateRequest', '$controller', '$modalStack', '$log', '$modalSuppressWarning',
```
wesleycho: Is there really no way we can't just reuse the `$uibModal` service here? This is quite a bit of code to duplicate.

chrisirhc: +1 , require $uibModal here and log warning.

Foxandxss: Will fix tomorrow.

src/modal/modal.js
```diff
@@ -689,3 +689,670 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
+    '$$stackedMap',
+    '$log',
+    '$modalSuppressWarning',
+    function($animate ,  $timeout ,  $document ,  $compile ,  $rootScope ,
+             $q,
+             $injector,
+             $$multiMap,
+             $$stackedMap,
+             $log,
+             $modalSuppressWarning) {
```
wesleycho: Can we investigate reusing `$uibModalStack` here as well? There is a $watch in this service, so I want to avoid a double $watch on $rootScope here as well.

Foxandxss:
I fixed all your issues.

wesleycho:
LGTM

PR #4512, opened 2015-10-02
fix(dropdown): restored deprecated directives

BobbieBarker:


Foxandxss:
Better, the if checks for deprecation, move them to the beginning of the link function (as we do in the others). Also, when this gets merged (/me looks to Wes), change fix for chore, we don't need this on changelog.

Foxandxss:
Also, there is an old commit in here, I am not sure if that is gonna blow or not. Damn git.

wesleycho:
That commit screws up history - this is not mergeable.

wesleycho:
I fixed everything in #4514.

Foxandxss:
Thanks Wes.

PR #4514, opened 2015-10-03
fix(dropdown): restore deprecated directives

wesleycho:
- Adds uib- prefix to `dropdownConfig`
- Restores functionality to deprecated directives

Supercedes #4512.


Foxandxss:
Here you are also moving controllers to the deprecated zones. In the others prefixes, we simply prefix controllers and we tell the deprecated directives to use it. That works too.

I like your way of doing that as well, but I think we shouldn't have two ways to deprecate and pick one. We can do it this way and update the others or modify this to be like the others.

wesleycho:
I've been doing this in general - I do it just in case we have some users using the controllers directly & building custom directives around them.

Foxandxss:
That means that the other ones should have the same thing.

wesleycho:
Merging this in for now, will investigate the others today.

PR #4515, opened 2015-10-03
feat(tooltip): add uib- prefix

wesleycho:
Note that this does it for the tooltip and popover (couldn't get both in the commit message due to conventional-changelog).


src/popover/popover.js
```diff
@@ -15,36 +14,108 @@ angular.module( 'ui.bootstrap.popover', ['ui.bootstrap.tooltip'])
...
+    }
+  };
+}])
+
+.directive('popoverTemplate', ['$log', '$popoverSuppressWarning', '$uibTooltip', function($log, $popoverSuppressWarning, $tooltip) {
+  if (!$popoverSuppressWarning) {
+    $log.warn('popover-template is now deprecated. Use uib-popover-template instead.');
+  }
+
+  return $tooltip('popoverTemplate', 'popover', 'click', {
```
Foxandxss: I think I would make this use the new tooltip.

Foxandxss:
Check my Modal PR to see how I managed to get all the deprecation messages out in tests.(you have lots of directives but few deprecation message tests).

src/popover/popover.js
wesleycho: This is the old `popoverTemplate` directive

wesleycho:
I think I just forgot to write the other tests

wesleycho:
This PR should be good to review again.

Foxandxss:
Missing doc updates, but I fixed that for you.

PR #4517, opened 2015-10-03
chore(buttons): update docs

wesleycho:


Foxandxss:
That can go on. The reason of why I haven't touched it on my prefix work is because I am going to delete it all. I am waiting to see what we decide to do with those two customizable things on the button (active class and toggle event) so I can properly make them public or delete that. With that done, I need to rewrite the doc from this component from zero.

wesleycho:
I'll merge for now for safety, then it can be addressed later.

PR #4521, opened 2015-10-03
chore(typeahead): remove unnecessary $digest

wesleycho:
- Removes unnecessary $digest inside $timeout


Foxandxss:
Yikes, weird code. Remove it with fire. :+1:

PR #4523, opened 2015-10-03
fix(dropdown): ensure class is present in dropdown-menu

### LINKED ISSUE #4442: DropdownMenu directive doesn't when it's evaluated as argument restriction
See the [plunker](http://plnkr.co/edit/3AM107KptafYK8oUR7Ww?p=preview).

Pay attention on line 18 in `index.html`. If change the `dropdown-menu` directive to the class the picker is working correctly. But accordingly to the [code](https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L254) it should work for the argument case either.

wesleycho:
- Ensure `dropdown-menu` class is present when using as an attribute
  directive
- Refactor `DropdownController` to be more agnostic of directive

This fixes #4442.


Foxandxss:
From ignorance on the issue, this doesn't need tests?

wesleycho:
I could write a test, but it seems pretty trivial of a change to me & not really function - what are your thoughts?

Foxandxss:
Alright, I had to ask. I was going to do that refactor too when I reached the D on my rewrite (which doesn't seem to move that much lately)

PR #4524, opened 2015-10-03
fix(accordion): re-expose AccordionController

wesleycho:
- Expose `AccordionController` with deprecation message


Foxandxss:
:+1:

PR #4131, opened 2015-08-07
feat(carousel): expose carousel controller via `controllerAs`

wesleycho:
- Expose the carousel controller in the component via `controllerAs`

PR #4527, opened 2015-10-03
fix(carousel): re-enable deprecated directives

wesleycho:
- Re-implement deprecated directives
- Re-expose `CarouselController` with deprecation notice


src/carousel/test/carousel.spec.js
```diff
@@ -551,27 +551,28 @@ describe('carousel deprecation', function() {
...
-        '</carousel>';
+        '<slide></slide>' + 
+      '</carousel>';
     element = $compile(element)($rootScope);
     $rootScope.$digest();
 
-    expect($log.warn.calls.count()).toBe(2);
-    expect($log.warn.calls.argsFor(0)).toEqual(['slide is now deprecated. Use uib-slide instead.']);
+    expect($log.warn.calls.count()).toBe(3);
+    expect($log.warn.calls.argsFor(1)).toEqual(['slide is now deprecated. Use uib-slide instead.']);
```
Foxandxss: Check all messages.

wesleycho:
Added other checks as noted

Foxandxss:
:+1:

chrisirhc:
Just wondering why this PR was created. Is it to resolve some issues? It should reference those issues, if that's the case.

Foxandxss:
@chrisirhc We are prefixing all the directives. There were a slip, and in the carousel, the deprecated directives were empty. They need to be fully functional. This PR puts back the functionality on those deprecated directives.

This shouldn't be a `fix(foo)` tho but chore.

src/carousel/carousel.js
```diff
@@ -421,11 +420,24 @@ angular.module('ui.bootstrap.carousel')
 
 .value('$carouselSuppressWarning', false)
 
+.controller('CarouselController', ['$scope', '$element', '$controller', '$log', '$carouselSuppressWarning', function($scope, $element, $controller, $log, $carouselSuppressWarning) {
+  if (!$carouselSuppressWarning) {
+    $log.warn('CarouselController is now deprecated. Use UibCarouselController instead.');
+  }
+
+  angular.extend(this, $controller('UibCarouselController', {
```
chrisirhc: Instead of extending `this`, you can probably return the `$controller('Uib…` directly.

src/carousel/carousel.js
```diff
@@ -421,11 +420,24 @@ angular.module('ui.bootstrap.carousel')
...
+  }));
+}])
+
 .directive('carousel', ['$log', '$carouselSuppressWarning', function($log, $carouselSuppressWarning) {
   return {
-    restrict: 'EA',
     transclude: true,
     replace: true,
+    controller: 'CarouselController',
+    controllerAs: 'carousel',
```
chrisirhc: I don't think this is needed? I don't see any usage of `scope.carousel` .

Foxandxss: We started to add this option to some directives to expose the controller to the template using `controllerAs`. It is not something I would add personally, but there it is.

src/carousel/carousel.js
wesleycho: @Foxandxss mind if I just patch this in all of the components?

src/carousel/carousel.js
chrisirhc: Hm.. okay, sounds like a separate feature, with feat, though. But okay then.

Foxandxss: Excuse me Chris, but after a couple of months away of the project, there is lots of new features done and I am afraid that you're barking to the wrong tree. I totally understand that.

Imagine we make a mistake a we push a commit where we get a directive and we remove it entirely. Then we want to create another PR to put back the content we deleted. We are simply doing a chore to put it back like it was before. We are not creating multiple PRs with all the features we deleted one by one. We are just restoring something that was there in a first place.

chrisirhc: You're right. It was added as a feature in #4131, I wasn't aware of that.

PR #4528, opened 2015-10-03
fix(progressbar): re-expose ProgressController

wesleycho:
- Re-expose `ProgressController` with deprecation notice


Foxandxss:
:+1:

PR #4531, opened 2015-10-03
fix(timepicker): re-expose TimepickerController

wesleycho:
- Re-expose `TimepickerController` with deprecation notice


Foxandxss:
:+1:

PR #4532, opened 2015-10-03
fix(carousel): reset $currentTransition when no slides

### LINKED ISSUE #4390: Carousel - empty slide list
When slides list is cleared and filled with new slides Carousel stops working.

Here is reproducer.
Scenario:
1. Load 4 images - works fine
2. After 5 seconds clear slides list and fill with new images - carousel stops working

Plunker: http://plnkr.co/edit/u6inF6vMIzdSVplgkJBr?p=preview

AngularJS: 1.4.5
UI Bootstrap: 0.13.4

wesleycho:
- Reset `$currentTransition` when there are no slides

[Here](http://plnkr.co/edit/oTI5FuWoGhHnYx0rhTMe?p=info) is a Plunker with the fix in action (as well as fixes to user error).

This fixes #4390.


Foxandxss:
When creating a Plunker, please add instructions of how to see it working. If I understood correctly, I need to delete all slides and add new ones, but I don't see how to do that in the plunker.

wesleycho:
The Plunker is resetting it at a setTimeout of 5000ms, which is coincidentally when the animation starts, which triggers the bug in the original Plunker in the referenced issue

Foxandxss:
I see. After some seconds (before doing the first change) it does a weird animation thing. Do you see that? (Chrome).

wesleycho:
Yep, that will be hard to fix I think - there is another issue open currently with the 0.-4 milestone on a weird animation problem when adding a new slide though, may be slightly related

Foxandxss:
Alright then.

PR #4534, opened 2015-10-04
feat(dropdown): Add dropup behavior

### LINKED ISSUE #1317: Make dropdown menu go upwards if it is at the bottom of the browser window

gorork:
Makes dropdown menu go upwards if it's at the bottom of the browser window and is being partially hidden. Moves DOM logic to directive.

Closes #1317

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -78,6 +78,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
     selectedOption = null;
 
 
+  self.element = $element;
   $element.addClass('dropdown');
 
   this.init = function() {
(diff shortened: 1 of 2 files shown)


src/dropdown/dropdown.js
```diff
@@ -264,9 +247,57 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
       if (!dropdownCtrl.dropdownMenu) {
         dropdownCtrl.dropdownMenu = element;
       }
+
+      var dropdown, appendToBody,
+          dropdownMenu = element;
+
+      scope.$watch(dropdownCtrl.isOpen, function( isOpen ) {
```
wesleycho: I think this $watch needs to be in the controller - it should be noted that `dropdownCtrl.isOpen` will evaluate immediately when the linking function is run, so this watch is basically
```js
scope.$watch(false, function(isOpen) {
...
});
```
which is not what we want.

gorork: @wesleycho This is purely DOM logic (element positioning and adding/removing classes) with nothing to do with $scope, model, etc. 

From the [official Directives guide](https://docs.angularjs.org/guide/directive) that provides example with `scope.$watch` in the link function: 
> "Directives that want to modify the DOM typically use the link option."

Why do you think it belongs in the controller instead of link function?

RobJacobs:
I'd like to start a discussion on how we could add a popup service that has smart/auto positioning logic for all popups - tooltip, popover, dropdown, datepicker, and typeahead rather than have one off solutions for each directive.

icfantv:
@RobJacobs, this is absolutely on the agenda for post-initial-release of the ng-bootstrap project as we intend to utilize interfaces such that the user can plug in their own library (like Tether).  I don't know how we might accomplish this for this project without also having to support ES6 since ES5 doesn't support interfaces and undoubtedly someone will not like our solution or will want some specialized use of it.  But I agree, this conversation is long overdue.

wesleycho:
That is too long to wait to have a solution - a solution for this problem needs to come before then I think, or at least needs to be in consideration independently as part of natural work in UI Bootstrap.

As for this PR, the $watch logic needs to return to the controller - that is where it belongs for modularity, especially if someone wanted to create a custom directive that made use of the dropdown controller. The PR needs documentation as well. Otherwise, I am ok with this making it in the library, and us having a discussion about implementing smart positioning functionality.

icfantv:
Closing in favor of #4762.

PR #4536, opened 2015-10-04
chore(pagination): add uib- prfix

BobbieBarker:


Foxandxss:
I think that my only concern now is that we decided to prefix configs (so good so far), but for local usage in the directives, we remove the prefix (check accordion for example).

Also, this is not a chore but a feat. And a prefix

Foxandxss:
We also decided to move the old controllers as well to the deprecated part, check how it is done in the other prefixes.

BobbieBarker:
You're killing me.

Foxandxss:
Yeah, sorry. We are iterating this really fast lately so the requirements changes fast too.

Foxandxss:
Check the PR #4527 where Chris suggested a better way to do the "old controllers". Basically you add the deprecation messages and then you call the original controller. That way we avoid having the controller code twice in the file. That with the proper deprecation message test.

wesleycho:
This LGTM, but I'll wait for @Foxandxss to do another pass to be sure.

Foxandxss:
You forgot to update the docs, but I fixed that for you :)

PR #4550, opened 2015-10-08
fix(tooltip): Tooltip causing scroll bars to quickly show

### LINKED ISSUE #4458: Tooltip causing scroll bars to quickly show
[Plunk](http://plnkr.co/edit/HdHyjJ?p=preview)

Resize the window so it comes up almost to the bottom of the button. Constantly move back and forth over the button to trigger the tool tip. Notice the scroll bar briefly showing sometimes.

cassilup:
When the element that has a tooltip attached is hovered upon, the tooltip is created and positioned accordingly. However, there is a slight flicker between the time that the tooltip element is added to the DOM and the moment when it is positioned where it needs to be displayed. This also causes scrollbars to appear/flicker on the browser window in some cases. The fix consists of hiding the tooltip by default (display: none) and then displaying it once it is in place (display: block).

Closes #4458


wesleycho:
So the `visibility: hidden` hack is so that the correct positioning can be calculated on the tooltip for positioning correctly - there are some situations where the position of the tooltip is calculated incorrectly when it is not present in the DOM.

RobJacobs:
The original issue #4195 where the display approach was added.

PR #4555, opened 2015-10-08
fix(tooltip): check for ttScope in $$postDigest

### LINKED ISSUE #4552: popover: Cannot read property 'isOpen' of null:TypeError: Cannot read property 'isOpen' of null
I have a few popovers on a page using popover-template. Each popover has one link in the template.

When I click on the link in the popover I go to the page and everything is fine.

When I click the BACK button and click on any popover and then click on a link I get an error:

>Cannot read property 'isOpen' of null:TypeError: Cannot read property 'isOpen' of null

I found this: #3347 and thought it might be related so I implemented the change in my code (I'm on v0.13.4) but that did not fix it.

I'm trying to create a plunkr but when I click on a link in a popover in plunkr I am not taken to the page. But when I right click and choose open in new tab I am taken to the correct page. Because I can't open a link in plunkr I can't reproduce this issue there.

RobJacobs:
In the tooltip $$postDigest function a check is done
on ttScope.isOpen but ttScope may have been set to null
in the scope.$on('$destroy'...) function.  Added a
check to make sure ttScope is defined and not null.

Fixes #4552

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -374,7 +374,7 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
                       repositionScheduled = true;
                       tooltipLinkedScope.$$postDigest(function() {
                         repositionScheduled = false;
-                        if (ttScope.isOpen) {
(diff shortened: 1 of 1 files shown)


icfantv:
:+1:

PR #4557, opened 2015-10-08
chore: add tooltip/popover classes by hand

Foxandxss:
Sigh round 2.


pkozlowski-opensource:
@Foxandxss what is the rationale behind this change? Asking since there is no link to an issue that it is closing. 

I though that we want to keep as much markup in templates as possible, so people can override templates with their own HTML / CSS. This change moves CSS class name to JS part, making full override harder. 

What is the issue that we are trying to solve here?

Foxandxss:
@pkozlowski-opensource It is a workaround. We had all the prefixing done but we still had a "ghost" `$tooltip is deprecated, use $uib-tooltip instead.`. Same with a `$modalStack is deprecated, use $uibModalStack instead`.

After 5 hours, I saw the culprit.

If we use a `uibTooltip`, that will eventually render a template with a `tooltip` class. Angular will send that template to the compiler. The compiler will try to match any directive that could exist on that template. Eventually it will try to match the `tooltip` class, will inject `tooltip` and with the injection and all dependencies are loaded as well. That will load `$tooltip` and print the deprecation message.

Then the compiler will discard it because `tooltip` is not restricted to classes and move on. Sadly, the damage is already done.

So TL;DR, we have a temporary workaround to avoid angular compiler to execute deprecated directives and log warning messages.

That is going away after we remove deprecated libs. Also Wesley and I decided to open an issue at angular repo https://github.com/angular/angular.js/issues/13057

pkozlowski-opensource:
@Foxandxss thnx for the info. If this is just temporary work-around I'm good, I see what is happening from the linked ng issue (thnx for opening it, btw!)

PR #4558, opened 2015-10-08
fix(datepicker): add custom class to year picker

### LINKED ISSUE #4546: custom-class not geting applied for Datepicker in year mode.
I tried to pass custom class for datepicker. I tried to pass custom class when datepicker is in year mode and it is not applying it. However the same works fine when datepicker is in day and month mode.

Please confirm me if it is a bug?

icfantv:
* add missing custom class to year picker

fixes #4546


icfantv:
Just a note that we're missing functional tests around all three templates custom class styling.  We test that the custom class expression was executed but not that the class was applied.

wesleycho:
LGTM

PR #4560, opened 2015-10-09
fix(dropdown): Don't allow using dropdownMenu and uibDropdownMenu

realityking:
Without this it's impossible to remove all deprecation warning without messing up the bootstrap styles.

BREAKING CHANGE: dropdownMenu and uibDropdownMenu have to be used as attributes, not classes.

As an aside, several directives are horribly broken when used without the `uib-` prefix, so this is actually a rather important fix.

### DIFF
--- a/src/dropdown/docs/demo.html
+++ b/src/dropdown/docs/demo.html
@@ -5,7 +5,7 @@
       <a href id="simple-dropdown" uib-dropdown-toggle>
         Click me for a dropdown, yo!
       </a>
-      <ul class="uib-dropdown-menu" aria-labelledby="simple-dropdown">
+      <ul class="dropdown-menu" uib-dropdown-menu aria-labelledby="simple-dropdown">
         <li ng-repeat="choice in items">
           <a href>{{choice}}</a>
         </li>
@@ -17,7 +17,7 @@
       <button id="single-button" type="button" class="btn btn-primary" uib-dropdown-toggle ng-disabled="disabled">
         Button dropdown <span class="caret"></span>
       </button>
-      <ul class="uib-dropdown-menu" role="menu" aria-labelledby="single-button">
+      <ul class="dropdown-menu" uib-dropdown-menu role="menu" aria-labelledby="single-button">
         <li role="menuitem"><a href="#">Action</a></li>
         <li role="menuitem"><a href="#">Another action</a></li>
         <li role="menuitem"><a href="#">Something else here</a></li>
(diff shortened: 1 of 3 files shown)


icfantv:
We'll take a look at this and get back to you.  Just a heads up:  when making PRs, they need to be as least invasive as possible (change as little as possible) - otherwise, sometimes it can be harder to see the relevant changes that are made - I'm speaking to the blank line changes, here.

Foxandxss:
I think this is ok. There is an issue with how Angular works which directly strikes our way of doing the prefixes. Probably this is just one case of it that we didn't see.

Will check.

Foxandxss:
Lets step back a bit. @realityking What is the problem in here? Forget about this PR and tell me.

Could you please create a plunker with what you want to fix? I can't see any prefix issue and I probably thinking about the wrong thing in here.

And as an aside, if you see that a directive is broken without prefix, don't hesitate to open as many issues as you need.

If there is any big issue, we won't hesitate on creating a hotfix release before the week ends.

wesleycho:
I am confused also what is the problem - [here](http://plnkr.co/edit/Pkcpyvj6BshQVndjtYI0?p=preview) is a working Plunker

RobJacobs:
This [plunk](http://plnkr.co/edit/lepvU46mKgA88Ro9ctzZ?p=preview) is a fork of @wesleycho with the bootstrap classes (dropdown, dropdown-toggle, & dropdown-menu) added.  Still not seeing the problem?

wesleycho:
I suspect that this was a misunderstanding with a subtle change we made to the dropdown to fix this.

Take a note of https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L81 , https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L257 , https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L318 , https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L399 , and https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L468 . We are adding the appropriate class as necessary to ensure it is present.

I am closing this issue as invalid unless there is more information.

realityking:
First of all sorry for the long delay, I didn't have time to check GitHub over the weekend.

I indeed missed that dropdownMenu/uibDropdownMenu add the required classes to element. That said, IMO it would be good to still get this merged eventually to be consistent about directives not matching on classes.

PR #4571, opened 2015-10-10
fix(modal): Backspace press while modal open and using ui-router

ivi-hamiti:
while the modal is open and no input element is focused, and the backspace is pressed the back app navigates back while the modal stays open


src/modal/modal.js
```diff
@@ -529,6 +539,17 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         return false;
       };
 
+      $modalStack.isFocusInFormItem = function (evt) {
```
wesleycho: Change to `$modalStack.formItemFocused`

src/modal/modal.js
```diff
@@ -529,6 +539,17 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
+
```
wesleycho: Remove blank line

src/modal/modal.js
```diff
@@ -529,6 +539,17 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         return false;
       };
 
+      $modalStack.isFocusInFormItem = function (evt) {
+          
+        for (var i = 0; i < focusableElementList.length; i++) {
+          if ((evt.target || evt.srcElement) == focusableElementList[i]) {
```
wesleycho: Use `===`

wesleycho:
I'm a bit skeptical of this - is there a particular reason one cannot work around this without modifying the library?

ivi-hamiti:
well i didn't found any work around for this case, but usually it is better to fix not to work around

ivi-hamiti:
do you have another idea how to handle this when using ui-router, to do a global event listener for that would not be a good aproach

wesleycho:
One thing you can do on the user side is listen for $stateChangeStart and then run $uibModalStack.dismissAll() I believe.

wesleycho:
I believe this is a problem to be solved on the user's side - managing the modal's presence should involve selecting what $scope the $modal service uses judiciously, as well as managing keypress behavior manually here. I am going to close this as a result, since this could be problematic if a developer did not want this behavior.

ivi-hamiti:
ok thanks anyway

ivi-hamiti:
@wesleycho I don't know i just thought about changing the behavior of the backspace press what if the implementation would be that on backspace while the modal is open it closes the modal and does not propagate to the browser to go back, it would be a more streamline experience to the user. What do you think, if you like the idea i can implement it.

Issue #4569, opened 2015-10-10
tabindex on timepicker does not work


getvictor:
When specifying **tabindex** on timepicker without spinners

`<uib-timepicker show-spinners="false" tabindex="8" />`

the surrounding table around timepicker is highlighted when getting to the corresponding tabindex, and not the individual inputs. Using Chrome on Mac.

When tabindex is specified, I'd like to see this same tabindex applied to each of the individual input fields of timepicker.

wesleycho:
Just for clarity, what would your preferred solution be? Would it be for the individual inputs to also inherit the same tabindex (i.e. 8 in this example)?

getvictor:
@wesleycho Yes, all the individual inputs should get the same tabindex. Elements with the same tabindex are processed in the order they appear. The outer element should not have a tabindex.

Reference: http://www.w3.org/TR/WCAG20-TECHS/H4.html#H4-ex3

wesleycho:
This means that the tabindex should also be on the links as well.

getvictor:
I'm not using the up/down links, which is what I believe you're referring
to. But yes, for accessibility purposes they should be on those links as
well.

PR #4573, opened 2015-10-11
feat(timepicker): add accessibility improvements

wesleycho:
- Add ng-disabled usage in template
- Add tabindex to support to timepicker controls

This addresses #4569


src/timepicker/docs/readme.md
```diff
@@ -50,3 +50,7 @@ All settings can be provided as attributes in the `<uib-timepicker>` or globally
  * `max`
     _(Defaults: undefined)_ :
      Maximum time a user can select
+
+ * `timepicker-tabindex`
```
pkozlowski-opensource: couldn't we use `tabindex` here? And remove it from the timepicker host element if needed?

wesleycho: I thought about it, but I had a concern...but I can't remember it now - I am willing to go that route though for simplicity.

Foxandxss:
Not really my topic.

The only bit to talk here is that we shouldn't be supporting the deprecated directives. The point is to force people to upgrade and not fixing them is a good way to do so.

wesleycho:
This has nothing to do with deprecated directive support :) .

Foxandxss:
I know I know, I just said that tabs and accessibility is not something I can talk about.

PR #4574, opened 2015-10-11
fix(accordion): make deprecated controller work with 1.3.x

Foxandxss:
Fix accordion controller to be compatible with 1.3.x again.


wesleycho:
LGTM

PR #4579, opened 2015-10-11
chore(datepicker): remove extra suppression service

Foxandxss:
Somehow we had it twice, wops.


Foxandxss:
Nevermind, somehow I am blind, yikes.

PR #4580, opened 2015-10-11
fix(pagination): make deprecated controller work with 1.3.x

Foxandxss:
Alright, this adds a bit of *temporary* overhead.

The problem with `angular.extend(this, $controller...)` is that it won't carry the stuff you assign to `self`.

So the thing is: 1.3.x support or a bit of temporary overhead?


wesleycho:
LGTM

PR #4588, opened 2015-10-11
Fix calculation of progressbar on removed bar

### LINKED ISSUE #4452: ProgressController recalculatePercentage bug

wesleycho:
This fixes the percentage calculation when a bar is removed.

This fixes #4452, and adds the unit test from #4471, as well as fixes the code style and commit message.

I am not sure why the release commit is listed in the history here, as that is from `master` and I ran `curl ... | git am -3`, which should have rebased the original commit on top of `master`.

Edit: History is fixed


Foxandxss:
Seems fine to me Wes. Progressbar is my cursed lib (tests that only fails to me) so /shrug.

PR #4595, opened 2015-10-12
refactor(typeahead): move logic to controller

wesleycho:
- Create `UibTypeaheadController` for decoupling the directive
  implementatino from the controller


Foxandxss:
:+1:

PR #4596, opened 2015-10-12
chore(docs): mention breakage of every api in 0.14.0

hoxu:
Pretty much every API changed in 0.14.0, and it was not mentioned under `BREAKING CHANGES`, so this is to fix that :)


icfantv:
That's because we retained the original functionality and marked it deprecated.  Thus, it's not a breaking change until we actually remove the old directives.  Because then, your code really won't work.

hoxu:
Admittedly I was a bit hasty with that commit message. Still, as a user of the library I would love to have a head-up of changes like this as early as possible - and the "breaking changes" (or similar) section of changelogs is exactly what people read. Don't you think a similar mention in the CHANGELOG is warranted?

Props for deprecation instead of direct removal, that's a clear improvement!

icfantv:
I don't.  Nearly every other properly developed library will insert deprecation warnings (that is, log the warnings to the console) before removing/changing functionality (unless it's critical).  If you're a developer and you're not looking at your developer console when you develop - well, that's just scary.

PR #4455, opened 2015-09-23
refactor(tooltip): rollup commit for several issues

RobJacobs:
This is a rollup commit intended to address several
issues around the positioning and parsing of
attributes.

- Fixes issue introduced under PR #4311 where setting
  height and width in tooltip position function
  messed up arrow placement.
- Fixes issue introduced under PR #4363 where setting
  visibility to hidden in tooltip position function
  caused elements in popover to lose focus.
- Fixes issue #1780 where tooltip would render if
  content was just whitespace.
- Fixes issue #3347 where tooltip isolate scope was
  being accessed after it was set to null.  Observers
  will now be created/destroyed as tooltip opens/closes
  which will also offer a performance improvement.
- Fixes issue #3557 by implementing evalAsync to set
  tooltip scope isOpen property.
- Fixes issue #4335 where if model isOpen property is
  undefined, tooltip would call show/hide toggle function.
- Closes PR #4429 where how the templated content
  was being evaluated could cause an infinite digest loop.

Closes #4400
Closes #4418
Closes #4429
Closes #4431

Fixes #1780
Fixes #3347
Fixes #3557
Fixes #4321
Fixes #4335

PR #4597, opened 2015-10-12
fix(tooltip): popup close delay not respected

RobJacobs:
Under PR #4455 the timeout for the popup close delay
was impacted by the evalAsync for setting the isOpen
value.  Moved the popup close delay timeout to the
hideTooltipBind method so the hide mehtod will now
get called after the close popup delay.  This will
also ensure if the show method is called and the
close timeout is cancelled, the isOpen property
will not get toggled.


wesleycho:
Can you explain what is going on here - how come it is not enough that all of the logic is in the hide function and why move it up to the `hideTooltipBind` function?

RobJacobs:
The hide function is called in the disabled and empty content watchers, by-passing the close delay. Maybe they shouldn't?

RobJacobs:
@tmcgee123 @wesleycho After looking into this further, I think the original implementation of this feature was wrong.  The transitionTimeout should be reserved for allowing the tooltip animation to run before removing the tooltip element from the dom.  The tooltip templates use ng-class="{ in: isOpen() }" to add and remove the 'in' class which sets the opacity (0 and .9).  The fade class has the opactiy transition with a time of .15s.

A close delay should be a different timeout that completes before the transitionTimeout.  I'm going to do some more work on this PR and get this working as it should.

tmcgee123:
@RobJacobs I agree. Originally in 13.x, the functionality was as you said. But, with the updates that were put through, it looks as though the timeout is no longer completing before the animation starts.

wesleycho:
This PR LGTM - that was a good find that the implementation was incorrect. We really need to port the timeout to use ngAnimate, but that's future work.

PR #4599, opened 2015-10-12
feat(modal): add container option

aroop:
Allows modal to be added to any element instead of body. If container option is not passed modal will be added to body

### DIFF
--- a/src/modal/docs/readme.md
+++ b/src/modal/docs/readme.md
@@ -19,6 +19,7 @@ The `$uibModal` service has only one method: `open(options)` where available opt
 * `windowTemplateUrl` - a path to a template overriding modal's window template
 * `size` - optional suffix of modal window class. The value used is appended to the `modal-` class, i.e. a value of `sm` gives `modal-sm`
 * `openedClass` - class added to the `body` element when the modal is opened. Defaults to `modal-open`
+* `container` - Appends the modal to a specific element. container must be an `angular.element`. Defaults to `body` element. Example: `container: $document.find('aside')`. 
 
 Global defaults may be set for `$uibModal` via `$uibModalProvider.options`.
 
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -276,8 +276,8 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
       });
 
       function removeModalWindow(modalInstance, elementToReceiveFocus) {
-        var body = $document.find('body').eq(0);
         var modalWindow = openedWindows.get(modalInstance).value;
+        var container = $document.find(modalWindow.container).eq(0);
 
         //clean up the stack
         openedWindows.remove(modalInstance);
@@ -285,7 +285,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         removeAfterAnimate(modalWindow.modalDomEl, modalWindow.modalScope, function() {
           var modalBodyClass = modalWindow.openedClass || OPENED_MODAL_CLASS;
           openedClasses.remove(modalBodyClass, modalInstance);
-          body.toggleClass(modalBodyClass, openedClasses.hasKey(modalBodyClass));
+          container.toggleClass(modalBodyClass, openedClasses.hasKey(modalBodyClass));
           toggleTopWindowClass(true);
         });
         checkRemoveBackdrop();
@@ -294,7 +294,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         if (elementToReceiveFocus && elementToReceiveFocus.focus) {
           elementToReceiveFocus.focus();
         } else {
-          body.focus();
+          container.focus();
         }
       }
 
@@ -413,13 +413,18 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
           backdrop: modal.backdrop,
           keyboard: modal.keyboard,
           openedClass: modal.openedClass,
-          windowTopClass: modal.windowTopClass
+          windowTopClass: modal.windowTopClass,
+          container: modal.container
         });
 
         openedClasses.put(modalBodyClass, modalInstance);
 
-        var body = $document.find('body').eq(0),
+        var container = $document.find(modal.container).eq(0),
             currBackdropIndex = backdropIndex();
+
+        if (!container.length) {
+          throw new Error('Container not found. Make sure that the container passed is in DOM.');
+        }
 
         if (currBackdropIndex >= 0 && !backdropDomEl) {
           backdropScope = $rootScope.$new(true);
@@ -430,7 +435,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
             angularBackgroundDomEl.attr('modal-animation', 'true');
           }
           backdropDomEl = $compile(angularBackgroundDomEl)(backdropScope);
-          body.append(backdropDomEl);
+          container.append(backdropDomEl);
         }
 
         var angularDomEl = angular.element('<div uib-modal-window="modal-window"></div>');
@@ -449,8 +454,8 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         var modalDomEl = $compile(angularDomEl)(modal.scope);
         openedWindows.top().value.modalDomEl = modalDomEl;
         openedWindows.top().value.modalOpener = modalOpener;
-        body.append(modalDomEl);
-        body.addClass(modalBodyClass);
+        container.append(modalDomEl);
+        container.addClass(modalBodyClass);
 
         $modalStack.clearFocusListCache();
       };
@@ -603,6 +608,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
             //merge and clean up options
             modalOptions = angular.extend({}, $modalProvider.options, modalOptions);
             modalOptions.resolve = modalOptions.resolve || {};
+            modalOptions.container = modalOptions.container || 'body';
 
             //verify options
             if (!modalOptions.template && !modalOptions.templateUrl) {
@@ -669,7 +675,8 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
                   windowClass: modalOptions.windowClass,
                   windowTemplateUrl: modalOptions.windowTemplateUrl,
                   size: modalOptions.size,
-                  openedClass: modalOptions.openedClass
+                  openedClass: modalOptions.openedClass,
+                  container: modalOptions.container
                 });
                 modalOpenedDeferred.resolve(true);
 
--- a/src/modal/test/modal.spec.js
+++ b/src/modal/test/modal.spec.js
@@ -833,6 +833,48 @@ describe('$uibModal', function () {
         expect($document.find('.modal-backdrop')).not.toHaveClass('fade');
       });
     });
+    
+    describe('container', function() {
+      it('should be added to body by default', function() {
+        var modal = open({template: '<div>Content</div>'});
+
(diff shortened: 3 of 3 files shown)


icfantv:
@aroop, I'm a little confused as to what the advantage would be here and what you hope to accomplish.

aroop:
we used modals and little bit of CSS to build the following components
* ASide - Similar to Aside component in [AngularStrap] (http://mgcrea.github.io/angular-strap/#/asides) project
* FullScreenDialog - Dialog which take all the available screen size. Used in mobile views

One of the new requirements that we have for Aside panel is to make it slide from the edge of a DIV. This PR will allow the Aside model to the added to any HTML element in the DOM hierarchy.

icfantv:
Ok.  I'll mark it as needs review, but I can't promise anything.

src/modal/test/modal.spec.js
```diff
@@ -833,6 +833,34 @@ describe('$uibModal', function () {
...
+    describe('container', function() {
+      it('should be added to body by default', function() {
+        var modal = open({template: '<div>Content</div>'});
+
+        expect($document).toHaveModalsOpen(1);
+        expect($document).toHaveModalOpenWithContent('Content', 'div');
+      });
+
+      it('should not be added to body if container is passed', function() {
+        var subElement = '<div clas="details-panel"></div>';
```
icfantv: typo here.  should be `class`.

src/modal/test/modal.spec.js
```diff
@@ -833,6 +833,34 @@ describe('$uibModal', function () {
...
+        var element = angular.element(subElement);
+        angular.element(document.body).append(element);
+
+        var modal = open({template: '<div>Content</div>', container: '.details-panel'});
+
+        expect($document).not.toHaveModalOpenWithContent('Content', 'div');
```
icfantv: it is not sufficient to test that the body doesn't have a modal.  you need to check that the modal is on the component you've specified.

src/modal/modal.js
```diff
@@ -553,7 +554,8 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
       options: {
         animation: true,
         backdrop: true, //can also be false or 'static'
-        keyboard: true
+        keyboard: true,
+        container: 'body'
```
icfantv: why is this duplicated when you've specified it above in `options`?

src/modal/modal.js
```diff
@@ -414,12 +414,13 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
           openedClass: modal.openedClass,
-          windowTopClass: modal.windowTopClass
+          windowTopClass: modal.windowTopClass,
+          container: modal.container
         });
 
         openedClasses.put(modalBodyClass, modalInstance);
 
-        var body = $document.find('body').eq(0),
+        var container = $document.find(modal.container).eq(0),
```
icfantv: what happens if `modal.container` refers to a container that doesn't exist?

src/modal/docs/readme.md
```diff
@@ -19,6 +19,7 @@ The `$uibModal` service has only one method: `open(options)` where available opt
 * `windowTemplateUrl` - a path to a template overriding modal's window template
 * `size` - optional suffix of modal window class. The value used is appended to the `modal-` class, i.e. a value of `sm` gives `modal-sm`
 * `openedClass` - class added to the `body` element when the modal is opened. Defaults to `modal-open`
+* `container` - Appends the modal to a specific element. Example: container: 'body'. Defaults to `body`
```
icfantv: should be "Defaults to `body` element."

src/modal/modal.js
icfantv: This is not going to work.  jqLite is strictly limited to finding by tag name so what's the use of being able to specify something other than `<body>` here?

wesleycho: `modal.container` should probably default to `$document.find('body')`, but allow the user to link to a jqLite wrapped DOM node for the api.

aroop: @icfantv 

Now I am throwing the following error is container is not found

```
Container not found. jqLite is limited to lookups by tag name. Make sure that the container passed is in DOM.
```

@wesleycho

Good idea. I can change to that approach if we all agree

src/modal/docs/readme.md
aroop: Thanks. Fixed.

src/modal/modal.js
aroop: Good catch. Removed duplicate.

src/modal/test/modal.spec.js
aroop: Thanks. Added couple of more test cases

src/modal/test/modal.spec.js
aroop: Removed usage of class in testcases so that this PR won't give the impression that any element can be found.

src/modal/modal.js
icfantv: I'm ok with this.  The documentation should state that the value of of the container attribute is an `angular.element`.

src/modal/docs/readme.md
```diff
@@ -19,6 +19,7 @@ The `$uibModal` service has only one method: `open(options)` where available opt
 * `windowTemplateUrl` - a path to a template overriding modal's window template
 * `size` - optional suffix of modal window class. The value used is appended to the `modal-` class, i.e. a value of `sm` gives `modal-sm`
 * `openedClass` - class added to the `body` element when the modal is opened. Defaults to `modal-open`
+* `container` - Appends the modal to a specific element. Example: container: 'body'. Defaults to `body` element.
```
icfantv: As we discussed below, please indicate that this must be an `angular.element`.

src/modal/modal.js
icfantv: We should think about an error check if `.eq(0)` doesn't return anything.  Do we default to `<body>` or do we log an error message?

wesleycho: Is .eq(0) even necessary?

src/modal/modal.js
```diff
@@ -413,14 +413,19 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
             currBackdropIndex = backdropIndex();
 
+        if (!container.length) {
```
icfantv: We shouldn't mention jqLite because if the user includes the full jQuery library before Angular, Angular will use the full jQuery library.  This goes back to how we want to do error handling as mentioned above.

src/modal/test/modal.spec.js
```diff
@@ -833,6 +833,50 @@ describe('$uibModal', function () {
...
+      it('should be added to body by default', function() {
+        var modal = open({template: '<div>Content</div>'});
+
+        expect($document).toHaveModalsOpen(1);
+        expect($document).toHaveModalOpenWithContent('Content', 'div');
+      });
+
+      it('should not be added to body if container is passed', function() {
+        var subElement = '<section>Some content</section>';
+        var element = angular.element(subElement);
```
icfantv: these two lines can be combined to one - there's not really any need to have them separate.

src/modal/test/modal.spec.js
```diff
@@ -833,6 +833,50 @@ describe('$uibModal', function () {
...
+        angular.element(document.body).append(element);
+
+        var modal = open({template: '<div>Content</div>', container: 'section'});
```
icfantv: based on comments above, this should be `..., container: element...`

src/modal/test/modal.spec.js
```diff
@@ -833,6 +833,50 @@ describe('$uibModal', function () {
...
+
+        expect($document).not.toHaveModalOpenWithContent('Content', 'div');
+      });
+
+      it('should be added to container if container is passed', function() {
+        var subElement = '<section>Some content</section>';
+        var element = angular.element(subElement);
```
icfantv: ditto.  combine to one line.

src/modal/test/modal.spec.js
```diff
@@ -833,6 +833,50 @@ describe('$uibModal', function () {
...
+        angular.element(document.body).append(element);
+
+        expect($document.find('section').children().length).toBe(0);
+        open({template: '<div>Content</div>', container: 'section'});
```
icfantv: ditto.  change `'section'` to `element`.

src/modal/test/modal.spec.js
```diff
@@ -833,6 +833,50 @@ describe('$uibModal', function () {
...
+        expect($document.find('section').children().length).not.toBe(0);
```
icfantv: should check for `length === 1` and that the child is our modal.

src/modal/test/modal.spec.js
```diff
@@ -833,6 +833,50 @@ describe('$uibModal', function () {
...
+      });
+
+      it('should throw error if container is not found', function() {
```
icfantv: Still not sure how we want to handle the error.  If we throw something, should we display a link to an error page like core Angular does (and other Angular libraries)?

src/modal/modal.js
icfantv: What if someone passes in an `angular.element` with more than one object?  E.g., `angular.element('div')`.

wesleycho: We can explicitly say we don't support that

icfantv: I still think we need to only grab the first one.  If I'm reading [this](http://api.jquery.com/append) right, if someone passed in multiple objects, the modal would get appended to each of them.

wesleycho: Does jqLite support it? That is my main concern

icfantv: Sorry.  Support what?  jqLite supports both `.eq()` and `.append()` with no strings attached (according to [this](https://docs.angularjs.org/api/ng/function/angular.element)).

wesleycho: Ah, thought for some reason Angular didn't support `.eq` in jqLite for a while - looks like I was wrong.

aroop: I guess logging error makes more sense here as user has explicitly passed the invalid container element. Let me know if you want me to default it to `body`.

src/modal/modal.js
aroop: Updated error message to `Container not found. Make sure that the container passed is in DOM.`

src/modal/test/modal.spec.js
aroop: Thanks. Combined into a single line.

src/modal/test/modal.spec.js
aroop: Thanks. Now tests are passing element instead of selector.

src/modal/test/modal.spec.js
aroop: Thanks. Combined into one line

src/modal/test/modal.spec.js
aroop: Thanks. Changed to `element`

src/modal/test/modal.spec.js
aroop: Thanks. Changed code. Now test is making more sense.

src/modal/test/modal.spec.js
aroop: I don't think we should handle this error as there is a need for the user to fix the expectation. Let me know if you want me to point the error to docs page.

wesleycho:
Please clean up the history - there should not be any merge commit. Please use rebase instead.

May need to do
`git checkout b7c56c2; git pull --rebase origin master`, where `origin` is pointing to this repository.

aroop:
This PR also fixes https://github.com/angular-ui/bootstrap/issues/2688 & https://github.com/angular-ui/bootstrap/issues/3686

wesleycho:
Changing api to be `appendTo` instead of container, but otherwise great work - merging shortly.

georgehrke:
Hey,
I'm using angular ui bootstrap via bower. Any ETA when this feature will be released?
Thx :blush: 
Georg

PR #4610, opened 2015-10-13
fix(tooltip): clean up stackedMap on scope destroy

### LINKED ISSUE #4604: $uibTooltip creates massive memory leak
Each usage of `<span uib-tooltip="Hey"></span>` creates a memory leak on each state change with `ui-router`.

When profiling the heap allocation of this [plnkr](http://plnkr.co/edit/Sq5KAfMcTzmCt2MfJlUa) memory leaks can be seen with tooltips:
screen shot 2015-10-13 at 2 32 46 pm
and less when there are no tooltips
screen shot 2015-10-13 at 2 33 00 pm

RobJacobs:
The tooltip was not being removed from the stackedMap
resulting in a memeory leak.  Tooltip will now be
removed from the stackedMap in scope destroy
function.

Fixes #4604

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -486,6 +486,7 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
               $timeout.cancel(positionTimeout);
               unregisterTriggers();
               removeTooltip();
+              openedTooltips.remove(ttScope);
               ttScope = null;
             });
           };


wesleycho:
LGTM

Foxandxss:
Please, add tests.

wesleycho:
Not sure we can add a test here, the stacked map is not exposed publicly.

Foxandxss:
You inject the stacked map in the test, you add a few tooltips, you destroy one, you check that the stack has one tooltip less.

wesleycho:
But the stack that needs to be checked is the internal one - that is the one that gets removed

Foxandxss:
@RobJacobs tell us your secrets to close a PR as merged like you did without creating extra unwanted history messages.

RobJacobs:
@Foxandxss When working in a PR branch, after I make the initial commit I always use 'git commit --amend' to keep my commit history clean.  Then when I merge to master:

    git checkout master
    git pull <fork-url> <branch>
    git push origin

PR #4621, opened 2015-10-14
fix(tooltip): delay timeouts

### LINKED ISSUE #4618: Tooltip not closing
Hi since last update (1.4.2) All my tooltips are not closing.
This is reproducible when you have 2 or more tooltips one near each other and you hover over them.

```html
<span uib-tooltip="Some tooltip" tooltip-popup-delay="300" tooltip-popup-close-delay="300">Some text</span>
```

RobJacobs:
The show and hide delay timeouts where not getting
cancelled correctly resulting in tooltips staying
open when both popup and popup-close delays were
in use.

Fixes #4618


RobJacobs:
Need to figure out why this [test](https://github.com/angular-ui/bootstrap/blob/master/src/tooltip/test/tooltip.spec.js#L278) did not fail

RobJacobs:
Tests added, should be good to go.

wesleycho:
I like this - also makes it more readable.

PR #4623, opened 2015-10-14
fix(tooltip): scrollbar flashing

### LINKED ISSUE #4458: Tooltip causing scroll bars to quickly show

RobJacobs:
When the tooltip is rendered for positioning scroll bars
may breifly appear depending on where the linked element
is positioned. This change adds a negative top and left
style to push the tooltip element out of view while it
is being measured and positioned.

Closes #4550
Fixes #4458


wesleycho:
LGTM

PR #4630, opened 2015-10-15
fix(buttons): double toggle on spacebar

### LINKED ISSUE #4474: Clicking space on a checkbox button
There is code to toggle a checkbutton with space. If you go to the demo and do it, it will toggle twice. I haven't checked it yet so I open this issue to not forget.

Foxandxss:
This is basically a revert of bdfb289.

A spacebar click is treated as a normal click event so it will trigger the toggle twice.

I don't think there is a proper way of testing this.

[Plunker](http://plnkr.co/edit/1hx7Hc5UBJdsrvwlYIDD?p=preview)

Fixes #4474


wesleycho:
LGTM

Only thing I might suggest is maybe update the docs on advice for accessibility, but I'm fine with this being merged as in.

Foxandxss:
What you mean with "on advice for accessibility"?

wesleycho:
Nesting an input tag so the input handles all of the accessibility requirements.

PR #4635, opened 2015-10-15
fix(datepicker): fix support for literal format on popup

### LINKED ISSUE #4616: uib-datepicker-popup formatting problem
Hi,

There seems to be a problem with uib-datepicker-popup and formatting. I am currently using version 0.14.2 but I can confirm that affected was also 0.14.1. The problem is when I declare in html datepicker popup like this:

`<input ...uib-datepicker-popup="dd/MM/yyyy"... />` - the date will not be in displayed

When I fallback to 

`<input ...datepicker-popup="dd/MM/yyyy"... />` - the date will be displayed correctly but I see the deprecation message.

It seems to be a regression in behavior for datepicker-popup. Also note that 
uib-datepicker-popup="yyyy-MM-dd" works as expected. But when you change the format to anything different the dates are not displayed anymore.

Foxandxss:
Since the controller is used by old and new, I need to add an extra checking. I will add this to "What to remove" issue for when we remove deprecated stuff.

Fixes #4616


wesleycho:
LGTM

PR #4641, opened 2015-10-15
feat(datepicker): Disabled Date New Functionality

egerrard:
There were two things I needed from the date picker that didn't exist.  I wanted to skip an entire month when shuffling through months if that entire month was disabled.  And I wanted to prevent people from scrolling further back then the first month with an available date and past the last month with an available date.  So the following changes resolve this for my specific date picker:
- Added functionality to allow blocking out and skipping of entire
months from the date picker.  I am passing in a day, month, and year range to calculate if a whole month is disabled and if I should skip the month.
- Disabled back month button when at the start and forward button when
at the end date.  I added an ng-disabled to the template and have it running through a new function to calculate weather it is the first or last month available.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -17,7 +17,13 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   yearRange: 20,
   minDate: null,
   maxDate: null,
-  shortcutPropagation: false
+  shortcutPropagation: false,
+  minMonth: 0,
+  minYear: 0,
+  maxMonth: 0,
+  maxYear: 0,
+  disabledStart: {},
+  disabledEnd: {}
 })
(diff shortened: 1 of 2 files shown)


wesleycho:
This needs tests - in addition, indentation in the project is 2 spaces, not 4.

Foxandxss:
Please no please, no more datepicker functionality, it is already a monster and it is a pain to maintain.

egerrard:
@Foxandxss this was something that came in very handy for my very specific use case but I totally understand if you don't think it is practical to add to the project out of the box.  I posted it here to share with others but I'm very new to the contributing world so if there is a more practical way to share this with people while not including it in the full product I'm happy to do that instead.  Thanks!

wesleycho:
Going to close this as something we do not want to support - this seems like a pretty complex feature that most users likely will not use.

PR #4643, opened 2015-10-16
Fix dropdown auto close in mobile phones for angular ~1.4.7

solispauwels:
Auto close dropdowns when click was broken in mobile phones with angular ~1.4.7

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -9,6 +9,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
 
   this.open = function(dropdownScope) {
     if (!openScope) {
+      $document.bind('touch', closeDropdown);
       $document.bind('click', closeDropdown);
       $document.bind('keydown', keybindFilter);
     }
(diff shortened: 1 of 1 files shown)


icfantv:
Please create a functional test that checks for this event handler being added, triggered, and removed.

src/dropdown/dropdown.js
```diff
@@ -23,6 +24,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
   this.close = function(dropdownScope) {
     if (openScope === dropdownScope) {
       openScope = null;
+      $document.unbind('touch', closeDropdown);
```
icfantv: Nice job on remembering this - it's REALLY easy to forget to do.

solispauwels: I close it because I realize that the thing that was wrong and nonfunctional were the $document.bind (mobilephone) in angular ~1.4.7 so my code does not fix anything... I am looking what is going on.

solispauwels: In order to be more precise, $document.bind is not working only in A tags in mobilephone

solispauwels: The problems seems to happend only in elements with ng-click attribute

solispauwels:
ngTouch rewrites ngClick and cause problems in dropdowns that are not auto close properly.

comment angular-touch.js line 350 and 351 fix the problem.

solispauwels:
https://github.com/angular/angular.js/pull/13122

PR #4646, opened 2015-10-17
chore(docs): add docs versioning

### LINKED ISSUE #3670: documentation versioning - hard to tell what version features apply to
Just spent way more time than was necessary trying to resolve why tooltip-class did not work in our application only to find out (via commit history here) that it was a brand new feature only in the latest release... 

After going back to the documentation I looked for indications of what version I might have been looking at and the only reference to the actual project version # was in the Download button at the top of the page. I know other projects have done it in the past  and it may not be the simplest thing, but being able to specify what version of documentation you are looking at would have been a huge boon to solving this in a more timely fashion. 

As per @Foxandxss here is a ticket to help track this issue

Foxandxss:
So here it is.

We will now have a dropdown on the navbar to have versioned docs. With a new autogenerated json, you will see the latest 10 versions on the dropdown.

To avoid previous docs from having a dropdown with "their own" versioned docs numbers, I put the json on the root so there is just one json. Also I am only activating the dropdown on the latest version.

The only downside I can see (apart from having to upload versioned docs by hand, which I will see how to solve in a future), is that we need to upload the actual build in the same directory of the demo. Not a big deal since github allows you to have 1GB~ of data in gh-pages and a set of 10 versioned docs is like 8mb.

After this get merged, I will update the `gh-pages` with the real stuff.

Fixes #3670


wesleycho:
So with this feature, we will need to update the wiki with the new process for adding the docs to gh-pages

Foxandxss:
Oh please, I named you "Sir release cutter and gh-pages updater".

Alright, will add it too.

Gruntfile.js
```diff
@@ -392,6 +392,29 @@ module.exports = function(grunt) {
                   grunt.config('meta.banner'), grunt.config('meta.cssFileBanner'));
   });
 
+  grunt.registerTask('makeVersionsMappingFile', function () {
+    var done = this.async();
```
wesleycho: I believe you can just pass done as an argument to the callback defining the task

Foxandxss: That is gulp. I saw this in the docs, but lemme try.

Foxandxss: Yeah, I was right.

Foxandxss:
The other thing is. Should I hack 0.14.2 demos to allow versioning and upload this asap or we put the versioning after next release?

Gruntfile.js
wesleycho: Ah, that is a weird api

Foxandxss: Oh please, I cursed grunt at least 10 times today and it was a silly task.

misc/demo/assets/app.js
```diff
@@ -47,6 +47,19 @@ angular.module('ui.bootstrap.demo', ['ui.bootstrap', 'plunker', 'ngTouch', 'ngAn
 .controller('DownloadCtrl', DownloadCtrl);
 
 function MainCtrl($scope, $http, $document, $uibModal, orderByFilter) {
+  $scope.latestVersion = false;
+
+  // Grab old version docs
+  $http.get('versions-mapping.json')
+    .then(function(result) {
+      $scope.latestVersion = true;
```
wesleycho: Does this imply the versions-mapping.json will only be in the root directory? Shouldn't the dropdown be available for the old docs too?

Foxandxss: Yes, it does that. I felt like it was a better idea to just have the menu on the latest version.

We don't want each version to have a .json because each versioned doc will have a dropdown with their own list of "old docs".

Let's invent numbers in here. We are in version 20 so we show docs from version 10 to version 19.  If we keep the .json with each update, they all will have their own ranges. So if we go to version 15, we will have a drop down from version 5 to version 14 and that is not what we want.

IIRC, we decided to version only the X latest versions. so as soon as we release a new version, we would remove the oldest one from versioning.

Anyway, .json is in the root because it will contain only one list of versioned docs and we don't want each copy to have their own.

The dropdown available (with the same .json file) is something to discuss. I would need to hack all versioned docs to have it.

misc/demo/index.html
```diff
@@ -50,6 +50,16 @@
                     </ul>
                 </li>
                 <li><a href="#getting_started">Getting started</a></li>
+                <li ng-show="latestVersion" class="dropdown" uib-dropdown>
+                    <a role="button" class="dropdown-toggle" uib-dropdown-toggle>
```
wesleycho: dropdown and dropdown-toggle classes should be unnecessary - same with dropdown-menu below as well I believe

Foxandxss: I am copying the same style as the "Directives" dropdown.

Gruntfile.js
```diff
@@ -392,6 +392,29 @@ module.exports = function(grunt) {
...
+
+    var versionsMappingFile = 'dist/versions-mapping.json';
+
+    exec('git tag --sort -version:refname', function(error, stdout, stderr) {
+      var versionedDocs = 10; // 10 is a safe number of versioned docs.
+      var versions = stdout.split('\n').slice(1, versionedDocs);
+      var jsContent = versions.map(function(version) {
+        return {
+          version: version,
+          url: '/versioneddocs/' + version
```
wesleycho: Maybe add a - in there, i.e. versioned-docs. IMO, it's easier to read

Foxandxss: Surething.

misc/demo/index.html
wesleycho: So my concern here is double registration of directives since we still support class-based directives

Foxandxss:
Ok, so now it will have dropdown in all the versioned docs. But we will only support 0.12+

wesleycho:
LGTM - let's start the doc versioning migration!

PR #4650, opened 2015-10-18
revert(collapse): remove $animateCss

### LINKED ISSUE #4561: Nested accordions do not expand properly
When an accordion is nested inside another, the child accordion does not expand the height of its parent.

You can see the issue in this plunkr: http://plnkr.co/edit/Vv0DWehGhyi6eYkQCqvN?p=preview. It works correctly in ui-bootstrap version 0.12.1, but if you change the version to anything higher (including 0.14.0), it breaks.

This is how it looks in version 0.12.1 (correct):

correct

This is how it looks in any version after 0.12.1 (broken):

broken

### LINKED ISSUE #4628: collapsing navbar issue
hi,

i noticed one interesting the collapse navbar.

http://embed.plnkr.co/FQWDqyFnDonUJdv1bPpr/preview
(latest bootstrap, angular-ui bootstrap, angularjs)

http://embed.plnkr.co/XVxhkRkSYPb1E8Tz6PDu/preview
(latest bootstrap, angularjs, angular-ui bootstrap 0.13.4)

in-and-out closing update navbar-collapse class: `collapsing` (latest angular-ui bootstrap)
in-and-out closing update navbar-collapse class: `collapse` (angular-ui bootstrap 0.13.4)

above the navbar 768px height has 0px in inline style.
the `collapsing` class does not work over 768px.
something is not right.

thanks for the help!

### LINKED ISSUE #4647: Accordion body incorrect height calculation
Accordion body height is calculated incorrectly when it has hidden content. Please check [plnkr](http://plnkr.co/edit/ZhPtRFB11zq3EXsClsmF?p=preview)

Initially it has height that exceed visible content. After toggling body is shown correctly, but when I'm toggling hidden content it shows cutted.

Foxandxss:
This reverts commit 6f9f1fc.

$animateCss is giving lots of side effects and I see no real reason to have it. Reading the official docs I can read:

> $animateCss is designed to be used inside of a registered JavaScript animation that is powered by ngAnimate.

So it seems that using $animateCss here is a bit overkill and it is adding lot of weird complexity and issues. I tried to fix them, but only more issues were coming out. It is better to leave it as it was before because it works perfect with all the angular versions we support.

The majority of issues comes without `ngAnimate` activated

Let's see those side issues.

First one: If there is a hidden element on screen, $animateCss will computate its height:

[Broken Plunker, no ngAnimate](http://plnkr.co/edit/ZhPtRFB11zq3EXsClsmF?p=preview)
[Broken Plunker, ngAnimate](http://plnkr.co/edit/hAUAXLRX21zlDcVBInhc?p=preview)
[no $animateCss, no ngAnimate](http://plnkr.co/edit/ORVslK51UjaRDllzq8Vc?p=preview)
[no $animateCss, ngAnimate](http://plnkr.co/edit/CLAkGK115UD4hLB83vru?p=preview)

Second one: I don't know how to explain this one, but with $animateCss, it is not showing the navbar correctly (just drag it for the responsive behavior):

[Broken Plunker, no ngAnimate](http://embed.plnkr.co/FQWDqyFnDonUJdv1bPpr/preview)
(Works with ngAnimate)
[no $animateCss, no ngAnimate](http://plnkr.co/edit/49p7sMkdNcvi4mYGeSh2?p=preview)
(Works the same with ngAnimate)

Third one: Nested accordions not being rendered correctly:

[Broken Plunker, no ngAnimate](http://plnkr.co/edit/Vv0DWehGhyi6eYkQCqvN?p=preview)
(Works with ngAnimate)
[no $animateCss, no ngAnimate](http://plnkr.co/edit/ga5EWSgV5OZfWe0HnquX?p=preview)
(Works the same with ngAnimate)

So the biggest issues are when we don't use `ngAnimate` (people not caring about animations) and in some other cases even with Animation.

I will probably raise an issue at the angular repo, but even if it ends being a bug, we cannot force all the users to move to the fixed version (not everyone can update) and the fix could take weeks. So I propose to revert this and if someday we see a real reason to use `$animateCss` we will use it, but the docs suggest me that we are barking to the wrong tree with this implementation.

Fixes #4647 
Fixes #4628 
Fixes #4561


Foxandxss:
Well, I am not happy if I don't really find the source. I tried to reproduce it and it seems like it was fixed in 1.4.5. Still, there is an issue with the current implementation so I will open a new PR.

hthetiot:
thx, fixed my issue.

PR #4659, opened 2015-10-19
fix(modal): Replace unnecessary querySelectorAll with querySelector

crisbeto:
In this case only the first element that is returned from `querySelectorAll` is being used, its better to use `querySelector` which stops looking after the first match.


wesleycho:
This is a good optimization - thanks for this!

PR #4660, opened 2015-10-19
chore(docs): fix uib-popover-tempate description

### LINKED ISSUE #4654: uib-popover-template unclear
The docs [say](https://angular-ui.github.io/bootstrap/#/popover):
> takes text that specifies the location of a template to use for the popover body. Note that this needs to be wrapped in a tag.

Questions:
1. What is meant by "wrapped in a tag"?
2. Is it the case that it evaluates the value of this attribute as an expression? That seems to be the case.
3. Can this be a link to an external file, or does it just work for inline templates? I tried to use it to point to an external template and it wasn't working. A demo of how to do that would be useful.

icfantv:
* clarify description for `uib-popover-template` attribute.

Fixes #4654


Foxandxss:
:+1:

Issue #4638, opened 2015-10-15
Change $modalInstance to $uibModalInstance


ravishivt:
Shouldn't this provider's name be updated to conform?

pkozlowski-opensource:
Yeh, we could do this for consistency.

wesleycho:
It should be noted that this is a breaking change - I'm fine with this breaking change. This could affect the regular modal service users since the code is not duplicated.

wesleycho:
Scheduling this for 1.0 since this will be a breaking change with a bit of potential pain.

wesleycho:
Figured out a way to get this to work in 0.14.x and play nicely - will have PR up shortly.

ravishivt:
Great, thanks!  I opened another similar request.  https://github.com/angular-ui/bootstrap/issues/4670

PR #4661, opened 2015-10-19
feat(modal): support $uibModalInstance

wesleycho:
- Add support for `$uibModalInstance` as a controller local
- Deprecate `$modalInstance` support

This implements #4638.


Foxandxss:
That is definitely clever Wes. :+1:

PR #4666, opened 2015-10-20
fix(alert): allow interpolations with dismiss-on-timeout

### LINKED ISSUE #4665: Alert is created and dissappears when dismiss-on-timeout contains a variable
I recently updated from version 0.13.3 to 0.14.1 and recognized that using a variable as timeout immediately hides the alert after showing. When you put a breakpoint to dom modification in chrome, you see that it appears and then gets removed immeditely after.

I made a little demo here http://plnkr.co/edit/6dEUew1athBdkrGDXFt8?p=preview

Foxandxss:
[Plunker](http://plnkr.co/edit/Ce3MulWm3bNltqtByBY2?p=preview)

Fixes #4665


src/alert/alert.js
```diff
@@ -1,12 +1,15 @@
 angular.module('ui.bootstrap.alert', [])
 
-.controller('UibAlertController', ['$scope', '$attrs', '$timeout', function($scope, $attrs, $timeout) {
+.controller('UibAlertController', ['$scope', '$attrs', '$interpolate', '$timeout', function($scope, $attrs, $interpolate, $timeout) {
   $scope.closeable = !!$attrs.close;
 
-  if (angular.isDefined($attrs.dismissOnTimeout)) {
+  var dismissOnTimeout = angular.isDefined($attrs.dismissOnTimeout) ?
+    $interpolate($attrs.dismissOnTimeout)($scope.$parent) : null;
```
wesleycho: Why not just use `$scope.$parent.$eval`? Is there something superior with interpolation here?

Foxandxss: That won't work.

wesleycho: Fair enough, PR LGTM then

bartrail:
thx for the quick fix

jmsunseri:
Thanks for fixing this!!!

PR #4668, opened 2015-10-20
chore: remove stale ngDocs folder

Foxandxss:
That thing that is in front of your eyes for so long that you never realized that it is not doing anything.

I checked the gruntfile and it seems to be just stale stuff.


wesleycho:
LGTM

PR #4674, opened 2015-10-21
chore(demo): use prefixed $modalInstance

Foxandxss:


wesleycho:
Never realized the demo used the modal too - LGTM

Foxandxss:
easy to skip because that is only triggered for the custom build or download modal.

PR #4676, opened 2015-10-21
feat(datepicker): preserve timezone with model

wesleycho:
- Default to using the date instance for any date manipulation, allowing
  for the timezone to automatically be preserved


wesleycho:
Note that this removes the `fixTimeZone` hack - the hack becomes unnecessary because we are using the existing date instances for all datetime manipulation.

PR #4681, opened 2015-10-21
Actually check if 'close' is in $attr

tnajdek:
Documentation states that "If the attribute close exists, a close button is displayed as well.", however this will not work in the following cases:

    <uib-alert close>

    <uib-alert close="">

Tiny change I'm proposing should fix that problem :)


wesleycho:
That will make the close button not do anything, which would be bad UI & UX - this doesn't make sense.

Closing this as invalid.

tnajdek:
Yeah you're absolutely right, sorry about this. I had a weird case where I tried to use that but ended up refactoring.

PR #4697, opened 2015-10-23
fix(tooltip): null scope check in isOpen watch

### LINKED ISSUE #4638: Change $modalInstance to $uibModalInstance

### LINKED ISSUE #4683: 'isOpen' of undefined $destroy popover
I thought this was fixed in #4552 but seems not, when destroying a directive this shows:
TypeError: Cannot set property 'isOpen' of undefined

            if (isOpenParse) {
              scope.$watch(isOpenParse, function(val) {
                /*jshint -W018 */
                if (!val === ttScope.isOpen) {
                  toggleTooltipBind();
                }
                /*jshint +W018 */
              });
            }

Shouldnt there be a ttScope check before?

RobJacobs:
The watch for the isOpen property from the model was not
checking to see if the internal tooltip scope was still
valid (null or undefined) before comparing the value to
the tooltip scope isOpen property.

Closes #4697
Fixes #4683

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -368,7 +368,7 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
             if (isOpenParse) {
               scope.$watch(isOpenParse, function(val) {
                 /*jshint -W018 */
-                if (!val === ttScope.isOpen) {
+                if (ttScope && !val === ttScope.isOpen) {
                   toggleTooltipBind();
                 }
                 /*jshint +W018 */


RobJacobs:
I'm not sure how to add a test for this.  The scenario is the element with the tooltip attached has it's scope destroyed and falls into the scope.$on('$destroy' function where ttScope is set to null.  The person reporting the problem (with no plunk) claims an error is being thrown (Cannot set property 'isOpen' of undefined) in the watch on the element scope isOpen [here](https://github.com/angular-ui/bootstrap/blob/master/src/tooltip/tooltip.js#L369).  How would I set the element scope isOpen property to trigger the call to the watch function after the element scope is destroyed?

PR #4698, opened 2015-10-23
chore(docs): popover close delay

RobJacobs:
The attribute for the close delay on the popover was incorrectly
documented as 'popover-close-popup-delay' and should be
'popover-popup-close-delay.


wesleycho:
I noticed that and was suspicious - LGTM

PR #4701, opened 2015-10-23
chore(buttons): update docs

### LINKED ISSUE #3562: Cleanup Demo Inconsistency and Readability

Foxandxss:
This was really needed.

Fixes #3562


wesleycho:
Ditto on the milestone questions here - LGTM

src/buttons/docs/readme.md
```diff
@@ -1,3 +1,36 @@
-There are two directives that can make a group of buttons behave like a set of checkboxes, radio buttons, or a hybrid where radio buttons can be unchecked.
+With the buttons directive, we can make a group of buttons to behave like a set of checkboxes (`uib-btn-checkbox`) or to behave like a set of radio buttons (`uib-btn-radio`).
```
icfantv: * make "buttons", `buttons` since it is a directive.
* remove the "to" in "...buttons to behave..."
* remove the "to" in "...or to behave like..."

src/buttons/docs/readme.md
```diff
@@ -1,3 +1,36 @@
...
+### uib-btn-radio settings
+
+* `uib-btn-radio` -
+  Value to assign to the `ng-model` if we check this radio button.
+
+* `ng-model` -
+  Model where we set the radio status. All radio buttons in a group should use the same `ng-model`.
+  
+* `uncheckable`
+  _(Boolean attribute)_ -
```
icfantv: is there a default value here?

src/buttons/docs/readme.md
Foxandxss: Didn't highlight because it is not talking about the directive in concrete, but html buttons. 

The other, yeah :)

icfantv: Ah yea.  Good point.

src/buttons/docs/readme.md
Foxandxss: Uhmmm, yeah... But since it is a "boolean" attribute, I felt like it was implicit.

PR #4734, opened 2015-10-26
chore(docs): downloads - promote npm

RobJacobs:
In the download modal on the demo page, changed Bower info to suggest using npm instead.


wesleycho:
LGTM

PR #4740, opened 2015-10-26
fix(modal): clean up animation when disabled

### LINKED ISSUE #4672: modal behaviour changed
there is a change in the modal behaviour. It can be seen at the modal example on https://angular-ui.github.io/bootstrap/

basically when animation is turned off, it appears the screen behind the modal is very obviously deleted and redrawn as the modal is closed.

wesleycho:
- When the animation is set to false, do not remove the `in` class to prevent black opaque background from showing

[Here](http://plnkr.co/edit/2dsgwarH9CEgyc5P7RNW?p=preview) is the Plunker in action - compare with the current behavior [here](http://plnkr.co/edit/B0CBWc2NvYkezYUX1e2I?p=preview) when the animations are disabled in both and the modal is dismissed.

This fixes #4672.


Foxandxss:
Nice, that error was ugly.

wesleycho:
I take that as a vote of approval? :)

Foxandxss:
Yes, sir.

PR #4742, opened 2015-10-27
refactor(carousel): a bit faster for

Foxandxss:
Not a big deal, but this had the ideas in place but they were never used. On the other hand, I used === instead of ==


wesleycho:
The `==` bothered me too - LGTM

PR #4749, opened 2015-10-27
chore(carousel): add more options to the demo

Foxandxss:
I added an option to see how the `index` attribute on the `slide` can be used. It adds lot of complexity to the Javascript code, but I added a comment so people can see the separation between "demo code" and "randomize logic".

I like to have the demos the more complete as possible, and this option is not that easy to deduce how it works and not easy to explain via simple docs.

Stole the shuffle algorithm from SO, serves as a good library of easy algorithms.


src/carousel/docs/demo.html
```diff
@@ -1,10 +1,10 @@
 <div ng-controller="CarouselDemoCtrl">
   <div style="height: 305px">
     <uib-carousel interval="myInterval" no-wrap="noWrapSlides">
-      <uib-slide ng-repeat="slide in slides" active="slide.active">
+      <uib-slide ng-repeat="slide in slides | orderBy: 'id'" active="slide.active" index="$index">
```
wesleycho: Wouldn't it be better to do `index="slide.id"`?

Foxandxss:
Updated.

PR #4752, opened 2015-10-27
fix(typeahead): clear typeahead input when value not from popup entered and editable is false

### LINKED ISSUE #4265: Clear typeahead input when value not from popup entered
When using `typeahead-editable="false"`, the model bound to the typeahead is undefined on blur when a value is not selected from the popup.  Unfortunately, the input still contains the value entered.  The expected/desired behavior for me is to clear the input in such a situation.

chenyuzhcy:
Closes #4265


BobbieBarker:
Lgtm thanks for the pr.

wesleycho:
My only question here is whether we should null the model as well in this situation. Otherwise, I'm fine with this PR.

chenyuzhcy:
The model is already undefined in this situation

wesleycho:
LGTM - thanks for the high quality PR!

kencaron:
This might not be the place for this -- found this PR when trying to debug an issue with typeahead-editable in the latest release (0.14.3).

Any ETA on the 14.4 release? Couldn't find anything specific in the roadmap wiki. I use `typeahead-editable="false"` heavily to prevent users from entering garbage in certain dropdowns.

icfantv:
@kencaron, there will not be a 0.14.4 release.  we are trying to get 1.0 out the door and the release will be very soon.  If you need this right away, please feel free to pull `master` and build a SNAPSHOT version.

bkirtika:
This will clear input with same match entered from popup also.

PR #4759, opened 2015-10-28
chore: add the current version of the docs

### LINKED ISSUE #4707: Fix version-mapping generation
It should be done after the tag is created.

Foxandxss:
The early problem is that on the current release process, you would generate `versions-mapping.json` with the old tags.

So if we release 1.0.0 today (heh), the tag wouldn't be there so the versions would be:

`0.14.3, 0.14.2, 0.14.1`...

Then rename the first one to:

`Current, 0.14.2, 0.14.1`...

It will skip `0.14.3` in this case.

Now, it just uses the tags as older versions and then it will add the current on top, so it would result in:

`Current, 0.14.3, 0.14.2`...

Which is what we want.

Fixes #4707


wesleycho:
LGTM

Issue #2428, opened 2014-07-07
Proposal: add semantic classes to datepicker directive (and possibly others as well)


TiddoLangerak:
I recently wanted to style the datepicker using custom css. Unfortunately I found that styling the datepicker cannot be done cleanly. Since the datepicker *only* has default bootstrap classes assigned to elements there are 2 options for styling the datepicker:

- Copy-paste the default template and change classes in the copy
- Override default styles using non-semantic selectors (e.g. `[datepicker-popup-wrap] table .btn` to target the calendar buttons).

The first option is fragile because it requires manually keeping the copied template up-to-date with upstream, and the second option is fragile since it gives little control and it has a very high coupling with the html structure, which again might be changed upstream.

Common practice is to include semantic class names in the html, such that these can be targeted very precisely using css only (proper separation of concerns). This technique should be applied to this directive as well.

This problem is likely not limited to the datepicker alone, but I simply have not checked the other directives.

danielbsig:
+1 for semantic class names. At the very least a few carefully chosen elements could be chosen, and then developers could style the rest using CSS child selectors. There is still some risk of breaking that CSS if the markup structure changes in future versions, but at least those changes could be isolated within these carefully selected elements.

brbradford:
+1

wesleycho:
Forgot to add in commit - this is addressed by 97c4333

PR #4761, opened 2015-10-28
feat(datepicker): add semantic classes

wesleycho:
- Add semantic classes to template to assist in styling

This addresses #2428.


Foxandxss:
Sounds good to me. But perhaps we should do the others as well.

wesleycho:
Agreed

PR #4763, opened 2015-10-28
feat(buttons): improve accessibility

wesleycho:
- Add modifiers to improve accessibility of button group component

Used these links as guides:
http://www.w3.org/WAI/GL/wiki/Using_grouping_roles_to_identify_related_form_controls
http://oaa-accessibility.org/example/29/
http://www.oaa-accessibility.org/examplep/checkbox4/


wesleycho:
It should be noted that this PR may have a slight issue due to the mixing of aria attributes and input elements. This needs a thorough test through.

wesleycho:
Going to close this due to potential issues.

PR #4765, opened 2015-10-28
fix(tooltip): race condition when setting position

### LINKED ISSUE #4757: tooltip race condition. throws TypeError: Cannot read property 'css' of null
Tooltip directive has a race condition [on this line](https://github.com/angular-ui/bootstrap/blob/master/src/tooltip/tooltip.js#L157) and [this line](https://github.com/angular-ui/bootstrap/blob/master/src/tooltip/tooltip.js#L315).

If you have two tooltips and hover from one element to another a few times you will see the error.

I've created this [plunker](http://plnkr.co/edit/nEPbSMvo5nZ9yRNaJMRf?p=preview).

And attached these two screenshots to show you the error:

captura de tela 2015-10-28 as 02 13 50
captura de tela 2015-10-28 as 09 38 48

I tested with tooltip and tooltip-html and both have the same issue.

```
var tooltip;
...
if (!tooltip || !tooltip.html()) { return; }
if (!positionTimeout) {
   positionTimeout = $timeout(function() {
      // Reset the positioning.
      tooltip.css({ top: 0, left: 0 }); // throws TypeError
...
```

You're checking tooltip is null or undefined and setting a timeout, but if before the timeout triggers, `removeTooltip()` is called, you set `tooltip = null;` [here](https://github.com/angular-ui/bootstrap/blob/master/src/tooltip/tooltip.js#L315) and the exception occurs after timeout triggers.

It doesn't break the ui, but I've setted `$exceptionhandler` in my app to send me an email with the error, and I'm getting this error a lot (like 50 times yesterday), so I think it happens quite frequently in a normal use case with "common users".

A "double-checked locking" would solve the error, I mean, just check `if (!tooltip || !tooltip.html()) { return; }` again inside the timeout, before set position.

If you agree, I can open a PR with this.

RobJacobs:
A race condition could occur when there is an open delay that
doesn't get cancelled after the transition delay resulting in
the position logic getting called.  This will then try to set
the css of the tooltip after is has been destroyed in the
removeTooltip function.

Fixes #4757


wesleycho:
LGTM

PR #4766, opened 2015-10-28
test(dateparser): missing tests, refactorized tests

Foxandxss:
I added 3 tests for missing formats, removed two that are not doing anything (one of them is more or less working, but I will rework those when I fix something else on dateParser).

Also used the helper we have in the others.


wesleycho:
LGTM

PR #4770, opened 2015-10-28
docs(datepicker): Add Validation Warning To Datepicker Documentation

### LINKED ISSUE #3440: Documentation - The datepicker documentation does not inform programmers about the date-disabled validation error produced if a manually-entered date is outside of the min/max-date range.
Functional reproduction Plunker can be found [here](http://plnkr.co/edit/XdUTSlx9koKFHctohdlg?p=preview)  

Additionally, refer to your public-facing [documentation](https://angular-ui.github.io/bootstrap/), in the Datepicker section.

The `min-date` and `max-date` property entries do not note that if a given date is outside of their date range, that the `date-disabled` validation error will be added to the form (refer to Bug #3438 and Bug #3439 for additional problems posed due to the current implementation.)  This is critical information for a UI programmer to have in order to raise hint messages to the user to help them enter valid information.

**Steps to Reproduce:**
 # Open the Angular UI Documentation link given above
 # Navigate to the Datepicker section
 # Find the entries for the `min-date` and `max-date` attributes
 # Observe that the information does not denote that `date-disabled` is raised on the form if the calendar's date lies outside of this range.

**Observed Behavior:**
 # Documentation does not denote the consequences of a user entering a date outside the min-/max-date range.

**Expected Behavior:**
 # Documentation should denote the consequences of a user entering a date outside the min-/max- date range.

yakirac:
Date-disabled validation error if date range falls out of min-/max-date range

Closes #3440

### DIFF
--- a/src/datepicker/docs/readme.md
+++ b/src/datepicker/docs/readme.md
@@ -28,7 +28,7 @@ All settings can be provided as attributes in the `uib-datepicker` or globally c
  * `date-disabled (date, mode)`
  	_(Default: null)_ :
  	An optional expression to disable visible options based on passing date and current mode _(day|month|year)_.
- 	
+
  * `custom-class (date, mode)`
  	_(Default: null)_ :
  	An optional expression to add classes based on passing date and current mode _(day|month|year)_. 	
@@ -79,7 +79,7 @@ All settings can be provided as attributes in the `uib-datepicker` or globally c
 
  * `year-range`
  	_(Default: 20)_ :
- 	Number of years displayed in year selection. 
+ 	Number of years displayed in year selection.
 
  * `shortcut-propagation`
   _(Default: false)_ :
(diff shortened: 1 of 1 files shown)


src/datepicker/docs/readme.md
```diff
@@ -154,4 +154,7 @@ Depending on datepicker's current mode, the date may refer either to day, month
  * `Enter`/`Space`: Select date.
  * `Ctrl`+`Up`: Move to an upper mode.
  * `Ctrl`+`Down`: Move to a lower mode.
- * `Esc`: Will close popup, and move focus to the input.
\ No newline at end of file
+ * `Esc`: Will close popup, and move focus to the input.
+
+**Known Issues**
```
wesleycho: I believe the tone of the header is wrong here - it is certainly known, but it isn't something that should be considered an issue.

yakirac: @wesleycho Perhaps 'Note' or 'Notes' would be better here?

Foxandxss:
The problem I see here is that this links to an issue that links to two open issues and links to more open issues regarding validation.

Adding docs in here is like saying: Intended behavior for all of them and we will just add a note in here.

I think we should see first if those issues are legit or not and if it is just an issue on the doc, fix the doc.

If we end fixing the doc, I would personally add a note in both min / max entries to say that a date below / after that one would trigger that validation error.

yakirac:
@Foxandxss Thanks for the explanation. I thought the developer may want a heads up on what the current behavior is, which prompted the header of 'Known Issue', until a resolution on the other issues is made.

PR #4774, opened 2015-10-29
fix(carousel): remove version checks

### LINKED ISSUE #4122: Remove the `angular.version.minor > 4` checks
These checks should be removed once we hit 0.14.0 & claim angular 1.4+ support.

- [ ] Carousel
- [x] Modal

wesleycho:
- Make use of `$animateCss` directly since a shell service was introduced in Angular proper

This closes #4122.


Foxandxss:
Sounds good to me.

PR #4775, opened 2015-10-29
fix: allow library to be loaded async

### LINKED ISSUE #3665: 0.13.0 does not work with async loading
I recently updated from 0.12.x to 0.13.0 and I have a few errors, the first I found is in the last line of both `ui-bootstrap.js` and `ui-bootstrap-tpls.js`: the function `angular.$$csp()` is called possibly before Angular itself is loaded. I fixed this wrapping it in a module `run` (and adding it to the dependencies in the first line):

    angular.module('ui.bootstrap.csp', []).run(function () {
      !angular.$$csp() && angular.element(document).find('head').prepend('<style type="text/css">.ng-animate.item:not(.left):not(.right){-webkit-transition:0s ease-in-out left;transition:0s ease-in-out left}</style>');
    });

I'd gladly produce a PR but I don't understand from which source code is that line built. I also noticed it's also present in `angular.js` so maybe it could be avoided altogether?

wesleycho:
- Fixed execution of style tag addition to inside module.run

This fixes #3665.


Foxandxss:
I have no idea about this, read the issue, never heard of `<script async>`.

wesleycho:
`<script async>` is actually part of the HTML spec - more on it [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script)

Issue #4760, opened 2015-10-28
Typeahead : attribute "typeahead-is-toggled" to know if dropdown is toggled


NicolasVargas:
Hi, 

I would like to know if the dropdown is open in order to change others elements behaviour during this time.
I didn't find any clean method to succeed.
The only solution I found was to check the dropdown attribute "aria-hidden" :
```
angular.element('.dropdown-menu').first().attr('aria-hidden')
```
but I don't really like it and the attribute change is not fired by angular when we click outside the dropdown.

What do you think about adding a new attribute to typeahead in order to bind a boolean to the ```isOpen()``` intern typeahead function.

Thank you,
Nicolas VARGAS

wesleycho:
This seems like a reasonable request.

PR #4779, opened 2015-10-29
feat(typeahead): add 'is-toggled' option

chenyuzhcy:
This is related to issue #4760


RobJacobs:
@chenyuzhcy I'm assuming you are adding a feature to indicate whether the typeahead dropdown is open or not?  In all the other directives that offer this feature we have been using 'is-open' so for consistency sake, could we name this the same?

src/typeahead/test/typeahead.spec.js
```diff
@@ -588,6 +588,78 @@ describe('typeahead tests', function() {
...
+          expect($scope.isToggled).toBeTruthy();
+          changeInputValueTo(element, 'not match');
+          expect($scope.isToggled).toBeFalsy();
+      });
+
+      it('should bind is-toggled indicator as false when a match is clicked', function () {
+          expect($scope.isToggled).toBeFalsy();
+          changeInputValueTo(element, 'b');
+          expect($scope.isToggled).toBeTruthy();
+          var match = $(findMatches(element)[1]).find('a')[0];
```
wesleycho: Better to do `var match = findMatches(element).find('a').eq(0)`

src/typeahead/test/typeahead.spec.js
```diff
@@ -588,6 +588,78 @@ describe('typeahead tests', function() {
...
+
+          $(match).click();
```
wesleycho: Given the above change, should be able to do just `match.click()`

wesleycho:
This LGTM, and I'm fine merging while making the changes as merging - any objections here?

chenyuzhcy:
@RobJacobs I had the same idea at first but then I looked all the 'is-open' in the other directives, it all does toggle to the dropdown, but this is only an indicator of open or not (it does not manually open the dropdown even we set it to true) so I chose the name that's mentioned in the issue. But if you still insist 'is-open' is a better name I can change it very quickly.

wesleycho:
I think `is-open` is a better name - I think the semantic is off in the other directives, but too late for those.

PR #4780, opened 2015-10-29
test(position): add more tests

Foxandxss:
Sadly not enough. For the `offset` ones I would love to test how it applies window's offsets, but mocking window is a real pain, it starts breaking other tests and you need to mock lot of it.

Would be good if jasmine were able to spy on variables returning a different value, but not possible.


wesleycho:
I didn't know we supported diagonal positioning - is this something we're using?

wesleycho:
Tests LGTM though.

Issue #4317, opened 2015-09-01
Right alignment is off for dropdown-append-to-body .dropdown-menu-right when there's a vertical scrollbar


ilameda-RA:
This is related to issue #3820 
https://github.com/angular-ui/bootstrap/issues/3820

The right alignment is off when there's a scroll bar. 
Demo link: http://plnkr.co/edit/y6qOfyjM54YE8Ft1Vxj6?p=preview

Both dropdowns have dropdown-menu-right.
The first dropdown has the dropdown-append-to-body white the second doesn't.

If the height of the "TRIGGER SCROLL" div is changed to 200px, for example, then the alignment is fine in both dropdowns.

The alignment should be the same in both scenarios

Thanks

wesleycho:
@RobJacobs given your scrollbar work, I'm pointing you to this issue for comment :) .

almeidap:
This fix solves the positioning problem with right-aligned dropdown menus but the menu position is not correctly updated when a window resize occurs and the Bootstrap grid collapses.

You can see the issue here if you open the first dropdown menu and shrink the window until grid collapses:
http://plnkr.co/edit/l2wW2ydgnRVuV86q8TEV?p=preview

PR #4782, opened 2015-10-29
fix(modal): prevent body content shifting when vertical scrollbar

### LINKED ISSUE #3714: Open modal is shifting body content to the right
When the page has a vertical scroll and opens a modal window, the vertical scroll on the page hiding. Because of this, the content moves to the right. In the Bootstrap this problem is solved by adding right padding to the body when scroll hiding.

http://plnkr.co/edit/ynWJUbnryVkSHdEjFRMw?p=preview

kunal999:
Prevents page body content shifting to the right when modal is open (when appendTo is 'body')

Closes #3714

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -19,6 +19,7 @@ module.exports = function(config) {
       'node_modules/angular/angular.js',
       'node_modules/angular-mocks/angular-mocks.js',
       'node_modules/angular-sanitize/angular-sanitize.js',
+      'node_modules/bootstrap/dist/css/bootstrap.css',
       'misc/test-lib/helpers.js',
       'src/**/*.js',
       'template/**/*.js'
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "1.0.0-SNAPSHOT",
   "homepage": "http://angular-ui.github.io/bootstrap/",
   "dependencies": {},
-  "scripts":{
+  "scripts": {
     "test": "grunt"
   },
   "repository": {
@@ -15,6 +15,7 @@
     "angular": "^1.4.4",
     "angular-mocks": "^1.4.4",
     "angular-sanitize": "^1.4.4",
+    "bootstrap": "^3.3.5",
     "grunt": "^0.4.5",
     "grunt-contrib-concat": "^0.5.1",
     "grunt-contrib-copy": "^0.8.0",
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -346,7 +346,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
           // reset any right padding set to body when opening the modal
           // make sure no other modal is open before resetting
           if(modalWindow.appendTo === 'body' && openedWindows.length() === 0){
-            var body = $document.find('body').eq(0);
+            var body = $document.find('body');
             body.css('padding-right', modalWindow.modalDomEl.originalBodyRightPadding);
           }
         });
@@ -533,7 +533,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
               $$scrollbarHelper.measureScrollbar();
             }
 
-            var body = $document.find('body').eq(0);
+            var body = $document.find('body');
             // set originalBodyRightPadding property to reset it when the last modal closes
             modalDomEl.originalBodyRightPadding = body.css('padding-right');
             body.css('padding-right', ($$scrollbarHelper.scrollbarWidth + parseInt(modalDomEl.originalBodyRightPadding || '0'))+'px');
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -151,33 +151,6 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         element.addClass(attrs.windowTopClass || '');
         scope.size = attrs.size;
 
-        // get topmost modal object
-        var bodyIsOverflowing;
-        var modal = $modalStack.getTop();
-        
-        // only when modal is attached to body
-        if(modal && modal.value && modal.value.appendTo === 'body'){
-          
-          // check bodyOverflowing property that was set when opening modal
-          if(modal.value.modalDomEl && modal.value.modalDomEl.bodyOverflowing){
-            bodyIsOverflowing = true;
-          }
-
-          // start - from adjustDialog method of modal.js of bootstrap
-          // check if modal is overflowing
-          var modalIsOverflowing = element[0].scrollHeight > document.documentElement.clientHeight;
-          
-          if(!$$scrollbarHelper.scrollbarWidth){
-            $$scrollbarHelper.measureScrollbar();
-          }
-
-          element.css({
-            'padding-left': (!bodyIsOverflowing && modalIsOverflowing ? $$scrollbarHelper.scrollbarWidth : '') + 'px',
-            'padding-right': (bodyIsOverflowing && !modalIsOverflowing ? $$scrollbarHelper.scrollbarWidth : '') + 'px'
-          });
-          // end - from adjustDialog method of modal.js of bootstrap
-        }
-
         scope.close = function(evt) {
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -15,6 +15,7 @@
     "angular": "^1.4.4",
     "angular-mocks": "^1.4.4",
     "angular-sanitize": "^1.4.4",
+    "bootstrap": "^3.3.5",
```
wesleycho: Undo this change.

karma.conf.js
```diff
@@ -19,6 +19,7 @@ module.exports = function(config) {
       'node_modules/angular/angular.js',
       'node_modules/angular-mocks/angular-mocks.js',
       'node_modules/angular-sanitize/angular-sanitize.js',
+      'node_modules/bootstrap/dist/css/bootstrap.css',
```
wesleycho: Does this affect the test?

src/modal/modal.js
```diff
@@ -463,6 +521,25 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
+          // set bodyOverflowing property
+          modalDomEl.bodyOverflowing = true;
+
+          // set right padding to body
+          if(openedWindows.length() === 1){
+            if(!$$scrollbarHelper.scrollbarWidth){
+              $$scrollbarHelper.measureScrollbar();
+            }
+
+            var body = $document.find('body').eq(0);
```
wesleycho: No need to do `.eq(0)` here, should only be one body tag.

karma.conf.js
kunal999: Yes. Without bootstrap.css, the tests always pass. The tests rely on the modal related classes in bootstrap css which remove the scroll bar from the body when the modal is opened.

src/modal/modal.js
kunal999: Removed the unnecessary `.eq(0)`

src/modal/modal.js
```diff
@@ -129,6 +151,33 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
+          // check if modal is overflowing
+          var modalIsOverflowing = element[0].scrollHeight > document.documentElement.clientHeight;
+          
+          if(!$$scrollbarHelper.scrollbarWidth){
+            $$scrollbarHelper.measureScrollbar();
+          }
+
+          element.css({
+            'padding-left': (!bodyIsOverflowing && modalIsOverflowing ? $$scrollbarHelper.scrollbarWidth : '') + 'px',
+            'padding-right': (bodyIsOverflowing && !modalIsOverflowing ? $$scrollbarHelper.scrollbarWidth : '') + 'px'
```
wesleycho: This means that the object could potentially be setting
```js
element.css({
  'padding-left': 'px',
  'padding-right': 'px'
});
```
It should be setting 0 at worse.

src/modal/modal.js
```diff
@@ -129,6 +151,33 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         element.addClass(attrs.windowTopClass || '');
         scope.size = attrs.size;
 
+        // get topmost modal object
+        var bodyIsOverflowing;
+        var modal = $modalStack.getTop();
+        
+        // only when modal is attached to body
+        if(modal && modal.value && modal.value.appendTo === 'body'){
```
wesleycho: Shouldn't `modal` exist if the `uib-modal-window` directive is used? Also shouldn't `modal.value' & `modal.value.appendTo` exist at all times?

src/modal/modal.js
kunal999: Fixed to set the `padding-left` and `padding-right` only when required

src/modal/modal.js
kunal999: Removed unnecessary null checks

src/modal/modal.js
```diff
@@ -106,9 +106,31 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
+    function measureScrollbar () { // thx walsh
+      var scrollDiv = document.createElement('div');
+      scrollDiv.className = 'modal-scrollbar-measure';
+      document.body.appendChild(scrollDiv);
+      var width = scrollDiv.offsetWidth - scrollDiv.clientWidth;
+      document.body.removeChild(scrollDiv);
+      this.scrollbarWidth = width;
+    }
+  })
+
```
RobJacobs: I'd like to see this in the $position service as scrollbarWidth(), we might need to use it for auto/smart positioning. Would be nice to cache the result to avoid manipulating the DOM every time we need this value as it should not change.

wesleycho: I agree - we have a related issue in #4317.

wesleycho:
Can you squash then rebase this, and then change to use the `$uibPosition.scrollbarWidth` method for calculating the scrollbar width?

wesleycho:
Also, please fix the code style to match the rest of the file.

src/modal/modal.js
```diff
@@ -202,6 +224,34 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
+            if(modal.value.appendTo === 'body'){
+              
+              // check bodyOverflowing property that was set when opening modal
+              if(modal.value.modalDomEl.bodyOverflowing){
+                bodyIsOverflowing = true;
+              }
+
+              // start - from adjustDialog method of modal.js of bootstrap
+              // check if modal is overflowing
+              modalIsOverflowing = element[0].scrollHeight > document.documentElement.clientHeight;
```
wesleycho: Should use `$document[0].clientHeight`

src/modal/modal.js
```diff
@@ -463,6 +522,25 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         appendToElement.append(modalDomEl);
         appendToElement.addClass(modalBodyClass);
 
+        // if modal is attached to body, then check if body is overflowing and set bodyOverflowing property on modal object and set right padding equal to scrollbarWidth
+        if (modal.appendTo === 'body' && document.body.scrollHeight > document.documentElement.clientHeight){
```
wesleycho: Ditto

src/modal/modal.js
```diff
@@ -463,6 +522,25 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
+
+          // set right padding to body
+          if(openedWindows.length() === 1){
+            if(!$$scrollbarHelper.scrollbarWidth){
+              $$scrollbarHelper.measureScrollbar();
+            }
+
+            var body = $document.find('body');
+            // set originalBodyRightPadding property to reset it when the last modal closes
+            modalDomEl.originalBodyRightPadding = body.css('padding-right');
```
wesleycho: That is an unnecessarily verbose key name - please choose a better name, such as `bodyRightPadding`.

src/modal/modal.js
```diff
@@ -291,6 +343,13 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
           openedClasses.remove(modalBodyClass, modalInstance);
           appendToElement.toggleClass(modalBodyClass, openedClasses.hasKey(modalBodyClass));
           toggleTopWindowClass(true);
+
+          // reset any right padding set to body when opening the modal
+          // make sure no other modal is open before resetting
+          if(modalWindow.appendTo === 'body' && openedWindows.length() === 0){
+            var body = $document.find('body');
+            body.css('padding-right', modalWindow.modalDomEl.originalBodyRightPadding);
```
wesleycho: This seems way too opinionated - this will make it near impossible for a user to override this, and is not a good solution.

src/modal/modal.js
RobJacobs: $position service now has a scrollbarWidth function.

wesleycho:
Going to close this due to the issues raised in the code review & due to the merge conflict - feel free to open a new PR fixing this along with a cleaner implementation.

PR #4783, opened 2015-10-29
fix(pagination): retain model on initialization

### LINKED ISSUE #2956: pagination module initializes ngModel to 1 if total-items starts out undefined
I'm grabbing metadata from an API where I know which page I would like to grab via the url parameters, but the total-items takes a few milliseconds to populate.

Because total-items starts out undefined, ngModel is changed from the correct page number to 1.

The only way I have gotten around this is to initialize total-items to an arbitrary number such as 1000. Then the page number is left alone.

Is there a better solution?

### LINKED ISSUE #3786: Watcher on $attrs.itemsPerPage added multiple times means all fire
I am using bootstrap pagination twice on a page, once above the content and once below. For both of these, the $attrs.itemsPerPage watcher in pagination.js on line 17 is added. When I modify itemsPerPage both watches fire. This is undesirable as update logic on my page is tied to the changing of the page number which calculateTotalPages() can cause to occur. This in turn causes ng-change to fire twice.

The current code is this.
```javascript
$scope.$parent.$watch($parse($attrs.itemsPerPage), function(value) {
        self.itemsPerPage = parseInt(value, 10);
        $scope.totalPages = self.calculateTotalPages();
});
```

Would this be a good solution?

```javascript
$scope.$parent.$watch($parse($attrs.itemsPerPage), function(newValue, oldValue) {
        if (newValue !== oldValue)
        {
                self.itemsPerPage = parseInt(newValue, 10);
                $scope.totalPages = self.calculateTotalPages();
        }
});
```
edit: what I thought was happening isn't actually happening (its still being called twice) but ng-change is not being fired twice due to that, that was something weird going on in the code I'm refactoring.

wesleycho:
- When `total-items` initializes as undefined, ignore value and do not update page information
- Remove extra watcher and call action manually

[Here](http://plnkr.co/edit/Wsp3aiboSGdAk8bUuu6f?p=preview) is a Plunker for reference.

Closes #3786
Fixes #2956


Foxandxss:
:+1:

Issue #2570, opened 2014-08-12
Feature: Typeahead match inside search box


bjornasm:
Can we have an option to display the first match (matches[0]) inside the search box, as you'd find in the Typeahead.js demo: http://twitter.github.io/typeahead.js/ (when searching for bower)

wesleycho:
For reference to anyone interested in implementing this - how this works is that a read only input is positioned in the same location as the regular input. This input has its value filled on focus of the typeahead input.

This would be a cool feature, but it may require the user to style it appropriately.

PR #4784, opened 2015-10-29
feat(typeahead): add 'show-hint' option

chenyuzhcy:
This is related to #2570

Add a read-only input that has the same postion as the original input

### DIFF
--- a/src/typeahead/docs/readme.md
+++ b/src/typeahead/docs/readme.md
@@ -82,3 +82,7 @@ The typeahead directives provide several attributes:
 * `typeahead-is-open` <i class="glyphicon glyphicon-eye-open"></i>
    _(Defaults: angular.noop)_ :
    Binding to a variable that indicates if dropdown is open
+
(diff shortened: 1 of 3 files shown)


Foxandxss:
I must say that I absolutely love your work. I wish all the PRs would be like this one and the others you made.

As a suggestion that would help us immensely is to create a plunker with your solution working. You can open the current typeahead demo on plunker, create a new file and paste the tpls bundle that contains your patch (as simple as `grunt after-test` to generate it).

Thank you again.

chenyuzhcy:
Thanks for your compliment!
This is a plunker for this pr: http://plnkr.co/edit/5oWIfw4z14uOVo6KzPu7?p=preview
I'll try to do this in the future too. Glad I can help!

wesleycho:
Seconded on the admiration of the great work - these are some high quality PRs you have put together, and I greatly appreciate the effort put into them. It is rare that we see work this good come out of the blue.

PR #4785, opened 2015-10-30
chore(karma): add coverage

Foxandxss:
The library was in place, it just needed a bit of love.

Used the dist folder for not creating yet another temporary root folder.


wesleycho:
LGTM - guess this means that I have to be a little careful when I copy/paste when doing a release, but that's ok.

wesleycho:
Actually, maybe it would be better to gitignore a `.coverage` folder and put it there, just to avoid polluting contents.

Foxandxss:
Sure.

Foxandxss:
Done.

Out of curiosity, we also add an entry in the gitignore :P

PR #4786, opened 2015-10-30
Change uib-typeahead to typeahead

msemenistyi:


pkozlowski-opensource:
@msemenistyi Why? This is not a valid change, as we've just introduced prefixes.

msemenistyi:
@pkozlowski-opensource Here are options to the right of examples:
e66a339cf4bb35f8599929d936f2a5ce
What version were these prefixes introduced in? I was fighting to get it working for an hour looking at uib-typeahead examples

Foxandxss:
0.14.0. If you need older docs, you can use the dropdown on the top of the page to change to the version you're currently working.

msemenistyi:
@Foxandxss thanks

Issue #2324, opened 2014-06-10
Typeahead no option to have all items display.


HerrPfister:
There is no clear way to display all items, stored in the typeahead, when there is no text in the input and the typeahead gains focus. Did I overlook something? Is there a way to do this? If so, how?

marcusg:
any updates here? I think this would be a nice feature ;)

sharpmachine:
I want the same thing!  Any updates?

dmitrijs:
+1

brunocoelho:
+1

brunocoelho:
I know that you have to support async load, but it would be a nice and very useful feature if you have a list pre-loaded.

flaurian:
Thanks for adding this

flaurian:
Would actually be even nicer with a scroll when displaying all elements but with a limit of (e.g.) 8 results only so that the user does not believe that the list is displaying all results.

icfantv:
@flaurian, that is beyond the scope of this library, but you can do it yourself quite easily with CSS.

flaurian:
thanks for your comment @icfantv !

Just wondering what you mean by scope... I thought the purpose of such a library is to 1) do things bootstrap isn't doing well (I suppose they also have defined a "scope"), 2) build best-practice versions of basic web functionalities. 

IMHO it seems rather obvious that when you are using a typeahead you should be able to a) show all elements (after all it's an enhanced dropdown), b) manage the use case smoothly if the list of elements is way too long (in my case I am selecting an employee out of ~100).
But indeed as you said I managed it easily like this :

`.dropdown-menu{
   max-height:200px;
   overflow-y:auto;
  }`

Cheers

icfantv:
While we do make certain UX decisions and indeed some CSS ones as well, determining on whether someone wants a scrollbar to appear it not our call.  To be clear, I have CSS in our app so in certain areas the scrollbar always shows on WebKit browsers, but again, this specific use case is not one we should be making.  As soon as we do, someone is going to file a bug asking for a way to turn it off.

PR #4789, opened 2015-10-30
feat(typeahead): display all items when focus with min-length=0

chenyuzhcy:
This relates to #2324 and #764 

Working test: http://plnkr.co/edit/JsEwGwoqckcnHMMOxVTN?p=preview


wesleycho:
Only thing I see here is that the docs need to be updated since it currently mentions that min-length must be >= 1.

Foxandxss:
Also, I would modify one of the demo's typeahead (just 1) to have this new option or perhaps add a new one. Not really sure.

You can skip that tho, I need to revamp all the doc and demos.

PR #4793, opened 2015-10-31
fix(tab): fix unexpected routing

### LINKED ISSUE #3266: For Tabs generated by ng-repeat, in tab-heading a call to a function that has $location.path(url) causes a weird page reload after redirect.
I was using angularjs 1.2 and ui-bootstrap-tpls-0.11.0. Works great!. I am generating tabs dynamically using ng-repeat. 
<tabset>
            <tab ng-repeat="item in items">
                <tab-heading>
                   <div class="rmv-item" ng-click="objMain.removeSelVal(item.id)">
                        <span aria-hidden="true">&times;</span><span class="sr-only">Close</span>
                    </div>
                </tab-heading>
                ...
            </tab>
</tabset>

In MainController.js I have a function
$scope.objMain.removeSelVal = function(id){
    // remove item from items collection based on the passed id and reload the page
    $location.path('/myroute/' + ids of remaining items);
};

When a user clicks on removeSelVal inside of each tab the selected id gets passed to a function in the main controller. This function removes one item and passes the new items collection to 'myroute'. Worked great in version 0.11.0. After the upgrade to 0.12.0 it stopped working. It does the redirect to 'myroute' but then does another redirect to an empty route. I can't use 0.11.0 with angular 1.3 since it breaks typeahead so I had to roll back to angular1.2 Hope the fix for 0.12.0 is coming soon. Thank you

wesleycho:
- Change anchor tag to `div` to prevent unexpected routing being triggered by the browser on generation of anchor tags with nested clickable controls

BREAKING CHANGE: This causes the cursor style to be removed from the tab - implement CSS on the `.uib-tab > div` selector, or similar, accordingly

This fixes #3266.


Foxandxss:
Same as accordion.

icfantv:
@chenyuzhcy, it's almost certainly because of that.  the issue we were having with the `<a>` is that the browser was handling the click event before Angular even got to it and as a result, the handler in the Angular directive was never getting called.  We didn't like this change, but we swallowed our pride and did it.  But it breaks the CSS...as you discovered.

And thanks for looking at existing issues first!  That's a huge help to us.

Foxandxss:
This broke more stuff indeed like all the tabs in the general demo (where you can choose between seeing the HTML or the javascript). That was fixed.

I can't repro your issue. Weird.

Foxandxss:
It is fixed in master, your plunker is missing some extra styles that were added for fixing this.

chenyuzhcy:
I saw the extra styles, thank you!

chenyuzhcy:
I copied the styles from the demo.html and found the disabled style is missing:
`.nav>li.disabled>div {
    color: #999;
}`

Foxandxss:
If there is something missing, sounds like a good PR deal for me :)

chenyuzhcy:
ok, I'll create one then

icfantv:
When we do the 1.0 official build, we need to thoroughly go through the breaking changes doc and make sure the CSS needed is placed for this item.

pkyad:
I know that this issue was fixed in 1.0 release by I am getting this behavior again. I inspected the generated html and there is anchor tag 



The UI is having dynamic tabs like a web browser and the user can cancel the tab by clicking on the close sign , I am using ng-click in the uib-tab-heading to splice the tabs array 

Can anyone point where I am going wrong 

thanks

PR #4794, opened 2015-10-31
feat(dropdown): add open class support

### LINKED ISSUE #4466: dropdown, adding "open" to body for append-to-body has side effects
This was mentioned in #4240.  With the changes made to the append-to-body dropdowns in 0.13.4 (specifically adding the class "open" to the body element for these dropdowns), other components on the page may be affected, particularly in a bootstrap heavy application.

In my case, I'm using dangrossman's date range picker: https://github.com/dangrossman/bootstrap-daterangepicker

Here's a plunkr demonstrating the issue:

http://plnkr.co/edit/5QoR6ZAgZZcORpLWghFu?p=preview

If you click on the append-to-body dropdown there, you'll notice that the date range picker opens as well.  I believe it only happens if the date range picker hasn't been opened by clicking on it yet.

This happens because bootstrap has this defined for `.open > .dropdown-menu`:
https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_dropdowns.scss#L121

The date range picker is appended to body as well, so matches that selector (and has the dropdown-menu class defined on it).  This might be fixable via user-land CSS, but overall I think adding the "open" class is too general in an application that makes use of bootstrap compatible components.

I don't know if it's possible to apply this "open" class to a different element, or possibly even store the dropdowns in a container that gets the class change.  That would probably be ideal.

However, if that's not possible, would it be possible to be able to specify what class is used?  I noticed that "open" is defined in a constant here:

https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L3

Could the library be modified so it's possible to define what this class is, perhaps through a provider?

wesleycho:
- Adds support for the `uib-dropdown-open` class to be toggled on the container element of a dropdown menu on dropdown toggle

This closes #4466.


Foxandxss:
Couldn't this be considered a BC?

wesleycho:
Technically, but I doubt anyone is using a class named uib-dropdown-open in their app

Foxandxss:
I was more thinking of a user relying on the open class to do something and now it is a different one. (If I understood the issue correctly.)

wesleycho:
This only affects appendTo and appendToBody

Foxandxss:
I see. Well, good to go.

PR #4796, opened 2015-10-31
chore(carousel): fix demo

Foxandxss:
I don't know how that got in there and why it worked in the first place? Magic.

Anyway, that fixes the demo problem. Also made a shorter button so it doesn't broke the demo visually.


wesleycho:
LGTM

PR #4797, opened 2015-10-31
feat(typeahead): change to `appendTo`

wesleycho:
- Change signature to `appendTo`, which takes a value that evaluates to a DOM node for API consistency

BREAKING CHANGE: This requires changing from using the string representation of the element id to select to passing in the element itself. A simple change on the user side would be to change it to a value on the element's parent $scope to a value that evaluates to the node, i.e. `angular.element($document[0].getElementById('elementId'))`


Foxandxss:
I am not sure if I understand that BC message, a bit complex. I would perhaps add what angular does. A before code and a after code. Other than that, fine for me.

PR #4800, opened 2015-11-02
fix(typeahead): allow parent to be required

### LINKED ISSUE #2679: typeahead custom templates with directives that have "require" option
If a template contains a directive that has a "require" option it will fail with the error stating that required controller can't be found. To fix I replaced `element.replaceWith($compile(tplContent.trim())(scope));` to

`var tplEl = angular.element(tplContent.trim());`
`element.replaceWith(tplEl);`
`$compile(tplEl)(scope);`

inside `typeaheadMatch` directive. This places the element into the DOM before compiling it, thereby resolving the issue of the missing dependent controller.

wesleycho:
- Change to insert content to DOM before compiling to preserve requiring parent controllers

This fixes #2679.


Foxandxss:
:+1:

PR #4806, opened 2015-11-02
docs(dateparser): add demo with docs

Foxandxss:


wesleycho:
LGTM - this is some great work here!

PR #4807, opened 2015-11-02
chore(demo): fix demo with tab change

wesleycho:
- Fix appearance of tabs with breaking change in demo


Foxandxss:
If that fixes it, :+1:

Issue #4805, opened 2015-11-02
Support optional leading zero in dateparser


wesleycho:
A pain for some users has been allowing support of an optional leading zero, i.e. 03/09/2015, or 3/9/2015 for the format of MM/dd/YYYY.

One approach for allowing support would be to add new supported regex keys `M*` and `d*` in the `formatCodeToRegex` object which allows optional leading zeroes.

PR #4809, opened 2015-11-02
feat(dateparser): add M! and d! support

wesleycho:
- Add M! and d! for optional support of leading zeroes in date string format

This implements #4805


Foxandxss:
:+1:

PR #4810, opened 2015-11-02
fix(dateparser): enforce order of regex construction

### LINKED ISSUE #4808: Dateparser is highly reliant on regex definition order
I discovered today that the dateparser is highly reliant on the order of key definition in the `formatCodeToRegex` object - this is highly flawed, as the JS spec itself specifies that the key order in iteration does not matter.

This fragility needs to be fixed.

wesleycho:
- Change `formatCodeToRegex` into an array to avoid potential iteration bug where the order the keys are listed in the object matters in the iteration order

This fixes #4808.


Foxandxss:
Better. Merge it.

Issue #2219, opened 2014-05-19
Add support of ng-disabled to timepicker


idrabenia:
Hi!

Timepicker is great! 
But will be very cool if timepicker will have support of ng-disabled directive.

Thanks!

carlospliego:
So, we need to set three things to correctly disable the input?

Siddeshgad:
Yes. If you set readonly-input="true", hour-step and minute-step to zero it will work.

xiphiaz:
While this is a feature that definitely should be implemented, there is a workaround - you can wrap the timepicker in a <fieldset ng-disabled="your.param"></fieldset> to disable all fields in the fieldset.

theaccordance:
Having to modify multiple attributes to disable the directive is not ideal, it would be much more preferable to have a single attribute that disables all aspects of this directive from user input.

amamut:
Meridian is still editable after disabling everything.

deepu105:
`<fieldset ng-disabled="your.param"></fieldset>` this is the best workaround disables everything thanks @xiphiaz  but yes ng-disabled directly would be better

deepu105:
PR seems to  have closed with a rollback

JoeMatulka:
Careful of the fieldset, it doesn't work in IE...

AliAdravi:
In our directive we can watch the ngDisabled value and disable the datepicker
scope.$watch(attr.ngDisabled, function (newVal) {
  if(newVal === true)
    $(elm).datepicker("disable");
  else
    $(elm).datepicker("enable");
});
See the complete detail here:
http://www.advancesharp.com/blog/1224/angularjs-jquery-ui-datepicker-for-any-condition

PR #4811, opened 2015-11-03
feat(timepicker): add ngDisabled support

wesleycho:
- Adds support for `ng-disabled` on timepicker element for disabling all controls programmatically

This implements #2219


Foxandxss:
Should this disable changes from model to view? Not sure if it does that already or if it makes sense.

wesleycho:
It should not disable changes from model to view - Angular doesn't do it, and nor should we.

Foxandxss:
Ok, good for me then.

stefan--:
@wesleycho  enabling `ng-disabled` causes my form to be invalid although the model contains a valid date. I would expect it to be valid. Please have a look:

```
{
  "$error": {
    "dateDisabled": [
      {
        "$viewValue": "2016-03-24T09:30:00.000Z",
        "$modelValue": "2016-03-24T09:30:00.000Z",
        "$validators": {},
        "$asyncValidators": {},
        "$parsers": [],
        "$formatters": [],
        "$viewChangeListeners": [
          null
        ],
        "$untouched": true,
        "$touched": false,
        "$pristine": true,
        "$dirty": false,
        "$valid": false,
        "$invalid": true,
        "$error": {
          "dateDisabled": true
        },
        "$name": "",
        "$options": null
      }
    ]
  },
...
}
```

PR #4812, opened 2015-11-03
fix(datepicker): correctly display pre-100 years

### LINKED ISSUE #4032: Cannot select pre-100 years
* Angular 1.3.13
 * Angular bootstrap 0.13.0

Cannot select date before 101. Manually typing year 1 (even with correct formatting) leads to an incorrect model value. This is due to the fact that 

```
new Date(1, 7, 7) // is 7th July 1901
new Date("0001-07-07") // is 7th of July 1
```

Can be replicated on the demo page. Try selecting any date pre-100.

wesleycho:
- Fix display of dates in datepicker for years less than 100

[Here](http://plnkr.co/edit/kNiZ4S4gKKGHpsuWd4NF?p=preview) is a Plunker in action.

This fixes #4032.


src/dateparser/dateparser.js
```diff
@@ -246,6 +246,7 @@ angular.module('ui.bootstrap.dateparser', [])
           dt = new Date(fields.year, fields.month, fields.date,
             fields.hours, fields.minutes, fields.seconds,
             fields.milliseconds || 0);
+          dt.setFullYear(fields.year);
```
Foxandxss: why is this needed?

Foxandxss:
Can you explain to me how this work? I made a note but then realized that all the PR was add stuff loike that and I don't follow the idea.

wesleycho:
The problem with setting a date with `new Date(year, month, day)` is that it interprets `new Date(3, 1, 2)` as `February 2, 1903` instead of `February 2, 3` - one has to manually call `setFullYear` in order to properly set the year.

Foxandxss:
I see. Good for me.

PR #4813, opened 2015-11-03
chore(debounce): create debounce helper

### LINKED ISSUE #3966: Split off debounce from Typeahead to separate component
Debouncing is a useful feature, and could likely be used elsewhere in the codebase - this should be separated out into a separate component.

The work done can be referenced in 86bfec1 .

wesleycho:
- Create internal debounce helper
- Move debounce logic from typeahead to use helper

This closes #3966.


Foxandxss:
I like what you did, let's start from there.

I don't know much about typeahead to be honest, but is this debounced related to ng-model? (to be solved by ng-model-options) if not, good for merging.

wesleycho:
This is not related to the model - this is related to repositioning on resize or scroll.

Foxandxss:
Ah, good to go then.

Issue #4059, opened 2015-07-31
`disabled` support for datepicker


wesleycho:
The datepicker should respect the `disabled` attribute.

The datepicker popup should respect the `disabled` attribute, and not open the popup when disabled.

This should also add support for ng-disabled.

PR #4814, opened 2015-11-03
feat(datepicker): add disabled and ngDisabled support

wesleycho:
- Adds support for `disabled` attribute
- Adds support for `ng-disabled`

This implements #4059.


Foxandxss:
I am not too fond of using `disable` in here because it is only meant for concrete elements. I think that ngDisabled is enough.

All things considered, not that Angular is respecting all conventions in HTML

wesleycho:
My reasoning is that we are treating this as a control, so we should try to hook into the descriptive native to implement the behavior desired. I know disabled has a particular semantic, but I think it's ok to use it, as long as we are disabling things sensibly.

I forgot to link it, but [here](http://plnkr.co/edit/yjlmc3mxT7zjTMdiZ73Q?p=preview) is the Plunker in action.

Foxandxss:
Yeah, I think it makes sense, to have a read only datepicker. Go for it.

PR #4818, opened 2015-11-04
fix(modal): cannot append to element outside of document

MerlierR:
- needed for appending modal's outside of an iframe


sparty02:
@wesleycho What's the release cadence for these kinds of things?  I looked at https://github.com/angular-ui/bootstrap/wiki/Milestones but it appears that isn't being updated.  Is the 1.0.0 release imminent?

wesleycho:
@sparty02 we're hoping to release by the end of the year - we've been bogged down by stuff going on in our own lives :) .

Issue #4592, opened 2015-10-12
typeahead-editable attribute does not update as per documentation


damonf:
Changing the value of typeahead-editable attribute should update the behaviour of the directive, but this is not implemented.
The documentation for this attribute has an eyeball icon next to it, and the doc says... "The eye means that the setting has an Angular $watch listener applied to it."

plunker here http://plnkr.co/edit/W2vneQGgiMDHMGC3uDGS?p=preview

wesleycho:
This looks like it was never supported - probably better to just remove the eye from the docs.

This would be an easy PR if you'd like to tackle it.

PR #4820, opened 2015-11-05
feat(typeahead): apply angular watch to the typeahead-editable setting

damonf:
Adds a watch for the typeahead-editable attribute.

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -46,7 +46,6 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
     var isEditable = originalScope.$eval(attrs.typeaheadEditable) !== false;
     originalScope.$watch(attrs.typeaheadEditable, function (newVal) {
       isEditable = newVal !== false;
-      modelCtrl.$setValidity('editable', isEditable);
     }); 
 
     //binding to a variable that indicates if matches are being retrieved asynchronously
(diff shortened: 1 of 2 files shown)


icfantv:
@damonf, what problem is this PR solving?  Also, you're missing functional tests.

Please give [this](https://github.com/angular-ui/bootstrap/blob/master/CONTRIBUTING.md) a read and in the future, it's generally a good idea to run an idea by the team first before spending time on a PR as we don't want you to possibly be wasting your time.  Thanks.

wesleycho:
There was a prior issue reported by @damonf related to this in #4592.

src/typeahead/typeahead.js
```diff
@@ -44,6 +44,10 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
 
     //should it restrict model values to the ones selected from the popup only?
     var isEditable = originalScope.$eval(attrs.typeaheadEditable) !== false;
+    originalScope.$watch(attrs.typeaheadEditable, function (newVal) {
+      isEditable = newVal !== false;
+      modelCtrl.$setValidity('editable', isEditable);
```
wesleycho: This is a breaking change, and editable state should not be related to ngModel's validity. This is a bad change.

wesleycho:
This needs tests.

The model $setValidity call needs to be removed as well.

damonf:
Thanks for your patience.
So if i removed the $setValidity call and added some unit tests you would
accept the PR?

wesleycho:
I would be willing to accept this feature with the necessary changes.

PR #4821, opened 2015-11-05
fix(tooltip): Correct Jasmine custom matcher message condition.

Incell:
Right now the custom matcher in tooltip2.spec.js would always show 'to have' message, as `result.message` would always evaluate to false. Probably a typo.


wesleycho:
Thanks for this, looks like you are correct

Issue #4823, opened 2015-11-05
Carousel corrupted if render slides with ng-repeat and cnahge slide set.


sergik:
Carusel corrupted if render slides with ng-repeat and cnahge slides collsection. Slides animations throttled with internal flag $scope.$currentTransition. This flag prevents moving to next slide if carusel is in slide change process. When changing slides collection this flag leaves in true state and carusel is corrupted. See the example. http://plnkr.co/edit/FNijUvcOjsXocqmtP3bh?p=preview.
Press "kill me" and wait for a second. This example emulates fast 2 time change of an active slide.

sergik:
could potentialy relate to https://github.com/angular-ui/bootstrap/issues/4667. Submitted a pull request for this: https://github.com/angular-ui/bootstrap/pull/4824

mchurichi:
Here is another example of the same issue happening when you add a bunch of slides after the carousel has been fully loaded: http://plnkr.co/edit/NyG9yX?p=preview. This pull request doesn't solve the issue in this case.

PR #4824, opened 2015-11-05
#4823 Carusel corrupted if render slides with ng-repeat and change slides set.

sergik:
turns off animation trottling in case of changing active slides during remove process.


wesleycho:
Closing as this is an invalid/untested code change - this overrides the buffering slide changes built into the component.

PR #4831, opened 2015-11-06
feat(dateparser): add Z support

wesleycho:
- Add support for the `Z` format, as referred to in the Angular
  documentation [here](https://docs.angularjs.org/api/ng/filter/date)


Foxandxss:
Sounds good to me.

Issue #3418, opened 2015-03-22
Add full date format support


wesleycho:
Currently dateParser only supports a limited number of date formats that [dateFilter](https://docs.angularjs.org/api/ng/filter/date) supports. We should aim to support a larger range of these options, at least supporting all options with the `a` format for AM/PM, which includes `medium`, `short`, `mediumTime`, and `shortTime`.

AngularJsChamp:
You can also create custom date picker using bootstrap date picker code.

PR #4833, opened 2015-11-07
feat(dateparser): add new format support

wesleycho:
- Adds `ww` support
- Adds `w` support
- Adds `G`, `GG`, `GGG` support
- Adds `GGGG` support

Should be noted, this almost finishes off #3418


Foxandxss:
I can see week, but the other are a bit weird, not sure if they are going to be used, ever.

About #3418, I think it is implemented (we have `a` and `short`, `medium`....)

wesleycho:
I'm not sure the era ones will ever be used, but it's nice to have the option, and it doesn't cost us much. It completes parity with the Angular dateFilter.

About `short`, `medium`, etc., how do we support that? I don't see anything obvious in the dateparser that allows for it.

Foxandxss:
Using: https://github.com/angular-ui/bootstrap/blob/master/src/dateparser/dateparser.js#L219-L220

that will translate `short`, `mediumDate`, etc into a string that can be tokenized later.

wesleycho:
Oh, so the issue will be finished then by this PR.

Foxandxss:
Yes sir. Merge this

PR #4834, opened 2015-11-07
feat(modal): change to use $animate

wesleycho:
- Change to use $animate

This implements #3418.


Foxandxss:
So animateCss got backported? And is the PR title ok? (use $animate or animateCss?)

wesleycho:
It did get backported - I am using $animate here for the $animate.enter

Foxandxss:
:+1:

PR #4835, opened 2015-11-07
feat(timepicker): add model state support

### LINKED ISSUE #3527: $touched not set to TimePicker
when hour control or minute control touched and value changed, we must could be able to get 
form.[TimePicker control's name].$touched == true

but it not set. mean ng-touched must be added to control's classes.

wesleycho:
- Adds support for the timepicker model $touched and $dirty states

This closes #3527.


Foxandxss:
:+1:

Issue #1205, opened 2013-10-24
Put templates in a subdirectory


webarnes:
This is not quite related to the asynchronous loading issue (I've seen #1188).

All the templateUrls are hard-coded to something like "template/progressbar/progress.html". This means that my template directory is filled with templates for Bootstrap components. It also means that the templateUrl template/(component) is taken for each of the components.

Unless I'm missing something, it would be friendlier to other code to put all the templates in template/bootstrap/*.

This would be a breaking change, but Bootstrap 3.0 support is going to break backwards compatibility soon anyway.

PR #4839, opened 2015-11-07
feat: prefix virtual templates with `uib/`

wesleycho:
- Change all virtual paths with built in templates to be prefixed by `uib/`

BREAKING CHANGE: All virtual templates in UI Bootstrap now are prefixed
with `uib/` - if one is overriding the templates via `$templateCache` path
or manually building the templates from the UI Bootstrap repository, one
needs to change the string used in the `$templateCache` representation
to have the new prefix

This implements #1205.


Foxandxss:
:+1:

Issue #4758, opened 2015-10-28
Project Code Style


wesleycho:
While we mostly now have a consistent code style, we should probably have it enforced at the linter level, either through eslint or through stronger rules in jshint.

igorlino:
I have found the project code style from John Papa's gulp patterns quite good. (https://github.com/johnpapa/gulp-patterns/)

It's based on jshint, and on my projects has worked quite well.

PR #4840, opened 2015-11-07
chore: add eslint

wesleycho:
- This adds eslint and integrates it to the build chain

This implements #4758.


Foxandxss:
I don't think we need eslint anymore.

Foxandxss:
Nice, so if we have the same goodies like catching issues before doing a PR, I am all for it.

PR #4841, opened 2015-11-07
fix(datepicker): min-date/max-date: parse fix for literals

### LINKED ISSUE #3437: min-date attribute does not disable the day prior to the given min-date
Reproduction Plunker can be found [here](http://plnkr.co/edit/XdUTSlx9koKFHctohdlg?p=preview)

**Steps to Reproduce:**
 # In the reproduction plunker, observe that '2000-01-02' is pre-set as the min-date for the datepicker.
 # Click the text field to open the calendar
 # Navigate to January 2000
 # Observe that 1/1/2000 is selectable

**Observed Results:**
1/1/2000 is selectable, with a min-date of 1/2/2000.

**Expected Results:**
1/1/2000 should not be selectable, with a min-date of 1/2/2000.

davious:
fixes #3437

original plunker, with pr code: http://plnkr.co/edit/jsfOVBZj51yajpZYkC9N?p=preview
maintainer plunker, with pr code: http://plnkr.co/edit/buQXUxndajMYFOZorzq9?p=preview

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/bootstrap/4841)

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -41,7 +41,7 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   angular.forEach(['minDate', 'maxDate'], function(key) {
     if ($attrs[key]) {
       $scope.$parent.$watch($attrs[key], function(value) {
-        self[key] = value ? new Date(value) : null;
+        self[key] = value ? angular.isDate(value) ? new Date(value) : new Date(dateFilter(value, 'medium')) : null;
         self.refreshView();
       });
     } else {
@@ -628,10 +628,10 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
       if (attrs[key]) {
         var getAttribute = $parse(attrs[key]);
         scope.$parent.$watch(getAttribute, function(value) {
-          scope.watchData[key] = value;
           if (key === 'minDate' || key === 'maxDate') {
-            cache[key] = new Date(value);
+            cache[key] = angular.isDate(value) ? new Date(value) : new Date(dateFilter(value, 'medium'));
           }
+          scope.watchData[key] = cache[key] || value;
         });
         datepickerEl.attr(cameltoDash(key), 'watchData.' + key);
 
--- a/src/datepicker/test/datepicker.spec.js
+++ b/src/datepicker/test/datepicker.spec.js
@@ -965,6 +965,20 @@ describe('datepicker', function() {
 
     });
 
+    describe('`min-date` attribute', function () {
+      beforeEach(function() {
+        element = $compile('<uib-datepicker ng-model="date" min-date="\'2010-09-05\'"></uib-datepicker>')($rootScope);
+        $rootScope.$digest();
+      });
+
+      it('accepts literals, \'yyyy-MM-dd\' case', function() {
+        var buttons = getAllOptionsEl();
+        angular.forEach(buttons, function(button, index) {
+          expect(angular.element(button).prop('disabled')).toBe(index < 7);
+        });
+      });
+    });
+
     describe('`max-date` attribute', function() {
       beforeEach(function() {
         $rootScope.maxdate = new Date('September 25, 2010');
@@ -1981,6 +1995,17 @@ describe('datepicker', function() {
 
             expect(buttons.eq(0).prop('disabled')).toBe(true);
           });
+
+          it('should disable today button if before min date, yyyy-MM-dd case', inject(function(dateFilter) {
+            var minDate = new Date(new Date().setDate(new Date().getDate() + 1));
+            var literalMinDate = dateFilter(minDate, 'yyyy-MM-dd');
+            var wrapElement = $compile('<div><input ng-model="date" uib-datepicker-popup="yyyy-MM-dd" min-date="\'' + literalMinDate + '\'" is-open="true"><div>')($rootScope);
+            $rootScope.$digest();
+            assignElements(wrapElement);
+            assignButtonBar();
+
+            expect(buttons.eq(0).prop('disabled')).toBe(true);
+          }));
 
           it('should disable today button if after max date', function() {
             $rootScope.maxDate = new Date().setDate(new Date().getDate() - 2);


wesleycho:
Fix formatting - an easy way to see what is off is to rebase off of current `master`, run `npm install`, and run `grunt eslint`.

src/datepicker/datepicker.js
```diff
@@ -618,7 +618,16 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
         scope.$parent.$watch(getAttribute, function(value) {
           scope.watchData[key] = value;
           if (key === 'minDate' || key === 'maxDate') {
-            cache[key] = new Date(value);
+            var date = parseDate(value);
+            if(!date) {
+              var tryDate = new Date(value);
+              if(angular.isDate(tryDate) && !isNaN(tryDate)) {
```
wesleycho: What happens when `isNaN(tryDate)` is `true`?

davious: `undefined` is gets set.... this is to cover for an invalid attribute; example: a scope variable that doesn't exist.

davious:
Not sure what formatting needs fixing. I rebased and ran `grunt eslint`: `Done, without errors`

wesleycho:
As a rule of thumb, the code should look like the surrounding code - the whitespace differences are pretty stark in this case, or lack of whitespace.

wesleycho:
Thinking about this problem for a bit, I am not sure we want to support this date format in the long run - it would be much better if it required a date object from the user's side as a breaking change, which would prevent the timezone issue, and allow the user the flexibility to specify the timezone of the min date.

src/datepicker/datepicker.js
```diff
@@ -629,7 +629,16 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
         scope.$parent.$watch(getAttribute, function(value) {
           scope.watchData[key] = value;
           if (key === 'minDate' || key === 'maxDate') {
-            cache[key] = new Date(value);
+            var date = parseDate(value);
+            if (!date) {
+              date = new Date(value);
+              if (angular.isDate(date) && !isNaN(date)) {
+                var dateArray = date.toISOString().split('T')[0].split('-');
```
wesleycho: One can get the information from the date via this mechanism:
```js
var year = date.getFullYear(),
  month = date.getMonth(),
  day = date.getDate();
```

src/datepicker/datepicker.js
```diff
@@ -629,7 +629,16 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
...
+                date = new Date(+dateArray[0], +dateArray[1] - 1, +dateArray[2]);
```
wesleycho: After this, do `date.setFullYear(+dateArray[0])` - this is due to surprising behavior where `new Date(year)` will treat years between 0 and 100 as 19xx, such as `new Date(94)` being the year 1994 instead of 94.

src/datepicker/datepicker.js
davious: the reason I'm using the `toISOString()` (I plan to switch to equivalent `toJSON()` next commit) is to get the UTC version of the date, otherwise there is the same timezone issue we are seeing in #3437... the date for EST looks something like 9PM the day before, and the date becomes off by 1.

davious:
Whoops... just realized this is an American biased solution and does work for East Europe and Asia

src/datepicker/datepicker.js
```diff
@@ -629,7 +629,17 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
         scope.$parent.$watch(getAttribute, function(value) {
           scope.watchData[key] = value;
           if (key === 'minDate' || key === 'maxDate') {
-            cache[key] = new Date(value);
+            var date = parseDate(value);
+            if (!date) {
+              date = new Date(value);
+              if (angular.isDate(date) && !isNaN(date)) {
+                var dateArray = date.toJSON().split('T')[0].split('-');
```
wesleycho: As stated prior, one can get the information from the date via this mechanism:

```js
var year = date.getFullYear(),
  month = date.getMonth(),
  day = date.getDate();
```
This is a better way of getting the date information

src/datepicker/datepicker.js
```diff
@@ -629,7 +629,17 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
...
+                date = new Date(+dateArray[0], +dateArray[1] - 1, +dateArray[2]);
```
wesleycho: I should have caught this earlier, but it would be better to do `date = new Date(date)` - this preserves the timezone of `date`, in the event that the value is a date object.

src/datepicker/datepicker.js
```diff
@@ -629,7 +629,17 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
...
+                date.setFullYear(+dateArray[0]);
```
wesleycho: Then here one can do
`date.setFullYear(year, month, day)`

src/datepicker/datepicker.js
davious: The `parseDate returns a date object when the value is a date: it most circumstances, including literals that match the dateFormat, this `toJSON` block is avoided. The `toJSON` block is exclusively for literals to avoid the timezone issues `new Date` gives.

wesleycho: `date` is already a date instance at this stage - `date = new Date(date)` just breaks the reference and makes a fresh copy, but with the timezone of the previous date object being the key difference. Whether the date passed into the component is from a literal is irrelevant.

davious: when I do a `console.log` for the plunkers' cases, they are coming in as their literal strings.

davious: perhaps an alternate logic is...
```
// don't call parseDate... not needed
date = new Date(value);
if (angular.isString(value) && angular.isDate(date) && !isNaN(date)) {
 ...tricks
}
```

wesleycho: When you do `console.log`, `Date.prototype.toJSON()` is called internally to convert the object into literal form. Note the `angular.isDate(date)` check - this ensures that you are working with a date instance.

davious: I meant to say... `console.log(value)` gives a literal when a literal that doesn't match the dateFormat is set on min-date

davious: so the story with these literals is
```
> a = new Date('2001-01-02')
> Mon Jan 01 2001 19:00:00 GMT-0500 (EST)
> a.getDate()
> 1
```

wesleycho: That is why I mentioned earlier that it might be a good thing to make a breaking change and require that the user has passed in a date instead - if the user does `$scope.minDate = new Date(2001, 1, 2)`, and then uses the directive with `<uib-datepicker min-date="minDate" ng-model="date"></uib-datepicker>`, this would be an adequate fix without having to guess at timezones or put in hacks.

davious: Oh, man. Sorry. I was on the wrong page with you. I was going for the edge-case support and incorrectly inferred that you still wanted to handle the edge-case when you added the `needs work` label. I don't know how the other way would be coded; my (not to strong) preference is to support this edge case.

If we decided to still allow literals, but required them to match the dateFormat, then we could just do a string check on the value here and return something like undefined or an invalid date in that case to just have it not work -- maybe that's what you are thinking?

davious:
Changed the code to simply parse dates:
* We now expect literal min-date and max-date to match the displayFormat; otherwise, it will not work at all
* If they want to use an alternate format, they can indicate it in `alt-input-formats`

This fixes the underlying problem in issue #3437, which is that javascript parses `yyyy-MM-dd` as GMT; see https://github.com/angular-ui/bootstrap/issues/3437#issuecomment-160008324

davious:
Just realized that the concepts of display format and alt-input-formats does not exist for the datepicker itself. So, this would still be an issue for the datepicker; and, the different behavior of min-date/max-date would be confusing.

davious:
OK. Switched to using `dateFilter` to feed `new Date` for these literal cases.

davious:
@wesleycho would you re-review?

wesleycho:
Only thing I hesitate on is the assumption of the `medium` format in the `dateFilter` usage - I guess I'm ok with that, but it needs to be mentioned in the documentation.

davious:
The medium format is an internal format picked "at random"... Any format that indicates all the possible date information would do the trick. It doesn't impact the parse work.

wesleycho:
Guess I'm fine with it - LGTM then

PR #4842, opened 2015-11-08
fix(build): uib/ dir for ui.bootstrap.tpls dependency list

davious:
[Review on Reviewable](https://reviewable.io/reviews/angular-ui/bootstrap/4842)


wesleycho:
The path is virtual - from the work in 342c576ea44e85f8406688109cdea12844341c3f, the virtual path has a uib prefix, to be in line with the rest of the project.

davious:
when I do a build, the `dist/ui-bootstrap-tpls-1.0.0-SNAPSHOT.js`
line 9 is `angular.module("ui.bootstrap.tpls", ["template/accordion/accordion-group.html", ...]);`

all the templates looks like
`angular.module("uib/template/accordion/accordion-group.html", [])...`

so, when I use `ui-bootstrap-tpls-1.0.0-SNAPSHOT.js` I get this error 
```
Uncaught Error: [$injector:modulerr] Failed to instantiate module app due to:
Error: [$injector:modulerr] Failed to instantiate module ui.bootstrap due to:
Error: [$injector:modulerr] Failed to instantiate module ui.bootstrap.tpls due to:
Error: [$injector:modulerr] Failed to instantiate module template/accordion/accordion-group.html due to:
Error: [$injector:nomod] Module 'template/accordion/accordion-group.html' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.
```

wesleycho:
I see - it would be better to maybe transform enquote to accept a function then & compose the functions, i.e. enquote(uibDir) instead of a double map.

davious:
How about mapping with this?
```
function enquoteUibDir(str) {
  return enquote('uib/' + str);
}
```

wesleycho:
That'd work too

wesleycho:
Thanks for this - it was a good catch.

Issue #3765, opened 2015-06-05
Modals that contain directives that require outer controller fail


raymMM:
Problem: when a modal template contains directives that require an outer controller, this outer controller must be present in the template otherwise it will fail.
I have created a test case here: http://embed.plnkr.co/UN5gqhrdZXHmWbqDA6Xg/preview
Clicking 'Open Popup' opens the modal. Then clicking 'ok' should call a function from a required outer controller. Unfortunately it will not be found.

Reason for failure: the $modalStack.open function compiles the template content directly on the page body. 

A solution would be to use a 'futureParentEl' as the parent instead of the html-body, when available.
See: https://github.com/angular-ui/bootstrap/commit/50a561a16bc538c8f895f6abfc268fe6e5084cbc

raymMM:
Hi Robin,
could you give me some insight on your decision to put the code in the purgatory?
BR, Ray

rvanbaalen:
From what I can tell in this short time it's basically an implementation issue. I'm still researching if this is an enhancement request or just a question that should be asked on StackOverflow.

That's why I put it there and didn't close it right away.

raymMM:
Thanks Robin. I think that the current implementation does not allow what I
am trying to do here. Is it a bug? Depends on whether or not the modal
should follow the same rules as ng-include for example.

Therefore I created a possible solution,
https://github.com/angular-ui/bootstrap/commit/50a561a16bc538c8f895f6abfc268fe6e5084cbc

rvanbaalen:
In my opinion, a modal should be self containing without parent controller dependencies. I'll ask some team members to take a look at this issue and make sure we're all on the same page.

wesleycho:
I do not believe we should support this. If a custom template is wanted, why not just specify this in a template and load the template with the $modal service?

This use case also has the additional issue that it is cumbersome to work with in general due to template clutter of mixed concerns.

I am inclined to reject this suggestion and close the issue, but I also would like to hear another opinion. @chrisirhc @RobJacobs @Foxandxss ?

chrisirhc:
Hm.. I considered supporting this but it'll be tricky.
If I understand this correctly, this is more of a feature request to support an arbitrary parent element for the modal directive.

The tricky part about this is is dealing with covering all the elements on screen. This is going to be very difficult to support.

PR #4844, opened 2015-11-08
feat(modal): support requiring from parent directive

wesleycho:
- Add support for requiring from parent directive of element being
  appended to

This implements #3765


Foxandxss:
Sounds good to me.

PR #4860, opened 2015-11-10
fix(debounce): Properly pass arguments to debounced function

### LINKED ISSUE #4859: debounce implementation doesn't pass arguments to debounced function
The current implementation incorrectly converts `arguments` collection into `args` array. As the result original function is invoked without arguments even if debounced function is called with some.

Here is the demo with the debounce code from original [commit](https://github.com/angular-ui/bootstrap/commit/73efc782e98123b6def5df85ec2875ab583cade2).
http://plnkr.co/edit/f1kCAlHOcVznlv3Yh5IG?p=preview 

Only the little thing is missing `Array.prototype.slice.call(arguments)` instead of `Array.prototype.slice(arguments)`. Try adding `call` in the demo above to see how it should work.

dfsq:
Original callback should be invoked with correct arguments passed to debounced function.

Closes #4859

### DIFF
--- a/src/debounce/debounce.js
+++ b/src/debounce/debounce.js
@@ -8,7 +8,7 @@ angular.module('ui.bootstrap.debounce', [])
 
       return function() {
         var self = this;
-        var args = Array.prototype.slice(arguments);
+        var args = Array.prototype.slice.call(arguments);
         if (timeoutPromise) {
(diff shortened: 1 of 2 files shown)


wesleycho:
That was careless of me - good catch

PR #4863, opened 2015-11-10
feat(datepicker): popup pass-through formatting attrs

### LINKED ISSUE #3338: Datepicker - year-range seems to be completely ignored
When I use the `year-range` attribute on my `datepicker-popup` directive, it seems to be completly ignored.

I set it to value `9` like this:

```html
year-range="9"
```

and also tried this:

```html
year-range="'9'"
```

But both don't work and this is always the result:

screen shot 2015-02-25 at 14 21 09

I would expect the datepicker to show the years in a 3x3 grid (9 total).

I'm using UI Bootstrap `0.12.1` and Angular JS `1.3.13`

davious:
fixes #3338


wesleycho:
Aha, you beat me to this - I was eyeing this over the weekend heh.

This is a solid job, thanks for the work!

Issue #4847, opened 2015-11-09
Allow dynamic panel-class on accordion


Fidel90:
Hi,

I'd like to update my angular-UI from 0.12.0 to 0.14.3 and just stumbled accross the new panel-class introduced in 0.13.4. I have accordion groups with a dynamic changing class name and wanted to use the new panel-class instead of ng-class like before. But it turns that panel-class will only be evaluated once. I need to use it with angular bindings {{}}. I don't see why there's a need for this attribute anyway as we had ng-class for such things...

Btw: I've tried ng-class instead but that shows some buggy behaviour (at leat in 0.13.4). Being on the site where my accordion is everything works fine at first. But then, changing route to another page and then changing back to my accordion, the ng-class wont be applied again.

wesleycho:
`ng-class` is not supported by angular on directives that use `replace: true`, and thus we do not support this usage.

If one really must bind this way, I recommend using `class="{{classBinding}}"` instead.

Issue #4172, opened 2015-08-10
Bug with accordion-group and ng-class


Polooo2:
Hi,
since 0.13.3 I am getting parse errors when using accordion-group in combination with ng-class. This is being caused by this change ead15e37 . 

You can test this with following example:

`... 
<accordion-group ng-class="{'test' : isTest}"> 
....`

This results in following generated code:
`... 
<accordion-group ng-class="{'test' : isTest} {'panel-open': isOpen}"> 
....`

stephanie-dm:
I can confirm this behaviour as well. And because of this problem, the class is not added when the panel is open.

wesleycho:
Using ng-class on a directive that uses `replace` is a bad idea.

The best solution is to use the `template-url` option on the `accordion-group` directive if you want to eliminate that ng-class usage.

icfantv:
@wesleycho what do you think about adding an `expandedClass` and/or `collapsedClass` option to the directive so the fix to #3419 isn't potentially "breaking?"  Seems pretty heavy-handed to force someone to make a new template because of a feature we added.

wesleycho:
I'm not against it necessarily, but in increases cumbersomeness of usage. replace is also a potentially dangerous thing in tandem with another directive that mutates the DOM.

If a PR is opened, I'll gladly review and merged though.

icfantv:
Agreed.  I recall a comment in an issue in the angular codebase by @caitp indicating that while `replace` isn't going anywhere, one needs to be cognizant that using it incorrectly can break the DOM.

I wonder if it would be best to establish best practices (or at least "standards") on a project level and just say "If we do _x_, then we won't support _y_."  In this case it could be:  "Because we use `replace`, to prevent breaking the DOM, we won't support certain, other directives, like `ng-class` on our components."

wesleycho:
I can get behind that, as ideally, you should avoid using directives that mutate the DOM in combination with a directive that uses `replace`.

One can wrap the element with a `div` that has the desired ng-class to work around this also.

icfantv:
@wesleycho found [this link](https://github.com/angular/angular.js/issues/5695) via #1741.  Look at @caitp's last comment.  If Angular isn't going to address this, I think we can close this issue and #1741 and add a comment to the docs about not supporting certain directives on UI Bootstrap directives that use `replace: true`.

wesleycho:
That settles it for me - closing as won't fix as per resolution in Angular.

[...]

PR #4870, opened 2015-11-10
fix(accordion): panel-open class not shown with custom class

### LINKED ISSUE #4849: Accordion wont apply panel-open class when initially opened and custom class binding is used
I use the accordion with dynamic class names. I'm using angular bindings for this. As in current version this is not supported via `ng-class` and the `panel-class` attribute is only evaluated once, I have to use the standard `class` attribute. So I do sth like this:

`<accordion-group class={{customClass}} is-open="true"></accordion-group>`

Where `{{customClass}}` will be replaced by `"myCustomClass"`. When the `is-open` attribute is set by default and I navigate to the page with the accordion, the class attribute evaluates to:

`<accordion-group class="myCustomClass panel panel-default panel panel-default" is-open="true"></accordion-group>`

But `"panel-open"` will only be applied if I close and reopen the affected group. You may look this up here and check the class value via debugger (I've just added a constant `class="testClass"` in my example as the result is the same):

http://plnkr.co/edit/cOJDzM2lYdVkA9TWmBDP?p=preview

chenyuzhcy:
Closes #4849 

Working plunker: http://plnkr.co/edit/9ZBDb8jvo0ds5lhTW5MH?p=preview

### DIFF
--- a/template/accordion/accordion-group.html
+++ b/template/accordion/accordion-group.html
@@ -1,4 +1,4 @@
-<div class="panel {{panelClass || 'panel-default'}}">
+<div class="panel" ng-class="panelClass || 'panel-default'">
   <div class="panel-heading" ng-keypress="toggleOpen($event)">
     <h4 class="panel-title">
       <div tabindex="0" class="accordion-toggle" ng-click="toggleOpen()" uib-accordion-transclude="heading"><span ng-class="{'text-muted': isDisabled}">{{heading}}</span></div>
--- a/template/accordion/accordion-group.html
+++ b/template/accordion/accordion-group.html
@@ -1,4 +1,4 @@
-<div class="panel" ng-class="panelClass || 'panel-default'">
+<div class="panel">
   <div class="panel-heading" ng-keypress="toggleOpen($event)">
     <h4 class="panel-title">
       <div tabindex="0" class="accordion-toggle" ng-click="toggleOpen()" uib-accordion-transclude="heading"><span ng-class="{'text-muted': isDisabled}">{{heading}}</span></div>
--- a/src/accordion/accordion.js
+++ b/src/accordion/accordion.js
@@ -76,7 +76,8 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
       accordionCtrl.addGroup(scope);
 
       scope.openClass = attrs.openClass || 'panel-open';
-      scope.panelClass = attrs.panelClass;
+      scope.panelClass = attrs.panelClass || 'panel-default';
+      element.addClass(scope.panelClass);
       scope.$watch('isOpen', function(value) {
         element.toggleClass(scope.openClass, !!value);
         if (value) {
--- a/src/accordion/test/accordion.spec.js
+++ b/src/accordion/test/accordion.spec.js
@@ -426,6 +426,29 @@ describe('uib-accordion', function() {
       });
     });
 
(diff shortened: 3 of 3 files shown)


icfantv:
@chenyuzhcy, so unfortunately, we can't use `ng-class` here.  In fact, this is what we used to have elsewhere in our code base (and perhaps even here...I don't remember off hand).

The issue is that the `uib-accordion-group` directive uses `replace: true` and Angular has decided to not support reconciling the `ng-class` issues when using this property.

wesleycho:
I don't think we used to use ng-class at all - we did manage to break ng-class support somehow in part due to replace: true being applied.

For reference, here was the prior commit that added the dynamic class addition: https://github.com/angular-ui/bootstrap/commit/5ee23a4c2668e6d1b1a5aef39a1804d7d5d9ec6a#diff-2c7728e12322ff30329ef487569c9435

icfantv:
@chenyuzhcy, [here's the link](https://github.com/angular/angular.js/issues/5695) to the angular issue addressing the `ng-class` issues.

icfantv:
@wesleycho, I couldn't remember exactly, only that we've decided to not support it.  #4847 from a few days ago says as much.

wesleycho:
I thought our decision applied to people using ng-class on the accordion-group element?

icfantv:
...and not the template used internally....yes.  That's totally fair.

template/accordion/accordion-group.html
```diff
@@ -1,4 +1,4 @@
-<div class="panel {{panelClass || 'panel-default'}}">
```
wesleycho: I think we do want the ng-class to replace the interpolated class, so
```html
<div class="panel" ng-class="panelClass || 'panel-default'">
```
This is to keep behavior consistent with dynamic class changing.

wesleycho:
I made the change - thanks for spending the time on this quick fix!

trickpattyFH20:
@wesleycho I very passionately disagree with this breaking change to use ng-class.

We had ng-class being used on a accordion-group element which now breaks after upgrading to the latest release.

I belive the ng-class merging should have either been done internally here till Angular releases a fix, or the implementation of ng-class should have just been avoided.

I was going to write a fix for this bug, but since the issue has already been closed I will begrudgingly rewrite my CSS to support ng-class being used on a wrapping parent element, even though this seems ludicrous.

wesleycho:
@trickpattyFH20 you are relying on unsupported behavior in Angular itself with regards to ng-class and directives that use `replace: true`. See #4172 for details.

trickpattyFH20:
@wesleycho wouldn't it be plausible to rework the uibAccordionGroup directive to *not* use replace:true ?

wesleycho:
That would result in breaking how Bootstrap's CSS works, and all theming libraries.

trickpattyFH20:
@wesleycho do you have any sources about what is considered "abuse" of ng-class?
I don't think that an aditional interpolated class should be considered abuse.

I could see it as abuse if you parsed attrs.ngClass and recompiled it in the directive...
but why not just avoid all of that and use the interpolated class like before? This seems like a perfectly reasonable workaround to (#5695)[https://github.com/angular/angular.js/issues/5695]

wesleycho:
A simpler workaround would be to just do `class="{{customClasses()}}"` than bloating library code for a situation explicitly not supported by Angular.

PR #4871, opened 2015-11-10
hide tooltips and popovers when a user clicks away (corrected)

thegreenpizza:
Please refer to #4419 for complete context. This pull request fixes the issues that cause the Travis CI build to fail.


SystemDisc:
@thegreenpizza and @wesleycho 

I used `XYZ == null` to check if `XYZ` is `null` or `undefined`. If the linter doesn't like the `==`, this should probably be changed from `XYZ == null` to `!XYZ`, instead of `XYZ === null`

I know this was already committed to master, but I just wanted to let you guys know that you broke my check - it's only checking whether or not the variable is strictly `null` or not.

wesleycho:
@SystemDisc can you file a PR with the fix?

SystemDisc:
@wesleycho Yeah, I can do that tomorrow.

PR #4874, opened 2015-11-11
test(datepicker): add a few tests

Foxandxss:
I added a few tests, one of them made me fix a bug also.

Bit of renaming here and there as well.

There are more tests to be done for better coverage, but that is for another PR.


icfantv:
:+1:

PR #4876, opened 2015-11-11
refactor(datepicker): simplify miscelanea stuff

Foxandxss:
The up / down... comments were really confusing and not needed at all.

The six-month one, totally my fault, I added that feature.

I merged the `self[key] = $scope[key]` stuff. I am not too fond of that syntax, but that is what we do everywhere already.

Also, the `$watch` doesn't really need the watchFn to be parsed, you can feed it with a string.


Foxandxss:
Wait on this one, saw another one to refactor, but in the middle of another PR atm.

icfantv:
:+1:

src/datepicker/datepicker.js
```diff
@@ -40,7 +40,7 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   // Watchable date attributes
   angular.forEach(['minDate', 'maxDate'], function(key) {
     if ($attrs[key]) {
-      $scope.$parent.$watch($parse($attrs[key]), function(value) {
```
wesleycho: Isn't this a breaking change? What if `$attrs[key]` is an expression?

src/datepicker/datepicker.js
```diff
@@ -51,17 +51,15 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
 
   angular.forEach(['minMode', 'maxMode'], function(key) {
     if ($attrs[key]) {
-      $scope.$parent.$watch($parse($attrs[key]), function(value) {
```

src/datepicker/datepicker.js
wesleycho: Hmm, I guess this is technically ok, but by passing in $parse, I think it actually saves slightly on performance since $parse already has cached the string parsing of the attribute value.

Issue #4873, opened 2015-11-11
Datepicker custom class on other modes


Foxandxss:
Custom classes are technically meant to be applied in all 3 modes (or that is what tests says) but it is only applied to `day` mode.

So it needs to be applied in the other two modes too.

Also tests to make sure that it is applied in popup mode too.

wesleycho:
I'm not able to reproduce this as an issue - see [here](http://plnkr.co/edit/7gpZd0EGmlTgbo6tGRRJ?p=preview). There are also tests against all 3 modes.

Foxandxss:
Seems like the other two (month and year) were fixed in separated PRs and I never saw them. My bad.

PR #4877, opened 2015-11-11
chore(datepicker): fix customClass demo

Foxandxss:
It was jumping one day on the algorithm.

This is not related with #4873.


icfantv:
:+1:

PR #4891, opened 2015-11-13
fix(dropdown): remove extra uib-keyboard-nav

Foxandxss:
BREAKING CHANGE: `keyboard-nav` for the dropdown is not longer a directive and to use it you have to use `keyboard-nav` instead of `uib-keyboard-nav`.


Foxandxss:
Tell me that I am not crazy and that the guy that implemented this, did it twice... Why do we need the whole keyboard nav both in a directive and also inside the main controller? Works directly as is.

Tweaked a bit the demo to include another keyboard-nav.

And don't worry about the docs, I have to rewrite them entirely.

RobJacobs:
LGTM

icfantv:
You're not crazy.  I noticed that before and it was the cause of a bunch of WTFs when trying to write tests because the code that you thought should be called was never getting called.  E.g. when you trigger the `keydown` event in jasmine it wasn't using the event handler on the directive - it was using the main one.  But in an actual web page, the opposite is true.  It took me a days to figure that out.  Very frustrating.

Foxandxss:
It took me a while too to figure it out.

Foxandxss:
Hold on this one, I saw a big stupidity on the tests.

Foxandxss:
Alright, fixed.

PR #4892, opened 2015-11-13
refactor(dropdown): remove regex to not match 138 and 140

Foxandxss:


wesleycho:
LGTM - on the flip side, the regex probably should have been `/^38|40$/` or something similar

Foxandxss:
You did this change in the past (for the directive) and it was argued that well, less performant as well.

PR #4893, opened 2015-11-13
refactor(dropdown): remove unneeded destroy

Foxandxss:
So the deal here is: On children destroy, children destroy again. Also: On parent destroy, destroy children. That is not needed, when a scope is destroyed, all children are as well.

Also, listen for this event is when to do some cleanup and there is none in there.


wesleycho:
That is a silly $destroy - LGTM

PR #4894, opened 2015-11-13
refactor(dropdown): remove rootScope from the service

Foxandxss:
All scopes shares the same $$phase string, so there is no need to inject $rootScope when you have a scope hanging around already.


Foxandxss:
Perhaps I am wrong? let me check

Foxandxss:
No, I am just stupid, fixed.

wesleycho:
I think $rootScope's was supposedly the only reliable one though - I am not sure if that changed though.

Foxandxss:
It is just a variable, I don't think it was never non reliable.

wesleycho:
Looking at https://github.com/angular/angular.js/blob/master/src/ng/rootScope.js#L1312-L1322 , this appears to confirm my memory that we should only check against $rootScope.$$phase ever.

Foxandxss:
That doesn't prove anything, that is just the logic to start / stop phases. What I mean is that there is `$$phase` object in the $rootScope and all children has access to it and it is the same object everywhere.

Check: http://plnkr.co/edit/Osh4PGOSOkPBkpRXF866?p=preview

wesleycho:
I absolutely would not trust that - the other mention of $$phase is https://github.com/angular/angular.js/blob/master/src/ng/rootScope.js#L169, and $rootScope.$$phase is also used in $evalAsync. IMO this is an invalid change & risky.

wesleycho:
As per discussion, closing as $$phase can be different if an isolate scope lies in between the inheritance chain.

PR #4898, opened 2015-11-13
test(dropdown): use existing matchers

Foxandxss:
Easier to read, and have nicer message error than `expected true to be false`.


wesleycho:
LGTM

PR #4899, opened 2015-11-13
feat(position): implement auto positioning

### LINKED ISSUE #4762: Add smart/fixed positioning option for components with dropdowns/popups and for tooltips/popovers
In an effort to give users more flexibility to dictate where popups and tooltips/popovers we need to add the ability let them specify where and how they want the item to be displayed.

**Supported Locations** - Top, Right, Bottom, Left, Top Left, Top Right, Bottom Left, and Bottom Right.

**Smart Positioning** - when used, implies that the library will determine the best location for the item.

**Fixed Positioning** - will display the item at the specified location.

**Hybrid Positioning** - exact behavior TBD, but maybe something like letting the user specify a fixed position and if it doesn't fit, use smart positioning.  Or perhaps, letting the user specify an array of positions to try before using smart positioning.

**Supported Components**:

- [ ] Datepicker
- [ ] Dropdown
- [ ] Popover
- [ ] Tooltip
- [ ] Typeahead

RobJacobs:
Implementing the ability to auto position elements in the position service - positionElements function.  This will allow tooltips and popovers to position where they fit within the closest scrollable ancestor.  As part of this PR I also documented the position service functions and refactored some of the code to make it more readable.  Partially closes #4762.

### DIFF
--- a/src/position/docs/readme.md
+++ b/src/position/docs/readme.md
@@ -1,0 +1,167 @@
+The `$uibPosition` service provides a set of DOM utilities used internally to absolute-position an element in relation to another element (tooltips, popovers, typeaheads etc...).
+
+### getRawNode(element)
+  Takes a jQuery/jqLite element and converts it to a raw DOM element.
+
+##### parameters
+
+* `element` _(Type 'object')_ - The element to convert.
+
+##### returns
+
+  _(Type 'element')_ - A raw DOM element.
+
+### parseSyle(element)
+  Parses a numeric style value to a number.  Strips units and will return 0 for invalid (NaN) numbers.
+
+##### parameters
+
+* `value` _(Type 'string')_ - The style value to parse.
+
+##### returns
+
+  _(Type 'number')_ - The numeric value of the style property.
+
+### offsetParent(element)
+  Gets the closest positioned ancestor.
+
+##### parameters
+
+* `element` _(Type 'element')_ - The element to get the offset parent for.
+
+##### returns
+
+  _(Type 'element')_ - The closest positioned ancestor.
+
+### scrollbarWidth()
+  Calculates the browser scrollbar width and caches the result for future calls.  Concept from the TWBS measureScrollbar() function in [modal.js](https://github.com/twbs/bootstrap/blob/master/js/modal.js).
+
+##### returns
+
+  _(Type 'number')_ - The width of the browser scrollbar.
+
+### scrollParent(element, includeHidden)
+  Gets the closest scrollable ancestor.  Concept from the jQueryUI [scrollParent.js](https://github.com/jquery/jquery-ui/blob/master/ui/scroll-parent.js).
+
+##### parameters
+
+* `element` _(Type 'element')_ - The element to get the closest scrollable ancestor for.
+
+* `includeHidden` _(Type 'boolean', optional - default is false)_ - Should scroll style of 'hidden' be considered.
+
+##### returns
+
+  _(Type 'element')_ - The closest scrollable ancestor.
+
+### position(element, includeMargins)
+  A read-only equivalent of jQuery's [position](http://api.jquery.com/position/) function, distance to closest positioned ancestor. Does not account for margins by default like jQuery's position.
+
+##### parameters
+
+* `element` _(Type 'element')_ - The element to get the position for.
+
+* `includeMagins` _(Type 'boolean', optional - default is false)_ - Should margins be accounted for.
+
+##### returns
+
+  _(Type 'object')_ - An object with the following properties:
+
+* width: _(Type 'number')_ The width of the element.
+* height: _(Type 'number')_ The height of the element.
+* top: _(Type 'number')_ Distance to top edge of offset parent.
+* left: _(Type 'number')_ Distance to left edge of offset parent.
+
+### offset(element)
+  A read-only equivalent of jQuery's [offset](http://api.jquery.com/offset/) function, distance to viewport.
+
+##### parameters
+
+* `element` _(Type 'element')_ - The element to get the offset for.
+
+##### returns
+
+  _(Type 'object')_ - An object with the following properties:
+
+* width: _(Type 'number')_ The width of the element.
+* height: _(Type 'number')_ The height of the element.
+* top: _(Type 'number')_ Distance to top edge of the viewport.
+* left: _(Type 'number')_ Distance to left edge of the viewport.
+
+### viewportOffset(element, useDocument, includePadding)
+  Gets the elements available space relative to the closest scrollable ancestor.  Accounts for padding, border, and scrollbar width.
+  Right and bottom dimensions represent the distance to the respective edge of the viewport element, not the top and left edge.
+  If the element edge extends beyond the viewport, a negative value will be reported.
+
+##### parameters
+
+* `element` _(Type 'element')_ - The element to get the viewport offset for.
+* `useDocument` _(Type 'boolean', optional - default is false)_ - Should the viewport be the document element instead of the first scrollable element.
+* `includePadding` _(Type 'boolean', optional - default is true)_ - Should the padding on the viewport element be accounted for, default is true.
+
+##### returns
+
+  _(Type 'object')_ - An object with the following properties:
+
+* top: _(Type 'number')_ Distance to top content edge of the viewport.
+* bottom: _(Type 'number')_ Distance to bottom content edge of the viewport.
+* left: _(Type 'number')_ Distance to left content edge of the viewport.
+* right: _(Type 'number')_ Distance to right content edge of the viewport.
+
+### parsePlacement(placement)
+  Gets an array of placement values parsed from a placement string. Along with the 'auto' indicator, supported placement strings are:
+
+* top: element on top, horizontally centered on host element.
+* top-left: element on top, left edge aligned with host element left edge.
+* top-right: element on top, right edge aligned with host element right edge.
+* bottom: element on bottom, horizontally centered on host element.
+* bottom-left: element on bottom, left edge aligned with host element left edge.
+* bottom-right: element on bottom, right edge aligned with host element right edge.
+* left: element on left, vertically centered on host element.
+* left-top: element on left, top edge aligned with host element top edge.
+* left-bottom: element on left, bottom edge aligned with host element bottom edge.
+* right: element on right, vertically centered on host element.
+* right-top: element on right, top edge aligned with host element top edge.
+* right-bottom: element on right, bottom edge aligned with host element bottom edge.
+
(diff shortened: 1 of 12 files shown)


icfantv:
@RobJacobs, is this for the 1.2 release or were you thinking 1.0?  @wesleycho?

What would you guys think about having it as a beta feature (if that's even possible).

wesleycho:
This is just the $uibPosition service

src/position/docs/readme.md
```diff
@@ -0,0 +1,180 @@
+The `$uibPosition` service provides a set of DOM utilities used internally to absolute-position an element in relation to another element (tooltips, popovers, typeaheads etc...).
+
+#### <strong>getRawNode(element)</strong>
```
wesleycho: Not hugely opinionated, but you can get the `<strong>` formatting from `*text*`

src/position/position.js
```diff
@@ -1,149 +1,539 @@
...
+       * @param {element} elem - The element to find the scroll parent of.
+       * @param {boolean=} [includeHidden=false] - Should scroll style of 'hidden' be considered,
+       *   default is false.
+       *
+       * @returns {element} A HTML element.
+       */
+      scrollParent: function(elem, includeHidden) {
+        elem = this.getRawNode(elem);
+
+        var overflowRegex = includeHidden ? /(auto|scroll|hidden)/ : /(auto|scroll)/;
```
wesleycho: Cache these regexes above the factory return

src/position/position.js
```diff
@@ -1,149 +1,539 @@
...
+       *
+       * @returns {array} An array with the following values
+       * <ul>
+       *   <li>**[0]**: The primary placement.</li>
+       *   <li>**[1]**: The secondary placement.</li>
+       *   <li>**[2]**: If auto is passed: true, else undefined.</li>
+       * </ul>
+       */
+      parsePlacement: function(placement) {
+        var autoRegex = /\s?auto?\s?/i;
```
wesleycho: Cache this regex as well.

src/position/position.js
```diff
@@ -1,149 +1,539 @@
...
+        var autoPlace = autoRegex.test(placement);
+        if (autoPlace) {
+          placement = placement.replace(autoRegex, '');
+        }
+
+        placement = placement.split('-');
+
+        placement[0] = placement[0] || 'top';
+        var primaryRegex = /^(top|bottom|left|right)$/;
```
wesleycho: Cache regex

src/position/position.js
```diff
@@ -1,149 +1,539 @@
...
+        if (!primaryRegex.test(placement[0])) {
+          placement[0] = 'top';
+        }
+
+        placement[1] = placement[1] || 'center';
+        var secondaryRegex = /^(top|bottom|left|right|center)$/;
```
wesleycho: Cache regex

src/position/position.js
```diff
@@ -1,149 +1,539 @@
...
+      *
+      * @param {element} elem - The tooltip/dropdown element.
+      * @param {string} placement - The placement for the elem.
+      */
+      positionArrow: function(elem, placement) {
+        elem = this.getRawNode(elem);
+
+        var isTooltip = true;
+
+        var innerElem = elem.querySelector('.tooltip-inner');
```
wesleycho: Is there any way we can make this tooltip/popover agnostic?

src/position/docs/readme.md
Foxandxss: actually `**text**`

src/position/position.js
```diff
@@ -1,149 +1,539 @@
...
+          top: 'auto',
+          bottom: 'auto',
+          left: 'auto',
+          right: 'auto',
+          margin: 0
+        };
+
+        switch (placement[0]) {
+          case 'top':
+            arrowCss.bottom = isTooltip ? '0' : '-' + borderWidth + 'px';
```
wesleycho: Couldn't you just do `-borderWidth + 'px'`?

wesleycho:
Were you able to test this on multiple browsers?

src/position/position.js
RobJacobs: The work done to position the arrow elements in this function are specific to the TWBS tooltip and popover arrow styles.  I didn't want it to work with anything else at this point.

src/position/position.js
RobJacobs: This should actually be:

    '-' + borderWidth;

As getComputedStyle will return units and be interpreted as a string.

RobJacobs:
@wesleycho PR updated per comments, couple of other tweaks worth noting:

* Added placementClassPrefix option to tooltip config.
* Formatted readme.md to resemble dateParser doc.
* Removed the 'auto' option as the default tooltip behavior so this will be an opt-in feature.

I've tested this change on the latest versions of Chrome, Firefox, Edge, IE11, IE10 & 9 using IE11 emulation mode, and on an iPad with iOS9.

icfantv:
I should say, awesome work @RobJacobs.

src/position/position.js
```diff
@@ -1,149 +1,546 @@
...
+        var documentEl = $document[0].documentElement;
+        var elemStyle = $window.getComputedStyle(elem);
+        var scrollParent = elem.parentElement || documentEl;
+
+        if (scrollParent === documentEl || elemStyle.position === 'fixed') {
+          return documentEl;
+        }
+
+        while (scrollParent.parentElement && scrollParent !== documentEl) {
+          var spStyle = $window.getComputedStyle(scrollParent);
```
wesleycho: Is this the best way to do this? `getComputedStyle` is a pretty expensive function, and inside a while loop could be quite pricey. What happens if this never finds an absolutely positioned or position static parent element?

How many times at worst will this loop fire?

RobJacobs:
@icfantv Thank you sir!  Once/if this PR is accepted, it should be pretty straight forward to implement on the datepicker and typeahead.

src/position/position.js
RobJacobs: @wesleycho  I was concerned about that as well, but it's the same approach the parentOffsetEl function uses in a while loop when calling the isStaticPositioned function.  getComputedStyle is the only way I know of to get the styles applied by CSS and not just inline styles that element.style provides.

I could refactor and call the getStyle function instead of calling getComputedStyle directly as I only need the position style value.

wesleycho:
We also want this implemented on the dropdown as well.

This is amazing work btw - I took a stab at this before, but positioning turns out to be quite formidable.

src/position/position.js
wesleycho: I did a bit of sleuthing, and it looks like we can kill `getStyle` - looks like it was necessary for IE8 support (IE8 used `currentStyle` instead of having the built-in `getComputedStyle` function), but it is now unnecessary.

RobJacobs:
Implementing this on the dropdown will present a different set of challenges as the TWBS dropdown relies more on CSS for positioning as I described in #4762, but I'm confident it is possible.  Next thing I want to look at is coming up with a generic 'popup' directive that would encapsulate some of the redundant code I'm seeing:

* Adding triggers
* placement attribute
* is-open attribute
* append-to-body attribute
* open/close delay

And switch datepicker, typeahead, popover, and dropdown directives to use that.  Then trim down the tooltip directive to serve just that purpose - tooltips.  It's getting pretty heavy trying to support all the different use cases.

wesleycho:
For the dropdown, we can restrict it to being positioned up, and maybe auto with only up being applied.

src/position/position.js
RobJacobs: @wesleycho PR updated with getStyle function removed.

src/position/docs/readme.md
```diff
@@ -0,0 +1,167 @@
+The `$uibPosition` service provides a set of DOM utilities used internally to absolute-position an element in relation to another element (tooltips, popovers, typeaheads etc...).
+
+### getRawNode(element)
+  Takes a jQuery/jQLite element and converts it to a raw DOM element.
```
wesleycho: I believe this should be `jqLite` (minor quibble)

wesleycho:
This PR LGTM - just a minor doc quibble, but this is good to merge otherwise.

icfantv:
Awwwww yea.  @RobJacobs you should do the honors.

Foxandxss:
You can merge it. I will check the docs in another moment and fix them if I feel like I have to. Sorry hadn't time to check it properly.

Foxandxss:
Hold tight. If this doesn't have a demo, the documentation won't show up on the page.

RobJacobs:
@Foxandxss Added demo page, @wesleycho Fixed jqLite typo.  Let me know if you see any other issues.

src/position/docs/readme.md
```diff
@@ -0,0 +1,167 @@
...
+##### parameters
+
+* `element` _(Type 'element')_ - The element to get the offset parent for.
+
+##### returns
+
+  _(Type 'element')_ - The closest positioned ancestor.
+
+### scrollbarWidth()
+  Caclulates the browser scrollbar width and caches the result for future calls.  Concept from the TWBS measureScrollbar() function in [modal.js](https://github.com/twbs/bootstrap/blob/master/js/modal.js).
```
Foxandxss: Calculates

Foxandxss:
About the demo, it is too technical in my opinion. I liked those demos you put to use where you can see tooltips in certain positions. That + all the technical information that you offers now is for me better.

Personally I don't like putting all that huge doc on the page, but we don't have a better mechanism for that either.

The doc itself has some nitpicks I want to change (Styling) so I would create a PR when this gets merged.

RobJacobs:
@Foxandxss Updated tooltip and popover examples to showcase all available positions.  Let me know if you see anything else, thanks.

Foxandxss:
It is good to me. There are some fixed docs / demo related I want to do but I better do it in another PR so we don't clutter this too much.

megabyzus:
Apologies, for the intrusion, does this comprehensively address popover (and tooltip) running over the edge of the page? If so, when will this be released into the wild (we are on 0.14.3 currently) and, if not, can we leverage this to address this current run-over issue once released?

icfantv:
@megabyzus, while it is comprehensive, there's no way we can test it like releasing it into the wild for you guys to hammer.

It is part of `master` and will be in the official 1.0 release.  This release is on our radar and we really want to release it by the end of the year, but as of right now we don't have a target date.  If you want to start using it, feel free to pull `master` and run `npm install` followed by `grunt` to build your own SNAPSHOT copy.  The built/generated files will be in the `dist` directory.

megabyzus:
@icfantv Much appreciated.  To address this immediately, do you have any links you can suggest to resolve this meanwhile?

icfantv:
Not sure I follow, @megabyzus.

Issue #4903, opened 2015-11-14
modal: If something within a modal gets itself the focus then that focus is lost after the animations finish.


arkarkark:
If you have directives inside your modal content that selectively get the focus then all that is lost when the modal finishes loading. (this is a regression somewhere between v0.12.1 and v0.14.2).

icfantv:
@arkarkark, when you say "selectively gets the focus" do you mean that you are setting the focus programmatically?  If so, where does that code live?  If not, what is your focus expectation?

The `$uibModal.open({...}).rendered` is a promise that is executed when the modal is finished rendering.  The correct behavior would be to call the focusing code in this block.

Please let us know.  Thanks.

arkarkark:
Hi @icfantv thanks for the quick reply.

Let's say my modal is an item picker - if there are no items I want the focus to be on the 'create new item' button but if there are items I want the focus to be on the select where I can choose one. 

I can't use autofocus attribute for this. but I can use a focus-when directive that looks at its attribute and if it's true gives the element focus.

That all happens as the modal is rendering, then it finishes as this code I wrapped in an if block steals the focus and gives it to the modal div.

I can imagine many other situations where I'm using components where I can't get the element I want to have focus to have an autofocus attribute, but I can programmatically find it and call focus() on it.

It's interesting that you mention the rendered promise, because even that isn't good enough anymore, because the focus isn't stolen until after the animation promise is resolved which is after the rendered promise is resolved (I was going to submit a PR to fix that too). There's no way outside of the modal to know when the animation promise is finished so even if you have no css animations, you still need to wrap your callback to the rendered promise in a $timeout to get it past the animation promise.

But using the rendered promise (even if it worked properly) or some surfaced animationComplete promise is much more frustrating for someone who only deals with the modal templates - right now they can use directives to say 'this' should have focus if these conditions are met. But if we're forced to use rendered we need to throw some rendered variable on the scope and wait until that becomes true before doing our logic. more watches, more focus flickering (which the comment near the modified code appears to indicate is very bad on mobile w.r.t. on screen keyboards).

languorous001:
Sorry. my English is not good.

Hello, I using modal too and i load ngAnimate.

In my modal I using <uib-carousel> for slide images and yes I using ngAnimate.
Sometime animate is success sometime is fail.
(Sometime I can slide 2 picture from 5 picture, Sometime I can slide 5 pictur from 5 picture)

[...]

PR #4904, opened 2015-11-14
fix(modal): leave focus with an element inside a popped up modal

arkarkark:
fixes https://github.com/angular-ui/bootstrap/issues/4903


wesleycho:
I am ok with this fix - the unit test does indeed fail without the patch.

PR #4913, opened 2015-11-16
refactor(modal): don't use reserved keywords for IE9 compatibility

BrianGenisio:
IE9 sees `.finally()` as a keyword.  This gets fixed with the ".min.js" version of the build, but precludes us from using the non-minified version in development.  Using the `["finally"]` form of this function fixes it and all is good with the world.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -26,7 +26,7 @@
     "block-scoped-var": 2,
     "complexity": 2,
     "curly": 2,
-    "dot-notation": 2,
+    "dot-notation": [2, {"allowKeywords": false}],
     "eqeqeq": 2,
     "guard-for-in": 2,
     "semi": 2,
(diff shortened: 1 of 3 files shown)


pkozlowski-opensource:
Could you please:
* use single quotes to be consistent with the rest of the file / library?
* change the commit message to follow our conventions. Sth like `refactor(modal): don't use reserved keywords for IE9 compatibility` or similar

pkozlowski-opensource:
And fix the build failure :-) https://travis-ci.org/angular-ui/bootstrap/builds/91434961

Foxandxss:
This hack is the build failure. I don't know if there are eslint rules (like in jshint) to disable it for some stuff. If not, this won't pass tests ever.

BrianGenisio:
Yeah, I think I'll need to add an exception comment or something.  I'm on it.

BrianGenisio:
Turns out you want the `{"allowKeywords": false}` option.  Turning that on found two more cases (in collapse.js) where this is happening.  http://eslint.org/docs/rules/dot-notation.html

Foxandxss:
Was going to suggest you to add it, but I see you did it already.

wesleycho:
LGTM - will merge this in shortly.

PR #4914, opened 2015-11-16
Corrected tab heading in example code

justin-john:
Incorrect closing tag for `<uib-tab-heading>` in example code for `tabs directive` in `tabs.js`. It is given as opening tag `<uib-tab-heading>` and closing tag `</tab-heading>`.


justin-john:
@Foxandxss That's all right :) I tried to copy the code from file for testing purpose in my app, but it doesn't render properly. Later I discovered it was because of this that tabs heading didn't render properly. So I just send pull request.

Foxandxss:
Yeah sorry.

That comment code is for documentation generation with ngdoc. We tried to do that really in the past (at least 1,5years ago) but we never ended doing it. I need to do some cleanup

justin-john:
@Foxandxss No worries. Please let me know if anything can be done that will help your cleanup of comments.

PR #4929, opened 2015-11-19
perf(dropdown): avoid extra digest cycle

harai:
This avoids an extra digest cycle when possible.


Foxandxss:
How does that avoid an extra digest cycle?

pkozlowski-opensource:
> How does that avoid an extra digest cycle?

Yeh, I had a look a this change - it won't prevent digest cycles - `$locationChangeSuccess` is already fired as part of the digest loop. So adding this check is not necessary - it adds bytes for nothing.

PR #4935, opened 2015-11-22
feat(pager): move to separate component

### LINKED ISSUE #4803: Move pager to its own file
Currently pager and pagination are "too close" together: mixed in one file and in documentation. There is no good reason for this, other than some code reuse.

Proposal: make pager top-level directive and factor-out shared code to a service (if any)

wesleycho:
- Separate pager into its own component

This partially resolves #4803.


Foxandxss:
Sure, good to me.

Issue #4880, opened 2015-11-11
Dateparser working with literals as o'clock


Foxandxss:
Should this work?:

![](http://puu.sh/lhImq/6fc1206abc.png)

wesleycho:
Not exactly elegant, but `yyyy/MM/dd o''''clock` works (4 x `'`)

Foxandxss:
That is weird. Needs more research.

PR #4936, opened 2015-11-22
docs(dateparser): add docs about template literals

wesleycho:
- Adds documentation about template literals

This addresses #4880.


Foxandxss:
I still think that that is really really really ugly.

wesleycho:
It is...perhaps we can change it later to another character? Not too many characters come to mind though...maybe backticks?

PR #4938, opened 2015-11-23
feat(dateparser): change template literal from ' to `

wesleycho:
- Changes from ' to ` for template literals
- Adds documentation about template literals

BREAKING CHANGE: This changes template literals from the single quote
`'` to the backtick character `` ` `` for more natural usage with certain
strings such as `o'clock`.

This supercedes #4936 and addresses #4880.


src/dateparser/docs/README.md
```diff
@@ -140,3 +140,5 @@ Certain format codes support i18n. Check this [guide](https://docs.angularjs.org
 \* The ones marked with `Leading 0`, needs a leading 0 for values less than 10. Exception being milliseconds which needs it for values under 100.
 
 \** It also supports `fullDate|longDate|medium|mediumDate|mediumTime|short|shortDate|shortTime` as the format for parsing.
+
+\*** It supports template literals as a string between the backtick `\`` character, i.e. `\`The Date is\` MM/DD/YYYY`. If one wants the literal backtick character, one must use `\`\`\`\``.
```
Foxandxss: I don't understand the last part, it ends like ````` or something like that.

wesleycho: It should be 4 backticks

Foxandxss: I added a random bunch in there. But I don't understand why you need 4 to put one literal.

wesleycho: It has to do with the implementation - probably could use some investigation here as to what is going on, but I am just documenting it.

Foxandxss:
Good for me then.

PR #4952, opened 2015-11-25
feat(datepicker): alt-input-formats

### LINKED ISSUE #4951: Datepicker popup: Support Multiple Input Formats
I'm using a fancy `MMMM dd yyyy` format on my datepicker popup. I like the fancy, but I'd also like my users to be able to type in a date using the `M!/d!/yyyy` format which, to them, is the go-to date format. Would angular bootstrap add a way to make this a reality? Maybe a list of acceptable date format inputs?

davious:
fixes #4951

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/bootstrap/4952)

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -561,7 +561,7 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
     onOpenFocus = angular.isDefined(attrs.onOpenFocus) ? scope.$parent.$eval(attrs.onOpenFocus) : datepickerPopupConfig.onOpenFocus;
     datepickerPopupTemplateUrl = angular.isDefined(attrs.datepickerPopupTemplateUrl) ? attrs.datepickerPopupTemplateUrl : datepickerPopupConfig.datepickerPopupTemplateUrl;
     datepickerTemplateUrl = angular.isDefined(attrs.datepickerTemplateUrl) ? attrs.datepickerTemplateUrl : datepickerPopupConfig.datepickerTemplateUrl;
-    altInputFormats = angular.isDefined(attrs.altInputFormats) ? attrs.altInputFormats.split(',') : datepickerPopupConfig.altInputFormats;
+    altInputFormats = angular.isDefined(attrs.altInputFormats) ? scope.$parent.$eval(attrs.altInputFormats) : datepickerPopupConfig.altInputFormats;
 
     scope.showButtonBar = angular.isDefined(attrs.showButtonBar) ? scope.$parent.$eval(attrs.showButtonBar) : datepickerPopupConfig.showButtonBar;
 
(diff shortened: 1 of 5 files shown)


src/datepicker/datepicker.js
```diff
@@ -811,6 +813,9 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
 
     if (angular.isString(viewValue)) {
       var date = dateParser.parse(viewValue, dateFormat, scope.date);
+      for(var i = 0; isNaN(date) && i < altInputFormats.length; i++) {
```
wesleycho: Would it make more sense to move the isNaN check to inside the loop after the dateParser is run? That way one can just `break` out of the loop if it is a valid date.

src/datepicker/datepicker.js
```diff
@@ -842,6 +847,9 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
 
     if (angular.isString(value)) {
       var date = dateParser.parse(value, dateFormat);
+      for(var i = 0; isNaN(date) && i < altInputFormats.length; i++) {
+        date = dateParser.parse(value, altInputFormats[i]);
```

src/datepicker/docs/demo.html
```diff
@@ -22,7 +22,7 @@
     <div class="row">
         <div class="col-md-6">
             <p class="input-group">
-              <input type="text" class="form-control" uib-datepicker-popup="{{format}}" ng-model="dt" is-open="popup1.opened" min-date="minDate" max-date="maxDate" datepicker-options="dateOptions" date-disabled="disabled(date, mode)" ng-required="true" close-text="Close" />
+              <input type="text" class="form-control" uib-datepicker-popup="{{format}}" ng-model="dt" is-open="popup1.opened" min-date="minDate" max-date="maxDate" datepicker-options="dateOptions" date-disabled="disabled(date, mode)" ng-required="true" close-text="Close" alt-input-formats="MMM d! yyyy" />
```
wesleycho: I thought `alt-input-formats` needs to be an array? It should probably be a variable that has a value in the controller...or maybe just take advantage of `datepicker-options` for this feature?

src/datepicker/datepicker.js
davious: OK. I plan to move the `isNaN(date)` to inside the loop.

src/datepicker/docs/demo.html
davious: For the attribute, it would be a comma-separated list. I don't consider it a datepicker option: it doesn't get passed to the datepicker.

davious: The attribute parsing should change to be evaled. I plan to change that. That will require it to be an array then.

src/datepicker/docs/readme.md
```diff
@@ -144,7 +144,11 @@ Options for the uib-datepicker must be passed as JSON using the `datepicker-opti
   _(Default: `yyyy-MM-dd`)_ -
   The format for displayed dates. This string can take string literals by surrounding the value with single quotes, i.e. `yyyy-MM-dd h 'o\'clock'`
 
-### Keyboard support ###
+ * `alt-input-formats`
+  _(Default: [])_:
+  A list of alternate formats acceptable for manual entry.
+
+### Keyboard Support ###s
```
Foxandxss: remove that extra `s` and lowercase that `support` (not sure why you even touched that anyway).

Also the `[]` on the default needs to be between backsticks.

PR #4961, opened 2015-11-26
feat(pager): change controllerAs to pager

wesleycho:
BREAKING CHANGE: As part of the split of the pager component from the
pagination component, this changes the controllerAs use to `pager` from
`pagination`


Foxandxss:
Sounds good to me.

PR #4967, opened 2015-11-27
(fix) Render and validate timePicker on min/max changes

fouzes:
Fixes wrong validation with dynamic min/max attributes.


wesleycho:
Please re-file against `master`.

Also this needs tests.

PR #4968, opened 2015-11-27
feat(paging): factor out common controller code

wesleycho:
- Create paging factory for creating common controller for the
  pagination and pager components

This finishes off #4803.


Foxandxss:
What I am not happy about is the tests. You refactorized it in different files but there is no `paging` tests.

wesleycho:
Do we require tests with this change? All this is doing is technically exposing the lack of tests on the old UibPaginationController, since the tests in this library is behavioral in general.

Foxandxss:
I hear you Wes. But I don't like to add new "components" without any kind of tests. I know that all we do is behavioral.

So here we are pretending that the `.create` method just do what you think it does.

wesleycho:
Added tests

src/paging/test/paging.spec.js
```diff
@@ -0,0 +1,257 @@
+describe('paging factory', function() {
+  var $rootScope, $scope, paging, ctrl, attrs;
+
+  beforeEach(module('ui.bootstrap.paging'));
+  beforeEach(inject(function(_$rootScope_, uibPaging) {
+    $rootScope = _$rootScope_;
+    paging = uibPaging;
```
Foxandxss: You're saving it for no reason.

Foxandxss:
It is good for me. I still need to sweep all of this, so I will do a second pass by then.

Issue #4764, opened 2015-10-28
Add semantic classes to components


wesleycho:
We should add semantic classes to other components as appropriate in the templates to assist in styling.

pkozlowski-opensource:
@wesleycho I wonder what is the motivation for this? Why people can't target custom elements we've got for our directives with their CSS selectors? 

Other thought: if people want to add a custom / semantic css class they could write they own directive to target our custom elements and add whatever CSS thing they might need.

or just override a template.

wesleycho:
For some of the components (the datepicker was the main one, although that has been addressed by 97c43334956c3876a57c4b6cde8b94080b80ab68), creating custom selectors is not ideal, and harder to read - this adds minimal overhead, and helps with CSS for users to override.

Overriding a template is a bit overkill when one just wants a particular class to hook into for styling, and same with a custom directive.

pkozlowski-opensource:
@wesleycho IMO this is all a question of balance. It is true that adding those classes might get things easier for _some_ people but it is not for free. It is increasing code size for _all_ users and add increases maintenance burden.

This is why I'm not sure we should be adding this.

wesleycho:
That is true, but in reality, I think we only need to worry about this for the datepicker, timepicker, typeahead, and rating components. I think for the datepicker and timepicker, this makes a lot of sense due to the relatively complex tables - for the typeahead and rating, maybe less so.

Thoughts?

PR #4971, opened 2015-11-27
feat(timepicker): add semantic classes

wesleycho:
- Adds semantic classes to assist with styling

This addresses #4764.


Foxandxss:
LGTM

PR #4976, opened 2015-11-29
chore(carousel): combine animation check

wesleycho:
- Combine animation check since $animate.enabled value is set in the
  $watcher


Foxandxss:
Indeed, LGTM.

Issue #3057, opened 2014-12-06
Modal result promise is resolved prior to animations completing when closing the modal


nauticalcoder:
Angular version: 1.2.16
Angular Bootstrap version: 0.12.0
Bootstrap version: 3.1.1
http://plnkr.co/edit/xNCeqlj9cXKjTtQOSoiN

I am trying to have a message box display after the modal is closed.  The code for opening the message box is in the result promise handler.  When close is called on the modal instance, the message box is displaying, then once the message box is closed, the modal is closing.  After investigating, the issue appears to be related to the close animation.  I have included a proposed fix for this issue that involves moving the promise resolution so it happens once the animation is complete.
Pull request a proposed solution:
https://github.com/angular-ui/bootstrap/pull/3056

chrisirhc:
Should investigate this only after #2724 is merged.

wesleycho:
I will take a look at implementing this sometime in the next couple of weeks - this requires branching out the logic depending on a configuration value.

wesleycho:
Closing as resolved via e9c4977dc4a326b19694ae4e12784d22a90f1f99

PR #4979, opened 2015-11-30
feat(modal): add closed promise

wesleycho:
- Add `closed` promise for when the modal animation completes on removal

This addresses #3057.


Foxandxss:
This gave me the idea that I would like to have promises in my toastr too. Perhaps for 2.0.0

Issue #4981, opened 2015-11-30
Support ng-model-options for typeahead


wesleycho:
Add support for all sensible ng-model-options configuration for the typeahead as noted [here](https://docs.angularjs.org/api/ng/directive/ngModelOptions).

- [ ] updateOn
- [ ] debounce
- [ ] allowInvalid
- [x] getterSetter

raghumadhavan:
@wesleycho Issue seems to be open from a long time, could you consider closing it.

noah2233:
is there any update on the subject?

PR #4982, opened 2015-11-30
feat(typeahead): add ng-model-options debounce support

wesleycho:
- Adds support for ng-model-options `debounce` option

Partly implements #4981


Foxandxss:
Docs pls and perhaps add it to a demo (not really needed)

Foxandxss:
LGTM

PR #4983, opened 2015-11-30
chore(carousel): remove dead code

wesleycho:
- Remove unnecessary $animate.enabled check

The carousel still works, as can be seen [here](http://plnkr.co/edit/LiQN270fX8F7zsLuiysP?p=preview). It turns out that the .animation code does not run if $animate.enabled is false.


Foxandxss:
LGTM

PR #4993, opened 2015-12-02
chore: update package.json

wesleycho:
- Update package.json with more information
- Improve .npmignore slightly


Foxandxss:
I have no idea what those package.json things does.

wesleycho:
https://docs.npmjs.com/files/package.json#files
https://docs.npmjs.com/files/package.json#directories-lib

Foxandxss:
I see, but I Got confusing with adding the files and then using npmignore too.

wesleycho:
I added those files to npmignore so they wouldn't get accidentally published when we do move to use the main repository for npm releases.

Foxandxss:
Yeah, well, in any case looks good, hope we don't break too much with this change.

PR #4994, opened 2015-12-02
Update tooltip.js

addctdto:
Tooltips/Popovers, when append-to-body is used: added hideOnUrlChange attribute in order to allow the tooltip/popover to be preserved on url location change. Old behaviour is preserved as to not cause any breakage.


wesleycho:
Rejecting this feature as something we do not want.

If one wants to implement it on the user side, you could add a custom event trigger, then on route change, fire that event.

PR #4995, opened 2015-12-02
Improve clarity of attribute explanation

mcbhenwood:
It wasn't immediately apparent that the uib-popover-template attribute
is an expression rather than a string attribute.


wesleycho:
The same modification should probably be done on the tooltip docs as well then - it should be noted that the docs are in the process of an overhaul to be more accurate & specific, so this may change by final release for 1.0, but hopefully we get more clarity that this is actually a watched expression from what I can tell.

icfantv:
technically, it's the same syntax as `ng-include`, no?

wesleycho:
This needs to be rebased on top of current `master`.

Foxandxss:
Fixed at 79fa363

cbdweb:
I don't know that it is fixed, at least in the current documentation that I can see. 

uib-popover-template $ - A URL representing the location of a template to use for the popover body. Note that the contents of this template need to be wrapped in a tag, e.g., `<div></div>`.

I have a `<script type="ng-template" id="some_name">` to define the HTML 
From the docs it looks like I could use `uib-popover-template="some_name"` but this doesn't work.

Instead I had to write a function:
`$scope.get_template = function(arg) {
     return arg;
 }`
and use 
`uib-popover-template='get_template("some_name")'`

Clumsy syntax.  Couldn't angular UI tell the difference between a function and a string?

Foxandxss:
That is not an issue of us, that is how angular works (which I agree with you nonetheless). Angular 2 makes it easier.

Problem is that it needs an expression (your function is an expression) and not a simple string. Try:

```
uib-popover-template="'some_name'"
```

Notice the double quotes.

cbdweb:
Thanks @Foxandxss, I should have realised it's an angular expression. I've implemented your solution.

pursual:
In bootstrap the template is the entire popover. In ang-bootstrap it is just the body? No way to add a custom class to the popover to remove padding, etc?

icfantv:
@pursual you can VERY easily do this via custom CSS classes.  This is what we do and I would argue is the proper way to do this.

pursual:
I'm not seeing any way to add a class to the .popover element.
In the end I had to target the popover container via the parent element it was attached inside of. If there is a method for adding a class please let me know. (not to the popover body...to the popover WRAPPER. The one with .popover class)

icfantv:
@pursual, since you're defining the template used for the popover, you know the HTML structure of the template.  You can add whatever classes you want to that structure to create CSS class selectors as you deem appropriate.  Additionally, you can use the appropriate `tooltip-class` or `popover-class` attribute to set a class on the template.

E.g., if I wanted to overrride the padding for the popover class (system wide):
```
.popover {
  max-width: 415px !important;
}
```

Or if I wanted to define a selector for when using, say, `popover-class="my-entity"`:
```
.popover.my-entity {
  padding: 0;
}
```

pursual:
That's my point. It seems that in the angular version, template references the BODY, not the TEMPLATE. Whatever you put in uib-popover-template gets injected into the default template, in the body section.

icfantv:
Ahhhh, I think I understand what you mean.  Yes.  The popover template is placed into another template for the tooltip/popover:  https://github.com/angular-ui/bootstrap/blob/master/src/tooltip/tooltip.js#L123-L136

pursual:
Yeah, I knew there was something like popover-class from past use, but couldnt find it documented in either project, so was kind of stuck.

icfantv:
You can safely make CSS class selectors (with hierarchy as you need) as we are generally VERY careful about changing DOM structure and if we do, we mark it with a BREAKING CHANGE so it's logged in our CHANGELOG.

PR #5003, opened 2015-12-03
refactor(modal): simplify backdrop creation

Foxandxss:
You know when you want to use `backdrop` but you can't use it because you already have it so you look for a synonym. No need to do that in this case, you can use it directly.


wesleycho:
LGTM

Issue #4990, opened 2015-12-01
uib-modal Trap focus doesn't work when keyboard: false option is provided


wawyed:
This is caused by this line https://github.com/angular-ui/bootstrap/blob/master/src/modal/modal.js#L357

Most likely that check (`modal.value.keyboard`) should be inside the `case 27:` (escape key) and not surrounding the whole switch that also handles trap focus.

wesleycho:
You mean the check should be inside case 27 - that is the escape key, 9 is the tab key.

Would you like to file a PR for this?

wawyed:
Derp, sorry I mean case 27 escape key yes. I might to submit a PR for it but I'm quite busy at the moment. I might get some time next week.

wesleycho:
Just let me know - this should be easy to fix and I can fix it, but I think it'd be a nice contribution if you'd like.

PR #5005, opened 2015-12-03
fix(modal): keyboard=false should not stop trapping tab focus.

wawyed:
Fix for https://github.com/angular-ui/bootstrap/issues/4990


wesleycho:
Can you add a test for this? Should be simple, just test when emitting the proper keypress event with `event.which` being 9, the esc key value, and testing whether it has been defaultPrevented.

wawyed:
What would you like me to test? keyboard = false is already been tested by https://github.com/angular-ui/bootstrap/blob/master/src/modal/test/modal.spec.js#L747

wesleycho:
Testing when hitting esc and making sure it isn't getting defaultPrevented normally is what I mean

wawyed:
I'm a bit confused with what you mean. What does preventDefault have anything to do with my PR? The only test that I can think of adding would be that the trap focus on tab works correctly even when keyboard=false is used.

wesleycho:
Ensuring that it isn't prevented is what I meant - the current behavior in `master` is that it will be prevented when tabbing (I got confused with the keycode heh).

wawyed:
Made another PR with unit tests, https://github.com/angular-ui/bootstrap/pull/5010

PR #5008, opened 2015-12-03
refactor(modal): migrate more animations to animateCss

Foxandxss:
Now that we are breaking stuff, lets clean the backdrop animation a bit.

We should create a milestone too for doing better animations tho.


src/modal/modal.js
```diff
@@ -76,24 +70,16 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
     function linkFn(scope, element, attrs) {
       if (attrs.modalInClass) {
-        if ($animateCss) {
-          $animateCss(element, {
-            addClass: attrs.modalInClass
-          }).start();
-        } else {
-          $animate.addClass(element, attrs.modalInClass);
-        }
+        $animateCss(element, {
```
wesleycho: Wrap this in a function to remove duplication

Foxandxss: where?

wesleycho: function animateIn(cssKey) {
  var animateObj = {};
  animateObj[cssKey] = attes.modalInClass;
  return $animateCss(element, animateObj).start();
}

Or something like that

Foxandxss: Yeah, the problem is that several directives need it and I am not sure I want another service just for this.

wesleycho: I just meant a helper function just for this directive.

Foxandxss: But the next one has the same need.

wesleycho: I guess I'm fine with that, but it looks a little ugly to have that duplication. This can just be merged in then.

PR #5010, opened 2015-12-04
fix(modal): keyboard=false should not stop trapping tab focus.

wawyed:
fix(modal): keyboard=false should not stop trapping tab focus.


wesleycho:
Thanks for the efforts - this looks good to me!

PR #5011, opened 2015-12-04
refactor(modal): mix both modal spec files

Foxandxss:
I can discuss having multiple spec files for different stuff, but in the meantime we don't and a separate file for 4 tests doesn't make sense.


wesleycho:
A few components have multiple test files - it isn't consistent though.

I'm fine with this change.

Foxandxss:
I'll reach those eventually.

PR #5014, opened 2015-12-04
fix(datepicker): alternate input formats: update datepicker on change

davious:
When and after typing an alternate format, the datepicker doesn't update, leading to confusion.

http://plnkr.co/edit/Qbx9hft21S9PAS0pqJ5i?p=preview

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/bootstrap/5014)


wesleycho:
This is the wrong approach - `parseDate` is a $parser for ngModel, one should never be calling $setViewValue there.

wesleycho:
This looks a lot better - good work!

PR #5016, opened 2015-12-05
chore(typeahead): change to on/off

wesleycho:


Foxandxss:
That bored that you're doing sweeping before me? :P

LGTM

PR #5020, opened 2015-12-06
fix(tabs): added controllerAs for tab controller

### LINKED ISSUE #5019: Unable to extend tab's template
I want to redirect user to particular page when he clicks on tab. As I have similar behavior in several places I don't want to write boilerplate code using `select + active` attributes. So that, I'd like to pass something like `href` `tab-href` attribute to the tab directive which is attached to link then:
```html
<tabset>
  <tab heading="Google" href="http://google.com"></tab>
  <tab heading="Gmail" href="http://gmail.com"></tab>
</tabset>
```
Currently it's not possible as I need to extend template and controller but I can't get access to controller inside `tab.html` as there is no `controllerAs`.

So, the simplest fix is to add `controllerAs: 'tab'` to tab directive so then it can be accessed inside view. So then I do this:
```js
angular.module('ui.bootstrap.overrides').directive('tab', function($parse) {
  return {
    restrict: 'EA',

    require: ['tab',
    
    compile: function(element, attrs) {
       var href = attrs.href || attrs.tabHref;

       if (href) { 
         element.find('a').attr('ng-href', '{{ tab.href() }}');
       }

       return this.link;
    },

    link: function(scope, element, attrs, tabCtrl) {
       var href = attrs.href || attrs.tabHref;

       tabCtrl.href = $parse(href).bind(scope, scope);
   }
  };
});
```

stalniy:
it fixes #5019


wesleycho:
This needs tests.

stalniy:
@wesleycho  Done

src/tabs/test/tabs.spec.js
```diff
@@ -98,6 +98,10 @@ describe('tabs', function() {
       titles().eq(1).find('> div').click();
       expect(scope.deselectFirst).toHaveBeenCalled();
     });
+
+    it('should expose controller as "tab" scope variable', function() {
+      expect(typeof elm.find('.uib-tab').first().isolateScope().tab).toEqual('object');
```
wesleycho: Using a typeof check is not great, and this test is a little too vague. It would be better to test against what it equals exactly.

stalniy: I believe the only way to do this is:
```js
it(..., inject(function(uibtabDirective) {
  expect(elm.find('.uib-tab').first().isolateScope().tab).toBeInstanceOf(uibtabDirective[0].controller);
}));
```
Let me know if you are ok with such solution

wesleycho: Check how some of the other tests for the other components is testing for controllerAs working - those are more useful guides as far as a readable test for this feature are written.

stalniy: @wesleycho Ok, thanks for pointing me to the right direction. Done

PR #5024, opened 2015-12-07
chore(grunt): set grunt-cli as a dependency

mastilver:


pkozlowski-opensource:
Well, it would work for TravisCI but what about regular developers? Now you would find yourself installing grunt-cli locally even if you have to have it installed globally for regular development.

What is the reasoning here? What would be the advantage of the proposed change?

mastilver:
running `grunt` will still work if you've got grunt-cli installed

As a potential contributors of a library, I follow those steps:
- clone repo
- `npm install`
- `npm test`

So in this case, it's giving me an error.
Because I'm not familiar with `grunt`, I install globally `grunt` instead of `grunt-cli`.
I keep getting an error and start feeling frustrated. ;)

I just think It will make it more friendly to new comers

pkozlowski-opensource:
> running `grunt` will still work if you've got grunt-cli installed

This is my point, I will have it installed _twice_ now :-)

> I just think It will make it more friendly to new comers

I'm not sure. Here you've made just one command to work. But to be able to contribute effectively you will need more commands (linting, running tests in tdd mode etc.). So it helps with the case of someone who jumps in and want to test a small fix. Which is an important use case, I agree. But at the same time we would see a pb on TravisCI.

I'm on the fence with this one, will let others decide

mastilver:
> But to be able to contribute effectively you will need more commands

We can still add more scripts (should I do it on this PR?)

> But at the same time we would see a pb on TravisCI.

Which problem?

> I'm on the fence with this one, will let others decide

:+1:  cool

pkozlowski-opensource:
> We can still add more scripts (should I do it on this PR?)

Personally I'm not big fan of supporting several ways of building a project.

wesleycho:
I'm ambivalent as well - it looks like grunt supports this situation, but it generally has been recommended to install it globally since if any other library uses grunt, it then supports a common CLI via global executable whereas this movement would add duplication of the CLI.

I think I'm leaning a little towards no on this PR as a result.

mastilver:
fair enough ;)
I kind of like to have a one command setup: `npm install`

Feel free to close it then!

wesleycho:
Closing as per discussion

Issue #5027, opened 2015-12-07
Datepicker closeOnDateSelection with ngFocus to open popup in IE


bencebalogh:
If you create an input with ngFocus which opens the datepicker popup then on date selection the popup is closed first then immediately reopened.

Reproduced here: http://plnkr.co/edit/NuKrsZG3t8TwbDvotCHp?p=preview

wesleycho:
This is a usage issue - [here](http://plnkr.co/edit/Dyh6knZ88XiXtVrRNfQS?p=preview) is a demonstration how to work around it for your use case.

bencebalogh:
I checked your solution and I can reproduce the issue there as well.
If I open the Plunker in Internet Explorer, focus on the input then the datepicker is shown. Then I click on a date field and the popup blinks one, but stays open.

wesleycho:
Sorry, I screwed up - my code is no different effectively.

I believe this was reported prior, I need to do some research.

bencebalogh:
Would it be possible to emit an event from the datepicker popup's dateSelection if closeOnDateSelection is true?

    if ( closeOnDateSelection ) {
      scope.$emit('closed-on-date-selection');
      scope.isOpen = false;
      element[0].focus();
    }

Then I could use this event in the controller to modify onFocusOpen (from your previous link) to:

    var closedOnDateSelection = false;
    $scope.onFocusOpen = function() {
      if (!closedOnDateSelection) {
         $scope.opened = true;
      }
    };

    scope.$on('closed-on-date-selection', function () {
      closedOnDateSelection = true;
      $timeout(function () {
        closedOnDateSelection = false;
      });
    });

This works.

KennedyIhe:
This issue is still hapenning in IE 11.
reproduce - https://plnkr.co/edit/YRjj91bmpEY5hBebyzlv?p=preview

christrude:
+1

luizbento:
@wesleycho The problem occurs because IE set focus to div containing the directive instead of input. To work on IE the line 220 in popup.js ```$element[0].focus();``` needs to be removed.

PS. Also the problem occurs on close button and esc key down event when popup is open. To fix this, the lines 228 and 252 needs to be removed as well, but with this, the behavior changes and focus doesn't returns to the input anymore. See plunker: [https://plnkr.co/edit/YRjj91bmpEY5hBebyzlv?p=preview](https://plnkr.co/edit/YRjj91bmpEY5hBebyzlv?p=preview)

PR #5028, opened 2015-12-07
Datepicker closeOnDateSelection with ngFocus

bencebalogh:
This solves #5027 
If you create an input with ngFocus which opens the datepicker popup then on date selection the popup is closed first then immediately reopened.

Reproduced here: http://plnkr.co/edit/NuKrsZG3t8TwbDvotCHp?p=preview


wesleycho:
Closing as invalid - we do not want to add an unnecessary timeout as that forces async behavior unnecessarily & could cascade into other problems.

Also see https://github.com/angular-ui/bootstrap/issues/5027#issuecomment-162546292 for how to solve the particular use case this PR was meant to address.

PR #5040, opened 2015-12-09
feat(datepicker): multiple months display support

ivi-hamiti:
added attribute month-count with default 1 for displaying multiple months on datepicker

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -191,8 +191,12 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   // Key event mapper
   $scope.keys = { 13: 'enter', 32: 'space', 33: 'pageup', 34: 'pagedown', 35: 'end', 36: 'home', 37: 'left', 38: 'up', 39: 'right', 40: 'down' };
 
-  var focusElement = function() {
-    self.element[0].focus();
+  
+  var focusElement = function () {
+    var element = self.element.children(0);
+    if (element && element.focus) {
+      element.focus();
+    }
   };
 
   // Listen for focus requests from popup directive
@@ -228,7 +232,7 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   var DAYS_IN_MONTH = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
 
   this.step = { months: 1 };
-  this.element = $element;
+  this.element = $element.parent();
   function getDaysInMonth(year, month) {
     return month === 1 && year % 4 === 0 &&
       (year % 100 !== 0 || year % 400 === 0) ? 29 : DAYS_IN_MONTH[month];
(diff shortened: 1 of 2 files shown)


icfantv:
@ivi-hamiti, when adding new functionality to an OSS library, it's generally a good idea to propose it first via an issue so that you receive the blessing of the team.  We don't want devs to waste their time on a feature that won't be incorporated for whatever reason.

That said, there are a few issues with your PR:

1.  The datepicker API is already bloated as it is and for that reason alone, we're VERY discerning about adding new features to it - especially for features that are not in high demand.
2.  Your code has no tests.
3.  Your code breaks the CI build because you've changed existing functionality without touching the existing tests.
4.  Your commit changes the format of code not affected by the functionality you added.  Code changes should be as minimal as possible.

I'm inclined to reject this PR, but please feel free to make your case.

ivi-hamiti:
Hi @icfantv, i didn't do the changes for pleasure, but it was a need for my projects. Basically the angular-ui bootstrap is a great library, and I as a developer see a loot of core functionalities missing or not handled. Starting from the popup which should have a certain behavior on backspace press which on a spa app ends with very strange behavior, continuing with the fact that you have the popup manager (the popup stack) already in your framework but do not make use of that and basically you overlay the multiple popups, and continuing to this tiny feature that for me is a must have on a datepicker directive or plugin or library (name it what you want). Have you ever tried to search on google about this feature.

Google Search

Following just see the images of what range of companies do need this feature, and if needed i can give you 1000+ cases when this is needed. I basically found one other directive but it had too many dependencies. Yes i dug around the code and saw that basically it's bloated with not the best solutions and techniques, but you should be open minded not overestimate what you have till now.

Lufthansa

Virgin

MakeMyTrip

Yes you have right the tests do not pass, basically they need to be changed while this will change a lot the directive logic and basically i changed only two files i don't get your point. i was expecting more collaboration from your team but simply rejecting as i see for you is the simplest way.

icfantv:
@ivi-hamiti, I never suggested it was a bad feature, only that you should run the feature requests through the proper channels first - I do think it could be a useful feature.  I did note that the datepicker API was fairly bloated, but didn't use that as the reason for rejecting your PR.

I would venture to say that most well-run OSS projects have a [CONTRIBUTING.md](https://github.com/angular-ui/bootstrap/blob/master/CONTRIBUTING.md) guide for users and ours is no exception.  It is clear from your PR that you did not read it.  I will say that criticizing a library and then saying you expect collaboration and help from its maintainers to is not the right way to go about things.

As I originally pointed out, you didn't discuss this feature with us ahead of time so the collaboration you expected, never took place.  This is akin to "it's better to ask for forgiveness than permission."  We absolutely don't want people to waste their time working on stuff we can't or won't use.  The onus has to be on contributors to be diligent in this regard.

I am happy to discuss adding this feature with the team if you are ok with addressing all the issues I raised and following the instructions set forth in our contribution guide.  Unfortunately, we all have very limited bandwidth through the end of the year between trying to get 1.0 out the door and our day jobs so our support may be limited to code reviews.

ivi-hamiti:
Well this is what i was thinking about collaboration, don't take the wrong meaning about what ment. So basically i didn't lose time cause for me these are features i needed and having no good posibility i did the changes to implement this. By the other way i do not have a time limit if there are high priority things it's not a problem. 

By the other side, i will like to discuss some ideas about some improvement and features which i think as a must have.

Foxandxss:
Leaving the PR thing aside. Couldn't you just put two inline datepicker next to each other? That is what I personally would do in a project if I get a case like that.

ivi-hamiti:
That was the first approach that i tried. For this there are several issues.
1. The next and previous buttons should not be always visible, this can be fixed with a line of css
2. Each calendar ends up having its own month and year picker which is a bit strange for a single control
3. The managing of the next and previous button clicks to rerender each datepicker as it should be is a nightmare
4. Each calendar has its own active date and it is totally strange and cunfusing to the user

Foxandxss:
Yeah, I guess that the default template is not handy for that use case. You could always override too.

ivi-hamiti:
@Foxandxss thanks for the response, from what @icfantv describes basically every idea or features should be preferably discussed with the team first. I went through the repository the CONTRIBUTING.md  file, etc, but there is no clear guideline how to discuss things, so please can you describe me which should be the best way. The second thing is that i was checking the milestones, and there is Angular 2 support which is the next step as i see from the repository wiki, are there any particular guidelines for the moment?

cfletcher:
@icfantv, if the issues with this PR could be solved (fixed build, new tests etc) would the team accept this feature?
My project is also in need of this feature, were otherwise using angular-ui bootstrap and are very happy with it. Obviously we'd like to stick with one library if possible.

Foxandxss:
It is a no for me. Datepicker is really unbearable already to add more features to it.

icfantv:
@cfletcher, as @Foxandxss has indicated and we've addressed numerous times over various issues in our forum, the datepicker API is already bloated and we are not willing to add more functionality to it at this point.

We have to strike a balance between features users want (or think they want) and being able to have a maintainable API.  We would literally have a hundred different configuration options if we supported everything people wanted and at that point, it would become as much a pain in the butt for the API developers as it would be for us as maintainers.

One could very easily write a custom directive to display two date pickers binding the models to two separate text fields in one's underlying form.

cfletcher:
Ok thanks @icfantv, I'll look into that.
Thanks for your time.

icfantv:
Cool.  Again, sorry we can't help you out here.  if you do come up with a slick directive and have it working in a plunker - you might consider sharing the love and paste the link here.  And/or create a github project for it?

PR #5048, opened 2015-12-10
fix(modal): fix copying from $scope

### LINKED ISSUE #5039: use objects as bindToController option for modals
We should be able to set `bindToController` as an object, which will be bound to the controller instead of the scope variables.

wesleycho:
- Fixes issue where `$scope` provided does not have properties present
  on controller instance due to $new resulting in the property on the
prototype of the $scope copied from, which causes it to not be
enumerable

This fixes #5039


Foxandxss:
LGTM but I would change the titlte to "fix bindToController"

PR #5052, opened 2015-12-11
fix(dropdown): do not close on right click

### LINKED ISSUE #5051: Dropdown menu closes on right click in Firefox
In Firefox, the dropdown menu closes immediately after opening if it is triggered by a right click (contextmenu event) because Firefox sends a click event to the document object despite no other browser doing this.

It would be preferable to check if the click event has evnt.which === 3 and if so not close the menu. That way it behaves the same on all browsers.

wesleycho:
- Do not close the dropdown on right click when open due to Firefox's
  behavior

Plunker in action [here](http://plnkr.co/edit/0gt6lQG9l4hGcKB2x62k?p=preview).

This fixes #5051


icfantv:
@wesleycho, is this _technically_ a breaking change?  Do you want to mark it as such?

wesleycho:
It should be noted that it currently behaves this way in all browsers excepting Firefox. This is a Firefox specific change.

icfantv:
yea, i got that.  i just wasn't sure if you felt it didn't explicitly need one.  i guess not?  :-)

Foxandxss:
Can we test it?

wesleycho:
Probably - I mainly wanted to see whether it was worth testing, but seems like it is, so I will

wesleycho:
Added test

Foxandxss:
LGTM. I don't follow you @thomassuckow

PR #5056, opened 2015-12-11
test(modal): test dismissAll

Foxandxss:
That is 1% more of covering :)

Didn't create a helper method to dismissAll because it is just one test.


wesleycho:
Strange we never had a test for this - LGTM

PR #5064, opened 2015-12-13
refactor(carousel): remove unneeded describe

Foxandxss:
All controller tests were inside the `addSlide` describe and that was incorrect.


wesleycho:
Works for me

PR #5065, opened 2015-12-14
refactor(datepicker): optimize datepicker

### LINKED ISSUE #3451: DatePicker - Performance issues with popup + other stuff

wesleycho:
- Add new class manipulation helper directive that uses 1 watcher as opposed to watchers for each element with ng-class
- Optimize disabled state
- Optimize class binding

This supercedes #3770, and resolves #3451 & #2613


Foxandxss:
A couple of times checking this PR and I have no idea what you're trying to accomplish in here.

If this are 3 changes, please, create 3 PRs because I am getting confused.

The new `uibIsClass`, I don't understand what you want to do with it or how it relates to the datepicker at all.

If possible, close this, create as many PRs as you need and for the one for the uibIsClass, give a good explanation of why we need it (apart from ng-class creating too many watches) and some real world examples in how to use it (because the PR is not touching the datepicker for that).

Cheers :)

wesleycho:
The original changes were in 3 commits - it was actually a bit tricky to parse it out, although they were in separate commits. `master` has changed a lot since then and conflicts need to be resolved.

The `isClass` change allows you to add only 1 $watch for ng-class behavior across an ng-repeat, as opposed to n $watches when using ng-repeat + ng-class.

The change to use Object.defineProperty removes updates of properties on the `$scope.watchData` object in the popup for use in the datepicker to avoid adding additional $watchers.

And lastly, a one-way binding on the datepicker disabled attribute since the isClass change requires elimination of use of track by.

Foxandxss:
But how is the isClass used in the datepicker?

wesleycho:
Something weird happened - let me push up the other changes necessary to the template.

Foxandxss:
I desist. This PR gives me headache. That isClass has an horrible syntax and combined with a ng-repeat, makes the code tests to understand.

I hope this cuts enough watchers to make it worthwile.

Issue #5066, opened 2015-12-14
Switch from node-markdown to marked


wesleycho:
[node-markdown has been deprecated for a while](https://github.com/andris9/node-markdown), so we should switch to [marked](https://github.com/chjj/marked) for our docs.

PR #5067, opened 2015-12-14
chore: change node-markdown for marked

wesleycho:
This addresses #5066 - verify this works locally first in generating proper docs.


Foxandxss:
If the result doesn't change, LGTM.

wesleycho:
Pretty much identical as far as I can tell.

Issue #4945, opened 2015-11-24
popover-append-to-body does not work


cdauth:
According to the documentation, the attribute `popover-append-to-body` can be added to elements using the `uib-popover` directive. I has no effect though, as can be seen in [this plunker](http://plnkr.co/edit/eV92XT5Rg1k9ZuEn1RtH?p=preview).

pkozlowski-opensource:
You need to use ` popover-append-to-body="true"`. I think that we've got an issue somewhere to trugger this functionality using a boolean attribute only as you've tried.

@Foxandxss as you can see this is not only ng2 thing :-)

cdauth:
Thanks for the tip, in fact it seems to work.

This seems to be an inconsistency, as `dropdown-append-to-body` works without `="true"`, for example. 

Also, the documentation should mention this.

cdauth:
@icfantv, thanks for taking care of the issue.

Either I am missing something, or I think there is a misunderstanding. You added to the docs that “the presence of the attribute implies true”. It is the issue of my report that in fact it does _not_, as opposed to the `dropdown-append-to-body` attribute.

RobJacobs:
There was a previous discussion on this where someone chimed in and said they use *-append-to-body="true" in a global config and turned it off on a case by case basis using *-append-to-body="false".  If we are going to change this, I think it needs to still evaluate the attribute value but default to true and be consistent across all directives.

PR #5071, opened 2015-12-14
feat(tooltip): add appendToBody only attribute support

wesleycho:
- Add support for using only `tooltip-append-to-body` without any attribute value to enable appending tooltip to body

This implements #4945


Foxandxss:
Sounds good to me.

PR #5074, opened 2015-12-14
fix(datepicker): stop propagation of esc in popup

### LINKED ISSUE #5013: Pressing ESC in ui.bootstrap.datepicker closes ui.bootstrap.modal Dialog
I am using both your ui.bootstrap.datepicker and ui.bootstrap.modal in a simple Angular app.  I can open the modal and display the datepicker when clicking into an input field.  (So far so good)

I can select a date or click outside the datepicker to close it.  Unfortunately when I use the ESC key to close the datepicker it also closes the modal dialog.  

I setup the input field like this:

```
<input type="text"
               class="form-control"
               name="startDate"
               ng-model="prospect.evalStartDate"
               uib-datepicker-popup="shortDate"
               is-open="datepickers.startDate"
               datepicker-options="dateOptions"
               show-button-bar="false"
               ng-click="open($event, 'startDate')"
               required />
```

and then in the controller I do this: 

```
        $scope.prospect = {};
        $scope.datepickers = {
            startDate: false,
            endDate: false
        }
        $scope.status = {
            opened: false
        };
        $scope.dateOptions = {
            startingDay: 1,
            showButtonBar: false,
            showWeeks: false
        };

        $scope.open = function($event, which) {
            $event.preventDefault();
            $event.stopPropagation();

            $scope.datepickers[which]= true;
        };

        $scope.ok = function () {
            console.log('OK was clicked: ' + $scope.prospect.name);
            $uibModalInstance.close($scope.prospect);
        };

        $scope.cancel = function () {
            console.log('Cancel was clicked');
            $uibModalInstance.dismiss('cancel');
        };
    };
```

I had hoped that the `$scope.open` function would have solved this by not propagating things but that appears to not be happening.  

Any thoughts would be appreciated...

wesleycho:
- Stop propagation of esc key in datepicker popup, which can propagate to undesirable elements/handlers such as the modal

This fixes #5013


Foxandxss:
LGTM.

icfantv:
Do we want a test for this?

Issue #4885, opened 2015-11-12
Strange behaviour of uib-collapse inside ng-repeat


BParnikel:
Faced with strange behaviour of uib-collapse inside ng-repeat with changing collection: animations calls with each collection update, despite the fact that uib-collapse is set to false;

See plunkr: http://plnkr.co/edit/rP1AlbLsi1qXuJOJL0D6?p=preview

Fidel90:
Maybe related? https://github.com/angular-ui/bootstrap/issues/4848

RobJacobs:
Closing as support question answered on SO, [link](http://stackoverflow.com/questions/33666761/strange-behaviour-of-angular-bootstrap-collapse)

BParnikel:
I got only a partial answer on SO about transition between primitives.
But what about first opening of uib-collapse="false" with animation? I suppose it shouldn't be happened, but it does...

RobJacobs:
This is a result of the expand animation being run immediately in the link function.  Adding the following code:

    if (!scope.$eval(attrs.uibCollapse)) {
      element.addClass('in')
        .addClass('collapse')
        .css({height: 'auto'});
    } 

at the top of the linking function seems to solve the problem.  Updated [plunk](http://plnkr.co/edit/qzHb2sNHBhb0vXKndNpK?p=preview) with code changes.

BParnikel:
Thank you! Is there a chance that this fix will be included in next releases? I think changing of vendor sources is a bad practice :)

wesleycho:
I wonder if this is related to all of our ngAnimate modal + collapse issues - can you check on whether the fix fixes those issues to @RobJacobs?

RobJacobs:
@wesleycho Will do, wanted to get the positioning PR in first.

wesleycho:
This is not related to the modal issues - that one turns out to be a gotcha with ng-app needing to be present on the `body` element or higher.

wesleycho:
Actually...nevermind, I could be mistaken.

wesleycho:
So to clarify why the collapse class is removed - Bootstrap uses the `collapsing` class to set the transition, and the `in` class to animate.

BParnikel:
Great job, @wesleycho , thank you!

PR #5075, opened 2015-12-15
fix(collapse): set initial state to avoid animation

### LINKED ISSUE #4848: Accordion shows expanding animation when navigating to page
Hello,

when my accordion group is initially opened with is-open attribute, navigating to the page shows the group expanding. I know this occurs since 0.13.4 or even earlier (I'm on 0.12.0 at the moment). I think this is is bad behaviour as something that is already opened, should not animate an opening animation when navigating to it.

wesleycho:
- Ensure animation does not occur when first loaded

Fix in action [here](http://plnkr.co/edit/HtK9eCHwezTmffzdf0ct?p=preview) and [here](http://plnkr.co/edit/xmNFfvYp31PIe5x2eCEq?p=preview)

This fixes #4848 and #4885


Foxandxss:
LGTM. To be honest, BS animations are not my thing.

PR #5079, opened 2015-12-15
docs(modal): update to latest version

Foxandxss:
I want to do a final sweep on this one when I finish all my tasks. Just for styling.


wesleycho:
LGTM

wesleycho:
Leaving as needs: work, but feel free to flip it and let us re-review when ready.

Foxandxss:
nah nah, it is mergeable. Work would be just add a header here or there. Content is good.

That work could not happen.

PR #5082, opened 2015-12-16
fix(datepicker): active date should be model if present

### LINKED ISSUE #5081: Datepicker popup's initial format is broken in angular 1.5.0
Please refer to https://github.com/angular-ui/bootstrap/issues/3025
Commented by rustygreen: "The datepicker popup's initial display value does not use the passed in format. After you select a value for the first time, it will display it property (using the passed in format)."

Angular 1.4.8 works fine but 1.5.0:
http://plnkr.co/edit/cg1Uj5kF8fNAaNuTvxVY?p=preview

wesleycho:
- When ng-model value is initially present, it should set the active date to be the model date as per documentation

Fix in action [here](http://plnkr.co/edit/HzJzpXJv1QsvjAhOby8G?p=preview)

This fixes #5081


Foxandxss:
works for me.

PR #5093, opened 2015-12-18
fix(datepicker): correctly set minMode/maxMode

### LINKED ISSUE #5090: min-mode / max-mode causing issues
Take a look at this plunker

http://plnkr.co/edit/cqs3EhAjo4oPGBKMgF1g?p=preview

The popup on the left doesn't show a calendar when setting min-mode or max-mode.  Am I just using this wrong?

wesleycho:
- Correctly set `minMode` and `maxMode` for the datepicker popup
- Fix documentation of `datepickerMode`, as it is bound to a property on $scope and not an attribute string value

Fix in action [here](http://embed.plnkr.co/IiWa5g95OpiCwp8tgx39/).

This fixes #5090.


Foxandxss:
The fix is nice, but I don't understand the doc change. I was thinking if I wanted to put quotes around values that are string. Perhaps I do.

For the time being, remove the doc change and I will see what I want to do with that in all the other docs.

wesleycho:
Alright, will merge with the removal now.

PR #5094, opened 2015-12-18
test(datepicker): optimize performance

wesleycho:
- Optimize performance of tests by compiling datepicker once max per test

Part of this purpose is to give us better insight as to what may be causing performance issues with the datepicker.


Foxandxss:
Changes LGTM, but I don't think that is gonna help too much with noticing performance issues. Still, faster tests are faster tests, datepicker ones are a pain.

PR #5097, opened 2015-12-18
fix(modal): stop IE throwing exception on focus

### LINKED ISSUE #5096: IE throws exceptions on removeModalWindow on close from javascript
When you open and close a modal from javascript, on a page that does not have any inputs or buttons.  body.focus() throws a 'focus not supported' error. 

The bug only occurs in IE, when body.activeElement is undefined.

Browser: IE11
Angular: 1.4.7
Angular-bootstrap: 0.14.3

Hadev-JHH:
check if the element the model is appended to supports focus before actually calling focus to stop IE from causing exceptions and breaking the angular app

fixes #5096


src/modal/modal.js
```diff
@@ -309,7 +309,9 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         if (elementToReceiveFocus && elementToReceiveFocus.focus) {
           elementToReceiveFocus.focus();
         } else {
-          appendToElement.focus();
+          if(appendToElement.focus) {
```
wesleycho: This is probably better put on the else, i.e.
```js
else if (appendToElement.focus) {
  appendToElement.focus();
}
```

Hadev-JHH: ah yes, good point indeed. completely overlooked that after console.log debugging i guess

PR #5099, opened 2015-12-18
docs(accordion): add extra badges

Foxandxss:
One of the hardest thing on angular is knowing when an attribute expects a string or something on the scope.

First I thought about adding a `{{ }}` badge for stuff that are expected to be string, but I wasn't able to escape those chars for the final result without doing ugly stuff.

So I decided to do it the other way around, add a `$` every time that you expect something from the scope (or booleans).

If you agree, I will add proper doc on the main page in how to interpret the badges.


wesleycho:
If we do this, we need an explanation somewhere that the $ is for $scope.

I'm fine with this.

wesleycho:
Although...they're technically Angular expressions.

Foxandxss:
Yes yes, I want to add that right now.

icfantv:
This is a really cool/good idea.  Nice work there @Foxandxss.

wesleycho:
Alright, this LGTM assuming #5100 is also merged.

PR #5104, opened 2015-12-21
fix(datepicker): min-date: parse using displayFormat

### LINKED ISSUE #3570: Datepicker min-date Ignores Format
**Angular**: 1.3.2
**Angular UI**: 0.12.1

I'm formatting my dates to the format ```dd/MM/yyyy``` (UK-style) and trying to use these for the ```min-date``` attribute on the DatePicker directive.

This directive is reading the date in the format of ```MM/dd/yyyy```, which causes (for example) April 5th to be read as May 4th.

I believe the directive should use the format provided in the ```datepicker-popup``` attribute for both ```min-date``` and ```max-date``` and default to the US style only when a custom format isn't provided.

davious:
Closes #3570

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/bootstrap/5104)

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -192,7 +192,7 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
             if (angular.isDate(value)) {
               self[key] = dateParser.fromTimezone(new Date(value), ngModelOptions.timezone);
             } else {
-              self[key] = new Date(dateFilter(value, 'medium'));
+              self[key] = dateParser.parseGuessFormat(value);
             }
           } else {
             self[key] = null;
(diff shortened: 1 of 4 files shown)


wesleycho:
This is in need of rebasing off of current `master`

src/datepicker/datepicker.js
```diff
@@ -861,15 +859,26 @@ function(scope, element, attrs, $compile, $parse, $document, $rootScope, $positi
...
     }
-    return date;
+  }
+
+  function parseDateStringWithCatchAll(viewValue) {
+    var date = parseDateString(viewValue);
+    if (!isNaN(date)) {
+      return date;
+    }
+    date = new Date(dateFilter(viewValue, 'MMM d, y h:mm:ss.sss a'));
```
wesleycho: Can you explain what format the viewValue could it be in this stage?

davious: this is a carryover of @07e0970. The format would be `yyyy-MM-dd`, for example.

davious:
This is rebased. I welcome a review.

wesleycho:
As per #5658, we intend on removing all literal support in 1.4.0 & intend to deprecate in 1.3.0 - on some datepicker work, we are finding quite a bit of bloat of having to duplicate unnecessarily complex conditionals/fallbacks for the fragile dual model of literals, and it is slowing down work significantly.

Going to close this in light of this decision & highly recommend transitioning to JS dates.

PR #5105, opened 2015-12-21
docs(carousel): tune docs

Foxandxss:
I am not entirely sure if we are putting eyes on the ones that have an internal $watch (AKA `=`).


wesleycho:
Works for me

PR #5106, opened 2015-12-21
chore(carousel): change images on demo

Foxandxss:
Don't tell me that I am the only one who is tired of those cats. Lorempixel has lots and lots of different images, so every refresh brings new images and they don't seem to be NSFW.

I simplified the text mechanism (it was overcomplicated for no reason). I accept suggestion on the texts I picked (yay imagination).


wesleycho:
Don't tell me you hate kittens :( .

Foxandxss:
I like them, but I grow tired of those 4 pics :P

PR #5119, opened 2015-12-23
docs(datepicker): docs tune up

Foxandxss:
More ideas in here.

Instead of having to write ALL options that are available for overriding in .config, I add a C badge to indicate that this option can be overrode globally. Also if the option is not named in the same way, I add a `Config` section between the parens.

If you OK, I add this badge on the top of the doc to explain how it works and also add it to previous docs.


wesleycho:
Sure, but before merging this, an update needs to be made explaining the C

PR #5126, opened 2015-12-26
chore(karma): install fixed version

Foxandxss:
Let's leave a working version for the time being, so we can process new pull request.


wesleycho:
LGTM

Issue #4739, opened 2015-10-26
Add npm support for main repository


wesleycho:
We should publish the main repository to npm so that users can conditionally include scripts as necessary to allow inclusion by feature instead of having to include the whole library in one bulk load & potentially loading excess scripts.

We need to figure out how to do this while supporting template distribution & inclusion.

AshCoolman:
@wesleycho could you expand on the problem of "template distribution & inclusion"?

Aside: I presume the default state should be to include all modules. This would allow the unwinding of modules to happen with time and experimentation.

Also I feel I should jump in early, and encourage any future readers to check the older issue focused on bower https://github.com/angular-ui/bootstrap/issues/2196 for motivations.

chad-autry:
If building with Browserify, then  browserify-shim + browserify-ng-html2js let you include the chosen modules given the raw source. Fairly minimal config, main thing is in figuring out the templates and javascript elements to pull in.

The developer is then responsible for minifying the javascript, but if going the Browserify route they're probablly already doing so anyways (and probably already using browserify-ng-html2js for their own html templates)

I could probably do a more in depth write up of how it works with an example that could be linked to. On the angular-ui/bootstrap side all that is needed is publishing into npm the /src and /template directories as is.

wesleycho:
@chad-autry thanks for that report - how about with Webpack?

chad-autry:
@wesleycho I've been keeping my build system extremely minimal, just calling Browserify commands stored in my package.json npm scripts. I don't have any experience with any other packaging frameworks.

wesleycho:
Looks like there are webpack solutions out there as well.

ctaepper:
+1

dfsq:
Meanwhile it's possible to specify path to tar archive in devDependencies which allows set up custom builds with just needed modules:

    "angular-ui-bootstrap": "https://github.com/angular-ui/bootstrap/archive/0.14.3.tar.gz",

ctaepper:
@dfsq thanks for the hint... i tried referencing the repo/commit, but that didn't get me anywhere

joh-klein:
@dfsq how do you reference the ui-bootstrap template files in your project?

dfsq:
@joh-klein In our case we use webpack to import templates. It is also possible to set up build task (html2js) to pack those templates into $templateCache for only necessary modules (from BootstrapUI).

joh-klein:
[...]

PR #5129, opened 2015-12-27
feat: add npm support in main repository

wesleycho:
- Add support for consuming the main repository via npm for releases

This addresses #4739.


wesleycho:
This may still require a docs update - any recommendation for where to put that @Foxandxss ?

wesleycho:
Hmm this is not quite right - need to define the `ui.bootstrap` module still.

wesleycho:
My thoughts - we will want to document usage of this, as this is a major change.

There is no good way to bundle template support, but we can point people to some loaders such as https://github.com/teux/ng-cache-loader and https://github.com/WearyMonkey/ngtemplate-loader for Webpack and https://www.npmjs.com/package/browserify-ng-html2js for Browserify. We could perhaps provide a `template.js` helper file that does a bunch of requires for the templates, creates a `ui.bootstrap.tpls` module and export that, and import it into the `index.js` to maintain consistency with the previous complete bundle. This would require people are using Browserify or Webpack, and have the proper config setup for requiring HTML templates.

Foxandxss:
Yeah, we have our hands a bit tied in regard of what we can provide. All we can do is to document the different options.

I will think in how can we make it as easy as possible.

wesleycho:
Cleared by @Foxandxss

PR #5138, opened 2015-12-31
docs(position): tune up

Foxandxss:
This one has some changes in size compared to modal and dateparser, will retrofit as well.


src/position/docs/readme.md
```diff
@@ -1,114 +1,182 @@
...
+  _(Type: `object`)_ -
+  The element to convert.
 
 ##### returns
 
-  _(Type 'element')_ - A raw DOM element.
+* _(Type: `element`)_ -
+  A raw DOM element.
+
+#### parseSyle(element)
```
wesleycho: Should be `parseStyle` according to the source

src/position/docs/readme.md
```diff
@@ -1,114 +1,182 @@
...
-### parseSyle(element)
-  Parses a numeric style value to a number.  Strips units and will return 0 for invalid (NaN) numbers.
+Parses a numeric style value to a number.  Strips units and will return 0 for invalid (NaN) numbers.
 
 ##### parameters
 
-* `value` _(Type 'string')_ - The style value to parse.
+* `value`
+  _(Type: `string`)_ -
+  The style value to parse.
```
wesleycho: Is there an example of what this accepts?

wesleycho:
This LGTM for the most part minus that one pre-existing typo.

src/position/docs/readme.md
Foxandxss: Indeed, I also fixed one like that.

src/position/docs/readme.md
Foxandxss: I have no idea. All I did was reestructuring this doc for the general styling on the homepage.

All the doc was written by @RobJacobs, so I am just putting it to the right style.

I agree that that needs an example, can you fill a PR? I am not up to date on what position does with the latest changes. I just reviewed the docs for now.

icfantv:
I can do an example tonight after the kids go to bed.  I've already used it.  I'm thinking we do a drop down that changes the location?

Foxandxss:
If I understood correctly, he just want that line of docs to contain an example of the input accepted.

icfantv:
Oh.  And not an example of what each position looked like?

wesleycho:
I'm fine with this being merged in and an update maybe afterwards explaining that being added.

src/position/docs/readme.md
RobJacobs: This function is used to parse any style value that may have a unit to a number, e.g. width: 5px to 5.

PR #5141, opened 2016-01-03
Revert "fix(progressbar): fix `min-width` for Bootstrap 3.2"

wil93:
This reverts commit 8dc13be92469329605ca3dd73478cf7898a864ec.

That commit was made to address #2511, because Bootstrap 3.2 had introduced a special `min-width` value for progress bar values 0%, 1% and 2%. However, this is no longer the case as that feature was later reverted: twbs/bootstrap#12848. See also: twbs/bootstrap#13953.

The currently hardcoded `min-width: 0;` appears to be an issue in some cases, see [here](http://stackoverflow.com/questions/34572286/angular-ui-bootstrap-progressbar-minimum-width).


wesleycho:
LGTM

PR #5157, opened 2016-01-07
feat(tooltip): change back to jqLite listeners

wesleycho:
- Change event listeners to be handled by jqLite

This change is because it is more difficult to use the tooltip with raw events, which sort of goes against the whole point of a unified API like jQuery, and due to the original rationale being a browser bug, thus not a worthwhile fix in this library itself.

This should close #4886


wesleycho:
Approved by @Foxandxss

PR #5160, opened 2016-01-07
fix(tab): revert change to linking function

areel:
reverting commit 20047577 : return to using  the compile function rather than the link function. If tab directive using ng=transclude to inject content, the injected html can not refer to injected variables e.g ng-repeat variables. Have raised issue with more detail.


wesleycho:
Can you link the issue you are referring to?

areel:
Sorry got side lined, the background for this PR is outlined in issue #5161

src/tabs/tabs.js
```diff
@@ -202,36 +202,38 @@ angular.module('ui.bootstrap.tabs', [])
...
+            }
+          });
 
-      scope.disabled = false;
-      if (attrs.disable) {
-        scope.$parent.$watch($parse(attrs.disable), function(value) {
-          scope.disabled = !! value;
-        });
-      }
+          scope.disabled = false;
```
icfantv: please fix the indents.  thanks.

src/tabs/tabs.js
```diff
@@ -202,36 +202,38 @@ angular.module('ui.bootstrap.tabs', [])
...
     },
     controllerAs: 'tab',
-    link: function(scope, elm, attrs, tabsetCtrl, transclude) {
-      scope.$watch('active', function(active) {
-        if (active) {
-          tabsetCtrl.select(scope);
-        }
-      });
+    compile: function(elm, attrs, transclude) {
+        return function postLink(scope, elm, attrs, tabsetCtrl) {
```
wesleycho: This should be a two space indentation here

wesleycho:
Going to close this as stale - in addition, I believe we will have a fix for the original issue when we enforce Angular 1.5+ support at 3.0.0. I believe this PR as is would trigger a breaking change.

PR #5165, opened 2016-01-07
feat(typeahead): add event object to on select callback

deeg:
It would be nice to have access to the event object in the typeahead on select callback.


src/typeahead/typeahead.js
```diff
@@ -202,7 +202,7 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
       return false;
     };
 
-    var getMatchesAsync = function(inputValue) {
+    var getMatchesAsync = function(inputValue, evt) {
```
deeg: There are some calls to this function which do not seem to have an event object, such as when the typeahead directive inits.

Is it okay for the event object to be undefined in these cases? Any other suggestions in order to always have an event object?

wesleycho: I think it is fine - only thing is to perhaps add a note in the documentation that this can be the case.

wesleycho:
This LGTM, just need that one documentation update and this should be good.

deeg:
@wesleycho, thanks just updated the documentation.

Just curious, how often do you release a new version? Do you see this getting added to a new stable version soon?

I noticed some errors happening with typeahead in master, before this change was put in. Should I open a bug for it, or are things in master not yet released still under active development?

wesleycho:
If there is something wrong, open an issue with reproduction - we may be releasing tomorrow or Saturday, so if there is a real issue, the earlier we can see what is going on, the better

deeg:
@wesleycho, sounds good thanks.

Created issue for it: #5167

PR #5168, opened 2016-01-08
fix(typeahead): use correct selector

### LINKED ISSUE #5167: Errors in console when interacting with demo for typeahead
With the latest master code, the last demo on the typeahead causes errors in the console:

```
ui-bootstrap.js:6267 Uncaught TypeError: Cannot read property 'scrollIntoView' of undefined
(anonymous function)	@	ui-bootstrap.js:6267
eventHandler
```

https://plnkr.co/edit/myFVFqnO2fHNo0aEednm?p=preview

Type some text in the last example, and then hit the arrow down key to scroll through results. This causes the error.

Might just be an issue with the demo and not the code itself, didn't have the time to look into it yet.

wesleycho:
- With the addition of the header, the selector to scroll the dropdown
  onto the correct element is incorrect - this fixes it to select all
list elements in the dropdown

Plunker with fix in action [here](https://plnkr.co/edit/iuy9PM9NVASjSIvlBZ14?p=preview).

This fixes #5167.


Foxandxss:
LGTM.

PR #5176, opened 2016-01-08
Fix grammar error

zachlysobey:
"not longer" -> "no longer"


wesleycho:
I saw this heh - thanks

PR #5185, opened 2016-01-08
fix(timepicker): Make inputs disabled for timepicker.

### LINKED ISSUE #5182: timepicker - Value can be changed via the text input fields even if ng-disabled evaluates to true
Thanks for adding the ng-disabled feature for the timepicker.
It works fine in disabling value changes via the up and down buttons.

Though even when setting it to true, the text input fields are not disabled and the value still can be changed (Tested in Firefox & IE)

deeg:
Closes #5182 

These lines were not being tested before it looks like, let me know if you need tests for this even though it is in HTML. I tried adding some tests, but was having issues with getting the disabled attribute to show up in the tests.


deeg:
[Plunker with fix](http://plnkr.co/edit/DZJEpnEP28ivYLm4vYCC?p=preview)

wesleycho:
This LGTM - I think it might have been the result of an accidental resolution.

PR #5188, opened 2016-01-09
fix(datepicker): fix usage of non-standard format

### LINKED ISSUE #5187: M! and d! formats show ! when selecting a new date
There is an extra ! when using the new optional leading zero formats, due to this being a format that the date filter in Angular not recognizing.

The format needs to be parsed double in order to support this feature.

[Here](http://plnkr.co/edit/UIc5NFTt1UUL7iXe49NC?p=preview) is the Plunker illustrating the issue.

wesleycho:
- Fix display of model in input when using a non-standard format with
  the dateparser (i.e. `M!` and `d!`)

Fix in action [here](http://plnkr.co/edit/jx83kYdGFtYxjR0a4YhD?p=preview).

This fixes #5187


Foxandxss:
LGTM

Issue #5101, opened 2015-12-19
Bug - uibDropdown + ngTouch - not closing after touch


radarfox:
I have found this bug in the dropdown module and created plunker to demonstrate it: http://plnkr.co/edit/cQ8zjbqFfYQ4KMlI4raJ

Conditions:

* angular v1.4.8
* angular-touch v1.4.8
* angular-bootstrap v0.14.3
* touch device or Chrome with "Emulate touch screen" option
* make a "uib-dropdown-menu" with anchor elements using ngClick directive

Result:

Dropdown menu is not closing after touch on any of its item with ngClick. If you exclude ngTouch or do not use ngClick directive everything works ok. Also clicking outside menu works as expected.

wesleycho:
Thanks a lot for this with detailed reproduction - one work around for now is to have your ng-click in each item set the `isOpen` to `false`.

This may be tricky to fix, but I'm in the process of trying to figure out a good fix.

wesleycho:
Took a look further, and I don't think there is anything we can do in the library itself - the problem here is that with ngTouch, ng-click is overrided with a custom version that calls `event.stopPropagation` on the event, as seen in https://github.com/angular/angular.js/blob/74eb17d7c8232f72f134bf2546f10fed7234d276/src/ngTouch/directive/ngClick.js#L159 .

I will accept updating of the documentation to mention this workaround for when one uses ngTouch though.

wesleycho:
cc @Foxandxss

radarfox:
Proposed work around should work, but it is not very handy to have to care about each action of each dropdown to make sure dropdown gets closed.

Are you sure, that the problem is in ngTouch? I see plenty of `return` before `event.stopPropagation`. That code should only prevent unwanted clicks. Also this function is called on non-touch devices and there is no bug.

I think that here is part of code where event is created, but somehow it does not trigger dropdown close https://github.com/angular/angular.js/blob/74eb17d7c8232f72f134bf2546f10fed7234d276/src/ngTouch/directive/ngClick.js#L242

wesleycho:
I did some experimentation console.logging, the click callbacks never gets fired because the ng-click swallows the event.

radarfox:
ngClick is supposed to prevent the original `click` event, but it should emit fake `click` just after `touchend` - without having to wait 200ms or so on touch devices.

Foxandxss:
I have a fix underway. Working on the tests.

radarfox:
[...]

PR #5189, opened 2016-01-09
docs(dropdown): add note about usage with ngTouch

wesleycho:
- Add note about an issue with the click event not bubbling up when
  using ng-click with ngTouch

This addresses #5101.


Foxandxss:
Oh, we missed that. I recall the issue. LGTM

PR #5195, opened 2016-01-10
fix(carousel): Frequently changed active slide breaks carousel

### LINKED ISSUE #5178: Carousel: Frequently changed active slide and press navigation button crash the control
Chrome 47.0.2526.106 m
Please, see http://plnkr.co/edit/mk5Pby?p=preview

User frequently clicks on links "go to slide ##", sometimes user see two or three slides at the same time. 
When user see few slides he clicks on navigation button that appears above the links (left arrow, that means go to previous slide).
If user repeat these action few times the control will be broken.

mazavr:
- Disable active state for slides in a buffer

fixes #5178


wesleycho:
LGTM, thanks for this catch.

PR #5197, opened 2016-01-10
fix(build): add ; after script

### LINKED ISSUE #5196: Encapsulate build files in IIFE
When using `ui-bootstrap-tpls.js` with `gulp-useref` into a build process resulting code is not working. This is due to a missing `;` at the end of the file. Adding the `;` or encapsulating into an IIFE solves the issue.

I suggest better to encapsulate all the lib into an IIFE.

wesleycho:
- Add extra `;` so there is no issue when concatenating scripts

This fixes #5196


Foxandxss:
LGTM

philBrown:
Might be a good idea to move the inline script string to a `.js` file so it can be linted / quality tested as part of your build.

PR #5198, opened 2016-01-10
fix(datepicker): fix error message

### LINKED ISSUE #5191: [Date picker] Calendar is not rendering when model is a timestamp
Calendar is not rendering when model is a timestamp. This issue happens only with angular 1.5 and only when the popup is opened first time.

http://plnkr.co/edit/i2y5RqCzvyro8O7NVryg?p=preview

wesleycho:
- Change error message to reflect what is officially supported as a model value

This fixes #5191.


Foxandxss:
Yes please, that error message was really weird.

Issue #5192, opened 2016-01-09
[Collapse] First animation is still ran


bogdanalexe90:
Collapse has a strange flickering when is opened by default at first rendering and some delayed content is loaded. I took a look in the collapse directive and the `expand()` logic is still ran.

```js
// This should add `collapse in` class
if (!scope.$eval(attrs.uibCollapse)) {
          element.addClass('in')
            .addClass('collapse')
            .css({height: 'auto'});
}

function expand() {
    ...
          if ($animateCss) {
            $animateCss(element, {
              addClass: 'in',
              easing: 'ease',

              // if the async content is loaded after this
              // it will cause an animation towards the new scrollHeight

              to: { height: element[0].scrollHeight + 'px' }
            }).start()['finally'](expandDone);
          } else {
            ...
          }
        }

// Still the watcher will trigger expand logic
scope.$watch(attrs.uibCollapse, function(shouldCollapse) {
          if (shouldCollapse) {
            collapse();
          } else {
            expand();
          }
        });
```

Because we need to skip `expand()` only for the first rendering we can add a simple check inside the watcher:
```js
scope.$watch(attrs.uibCollapse, function(shouldCollapse) {
          if (shouldCollapse) {
            collapse();
          } else {
            if(arguments[0] !== arguments[1]){
                expand();
            }
          }
        });
```

icfantv:
@bogdanalexe90, per [these instructions](https://github.com/angular-ui/bootstrap/blob/master/CONTRIBUTING.md#you-think-youve-found-a-bug), can you please post a plunker that reproduces the issues?  Doing this for bugs usually greatly reduces the time it takes for us to diagnose and fix issues.  Thanks.

bogdanalexe90:
Please have a look on https://github.com/angular-ui/bootstrap/issues/4414 it's the same issue. Also you can have a look on my pull request https://github.com/angular-ui/bootstrap/pull/5199

icfantv:
Awesome, thanks.

Issue #4414, opened 2015-09-14
Collapse usage inside modal: section "pops" when not collapsed


levin81:
http://plnkr.co/edit/x7FIiKirhANONuAwgKvH?p=preview

Just added to the unordered-list element a 'collapse="false"' (so everything should be business as usual), but when the modal is loaded it first displays the rest of the elements and then the element with the collapse directive pops into view.

wesleycho:
I think this might be a duplicate of an existing issue - not quite sure what is causing this bug, it needs some investigation.

wesleycho:
Just noting, I suspect this could be a ngAnimate regression in Angular.

icfantv:
Given the age of this in OSS days we should see if it's reproducible with the latest code versions.  When I was going through the backlog this morning I was able to close several animation issues just by doing this.

wesleycho:
This is still an issue in `master`, as can be seen [here](http://plnkr.co/edit/9rv2nNLxqq4Ppkgx1abc?p=preview).

icfantv:
Closing in favor of #5192.  Also see PR #5199.

PR #5199, opened 2016-01-10
fix(collapse): avoid initial animation

bogdanalexe90:
- Ensure animation does not occur when first loaded

Fixes: https://github.com/angular-ui/bootstrap/issues/5192
Fixes: https://github.com/angular-ui/bootstrap/issues/4414


Foxandxss:
Can you create a plunker with the old broken behavior and another plunker on how it end with your fix?

The test stuff is really nice (that cleanup always is handy) but this would need another test if we end deciding in merging this.

Thank you!

bogdanalexe90:
I was able to reproduce this only with accordion plugin. For this reason i don't have any idea how to test the fix. Any suggestion are welcomed.

Before: http://plnkr.co/edit/l0Qc5rWfgCNhx4hU4JZx?p=preview
After: http://plnkr.co/edit/bp5PlTC5g6jjWDoLcleR?p=preview

Foxandxss:
What should I do / see there? I don't see any difference when loading.

bogdanalexe90:
ahh it was a false result. i will try to include some 3rd party libs. I'm experiencing a flickering a lot when using `ui-select` inside. Also i will try to have a look if i have some custom styles in my app.

I will try to create another plnkr soon.

bogdanalexe90:
I identified the following problems:

1. An fixed height is set to the element
```
$animateCss(element, {
              addClass: 'in',
              easing: 'ease',
              to: { height: element[0].scrollHeight + 'px' }
            }).start()['finally'](expandDone);
```
2. Some content arrives and `element[0].scrollHeight` changes

3. The height is set to auto and the rest of the content is showed
```
element.removeClass('collapsing')
            .addClass('collapse')
            .css({height: 'auto'});
```


The problem (flickering) is only visible when you have some complex styles to render. I created a plnkr and you can see the describe behaviour with some breakpoints (on line 31 and line 44).

http://plnkr.co/edit/7dyZ3InIQ3gg69SmVAgT?p=preview

bogdanalexe90:
I had a look on https://github.com/angular-ui/bootstrap/issues/4414 . It seems it's the same problem, i created a plunker demonstrating the fix:

http://plnkr.co/edit/Y8Uf0KKDq2ubMxAa71dg?p=preview

wesleycho:
This needs a test still in order for us to accept it.

src/collapse/test/collapse.spec.js
```diff
@@ -1,55 +1,76 @@
-describe('collapse directive', function() {
+describe('collapse directive', function () {
```
wesleycho: This whitespace change is unnecessary - this will need to all be changed back on merge.

wesleycho:
Minus all of the whitespace changes that should be undone, what are your thoughts on this @Foxandxss ?

wesleycho:
Can you squash your commits into 1?

PR #5200, opened 2016-01-11
feat(timepicker): set templateUrl from uibTimepickerConfig

### LINKED ISSUE #5087: Add ability to set templateUrl with uibTimepickerConfig
new timepicker has uibTimepickerConfig but it doesn't support templateUrl option
It's still possible to set it via attributes, but it would be useful to configure it with standard config

you may check here: http://plnkr.co/edit/Uvw1ECUNK4S6mviz3Qak?p=preview

deeg:
Closes #5087


wesleycho:
LGTM

PR #5201, opened 2016-01-11
fix(timepicker): Don't allow mixture of numbers and letters.

### LINKED ISSUE #5085: Timepicker Bug:minute accepts alphabetic character and doesnt show validation error
There is bug with when u enter 10:3a it accepts the input. It should have thrown an error. The following example is from the website demo.

error

 Now I need to Validate that how can i do that??

i want the 10:3a to throw an error. and not accept it has 10:03 PM

deeg:
Closes #5085

[This stack](http://stackoverflow.com/questions/4564158/what-is-the-difference-between-parseintstring-and-numberstring-in-javascript) says you should use [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) when you want to make sure the entire string is a number.

[Plunker showing fix.](http://plnkr.co/edit/WxoVcEBEhOV0xIcAn7kS?p=preview)


src/timepicker/timepicker.js
```diff
@@ -51,14 +51,14 @@ angular.module('ui.bootstrap.timepicker', [])
   var hourStep = timepickerConfig.hourStep;
   if ($attrs.hourStep) {
     $scope.$parent.$watch($parse($attrs.hourStep), function(value) {
-      hourStep = parseInt(value, 10);
```
wesleycho: I think `+value` would be better if `parseInt` is not sufficient to ward off this issue.

wesleycho:
Linking a [thorough list of different number parsers](http://stackoverflow.com/questions/17106681/parseint-vs-unary-plus-when-to-use-which/17106702#17106702) for reference.

IMO the unary operator is the best choice given it behaves identically to the `Number` constructor and cannot be overridden by the user.

deeg:
Sounds good, will change to that method.

deeg:
@wesleycho, should be all set.

Updated plunker with new code.

src/timepicker/timepicker.js
```diff
@@ -51,14 +51,14 @@ angular.module('ui.bootstrap.timepicker', [])
...
+      hourStep = parseInt(+value, 10);
```
wesleycho: Oh I meant just changing it all to `hourStep = +value` and all the rest accordingly. No need to run `parseInt` after `+`, it's already a number at that stage.

deeg: Doh, duh, will change.

deeg:
Alright, third time a charm!

Issue #5193, opened 2016-01-09
Upgrade from 0.13.0 to 1.0 doesn't render tabs


andyczerwonka:
I'm doing an upgrade, followed the guide. I no longer get my tabs rendered. They work, but they're just strings. In 0.13.0, they rendered nicely.

Is there a new requirement for new CSS other than Bootstrap?

wesleycho:
In 1.0, there is - see the changelog: https://github.com/angular-ui/bootstrap/blob/master/CHANGELOG.md#breaking-changes

The necessary CSS can be found [here](https://github.com/angular-ui/bootstrap/blob/master/src/tabs/docs/demo.html).

andyczerwonka:
Ok. That's quite strange given the whole point is to integrate Bootstrap, so I didn't anticipate a CSS requirement.

wesleycho:
We made this decision because we've had a fair number of issues where users were running into issues where the anchor tag would cause users to be routed whenever they nested other clickable elements inside the tab, such as another anchor element or a button element. It is our belief that the library should not have an opinion on what HTML is used there.

A similar issue happened with the accordion, which resulted in the same modification made there.

philBrown:
Aren't issues with user anchors the user's problem? FWIW, we didn't have a single issue with using `<a>` tags in tabs and now we need to add custom CSS to try and keep this feature in line with Bootstrap.

Personally, I think this was a very bad idea. A better solution would have been to explicitly state in the documentation that clickable sub-elements will need to cancel event propagation or something to that effect.

> It is our belief that the library should not have an opinion on what HTML is used there.

Unfortunately, Bootstrap doesn't appear to hold this belief and they're the ones **you** are meant to be integrating with.

wesleycho:
That isn't sufficient to fix that issue - the browser behavior results in the click triggering on the anchor tag, and not any children elements even though the user clicked on the children elements.

This is the solution we ultimately decided on due to another alternative violating CSP, and this being behavior that should never happen when using tabs as the option of least problems on a library level.

The custom CSS is minor, and unless there is a stronger argument made given the known routing and CSP problems that have to be consistent, we are not going to change this. If you do not want to override the CSS, then you could override the template and replace the div elements with anchor elements, but this is a decision arrived at with consensus with multiple team members.

philBrown:
Minor as the CSS may be, it's another point you're leaving to your users to keep in sync with the Bootstrap styles that this library is meant to integrate with.

[...]

PR #5202, opened 2016-01-11
style(tabs): Match Bootstrap tabs / pills styles

philBrown:
Adds all style declarations for Bootstrap (v3.3.6) tabs and pills.

See http://getbootstrap.com/components/#nav


wesleycho:
LGTM

georgmu:
Wouldn't it be better to alter the template instead of the demo?

This way, everybody has to implement it. Changing the default template from "<div ..." to "<a ..." with an empty href would solve it for everybody

wesleycho:
@georgmu that has an issue as well, although I can't remember exactly what it was - it might have to do with the element the browser triggers the click on as the anchor element instead of any nested clickable element.

georgmu:
I changed the following and it works fine for me. But if you have other experience: ok.
```diff
diff --git a/template/tabs/tab.html b/template/tabs/tab.html
index b7b83e5..d93a2d3 100644
--- a/template/tabs/tab.html
+++ b/template/tabs/tab.html
@@ -1,3 +1,3 @@
 <li ng-class="{active: active, disabled: disabled}" class="uib-tab">
-  <div ng-click="select()" uib-tab-heading-transclude>{{heading}}</div>
+  <a ng-click="select()" uib-tab-heading-transclude href="">{{heading}}</a>
 </li>
```

Maybe I should raise an issue within bootstrap to not depend on "a" inside of the li element, but support div as well...

The other idea would be to provide the css somewhere in angular-ui-bootstrap. But without this, the out-of-the-box tpls solution is not usable without major css additions.

philBrown:
@georgmu the original issue is [here](https://github.com/angular-ui/bootstrap/issues/5193) and at the moment, I agree that the best approach is to provide your own CSS or override the template.

> Maybe I should raise an issue within bootstrap to not depend on "a" inside of the li element, but support div as well...

I too am not a fan of all the `.bootstrap-component > tag > tag` stuff in Bootstrap's CSS but I don't like your chances of getting that changed ;)

wesleycho:
@philBrown take a look at #5211 - I figured out a solution to bundle the CSS, will go out in a few hours when we release 1.0.1 (eta ~6-8 hours from now)

philBrown:
@wesleycho it's just a shame you'll have to keep your CSS up-to-date with any Bootstrap changes. I might try creating a pull request over at https://github.com/twbs/bootstrap to change their CSS to

    .nav-tabs > li > a, .nav-tabs > li > .nav-tab
    
then you'll just need to add a `class="nav-tab"` to whatever element acts as the trigger.

Issue #1988, opened 2014-03-26
Typeahead issues


vitaly-t:
Just looking at the Typehead demo that's on the library's official website, it doesn't seem to work right.

 - Example 1, for "Asynchronous results"
Type in "Mosc", nothing comes up. Then press "o", suddenly 2 items come up, then "w", for "Moscow", it is found, but strangely, wasn't shown before on the list. That is wrong.

- Example 2, for "Custom templates for results"
Type in something and select any item found. Now hit Backspace to remove the last symbol. The select list appears totally out of place.

Tested in Chrome version 33

wesleycho:
This does not appear to be issues with typeahead itself.

The second one is a demo site issue, and PRs are welcome fixing it.

wesleycho:
Closing, as it appears this has been fixed.

PR #5212, opened 2016-01-11
fix(typeahead): Scroll only parent element, not apge

### LINKED ISSUE #5180: Using arrow keys to navigate typeahead response causes page to move when typeahead below the fold
https://plnkr.co/edit/vk7MZB7JXnE40U1AeU0z?p=preview

Go to the first example and type in a letter so some values show up. Then use the down arrow to try to scroll through those values. This moves the page so the highlighted value is at the very bottom of the page rather than keeping the page where it is and moving just the highlighted item.

This is also apparent on the ui-bootstrap homepage demo as the Plunk is just a copy of that example.

deeg:
Closes #5180 

[Plunker demonstrating fix](https://plnkr.co/edit/b3AWsVVul8hadUJwEoVC?p=preview)

Let me know if you want tests if this is a good solution. I can figure out a way to test it. FWIW does not look like this was being tested before.


wesleycho:
I'm fine with this approach - @Foxandxss what is your thoughts?

Foxandxss:
Fine for me, better if tested.

deeg:
Sounds good, will add tests within an hour or so after I finish some work up.

Hope to make it into 1.0.1 if you don't mind waiting for an hour or two.

Foxandxss:
That is fine by me.

deeg:
Just added tests.

Thanks @wesleycho & @Foxandxss for the feedback and getting this into 1.0.1

frfancha:
The behavior of the last typeahead in the plunker with fix is quite strange: sometimes it opens several pixels under the input instead of just below as usual, and the "candidate" row is not highlighted in blue.
See attached print screen (windows 7, firefox ESR 38.4.0)
typeahead_printscreen

wesleycho:
That is actually a bug that has been existing for about 2-3 years on the demo page, but no fix yet.

frfancha:
Ok but note that in the plunker without the fix (demo in 5180) I can't reproduce this last problem (at least not after 20 tries) (But I see the bug described in 5180 of course)

deeg:
I am planning to fix another bug out there while making the behavior a bit better.

I will take a look at the issue you are describing @frfancha and see if I can't fix that also.

@wesleycho, do you know if there is a ticket for that one?

wesleycho:
The positioning issue with the typeahead demo is the second mentioned in #1988

PR #5216, opened 2016-01-12
chore(license): update copyright year range

nunoarruda:


wesleycho:
Heh, good catch - thanks!

PR #5224, opened 2016-01-12
fix(modal): compile modal before inserting into DOM

### LINKED ISSUE #5183: display issues when opening modal
reproducible on demo: http://angular-ui.github.io/bootstrap/#/modal

issue 1:
open modal and pay attention to the page in the background. notice it shifts down briefly before returning to normal.

issue 2:
turn off animation. open a large or small modal and pay attention to how the modal content area displays as its opened. notice it shrinks or expands before displaying as expected.

onumossn:
I think this fixes #5183

### DIFF
--- a/src/modal/test/modal.spec.js
+++ b/src/modal/test/modal.spec.js
@@ -213,12 +213,16 @@ describe('$uibModal', function () {
     element.trigger(e);
   }
 
-  function open(modalOptions, noFlush) {
+  function open(modalOptions, noFlush, noDigest) {
     var modal = $uibModal.open(modalOptions);
-    $rootScope.$digest();
-    if (!noFlush) {
-      $animate.flush();
+
+    if (!noDigest) {
+      $rootScope.$digest();
+      if (!noFlush) {
+        $animate.flush();
+      }      
     }
+
     return modal;
   }
 
@@ -259,6 +263,46 @@ describe('$uibModal', function () {
       expect($document).toHaveModalsOpen(0);
 
       expect($document).not.toHaveBackdrop();
+    });
+
+    it('should compile modal before inserting into DOM', function() {
+      var topModal;
+      var modalInstance = {
+        result: $q.defer(),
+        opened: $q.defer(),
+        closed: $q.defer(),
+        rendered: $q.defer(),
+        close: function (result) {
+          return $uibModalStack.close(modalInstance, result);
+        },
+        dismiss: function (reason) {
+          return $uibModalStack.dismiss(modalInstance, reason);
+        }
+      };
+      var expectedText = 'test';
+
+      $uibModalStack.open(modalInstance, {
+        appendTo: angular.element(document.body),
+        scope: $rootScope.$new(),
+        deferred: modalInstance.result,
+        renderDeferred: modalInstance.rendered,
+        closedDeferred: modalInstance.closed,
+        content: '<div id="test">{{\'' + expectedText + '\'}}</div>'
+      });
+
+      topModal = $uibModalStack.getTop();
+
+      expect(topModal.value.modalDomEl.find('#test').length).toEqual(0);
+      expect(angular.element('#test').length).toEqual(0);
+
+      $rootScope.$digest();
+
+      expect(topModal.value.modalDomEl.find('#test').text()).toEqual(expectedText);
+      expect(angular.element('#test').text()).toEqual(expectedText);
+
+      $animate.flush();
+
+      close(modalInstance, 'closing in test', true);
     });
 
     it('should not throw an exception on a second dismiss', function() {
@@ -364,7 +408,6 @@ describe('$uibModal', function () {
       expect(document.activeElement.tagName).toBe('A');
 
       var modal = open({template: '<div>Content<button>inside modal</button></div>'});
-      $animate.flush();
       $rootScope.$digest();
       expect(document.activeElement.tagName).toBe('DIV');
       expect($document).toHaveModalsOpen(1);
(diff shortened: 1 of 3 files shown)


wesleycho:
This looks like a good fix to me - thoughts @Foxandxss and/or @icfantv ?

Foxandxss:
If it fixes it, I am all in, but we need to be sure (AKA plunker).

icfantv:
Looks like there's 13 broken tests in the build.  Need to see if the tests are no longer valid or they need to be updated.  @onumossn, if you would like this merged, your work is not yet complete.

onumossn:
The original commit, actually, only fixed issue 1 of the two listed on #5183, so I made additional updates to fix the 2nd issue.

@Foxandxss The cases can be tested from the demos, so is a plunker necessary?

@icfantv Sorry about that. I had used the wrong command to test initially. I updated the tests. Previously an extra $animate.flush was used because the $compile would use more $animates after the first $animate.flush in the open function, but since the compile no longer waits for the first animation, its no longer necessary.

Foxandxss:
A test to prove that your PR fixes the issue.

icfantv:
@onumossn, what @Foxandxss means is take the proper file from the `dist` folder and create a plunker using it.  Easiest thing to do would be to use the modal plunker demo (from the docs page) remove the script include for UIBS, and add yours locally instead.

@Foxandxss, @wesleycho, IIRC we added those `flush()` calls when angular-animate changed between 1.4.4 and 1.4.5, right?

wesleycho:
The `flush()` has to do with ngAnimate.

What you could do is test for the newly created element's presence in the element being appended to (`body` by default) and then flush.

onumossn:
`flush` is being run in the `open` helper. However, `flush` needed to be run twice before because after the first `flush` the `uib-modal-window` directive gets compiled which adds more animation to the buffer which needs to be flushed again. In this version, the compile occurs before the first `flush` so all the animations are in the buffer when the first `flush` occurs in the `open` helper. I may not have been very clear in my last comment.

before plunker:
http://embed.plnkr.co/YbWBhcwoFECDFnpP4rnW/

after plunker:
http://embed.plnkr.co/pCBkbYY1BIkBeQ8lUR0q/

onumossn:
before plunker: 
https://embed.plnkr.co/YbWBhcwoFECDFnpP4rnW/

after plunker:
https://embed.plnkr.co/pCBkbYY1BIkBeQ8lUR0q/

template/modal/window.html
```diff
@@ -2,5 +2,5 @@
     uib-modal-animation-class="fade"
     modal-in-class="in"
     ng-style="{'z-index': 1050 + index*10, display: 'block'}">
-    <div class="modal-dialog" ng-class="size ? 'modal-' + size : ''"><div class="modal-content" uib-modal-transclude></div></div>
+    <div class="modal-dialog {{size ? 'modal-' + size : ''}}"><div class="modal-content" uib-modal-transclude></div></div>
```
wesleycho: Why is this change made? This seems unnecessary

wesleycho:
This needs a test still that the element is compiled before the $animate.enter is fired.

template/modal/window.html
onumossn: ngClass was going through the animation processes of all the before and after classes which was causing the size when picking small or large to be regular size when initially rendered and the jump to the correct size.

wesleycho: Ugh, that's a negative of it being tied to $animate...but I guess it can't be avoided then.

onumossn:
Added tests

src/modal/test/modal.spec.js
```diff
@@ -261,6 +265,46 @@ describe('$uibModal', function () {
...
+        scope: $rootScope.$new(),
+        deferred: modalInstance.result,
+        renderDeferred: modalInstance.rendered,
+        closedDeferred: modalInstance.closed,
+        content: '<div id="test">{{\'' + expectedText + '\'}}</div>'
+      });
+
+      topModal = $uibModalStack.getTop();
+
+      expect(topModal.value.modalDomEl[0].querySelector('#test')).toBeNull();
```
wesleycho: You could just do `topModal.value.modalDomEl.find('#test')` here I believe - we have jQuery in the testing situation.

src/modal/test/modal.spec.js
```diff
@@ -261,6 +265,46 @@ describe('$uibModal', function () {
...
+      expect(document.querySelector('#test')).toBeNull();
```
wesleycho: You can just do `$('#test').length`

src/modal/test/modal.spec.js
```diff
@@ -261,6 +265,46 @@ describe('$uibModal', function () {
...
+
+      $rootScope.$digest();
+
+      expect(topModal.value.modalDomEl[0].querySelector('#test').textContent).toEqual(expectedText);
```
wesleycho: Ditto on the query selector

src/modal/test/modal.spec.js
```diff
@@ -261,6 +265,46 @@ describe('$uibModal', function () {
...
+      expect(document.querySelector('#test').textContent).toEqual(expectedText);
```
wesleycho: Could just do `$('#test').text()`

wesleycho:
Once the query selectors are changed to just use jQuery, can you squash your commits into one?

This LGTM otherwise.

onumossn:
Update the tests to use jQuery. Sorry for the delay.

wesleycho:
LGTM - thanks for this!

PR #5226, opened 2016-01-13
feat(collapse): add collapsing/collapsed/expanding/expanded callbacks

### LINKED ISSUE #5194: Collapse: Support for pre/post collapse/expand callbacks
In my application, before expanding a collapsed section, I need to populate the contents.  Similarly, after collapsing an expanded section, I'd like to do some cleanup.

It would be great if the `uib-collapse` supported additional attributes that would be evaluated before and after the collapse animation, e.g. `expanding`, `expanded`, `collapsing`, `collapsed`.  If the "before" attribute (`collapsing`/`expanding`) returns a promise, the directive should wait for it to resolve before beginning the collapse/expand animation.

Specifically, I'd like to be able to do something like this:
```html
<div uib-collapse="isOpen" expanding="setupContent()" collapsed="cleanupContent()">
    <!-- complex & expensive content goes here -->
</div>
```

FWIW, I have an implementation and can create a PR if there is general interest in this...

nonplus:
Closes #5194

### DIFF
--- a/src/collapse/collapse.js
+++ b/src/collapse/collapse.js
@@ -1,9 +1,14 @@
 angular.module('ui.bootstrap.collapse', [])
 
-  .directive('uibCollapse', ['$animate', '$injector', function($animate, $injector) {
+  .directive('uibCollapse', ['$animate', '$q', '$parse', '$injector', function($animate, $q, $parse, $injector) {
     var $animateCss = $injector.has('$animateCss') ? $injector.get('$animateCss') : null;
     return {
       link: function(scope, element, attrs) {
+        var expandingExpr = $parse(attrs.expanding),
+            expandedExpr = $parse(attrs.expanded),
+            collapsingExpr = $parse(attrs.collapsing),
+            collapsedExpr = $parse(attrs.collapsed);
+
         if (!scope.$eval(attrs.uibCollapse)) {
           element.addClass('in')
             .addClass('collapse')
@@ -11,28 +16,32 @@ angular.module('ui.bootstrap.collapse', [])
         }
 
         function expand() {
-          element.removeClass('collapse')
-            .addClass('collapsing')
-            .attr('aria-expanded', true)
-            .attr('aria-hidden', false);
+          $q.resolve(expandingExpr(scope))
+            .then(function() {
+              element.removeClass('collapse')
+                .addClass('collapsing')
+                .attr('aria-expanded', true)
+                .attr('aria-hidden', false);
 
-          if ($animateCss) {
-            $animateCss(element, {
-              addClass: 'in',
-              easing: 'ease',
-              to: { height: element[0].scrollHeight + 'px' }
-            }).start()['finally'](expandDone);
-          } else {
-            $animate.addClass(element, 'in', {
-              to: { height: element[0].scrollHeight + 'px' }
-            }).then(expandDone);
-          }
+              if ($animateCss) {
+                $animateCss(element, {
+                  addClass: 'in',
+                  easing: 'ease',
+                  to: { height: element[0].scrollHeight + 'px' }
+                }).start()['finally'](expandDone);
+              } else {
+                $animate.addClass(element, 'in', {
+                  to: { height: element[0].scrollHeight + 'px' }
+                }).then(expandDone);
+              }
+            });
         }
 
         function expandDone() {
           element.removeClass('collapsing')
             .addClass('collapse')
             .css({height: 'auto'});
+          expandedExpr(scope);
         }
 
         function collapse() {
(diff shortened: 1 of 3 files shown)


icfantv:
@nonplus, thank you for taking the initiative to do this.  Due to the changes that aren't actually changes it's really hard to see what's changed and what has not.  Please fix your commit such that only the lines you've changed are committed.  I.e., don't change the whitespace, don't touch lines not relevant to your changes, etc...  Thanks.

src/collapse/docs/readme.md
```diff
@@ -7,3 +7,27 @@
   <i class="glyphicon glyphicon-eye-open"></i>
   _(Default: `false`)_ -
   Whether the element should be collapsed or not.
+
```
icfantv: Look at the `uib-tab` documentation for what to put here:  default should be `null`, callback could also be an expression and not a function.

icfantv: Check that, @wesleycho, what do you think about that last bit?  Function vs. expression?

nonplus:
All the changes are relevant.  The indentation changes in collapse.js (`collapse()` and `expand()` functions) are due to the fact that the existing (unchanged) code is now inside of a `$q.when()` callback.

Not sure what else I ccould do about that.  Or maybe I'm misunderstanding your request?

src/collapse/collapse.js
```diff
@@ -11,28 +11,31 @@ angular.module('ui.bootstrap.collapse', [])
         }
 
         function expand() {
-          element.removeClass('collapse')
-            .addClass('collapsing')
-            .attr('aria-expanded', true)
-            .attr('aria-hidden', false);
+          $q.when(scope.$eval(attrs.expanding), function() {
```
wesleycho: Might be better to use `$q.resolve`.

In addition, perhaps it is better to do
```js
$q.resolve(scope.$eval(attrs.expanding))
  .then(function() {
    ...
  });
```
since it is more in line with standard promise usage.

src/collapse/collapse.js
```diff
@@ -40,34 +43,37 @@ angular.module('ui.bootstrap.collapse', [])
             return collapseDone();
           }
 
-          element
+          $q.when(scope.$eval(attrs.collapsing), function() {
```
wesleycho: Ditto for here too

src/collapse/docs/readme.md
wesleycho: I think it should read expression.

src/collapse/collapse.js
wesleycho: In addition, it would be better to cache the expression via `var expandingExpr = $parse(attrs.expanding)` earlier (and similarly for the other expressions), and then one can execute the expressions by doing `expandingExpr(scope)`.

src/collapse/docs/readme.md
nonplus: I went off the uib-alert documentation which uses:
```
* `close`
  <small class="badge">$</small>
  _(Default: `none`)_ -
  A callback function that gets fired when an `alert` is closed. If the attribute exists, a close button is displayed as well.
```
But I'm happy to go with the format used by the uib-tab directive, if you prefer.

src/collapse/collapse.js
```diff
@@ -11,28 +11,31 @@ angular.module('ui.bootstrap.collapse', [])
...
+              }).then(expandDone);
+            }
+          });
         }
 
         function expandDone() {
           element.removeClass('collapsing')
             .addClass('collapse')
             .css({height: 'auto'});
+          scope.$eval(attrs.expanded);
```
wesleycho: Caching the parsed expression via $parse would be good here.

src/collapse/collapse.js
```diff
@@ -40,34 +43,37 @@ angular.module('ui.bootstrap.collapse', [])
...
+              }).then(collapseDone);
+            }
+          });
         }
 
         function collapseDone() {
           element.css({height: '0'}); // Required so that collapse works when animation is disabled
           element.removeClass('collapsing')
             .addClass('collapse');
+          scope.$eval(attrs.collapsed);
```
wesleycho: Ditto here for caching via $parse

wesleycho:
Thanks for the PR here!

This looks mostly solid - just would like to see those changes made with those minor optimizations and documentation updates, and should be good to get in afterwards.

src/collapse/docs/readme.md
nonplus: Hmm, looks like `uib-alert` documentation is the exception.  How does the following look?
```md
* `collapsing`
  <small class="badge">$</small> -
  An optional expression called before the element begins collapsing.
  If the expression returns a promise, animation won't start until the promise resolves.
  If the returned promise is rejected, collapsing will be cancelled.
  
* `collapsed`
  <small class="badge">$</small> -
  An optional expression called after the element finished collapsing.
  
* `expanding`
  <small class="badge">$</small> -
  An optional expression called before the element begins expanding.
  If the expression returns a promise, animation won't start until the promise resolves.
  If the returned promise is rejected, expanding will be cancelled.
  
* `expanded`
  <small class="badge">$</small> -
  An optional expression called after the element finished expanding.
```

src/collapse/collapse.js
nonplus: Good point about caching the expression.

FWIW, `$q.resolve(x, callback)` is more performant than `$q.resolve(x).then(callback)` since the latter waits for another tick.  But I can make the change if you'd prefer.

nonplus:
Should be good to go.

wesleycho:
Can you squash this into one commit?

src/collapse/docs/readme.md
Foxandxss: Let's be honest in here. Some callbacks has default and other doesn't. Leave it as is, I prefer this way.

wesleycho:
This is good work - thanks for this PR!

PR #5244, opened 2016-01-14
test(tooltip): add test for class change

wesleycho:
- Add test for adding the appropriate classes to the tooltip

This is the test that passes with the addition of #5235


RobJacobs:
LGTM

PR #5245, opened 2016-01-14
fix(tabs): add styles for -justified tab classes

### LINKED ISSUE #5243: Missing bottom border on uib-tabset inactive tabs with justified=true
When the uib-tabset is configured with justified=true, the bottom border of the inactives tabs is not visible. This is reproductible on the demo page.

For justified tabs, Bootstrap display the bottom border on each individual tabs but this CSS is missing as div are used for tabs instead of a.

wesleycho:
- Add styles for when the user is using `.nav-justified` and
  `.nav-tabs-justified`

This fixes #5243.


src/tabs/tabs.css
```diff
@@ -44,4 +44,33 @@
...
+  }
+}
+.nav-tabs.nav-justified > li > div {
+  margin-right: 0;
+  border-radius: 4px;
+}
+.nav-tabs.nav-justified > .active > div,
+.nav-tabs.nav-justified > .active > div:hover,
+.nav-tabs.nav-justified > .active > div:focus {
+  border: 1px solid #ddd;
```
icfantv: do we want `cursor: pointer;` here as well?

icfantv: nevermind.  i think we're ok...

wesleycho:
This PR may be rejected by discussion in #5254

wesleycho:
Closing as superceded by the culmination of #5262.

PR #5246, opened 2016-01-14
fix(position): positionArrow descendant selector

### LINKED ISSUE #5230: Arrow placement of a tooltip when inside a popover gets the placement of the popover arrow
Hi

I'm encountering an issue when using a tooltip inside a popover where the tooltip arrow gets the same placement property as the popover one when placement is _beyond the standards_ like bottom-left. See
[this plunker](http://plnkr.co/edit/JKfiCAJcIxIRGDmEKOHi?p=preview) for example.

This comes from the method [positionArrow in the position.js file](https://github.com/angular-ui/bootstrap/blob/master/src/position/position.js#L469) that manage the _beyond the standards_ placement. It finds the `.tooltip-inner` arrow element when positionning the popover arrow...

I did not find any quick fix to contribute sorry

Happens with Angular 1.4.8 and Angular-Bootstrap 1.0.3.

RobJacobs:
- The element selector was not limited to direct descendants so
nested tooltip/popover arrow elements were getting selected instead
of the direct descendant arrow element.

Fixes #5230

### DIFF
--- a/src/position/position.js
+++ b/src/position/position.js
@@ -464,16 +464,12 @@ angular.module('ui.bootstrap.position', [])
       positionArrow: function(elem, placement) {
         elem = this.getRawNode(elem);
 
-        var isTooltip = true;
-
-        var innerElem = elem.querySelector('.tooltip-inner');
-        if (!innerElem) {
-          isTooltip = false;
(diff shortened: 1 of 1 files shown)


wesleycho:
LGTM

PR #5250, opened 2016-01-14
fix(tabs): fix tab css cursor

### LINKED ISSUE #5247: Mouse cursor doesn't show as hand hovering on tab header
"name": "angular-bootstrap",
"version": "1.0.3",

See the example code here: http://plnkr.co/edit/?p=preview

icfantv:
* fix tab cursor for hover state on non-disabled tab.

Fixes #5247


icfantv:
just a note that i changed the order of the css styles to match the ones for `uib-tab.disabled` - i think Bootstrap actually provides a recommendation for style ordering but I don't know what it is off the top of my head.

icfantv:
Crap, I just saw this:  https://github.com/angular-ui/bootstrap/blob/master/CHANGELOG.md#breaking-changes.  If we don't want to merge this, I understand.

wesleycho:
I'm fine with this being merged - less differences from prior tab behavior.

PR #5251, opened 2016-01-14
fix(typeahead): Fix css styles

deeg:
Typeahead was broken as part of https://github.com/angular-ui/bootstrap/commit/e8201d1f34cb3afc2d5de4ef6c2d0e591347f2b5

This fixes the CSS so it shows up.

@wesleycho, this is important before you release so typeahead does not break.


wesleycho:
Oh, whoops...that was a mistake - thanks for this catch.

PR #5252, opened 2016-01-14
feat(typeahead): Add title to matches

deeg:
This adds a title attribute to the typeahead matches.

This helps if users want to truncate text and set a max width of the typeahead.

This allows the user to hover over a match and get the full text of the label.

Let me know if you want me to make this configurable in the options to allow the user to chose if they want the title attribute.


template/typeahead/typeahead-match.html
```diff
@@ -1 +1,4 @@
-<a href tabindex="-1" ng-bind-html="match.label | uibTypeaheadHighlight:query"></a>
+<a href
+   tabindex="-1"
+   ng-bind-html="match.label | uibTypeaheadHighlight:query"
+   ng-attr-title="{{match.label}}"></a>
```
wesleycho: Any reason for `ng-attr` here instead of just `title="{{...}}"`?

deeg: [These guys](http://www.thinkingmedia.ca/2015/03/conditionally-add-a-html-element-attribute-value-in-angularjs/#understanding-the-ng-attr-prefix) explain it pretty well.

It is basically so you never see the text ```{{match.label}}``` in the tooltip. This will only add the title attribute with the actual value of match.label.

wesleycho:
LGTM, thanks for this!

PR #5258, opened 2016-01-15
fix(datepicker): Fix datepicker CSS

deeg:
@wesleycho, this is important for the next release for the datepicker popup to not break.

The issue was the dropdown-menu display:none css was taking precedence before this change.


wesleycho:
LGTM

PR #5260, opened 2016-01-15
chore(datepicker): Unregister parent watchers on $destroy

deeg:
Closes #5242

@wesleycho, looking for advice on how to test.

I basically want to spy on the $parent.$watch function and return a spy and see how many times the spy is called. I had this

```
    describe('clean parent watchers', function(){
        it("should clean watchers in controller", function(){
          var scope;
          var clearSpy;
          inject(function($controller, $rootScope, $parse, $interpolate, $log, dateFilter, uibDatepickerConfig, $datepickerSuppressError, uibDateParser) {
            scope = $rootScope.$new();
            scope.$parent = jasmine.createSpyObj("scope.$parent", ["$watch"]);
            clearSpy = jasmine.createSpy("clearSpy");
            scope.$parent.$watch.and.returnValue(clearSpy);
            $controller('UibDatepickerController', {
              $scope: scope,
              $attrs: {},
              $parse: $parse,
              $interpolate: $interpolate,
              $log: $log,
              dateFilter: dateFilter,
              datepickerConfig: uibDatepickerConfig,
              $datepickerSuppressError: $datepickerSuppressError,
              dateParser: uibDateParser
            });
          });

          $templateCache.put('uib/template/datepicker/datepicker.html', '<div>{{datepicker.text}}</div>');
          $rootScope.date = new Date('September 12, 2010');
          $rootScope.mode = 'month';
          element = $compile('<uib-datepicker ng-model="date" min-date="date" max-date="date" min-mode="mode" max-mode="mode" init-date="date" ng-disabled="true"></uib-datepicker>')($rootScope);
          $rootScope.$digest();

          var ctrl = element.controller('uib-datepicker');
          ctrl.$destroy();

          expect(clearSpy.calls.count()).toBe(6);
        });
```

but was getting this error when running

```
TypeError: Cannot read property 'uib:datepicker.focus' of undefined
	    at Scope.$on (/Users/dgornstein/Workspace/ui-bootstrap-fork/node_modules/angular/angular.js:16231:39)
	    at new <anonymous> (/Users/dgornstein/Workspace/ui-bootstrap-fork/src/datepicker/datepicker.js:9:15478)
	    at invoke (/Users/dgornstein/Workspace/ui-bootstrap-fork/node_modules/angular/angular.js:4523:17)
	    at Object.instantiate (/Users/dgornstein/Workspace/ui-bootstrap-fork/node_modules/angular/angular.js:4531:27)
[...]

PR #5261, opened 2016-01-15
chore(datepicker): unregister parent watchers on $destroy

### LINKED ISSUE #5242: Unregister watches on the parent scope
Several watches are registered on the parent scope in the datepicker.js. Those watches are going to be unregistered the moment the parent scope will be destroyed. Wouldn't it be wiser to unregister them when the directive's scope is being destroyed?

```javascript
    if ($attrs[key]) {
      $scope.$parent.$watch($attrs[key], function(value) {
        self[key] = value ? angular.isDate(value) ? dateParser.fromTimezone(new Date(value), ngModelOptions.timezone) : new Date(dateFilter(value, 'medium')) : null;
        self.refreshView();
      });
    } else {
      self[key] = datepickerConfig[key] ? dateParser.fromTimezone(new Date(datepickerConfig[key]), ngModelOptions.timezone) : null;
    }
  });
```

wesleycho:
Closes #5242 

This also supercedes #5260


deeg:
Thanks for doing that!

wesleycho:
Forgot to specify this commit - closed by c4171a2de2c1f972c09b485650671f0248e8a170

PR #5262, opened 2016-01-15
fix(tabs): revert template change

### LINKED ISSUE #5254: Bootstrap themes no longer apply to ui-bootstrap elements
Hello

Up until 1.0 I could use something like Bootswatch to override the general theme on my page. If I upgrade to 1.0, the the uib templates have inline styling that override my CSS and hence adopt the default bootstrap theme.

Is this intended behavior? I cannot find anything about it in the documentation for breaking changes.

Edit: Unless it's related to this: "All virtual templates in UI Bootstrap now are prefixed with uib/ - if one is overriding the templates via $templateCache path or manually building the templates from the UI Bootstrap repository, one needs to change the string used in the $templateCache representation to have the new prefix". I'm not sure though?

wesleycho:
- Reverts template change to tab
- Adds warning if one desires more complex HTML inside the tab

This fixes #5254


icfantv:
LGTM after fixing tests.  Do accordions have the same issue?

wesleycho:
Not really, although we could also do the same reverting on the accordion - the only thing we lose is any accessibility benefits and the cursor: pointer style.

I'll open a PR for that.

PR #5263, opened 2016-01-15
refactor(accordion): change back to anchor element

wesleycho:
- Change back from div element to anchor element

BREAKING CHANGE: This changes the div element in the accordion-group
element back to an anchor element. If one wishes to use more complex
templates inside the accordion group heading, one must override the
template and add the `cursor: pointer` CSS targeting that element.


icfantv:
LGTM

PR #5264, opened 2016-01-15
Bug fix for `uib-datepicker-popup` as attribute

a-t:
The `uibDatepickerPopup` directive's settings `max-mode` and `min-mode` are expected to take AngularJS expressions as values. In particular it's mentioned in the [documentation](http://angular-ui.github.io/bootstrap/#/uib-datepicker-settings). It has worked fine before the version 1.0.0 of UI Bootstrap.

In the current version though if `uibDatepickerPopup` is used as a HTML tag attribute, datepicker breaks with an error on init. [Here](http://plnkr.co/edit/WCRfPr8EbPlpLJbo2cO2?p=preview) is a Plunker example.

```html
<input type="text" uib-datepicker-popup max-mode="month" ...>
```
The above code works, but is incorrect. The correct is like this:
```html
<input type="text" uib-datepicker-popup max-mode="'month'" ...>
```
but it causes an error in version 1.0.0 an above.


wesleycho:
cc @Foxandxss - you mentioned wanting to take a look here since there was a big problem you noticed.

Foxandxss:
I am confused. You talk about `maxDate` and `minDate` but your fix is for `minMode` and `maxMode` (also your plunker).

The mode one is expected to get a string like in your first example, not an expression.

a-t:
I'm sorry. Mode, not date of course. Edited above. The docs say expression. Look at the ($) sign near the setting description.

Foxandxss:
Oh, this attribute accept both modes. I plan some major rewrites but not sure if for today or a future version.

For the time being, that `$parse` on your PR is not needed at all.

a-t:
> this attribute accept both modes

Well, actually it _did_ before the version 1.0.0. But now it doesn't (see my [Plunker example](http://plnkr.co/edit/WCRfPr8EbPlpLJbo2cO2?p=preview)). The worst thing is that the backwards compatability is broken. Will a fix be provided?

Foxandxss:
We want to release tomorrow, so it will be fixed there.

a-t:
Thank you. Sorry for a duplicate issue.

PR #5265, opened 2016-01-15
Fix dynamic min-mode and max-mode in uib-datepicker-popup

luber:
Making it possible to dynamically change min-mode and max-mode properties on uib-datepicker-popup


wesleycho:
Revert the change to bower.json - there should not be a merge commit at all either.

Issue #5190, opened 2016-01-09
[Date picker] Calendar not rendering when initDate is undefined


bogdanalexe90:
Calendar is not rendering when `initDate` and model are `undefined`.

http://plnkr.co/edit/Dy9UvDb5z8vRt8phYypD?p=preview

wesleycho:
Looks like a bug (fixed the Plunker [here](http://plnkr.co/edit/1BsP29JRcaR15pKllUQd?p=preview)), but should be simple to fix by setting an initial model value of the current date.

wesleycho:
Linking this investigation with the inline datepicker [here](http://plnkr.co/edit/fSRjIy0cVDorWlT6Wwdp?p=preview) to note that this is only an issue with the popup version, so something in between is being set incorrectly.

PR #5266, opened 2016-01-15
fix(datepicker): init-date: check if parsed value is valid

AndriIushchuk:
Closes https://github.com/angular-ui/bootstrap/issues/5190

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -207,10 +207,12 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
         $log.warn('uib-datepicker initDate attribute usage is deprecated, use datepicker-options attribute instead');
       }
 
-      this.activeDate = dateParser.fromTimezone($scope.$parent.$eval($attrs.initDate), ngModelOptions.timezone) || new Date();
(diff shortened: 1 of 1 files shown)


wesleycho:
Can you take a look at what broke the unit test here in this change?

src/datepicker/datepicker.js
```diff
@@ -71,10 +71,12 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   $scope.uniqueId = 'datepicker-' + $scope.$id + '-' + Math.floor(Math.random() * 10000);
 
   if (angular.isDefined($attrs.initDate)) {
-    this.activeDate = dateParser.fromTimezone($scope.$parent.$eval($attrs.initDate), ngModelOptions.timezone) || new Date();
+    var initDate = dateParser.fromTimezone($scope.$parent.$eval($attrs.initDate), ngModelOptions.timezone);
+    this.activeDate = !isNaN(initDate) ? initDate : new Date();
```
wesleycho: All of this with the initDate and setting to this.activeDate can be abstracted out to a common helper function to avoid duplicated code.

AndriIushchuk: in this case `this.activeDate ` changes and below - `self.activeDate `

PR #5267, opened 2016-01-15
fixes TypeError: Cannot set property 'isOpen' of null

kwiateusz:
Sometimes $evalAsync is fired when scope is null and I get error in console. This fixes that specificc situation


wesleycho:
@RobJacobs is this possible when using $evalAsync? Seems strange to me...but not necessarily against this fix. Thoughts?

RobJacobs:
@wesleycho I agree, would be nice to see a plunk reproducing the problem or a description of how this occurred. I'm not against the change but there may be a better approach.

wesleycho:
I'll merge this for now, and perhaps an investigation could be done to figure out what is happening here - this seems like a perpetual problem :( .

PR #5268, opened 2016-01-15
fix(datepiker): dynamic min-mode and max-mode fix

### LINKED ISSUE #5240: uib-datepicker-popup: the min-mode attribute doesn't work when it is bound to the $scope property
Hi,
Found a problem with **min-mode** attribute when it's bound to the $scope.datePickerMode.
It works as expected when hardcoded directly in html.

I have reproduced it here:
http://plnkr.co/edit/hkavDudn6aZuAtRVsHKx?p=preview

luber:
- Reapply changes that was made in 0.3.13 to make min-mode and max-mode work correctly in popup

Closes #5240


wesleycho:
This LGTM - thanks for fixing up the commit history!

PR #5274, opened 2016-01-15
fix(datepicker): Fix destroy event for popup

### LINKED ISSUE #5058: datepicker-popup remains on screen when launched from modal that goes away
Open a datepicker-popup from a modal, with datepicker-append-to-body set to true. Click outside the modal to close it. The datepicker stays around even though its scope was destroyed.

See http://plnkr.co/edit/YdWmqYPFBNbv4vhQZc4t?p=preview

deeg:
Closes #5058 

[Plunker with fix](http://plnkr.co/edit/DwvlMHyqHSw3VfNTIf8h?p=preview)

Click destroy button to see popup go away on first input example.

The div keeps the reference to the element correct because it isn't added and removed to the dom as a clone due to ng-if


wesleycho:
This LGTM, but since I made the original suggestion, someone else needs to sign off.

template/datepicker/popup.html
```diff
@@ -1,10 +1,12 @@
-<ul class="uib-datepicker-popup dropdown-menu" dropdown-nested ng-if="isOpen" ng-style="{top: position.top+'px', left: position.left+'px'}" ng-keydown="keydown($event)" ng-click="$event.stopPropagation()">
-	<li ng-transclude></li>
-	<li ng-if="showButtonBar" class="uib-button-bar">
+<div>
```
icfantv: fix indent

deeg:
@icfantv, spaces should be all set. Sorry about that.

icfantv:
Ok, please walk me through how wrapping in a `<div>` fixes the issue.

wesleycho:
Just leaving for posterity, wrapping it in a `div` is a hack to prevent `popup[0]` in the popup from being a comment tag due to how `ng-if` works.

jonasem:
Thanks for clearing that up wesleycho, our code broke when updating deps due to a missing <div> in our version of the template. I would expect the need to rewrite to be listed as a breaking change though (the required <div>-wrapper part) as we had no clue that we had to update our template.

wesleycho:
Sorry about that, I should have marked this as a breaking change technically.

icfantv:
@wesleycho do we need to update the changelog?

wesleycho:
Technically, we probably should.

icfantv:
@jonasem, make sure you use the backtick characters (or escape the angle brackets with a backslash) around any code that could be interpreted by the editor as an actual DOM event.  This is why your `<div>` didn't actually show up.  I've edited your comment to escape them.

jonasem:
Thanks :-)

PR #5275, opened 2016-01-16
fix(datepicker): pass through null

### LINKED ISSUE #5238: Datepicker - max-date
When I put the max-date to null is not expected behaves.

Example: http://plnkr.co/edit/4hJWbKjKvD5NGIlDoZvJ?p=preview
The example shows that when the calendar is opened, the date that lets you select is the 31/12/1969 (1969-12-31)

The expectation is that it has no maximum limit. Since max-date is null.

wesleycho:
- When `minDate` or `maxDate` is `null` in the popup, pass through value to inline datepicker

This fixes #5238.


Foxandxss:
Sounds good to me.

wesleycho:
Just waiting on tests to pass on Travis to merge - should be good though.

Issue #5144, opened 2016-01-04
date.setTime doesn't update timepicker 'selected' variable


jackstrapp:
Hello, 

maybe it's a design choice more than a bug? you're probably gonna answer my question later.
here's the plunk:
http://plnkr.co/edit/QRXZaDWI5DWfQpIKvv1g

if i use the <code>date.setDate</code> or <code>date.setHours</code>, the timepicker isn't updated but if the variable actually gets a new reference by doing <code>date = new Date(bla bla)</code> it works well. You can see that my problem is that if i change the date then the time, the date modification is canceled.

Happy new Year!!!

wesleycho:
So this is something that would not be simple to detect without degrading performance - we would have to watch for changes in `[getHours(), getMinutes(), getSeconds()]`.

IMO if one wants to trigger the change via that method, one should instantiate a new date instance, i.e. `date = new Date(date)`.

What are your thoughts on this given that the workaround seems easy?

jackstrapp:
I aggree, the workaround is pretty easy.  I was just surprised of this behaviour and i thought this had to be mentionned. 

I don't know how to bring this out but this could be an interesting detail for other users.

Maybe a small note in the documentation?

PR #5278, opened 2016-01-16
docs(timepicker): add note about updating model

wesleycho:
- Add note on updating the model value programmatically

This addresses #5144.


Foxandxss:
LGTM

PR #5280, opened 2016-01-16
docs(datepicker): add clarifications for the popup

Foxandxss:
I hope people understand the `ng-model[-options]` as `ng-model` and `ng-model-options`.


src/datepicker/docs/readme.md
```diff
@@ -215,6 +219,9 @@ Options for the uib-datepicker must be passed as JSON using the `datepicker-opti
   _(Default: `yyyy-MM-dd`, Config: `datepickerConfig`)_ -
   The format for displayed dates. This string can take string literals by surrounding the value with single quotes, i.e. `yyyy-MM-dd h 'o\'clock'`.
 
+**Note:** With the exception of `ng-model[-options]` and `templateUrl`, you can configure the wrapped datepicker using its attributes in the popup as well.
+But beware **this possibility may be deprecated on the near future.**
```
wesleycho: We're not deprecating it now?

Foxandxss: Uhm, I don't know. We haven't even documented those options yet (that is what this PR covers) so I don't think we should deprecate it right away.

The main problem for me now is that our current implementation doesn't allow a nice way of **not
** using this attributes on the popup (you can use the -options but you lose the ability of changing the attributes dynamically).

So I think that until we lay down some infrastructure to let the users work without the attributes correctly, we cannot deprecate them.

wesleycho:
Works for me then - LGTM

Issue #4954, opened 2015-11-25
datepicker should support new angular $locale.DATETIME_FORMATS.FIRSTDAYOFWEEK parameter


yurigenin:
Angular's $locale service now supports FIRSTDAYOFWEEK parameter that is supposed to control which week of day appears first in the calendar (daypicker template in Angular UI). Currently, Sunday is always the first one even though most of the locales use Monday as the first day of the week. Please work this into your next release.

pkozlowski-opensource:
>  Please work this into your next release.

Sure, feel free to send us a PR

PR #5281, opened 2016-01-16
feat(datepicker): use $locale for starting day

wesleycho:
- Change to fallback to $locale.DATETIME_FORMATS.FIRSTDAYOFWEEK for
  first day column

This implements #4954.


Foxandxss:
Does the previous forEach need startingDay as well? And I would like a docs update.

wesleycho:
Forgot to remove that startingDay - good catch.

Foxandxss:
Works for me now.

wesleycho:
Updated PR to make this a non-breaking change - please re-review.

PR #5290, opened 2016-01-18
docs: add angular-animate to dependencies

Foxandxss:
Someone said that there is no indication of needing angular-animate, so I add it.


wesleycho:
LGTM

Issue #5229, opened 2016-01-13
Modal does not trap focus if shift+tab is immediately pressed


d-lane:
By opening a ui.bootstrap.modal which does not have an initially focused element and then pressing shift+tab the keyboard focus will not be trapped within the modal. For instance, if you press the button entitled 'Open me!' on the [demo page](https://angular-ui.github.io/bootstrap/) and press shift+tab a few times the keyboard focus will be behind the dialog.

wesleycho:
This seems strange, as we have code specifically for this situation, as well as a unit test against this.

wesleycho:
What is happening here is that the modal does not have focus when it is opened.

The question is, should the modal have focus? On research, the answer appears to be yes. PRs are welcome for moving the focus to the modal window element on instantiation - a simple unit test checking for focus is necessary. This also should not interfere with the `autofocus` attribute.

PR #5294, opened 2016-01-18
fix(modal): shift+tab without focused element within modal

AndriIushchuk:
Closes https://github.com/angular-ui/bootstrap/issues/5229

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -404,7 +404,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
               $modalStack.loadFocusElementList(modal);
               var focusChanged = false;
               if (evt.shiftKey) {
-                if ($modalStack.isFocusInFirstItem(evt)) {
+                if ($modalStack.isFocusInFirstItem(evt) || $modalStack.isModalFocused(evt, modal)) {
                   focusChanged = $modalStack.focusLastFocusableElement();
                 }
               } else {
@@ -543,6 +543,16 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         if (focusableElementList.length > 0) {
           focusableElementList[focusableElementList.length - 1].focus();
           return true;
+        }
+        return false;
+      };
+
+      $modalStack.isModalFocused = function(evt, modalWindow) {
+        if (evt && modalWindow) {
+          var modalDomEl = modalWindow.value.modalDomEl;
+          if (modalDomEl && modalDomEl.length) {
+            return (evt.target || evt.srcElement) === modalDomEl[0];
+          }
         }
         return false;
       };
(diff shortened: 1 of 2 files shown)


icfantv:
@AndriIushchuk, please fix the broken tests.  Thanks.

icfantv:
Also, just a tip, you can run the tests locally without committing them simply by executing the `grunt` command from the top level directory of the repo on your local system.

AndriIushchuk:
@icfantv thanks for the tip, tests fixed

icfantv:
Thanks.  I'll mark for review as I admit I don't know this code well enough to sign off on it.

src/modal/modal.js
```diff
@@ -547,6 +547,16 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         return false;
       };
 
+      $modalStack.isFocusInModal = function(evt, modalWindow) {
+        if (evt && modalWindow) {
+          var modalDomEl = modalWindow.value.modalDomEl;
+          if (modalDomEl && modalDomEl.length) {
+            return (evt.target || evt.srcElement) === modalDomEl[0];
```
wesleycho: This might need to check whether it contains the element using [contains](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains).

AndriIushchuk: But by using [contains](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains), method will return `true` for each tabbable element within modalWindow

wesleycho: The name of the function is not quite right if that is the case then.

AndriIushchuk: How about `$modalStack.isModalFocused`?

wesleycho: That sounds reasonable

src/modal/test/modal.spec.js
```diff
@@ -606,11 +606,15 @@ describe('$uibModal', function() {
         template:'<a href="#" id="tab-focus-link"><input type="text" id="tab-focus-input1"/><input type="text" id="tab-focus-input2"/>' +
         '<button id="tab-focus-button">Open me!</button>'
       });
+      $rootScope.$digest();
```
wesleycho: Is this extra $digest necessary? This seems a little suspicious to me.

AndriIushchuk: it's necessary to update focused element after modal opening

wesleycho:
Verified [here](http://plnkr.co/edit/DL337wA0tQP7O5j7FAE1?p=preview) this looks good - thanks for the fix!

Hardhik:
Hey, 

I can still this issue:
http://getbootstrap.com/javascript/ 

Click on "Launch demo modal" and try shift+tab. No action done. The website seems to use 3.3.7 at the time I am seeing this issue.
Also while the focus is one some of the tabbed element, shift tab works till the first element in the tab order and doesn't cycle back to the last element.

Is this expected? From which version onwards is this fix included?

PR #5297, opened 2016-01-18
Typeahead: Make active element always visible

### LINKED ISSUE #5023: Typeahead scroll to selected element with overflow
Also commented [here](https://github.com/angular-ui/bootstrap/pull/4463)

Downloaded and built the master repo (1.0 snapshot version ) and while the issue is fixed with keyboard-nav there is another bug left.

Scenario.
1) Have multiple suggestions available.
2) Suggestions must be the same, at least in the first letters which are typed. The rest characters are irrelevant
3) Define height with overflow scroll that is smaller than the total suggestions.
4) Scroll down to a suggestion and mouse over it or do the same with the keyboard
5) Type the letter sequence which exists in every/almost every suggestion in the list.

Typeahead should focus the first suggestion and scroll to that suggestion but it doesn't.
Demo:
http://codepen.io/Duskfall/pen/adowpL

For the demo to reproduce:
Type "t"
Scroll down 
Type "e"
You can see the problem

deeg:
Closes #5023.

[Plunker demonstrating the fix.](https://plnkr.co/edit/GPvS9FBnT4YneguOKauN?p=preview)

Also updated the rules for when to scroll the typeahead list. Before it would always keep the active one at the top. Now it is a bit smarter and either keeps it at the top or bottom depending on which way it is going 'off screen' in the typeahead list.

There is going to be one test failing. I am looking for advice on how to test scrollTop when testing within phantomjs. @wesleycho, @Foxandxss, ideas?

scrollTop is always 0. Even if I do in the test

```parentNode.scrollTop = 10; console.log(parentNode.scrollTop);```

the log is 0.


deeg:
Also now realizing this can break with custom templates. (last example on plunker) Will try to fix calculations even with custom template, but if not might have to disable this feature if using a custom template.

wesleycho:
My thoughts - position logic should be abstracted to $position. There may be logic that may be useful for the dropdown component. This also would the benefit of testing more cleanly in isolation.

deeg:
Sounds good, will try to clean this up in the next day or two.

deeg:
Going to close this, will re-open when I have some time to clean it up.

PR #5299, opened 2016-01-19
feat(timepicker): added ability to clear timepicker control

### LINKED ISSUE #5293: Can't clear timepicker without adding extra button
There is no ability to reset timepicker's model using keyboard, without adding extra button. Input value sets to '00' on blur.

asurinov:
Closes #5293

### DIFF
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -165,7 +165,7 @@ angular.module('ui.bootstrap.timepicker', [])
     var hours = +$scope.hours;
     var valid = $scope.showMeridian ? hours > 0 && hours < 13 :
       hours >= 0 && hours < 24;
-    if (!valid) {
+    if (!valid || $scope.hours === '') {
       return undefined;
     }
 
@@ -182,7 +182,11 @@ angular.module('ui.bootstrap.timepicker', [])
 
   function getMinutesFromTemplate() {
     var minutes = +$scope.minutes;
-    return minutes >= 0 && minutes < 60 ? minutes : undefined;
+    var valid = minutes >= 0 && minutes < 60;
+    if (!valid || $scope.minutes === '') {
+      return undefined;
+    }
+    return minutes;
   }
 
   function getSecondsFromTemplate() {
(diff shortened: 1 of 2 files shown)


wesleycho:
@Foxandxss what is your thoughts? I'm a little wary since this is a breaking change with UX ramifications.

Perhaps this should be hidden behind a config flag to prevent it from becoming a breaking change.

src/timepicker/timepicker.js
```diff
@@ -464,6 +474,12 @@ angular.module('ui.bootstrap.timepicker', [])
     return newDate;
   }
 
+  function modelIsEmpty(){
```
wesleycho: Add space before `{`.

src/timepicker/timepicker.js
```diff
@@ -165,7 +165,7 @@ angular.module('ui.bootstrap.timepicker', [])
     var hours = +$scope.hours;
     var valid = $scope.showMeridian ? hours > 0 && hours < 13 :
       hours >= 0 && hours < 24;
-    if (!valid) {
+    if (!valid || $scope.hours === '') {
```
icfantv: Just a tip, the empty string is falsey so this and all your other checks of this nature can be simplified.

wesleycho: This is fine - `valid` is a boolean, as can be seen from the two lines prior. I did do a double take too, but it checks out.

icfantv:
On one hand, I'm ok with this being a BC behind an optional config flag.  On the other, we talk about API bloat.  I'm on mobile right now so I can't look closely enough, but what if we let the user optionally specify their own validation function so it's totally up to them?

asurinov:
I think it's a normal behavior of any control to have an empty state and this state must be valid

src/timepicker/timepicker.js
icfantv: ahhh.  thanks.

icfantv:
@asurinov, while I tend to agree with that statement, we still need to balance feature vs. usability.  there is such a thing as too much control and if a component gets to that point, it becomes unwieldy and confusing and actually creates more problems than it solves.  See the datepicker for what I mean.

wesleycho:
After letting this sit a little while, I'm ok with this breaking change.

This PR LGTM - the only question left now is when should we schedule this change - 1.2.3 or 1.3.0?

asurinov:
So when will it be merged?

icfantv:
@asurinov, please remember we are all unpaid volunteers.  while i understand and can appreciate some level of impatience, please be a little more discerning with your word choice.  we'll get this merged.

wesleycho:
There was some objection by @Foxandxss on this PR, as to this particular feature, which is the primary reason for holding up this being merged.

Foxandxss:
I just don't like breaking changes for such feature, but also don't bloat the API. So I am not entirely sure but I don't mind it either.

wesleycho:
Alright, then I'll merge it - one could write a directive with an appropriate $parser/$formatter if one wants the previous behavior.

PR #5308, opened 2016-01-20
Make all versions the same level in the ChangeLog.

svetlyak40wt:
Please, make all versions the same level headers, because otherwise it seems that 1.0.1, 1.0.2 and 1.0.3 are subsections of 1.1.0.
And this is breaks automatic changelog parser at https://allmychanges.com/p/web/angular%20bootstrap/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 # [1.1.0](https://github.com/angular-ui/bootstrap/compare/v1.0.3...v1.1.0) (2016-01-18)
 
 
-### Bug Fixes
+## Bug Fixes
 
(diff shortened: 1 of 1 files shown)


icfantv:
Thanks for taking the time to do this.  Unfortunately, we switched changelog generators (or a version changed somewhere) and this got missed.

That said, you should rely on the format of the last few entries as the official format or this is going to break again with the next release.  We'll need to look at the changelog generator to see if/how we can control this.

Finally, I would be VERY wary of third party changelog parsers for exactly this reason - you will always (ok, usually always) get the correct information from our official copy of the changelog and we cannot guarantee the information hosted by any other site.

wesleycho:
This needs to make the appropriate changes to https://github.com/angular-ui/bootstrap/blob/master/misc/changelog.tpl.md as well.

icfantv:
Oh cool. So it is a template.  That makes things easier.

svetlyak40wt:
Should I reformat changelog to make all versions level 1 header and sections in description, level 2?
As I see, template uses these levels.

svetlyak40wt:
Ok, I've fixed formatting according to template.

icfantv:
This LGTM, but given the sheer number of changes I'd like another set of eyes.

wesleycho:
LGTM

svetlyak40wt:
Thank you, guys!

PR #5317, opened 2016-01-20
docs(accordion): add note about ng-class

wesleycho:
- Adds note about custom class usage for the accordion-group


icfantv:
LGTM

PR #5321, opened 2016-01-21
fix(uibModal): fix uibModal backdrop click bug

feyy:
when uibModalInstance without defining "result.then()", backdrop click not work

As the [demo](http://plnkr.co/edit/rQioJP58aNwpIvTiDhps?p=preview) see, the last button opening modal without defining the `modalInstance.result.then()`, then the "backdrop click" doesn't work right. 
Attention: Angular is v1.3.20.


src/modal/modal.js
```diff
@@ -130,7 +130,9 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
           if (modal && modal.value.backdrop && modal.value.backdrop !== 'static' && (evt.target === evt.currentTarget)) {
             evt.preventDefault();
             evt.stopPropagation();
-            $modalStack.dismiss(modal.key, 'backdrop click');
+            $rootScope.$apply(function() {
```
wesleycho: Instead of wrapping with $apply, just do
```js
$modalStack.dismiss(...);
$rootScope.$digest()
```

feyy:
｡◕‿◕｡I writed this referring to the 'Esc' key callback function.

wesleycho:
Just realized this was against 0.14.x - can you change your fix to be based on `master` and to `master?

Going to close this - the merge conflict to `master` is too great.

PR #5327, opened 2016-01-22
fix(modal): fix modal backdrop click bug

feyy:
when uibModalInstance without defining "result.then()", backdrop click not work, especially with angular 1.3.x

chaged for [pr#5321](https://github.com/angular-ui/bootstrap/pull/5321)


wesleycho:
Should note that we only support Angular 1.4+ now

wesleycho:
Wondering whether this is still an issue given that fact - this appears to work in 1.1.0, as seen [here](http://plnkr.co/edit/wj7Pwc2sD71nivKl7CzB?p=preview).

feyy:
It's work in 1.1.0, bug we still use angular 1.3.x with ui-bootstrap 0.14.x....

wesleycho:
We are not planning on doing another 0.14 release - your best bet would be patch this in your version of the library until you are ready to upgrade to 1.1.0 & angular 1.4. At the least, an angular 1.4 upgrade should not be difficult.

feyy:
ok, thanks!

PR #5334, opened 2016-01-23
fix(datepicker): allow using min/max mode/date with datepickerOptions

### LINKED ISSUE #5315: minDate/maxDate not working when passed via datepicker-options
When I use uib-datepicker-popup and pass minDate/maxDate as properties via datepicker-options I get this error:
```
[$parse:lexerr] Lexer Error: Unexpected next character  at columns 46-46 [ä] in expression [Wed Jan 20 2016 19:02:20 GMT+0100 (Mitteleuropäische Zeit)].
```
The problem is caused in this loop:
```
    if ($attrs[key]) {
      watchListeners.push($scope.$parent.$watch($attrs[key], function(value) {
...
      }));
    } else {
...
    }
  });
```
When using the min-date/max-date attributes, $attrs[key] resolves to `watchData.min/maxDate`. When using the way described above, $attrs[key] is the actual Date instance passed via the options.

Here's a fiddle:

https://jsfiddle.net/qz34osvr/

Foxandxss:
So I changed `datepickerOptions` to be processed on the scope directly, that way we can achieve two-way databinding for free.

Now, all the options that are not watchable are put directly and the watchables are put in a way that if they change, the change will be propagated.

As today, attrs overrides options, but that if another PR and another issue.

Fixes #5315


wesleycho:
LGTM

PR #5338, opened 2016-01-25
feat(accordion): use appropriate aria tags for accessibility

jeremytowne:
The id generation in accordion.js, seems a little sketchy - open to ideas about this.

### DIFF
--- a/src/accordion/accordion.js
+++ b/src/accordion/accordion.js
@@ -93,8 +93,8 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
       };
 
       var id = 'accordiongroup-' + scope.$id + '-' + Math.floor(Math.random() * 10000);
-      scope.headingId = id + '_tab';
-      scope.panelId = id + '_panel';
+      scope.headingId = id + '-tab';
+      scope.panelId = id + '-panel';
     }
   };
 })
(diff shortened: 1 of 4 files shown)


wesleycho:
Something similar to https://github.com/angular-ui/bootstrap/blob/3dac5e34504a391ff9e95eb1b0e9772bae9d7448/src/typeahead/typeahead.js#L113 would work I think for generating the id attribute - the Math.random usage is not the greatest solution, but is safe IMO.

jeremytowne:
Ok, I will update it to be similar to that. Should I be adding the aria tags like https://github.com/angular-ui/bootstrap/blob/3dac5e34504a391ff9e95eb1b0e9772bae9d7448/src/typeahead/typeahead.js#L114, or is putting them in the template OK?

wesleycho:
Template is fine

jeremytowne:
It looks like the travis ci build failed because of a timeout issue, is there anyway to run a new build against this?

icfantv:
Looks ok now.

src/accordion/accordion.js
```diff
@@ -91,6 +91,10 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
           }
         }
       };
+
+      var id = 'accordiongroup-' + scope.$id + '-' + Math.floor(Math.random() * 10000);
+      scope.headingId = id + '_tab';
+      scope.panelId = id + '_panel';
```
wesleycho: Can you switch these to `-` instead of `_`?

wesleycho:
Just that switch in those two ids to dashes & this should be good.

jeremytowne:
Awsome! Thanks for the feedback.

PR #5340, opened 2016-01-25
feat(datepicker): add datepickerOptions support

wesleycho:
- Add support for options object to concisely configure datepicker


src/datepicker/datepicker.js
```diff
@@ -31,66 +31,173 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   // Modes chain
   this.modes = ['day', 'month', 'year'];
 
-  // Interpolated configuration attributes
-  angular.forEach(['formatDay', 'formatMonth', 'formatYear', 'formatDayHeader', 'formatDayTitle', 'formatMonthTitle'], function(key) {
-    self[key] = angular.isDefined($attrs[key]) ? $interpolate($attrs[key])($scope.$parent) : datepickerConfig[key];
-  });
+  if ($attrs.datepickerOptions) {
+    var datepickerOptions = $scope.$parent.$eval($attrs.datepickerOptions);
```
wesleycho: There is a potential flaw with this - if the user breaks the object reference, this will have the old reference cached.

Foxandxss: Can't we simply put in on the scope object? Extra watch but fixes that.

wesleycho: Guess I could do `datepickerOptions: '="'`

src/datepicker/docs/readme.md
```diff
@@ -12,6 +12,29 @@ The datepicker has 3 modes:
 
 ### uib-datepicker settings
 
+* `datepicker-options`
```
Foxandxss: Alphabetical order please.

PR #5341, opened 2016-01-25
Dropdown performance: don't call removeClass for the initial load

jburson:
Dropdowns: Radically speed up the initial load when there are multiple dropdowns on a page, by only calling add/removeClass when the open state is actually changed.

@wesleycho


wesleycho:
LGTM - thanks for this improvement!

Issue #5349, opened 2016-01-26
Double click breaks modal


sime:
I've discovered that double clicking a button which toggles a modal open will execute undesirable event on the modal.

Examples:
1. Double clicking will click a button within a modal: http://plnkr.co/edit/ZLyjaysD2YlXWSTqfpvn?p=preview
2. Double clicking will click on the backdrop and close the modal before the user has had a chance to interact with it. http://plnkr.co/edit/xsR1ij6W8tp77EPhhmAZ?p=preview

Double clicking 'live demo modal' on the official Bootstrap docs does not behavior in the above mentioned way. http://getbootstrap.com/javascript/#live-demo

icfantv:
@sime, since this is all in user-land, there's not really anything we can do here as you, the caller of the API, control when the modal is opened.  You can easily do this by setting a flag when the button is clicked the first time and then ignoring it on subsequent times until the modal is closed (and the flag reset).  This is the same thing you would do on a submit/save button for a form.  I understand this is annoying, but as with the save/submit button - I don't think there's a whole lot we can do.

I'm going to close this as a wontfix, but I'm open to any ideas you might have as to how to solve this and am happy to reopen if we can figure something out that's clean.

sime:
I reluctantly accept that this is an user land problem.

But I don't understand your suggested workaround.

The button is not clicked twice, the second click is on the modal.

icfantv:
@sime, i'm sorry, I misunderstood you then.  I do understand now.  I don't know that there's a way for us to solve this in the library short of timing web application mouse clicks which, even then, I don't know would work and I would consider to be a hack.  I think the only real solution here is to have a flag around the modal open code.  I don't like the statefulness of this but I have no other ideas.

You might try posing a well-written question to StackOverflow to open this to a wider audience.

PR #5351, opened 2016-01-26
Delay binding of click for closing

sime:
Potential counter measure for #5349

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -161,9 +161,6 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
           }
         };
 
-        // moved from template to fix issue #2280
-        element.on('click', scope.close);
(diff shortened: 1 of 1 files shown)


src/modal/modal.js
```diff
@@ -176,6 +173,11 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
         attrs.$observe('modalRender', function(value) {
           if (value === 'true') {
             modalRenderDeferObj.resolve();
+
+            // moved from template to fix issue #2280
+            // and delay potential double click #5349
+            element.on('click', scope.close);
```
wesleycho: This has major issues since every time the value toggles true, it will trigger.

sime: Okay, thanks. I will look at other potential promises.

PR #5352, opened 2016-01-26
Added a way to get the number of opened windows

martimarkov:
Using the `$uibModalStack.getOpenedWindowsNum()` we can get the number of opened windows/modals. This is great if we want to integrate with the angular-material design and edit the dropdown menu's z-index using the same formula as in the modal in angular-ui.

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -524,6 +524,10 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
       $modalStack.getTop = function() {
         return openedWindows.top();
       };
+      
+      $modalStack.getOpenedWindowsNum = function() {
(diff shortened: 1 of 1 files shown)


wesleycho:
This needs a unit test.

martimarkov:
@wesleycho it only exposes a variable. why would that need any kind of unit testing. Surely it should have been done on the variable itself beforehand.

RobJacobs:
I would prefer this be called openedCount

wesleycho:
Any method exposed publicly needs a unit test as a guarantee the API behaves as expected. Without one, this will be rejected.

wesleycho:
Going to close this due to non-response - feel free to open a new PR with the appropriate test

PR #5355, opened 2016-01-27
feat(datepicker): pass options from popup to inline

wesleycho:
- Pass through options object wholesale from popup to inline datepicker

BREAKING CHANGE: This breaks any snake-cased key usage, i.e.
`show-weeks`. Convert all keys used to camelCase.


Foxandxss:
What is the reasoning of having datepickerMode in a separate way?

wesleycho:
The reason is because it is already optionally two-way bound in the inline datepicker as a separate attribute. Probably need to clean that up, but doing that now would be a slightly bigger change.

I'm fine with making it - what do you think?

Foxandxss:
What I don't see is how the inline datepicker picks that mode from how you put it here.

wesleycho:
It's mainly meant for reading

Foxandxss:
LGTM

PR #5358, opened 2016-01-28
fix(typeahead): add guard for no $viewValue

### LINKED ISSUE #5357: [typeahead]when I use 'template-url' with 'typeahead-show-hint' makes conflict.
Here is [plnkr](http://plnkr.co/edit/Iz4yxKJRxMwnKQfhVeM2?p=preview).

when i click to try to use typeahead.

makes a some error.

`Cannot read property 'length' of undefined , ui-bootstrap-tpls-1.1.1.js:6049`

I think the error because of  'template-url' not load yet. but show-hint option going to show some data before load finishing. 

so i fix my code.

`  if (showHint) {var firstLabel = scope.matches[0].label;
  if(!inputValue)return;......}`

wesleycho:
- Add guard when $viewValue is undefined, and when showHint & min-length
  0 is used

Plunker with fix in action [here](http://plnkr.co/edit/btC8b63t0qwIcUJKGQS3?p=preview)

This fixes #5357.


deeg:
This LGTM

PR #5359, opened 2016-01-28
docs(carousel): make sure request is made to http

wesleycho:
- Ensure request uses http protocol and not https potentially


deeg:
Ship it!

PR #5360, opened 2016-01-28
fix(tooltip): Remove the $locationChangeSuccess close behaviour

### LINKED ISSUE #5337: BUG: Popover closes itself even though popover-is-open is set to true
Following are the options set:
```
    uib-popover="I am activated manually" 
    popover-is-open="true" 
    popover-trigger="none"
    popover-append-to-body="true"
```

When a route change occurs(ui-router) the popover closes itself.

- Version Info:
- Angular UI Bootstrap version: 1.1.0
- UI-Router version: 0.2.15
- Angularjs version: 1.4.7

Demonstration in [PLUNKER](http://plnkr.co/edit/tNwexNKsznSucoZrYxxH?p=preview)

vai:
This fixes #5337 

Removed the code for handling location change, and removed the corresponding test.

### DIFF
--- a/src/tooltip/test/tooltip.spec.js
+++ b/src/tooltip/test/tooltip.spec.js
@@ -998,25 +998,6 @@ describe('$uibTooltipProvider', function() {
       expect($body.children().length).toEqual(bodyLength);
     }));
 
-    it('should close on location change', inject(function($rootScope, $compile) {
-      elmBody = angular.element(
-        '<div><span uib-tooltip="tooltip text">Selector Text</span></div>'
-      );
-
-      scope = $rootScope;
-      $compile(elmBody)(scope);
-      scope.$digest();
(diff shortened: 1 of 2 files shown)


Foxandxss:
Weird, I checked that to be the problem on that issue and was negative for me. I tried this PR with the plunker and works: http://plnkr.co/edit/EF0uJQijkuBt3InRNkUW?p=preview

Clearly I checked wrong.

wesleycho:
On the original Plunker, if you click the link, the popover will go away - this PR fixes it. It may have been an old hack to deal with some garbage collection issue that has since been fixed.

Foxandxss:
yeah yeah, I checked if clicking on the new state would trigger that code and it didnt so I thought it wasn't the problem. Now with it gone, it works as expected. Weird but LGTM.

PR #5363, opened 2016-01-29
add support for changing typeaheadMinLength

germannj:
One might want to change minLength no of characters condition based. 
With the old implementation minLength is not changed when typeaheadMinLength is changed from outside, now with a watch it does change.

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -38,7 +38,14 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
     if (!minLength && minLength !== 0) {
       minLength = 1;
     }
-
+    originalScope.$watch(attrs.typeaheadMinLength, function (newVal) {
+        if (!newVal && newVal !== 0) {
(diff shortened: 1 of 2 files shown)


src/typeahead/typeahead.js
```diff
@@ -38,7 +38,14 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
     if (!minLength && minLength !== 0) {
       minLength = 1;
     }
-
+    originalScope.$watch(attrs.typeaheadMinLength, function (newVal) {
+        if (!newVal && newVal !== 0) {
```
wesleycho: Can just make a ternary, i.e.
```js
minLength = !newVal && newVal !== 0 ? 1 : newVal;
```

wesleycho:
This needs tests and a docs update.

germannj:
Test was added.
I think the documentation is already ok.

> typeahead-min-length $ (Default: 1) - Minimal no of characters that needs to be entered before typeahead kicks-in. Must be greater than or equal to 0.

Issue #5373, opened 2016-02-01
Progressbar modifies outside object of max attribute


MartinNuc:
Today we dealt with issue that in one objects in our app appeared value `100` which triggered our watch although data were not loaded from server yet. We tracked this to `progressbar` where property of this object was used in `max` attribute. **Progressbar modified object provided in `max` attribute.**

Here is plunker reproducing issue: http://embed.plnkr.co/m5f9yz/

Progress bar uses object *loaded* from the server to set value and max. However referenced object is modified by default parameter (100) of progress bar until data are loaded from server.

Expected behaviour: Object used in max attribute should not be modified. Progress bar should use internal value to represent current max value.

PR #5374, opened 2016-02-01
fix(progressbar): Progressbar doesnt modify object provided in max attribute

MartinNuc:
related issue: https://github.com/angular-ui/bootstrap/issues/5373


icfantv:
@MartinNuc, in general, it's a good idea to run the tests locally before committing anything to make sure you didn't mess something up...which is the case here.  You can do that simply by running the `grunt` task from your root project folder.

MartinNuc:
@icfantv My mistake. I had tests running localy but I made some changes after I made PR and forgot to run them again. Should be ok now.

wesleycho:
This is subtle - good catch!

PR #5375, opened 2016-02-01
chore: add support for webpack

Foxandxss:
This should support both webpack and browserify (I haven't tried the latter so I appreciate a test).


wesleycho:
LGTM - this should be really nice for end users.

icfantv:
Woot!

PR #5376, opened 2016-02-01
fix(datepicker): fix validation for M! & d!

### LINKED ISSUE #5225: M!/d!/yyyy does not Validate in the Dateparser Example
http://angular-ui.github.io/bootstrap/

Format defined: M!/d!/yyyy

Failed
image

Expecting:
image

wesleycho:
- Fix display & validation for M! & d!

This fixes #5225 (tested on local docs).


Foxandxss:
LGTM

timonwong:
@Foxandxss @wesleycho 
I don't know whether or not this is a bug, it seems that it will drop the text after last (right-most) formatting token, for example:
```
dateParser.filter(date, 'yyyy -');
```
returns only "2016", dropped " -", is that intended?

and:

```
dateParser.filter(date, 'yyyy - MM -');
```
returns only "2016 - 02".

wesleycho:
Hm, that sounds like a bug - can you open a new issue on that?

timonwong:
@wesleycho 
Ok: #5387

PR #5383, opened 2016-02-01
chore: add lodash as devDependency

Foxandxss:
I used lodash that was a dependency someone pulled (grunt) but I am using npm3 so lodash was on the root. Better to add it as a proper dependency.

Also my "npm i -S lodash" decided to tweak the package.json, meh, I like the changes tho.


wesleycho:
Works for me

PR #5385, opened 2016-02-02
fix(dateparser): append end of format

### LINKED ISSUE #5387: dateParser.filter issue which drop text after right most formatting token
After #5376, it seems that now it will drop the text after last (right-most) formatting token, for example:
```
dateParser.filter(date, 'yyyy -');
```
returns only `"2016"`, without `" -"`
and:
```
dateParser.filter(date, 'yyyy - MM -');
```
returns only `"2016 - 02"`.

http://plnkr.co/edit/h7lgrX2mpHNTRHSEOvyD?p=preview

wesleycho:
- Ensure trailing format string is appended to generated date string

This fixes #5387.


Foxandxss:
LGTM

Issue #5392, opened 2016-02-02
extra space next to the month selector


kevinlbatchelor:
In chrome, there is extra space on the left side of the right arrow, next to the month selector.

icfantv:
I'm not seeing this on [our demo page](http://angular-ui.github.io/bootstrap/#/datepicker).  Can you post a screenshot along with your code snippet?

kevinlbatchelor:
I'm just looking at your demo page, not my own code.

![space](https://cloud.githubusercontent.com/assets/5952444/12756291/60821062-c990-11e5-8f11-bcc785e41eb6.PNG)

icfantv:
Oh.  On the left side of the right arrow _button_.  Thanks.

icfantv:
@wesleycho, I think this is a browser bug.  Chrome is allocating an extra 2px to the `<th>` cell.  We have no styling here and there's nothing I can see that we can do.

RobJacobs:
It's the pull-right class on the template [here](https://github.com/angular-ui/bootstrap/blob/master/template/datepicker/month.html#L6)  Year, Month, and Day templates all have the pull-right class on the right button.

wesleycho:
I think the fix for this should be
```css
.uib-datepicker .uib-right {
  width: 100%;
}
```

Would need to test across browsers though.

icfantv:
That removes all the space entirely.  Is that what we want?

agrahamlincoln:
```
table.uib-daypicker button.uib-right {
    width: 100%;
}
```
seems to fix it for all of the ocurrences

icfantv:
It needs to be
```
.uib-datepicker .uib-right, .uib-datepicker .uib-left {
  width: 100%;
}
```

or the spacing on the left won't match.

agrahamlincoln:
in that case, why not 
```css
.uib-left, .uib-right {
  width: 100%
}
```

Both CSS classes are only used in the template, I see no reason to complicate it

wesleycho:
I think the concern is if we ever use those classes for another component. I'm not hugely opinionated on which in this case.

PR #5393, opened 2016-02-02
Add css class per issue #5392

agrahamlincoln:
Expands width of both arrow selectors for the top of the datepicker

See Issue #5392 for more details.


wesleycho:
LGTM - any objections @icfantv ?

icfantv:
None here.  As you pointed out on slack we should test cross browser/platform just to make sure.

PR #5394, opened 2016-02-02
docs: add info to CSP CSS

jengeb:
See, e.g. analogous information about CSP at https://docs.angularjs.org/api/ng/directive/ngCsp


wesleycho:
LGTM, thanks for this!

wesleycho:
Hm, can you also modify https://github.com/angular-ui/bootstrap/blob/master/misc/demo/index.html ?

jengeb:
See #5397.

PR #5401, opened 2016-02-03
fix(modal): fix modal rendered promise

### LINKED ISSUE #5331: Closing modal quickly leaves modal-open style on body
Exists in: angular-ui-bootstrap 1.1.0 + Angular 1.4.8, 1.4.9, 1.5.0-rc.1, whether ngAnimate is present or not.
Does not exist in at least: angular-ui-bootstrap 0.11.0 + Angular 1.2.28 without ngAnimate

Closing the modal before the open animation finishes leaves the modal-open style on body, which eliminates the scroll bar.

http://plnkr.co/edit/3712x2?p=preview

I discovered this regression when upgrading from Angular 1.2.28 and angular-ui-bootstrap to the latest. The presence (or not) of ngAnimate does not seem to matter.

One workaround is to use $timeout to close, but that is brittle and it would have visible artifacts, and I have to just guess at the right amount of time to wait.

Relevant pattern:

```js
		var mi = $uibModal.open({
			//whatever
		});

		//very shortly after:
		mi.opened.then(function() {
			mi.close();
		});
```

Here is my use case. I issue an AJAX call to perform a potentially long operation in my app. After 500ms, if the operation is not complete I open a "please wait" modal. Once the operation completes I close the modal as seen above (calling close won't work if the modal isn't finished opening yet, so I use the promise). However if operation completes in a total of 501ms for example, the modal is requested but then I want to close it right away.

wesleycho:
- Fix timing of rendered promise to be after animation is complete

Fix in action [here](http://plnkr.co/edit/0gG0Im?p=preview).

This fixes #5331


Foxandxss:
LGTM

PR #5410, opened 2016-02-03
Update index.html

ndelvalle:
Added uib prefix to $modalInstance


wesleycho:
Can you open a second PR fixing it in https://github.com/angular-ui/bootstrap/blob/master/src/modal/docs/demo.js ?

ndelvalle:
:+1:

PR #5419, opened 2016-02-05
test(tab): fix ng-repeat select twice callback

RobJacobs:
The select callback for the ng-repeat select twice test was bound to
the scope callback instead of the select property in the tabs
array.


wesleycho:
So this changes behavior when it comes to angular 1.4 - the callback will not fire in that situation.

What are your thoughts on this @Foxandxss ?

Issue #759, opened 2013-08-05
Trigger typeahead manually? Is that possible?


artworkad:
In my app the user may send a request for his position. When the position information is retrieved the data is populated to an input field which has the typeahead directive

```
document.getElementById('city').value = input;
```

But I want the typeahead to start and to search if the city is available in the typeahead resource. Is this possible?

segevofer:
focus event was a overhead for me, and i used click event - because focus occurred also after selection.. 
(i also overloaded the 'typeahead' directive itself)

app.directive('typeahead', function () {
    return {
        require: 'ngModel',
        link: function (scope, element, attr, ctrl) {
            element.bind('click', function () {
                ctrl.$setViewValue();
            });
        }
    };
});

segevofer:
focus event was a overhead for me, and i used click event - because focus occurred also after selection.. 
(i also overloaded the 'typeahead' directive itself)

```javascript
app.directive('typeahead', function () {
    return {
        require: 'ngModel',
        link: function (scope, element, attr, ctrl) {
            element.bind('click', function () {
                ctrl.$setViewValue();
            });
        }
    };
});
```

jbmyid:
I tried above but could not make it work

robbse:
For everyone having the same issue, the following example works for me.

```javascript
angular.module('app').directive('typeahead', typeaheadWatchChanges);

  typeaheadWatchChanges.$inject = ["$timeout"];

  function typeaheadWatchChanges($timeout) {
    return {
      require: "ngModel",
      link: function (scope, element, attr, controller) {
        scope.$watch('vm.model', function (value) {
          if (value) {
            element.focus();
            var old = controller.$viewValue;
            controller.$setViewValue('');
            $timeout(function () {
              controller.$setViewValue(old);
            });
          }
        });
      }
    };
  }
```

jusopi:
I finally got this working after quite a bit of hassle with the aforementioned proposed solutions.  These were the scenarios that I had to have this working for:

- user clicks inside input to trigger opening of dropdown for:
  - existing values on model
  - when model is undefined/blank
  - when existing model value is a number

Sorry this is in Coffeescript.... should be self explanatory however.

```
.directive 'typeaheadClickOpen', ($parse, $timeout)->
    dir =
        restrict: 'A'
        require: 'ngModel'
        link: ($scope, elem, attrs)->

            triggerFunc = (evt)->
                ctrl = elem.controller 'ngModel'
[...]

PR #5427, opened 2016-02-07
Added reopen-on-click typeahead option (Issue #759)

kazeno:
Added code to allow to reopen the typeahead list on click, as per issue #759. This requires adding the "typeahead-reopen-on-click" boolean attribute to the typeahead input, example:
<input type="text" uib-typeahead="obj for obj in objs" typeahead-reopen-on-click="true">

Only works on typeahead-editable="true" since otherwise the view value is erased on blur. The `element[0] === evt.target` comparison makes sure only the clicked typeahead's list is reopened.


piyush-dholariya:
Hi All,
I want to use Typeahead functionality to get ajax call on time interval but I do not want to show list of items's pop up. How can I prevent to show that popup?

PR #5428, opened 2016-02-07
Added reopen-on-click typeahead option (Issue #759) 

kazeno:
Added code to allow to reopen the typeahead list on click, as per issue #759. This requires adding the "typeahead-reopen-on-click" boolean attribute to the typeahead input, example:
`<input type="text" uib-typeahead="obj for obj in objs" typeahead-reopen-on-click="true">`

Only works on typeahead-editable="true" since otherwise the view value is erased on blur. The `element[0] === evt.target` comparison makes sure only the clicked typeahead's list is reopened.

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -81,6 +81,8 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
 
     var showHint = originalScope.$eval(attrs.typeaheadShowHint) || false;
 
+    var reopenOnClick = originalScope.$eval(attrs.typeaheadReopenOnClick) || false;
+
     //INTERNAL VARIABLES
 
     //model setter executed upon match selection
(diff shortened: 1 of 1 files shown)


wesleycho:
This can't be done currently with `min-length="0"`?

kazeno:
No, min-length only determines how many characters are needed for the typeahead to make a query. Even if it is 0, normally the typeahead list will not reopen until the input text is modified.

wesleycho:
What about programmatically controlling it with `typeahead-is-open`?

kazeno:
`typeahead-is-open` is not bound two-way, changing it programmatically doesn't affect the typeahead

wesleycho:
My preference would be for a PR implementing that as a two-way bound variable - that would then make any custom programmatic opening/closing possible on the user side.

kazeno:
That seems more complicated to implement, but sounds like a great idea. I'll see if I can make it work.

AntiAsko:
Is this on the npm repository?

deeg:
@AntiAsko, this is not currently on npm as it is an open PR which has not been merged.

The PR still needs work before it can be merged. Feel free to take a look through the comments for the proposed changes, and finishing up the PR if you have time so that we can get it merged in.

shyamal890:
Any progress on this. Much needed!

`typeahed-min-length = 0` and `typeahead-is-open` doesn't work when a user types something like 'a' in the input clicks outside of the element> Dropdown closes > User again clicks on the typeahead input > NO RESULT!

damianohcm:
Same here. Would love the typeahead-reopen-on-click

Issue #5429, opened 2016-02-07
Link to commit conventions broken


amahfouz:
The link at the bottom of the main page at the end of the instructions for contributing is broken and yields a 404.

wesleycho:
This has been reported to us before - the link should be changed to https://github.com/stevemao/conventional-changelog-angular/blob/master/convention.md

PR #5434, opened 2016-02-07
docs(contributing): fix conventions link

zachlysobey:
Addresses #5429

That issue mentioned a "link at the bottom of the main page" as well, but wasn't easily able to figure out where that was.


wesleycho:
I think this was it - thanks!

PR #5437, opened 2016-02-08
chore: update with version support

### LINKED ISSUE #5436: Support Angular 1.5
Angular 1.5 is now stable, we should test against it. If it already works, we should indicate it in the readme.

wesleycho:
- Update to note support for 1.5
- Bump up the ngversion for Plunker

This closes #5436.


Foxandxss:
There is something I wanted to do and this PR is perfect. Lock dependencies in package.json so we don't get caught into random update issues anymore.

wesleycho:
What should I modify the package.json to? We're fairly loose with the dependency versioning there, since some users like to test the edges of Angular as it's released.

Foxandxss:
what we put on the package.json  is irrelevant to those users because those are devDependencies.

I vote for hard locking it to "1.5.0" instead of "^1.x.x" that way travis won't use an updated angular when it gets out and freaking us out.

wesleycho:
I'm still confused - where do we want to do this?

Issue #4836, opened 2015-11-07
Investigate fixing the tab api


wesleycho:
We need to make a breaking change to fix how the tab API is handled. It is a much more awkward component to use than the accordion, which allows for more use cases.

In particular, we have bugs against the select callback api, and the active state should be kept at a higher level since we can only have one tab active at any given time.

API needs to
- [x] Expose active state separate from the tab object and into a single two-way bound value
- [x] Select in a more flexible manner
- [x] Allow passing in of `event` into select callback
- [x] Handle dynamically generated tabs gracefully with selection of first tab
- [x] Handle ng-if, ng-show, ng-hide gracefully
- [ ] Allow animate support with showing/hiding new tab (investigate content transcluding re: #4339)

chenyuzhcy:
I have fix for one of the issue, should I just creat a pr and refer to the item or should I try to fix all of them and creat one pr for all?

icfantv:
Using one PR for each and referencing this issue is fine - we won't close the issue until all the checkboxes are addressed.

We use the checkboxes in the description allow us to track progress separately and we've done this for other issues - adding the `uib-` prefix was one such issue.

daniel-jann:
I guess "Allow passing in of event into select callback" was meant to address #2715 (A way to intercept tab change), but #2715 also mentions being able to prevent tab selection (I believe that's not been addressed yet, has it ?)

icfantv:
@daniel-jann, #5720 was entered for this feature.

wesleycho:
[Here](http://plnkr.co/edit/EAlCBsdE6fJLizNBA5NO?p=preview) is a demonstration of being able to animate using the current api.

Closing this as resolved.

PR #5438, opened 2016-02-08
feat(tab): add select expressions

wesleycho:
- Add support for event and tab being available in select/deselect
  callbacks

This implements the third task of #4836.


Foxandxss:
Tests? :P

wesleycho:
I tried adding some tests, but it actually ended up locking up Karma - I think I might have to get clever about how to check the select/deselect arguments (`jasmine.any(Object)` caused issues, and I suspect a direct $scope comparison would too - might need to do a $scope.$id comparison).

wesleycho:
I'm not sure this is testable - having circular reference issues since an object containing a $scope is passed into one of the arguments, jasmine seems to handle this poorly.

wesleycho:
Removed $tab expression support, left as just $event support.

icfantv:
Do we want a test for this?  If not, LGTM.

wesleycho:
Updated with expectations!

deeg:
LGTM

PR #5442, opened 2016-02-08
refactor(paging): remove extra $parse

Foxandxss:
$watch already $parse the parameter, there is no need to re-$parse.


wesleycho:
LGTM

Issue #5414, opened 2016-02-04
The dropdown typeahead box is NOT right justified


anasqadrei:
My website is in an RTL language. My `<body>` is right justified `<body dir="rtl" class="text-right">`, but the typeahead dropdown box is always aligned to the left. I want it to be aligned/justified to the right. Here is a plunker to an example http://plnkr.co/fs79igc1fJWTqY4xrWTt. Type "on" or "one" in the text box.

icfantv:
Technically this is a duplicate of #2753 and superseded by #4762, but comments weren't followed up on.  Need to decide whether or not to tackle this separately.

icfantv:
@RobJacobs, do you have a sense of whether or not this is a simple fix involving a check for RTL and placing the dropdown appropriately?

RobJacobs:
@icfantv Is this in regards to the text displayed in the dropdown options or the popup not getting right aligned?  I've been working on adding the smart positioning to the datepicker and dropdown popup, but that would not cover the text in the options.

icfantv:
I'm pretty sure it's just the drop down, but since I can't read Arabic I don't know for sure.  I'd need to check with an RTL font with which I'm familiar.  I'll see if u can check with Hebrew this morning.

wesleycho:
Couldn't one override the alignment with user CSS?

evilaliv3:
@anasqadrei have you managed to solve this somehow?

for sure using https://github.com/MohammadYounes/RTL-bootstrap/tree/inline-rtl would fix some of the issues; but the tooltip.js implemented in angular-ui remains the main problem.

anasqadrei:
@evilaliv3 the way I've solved it is by hard coding right alignment in my version of ui-bootstrap. That wasn't a good solution and that's why I'm asking for ui-bootstrap to be alignment aware.

wesleycho:
Ah, I see what the problem is - this needs to be modified in the $uibPosition service. This likely also requires configuration in the typeahead for this.

evilaliv3:
Related to https://github.com/angular-ui/bootstrap/issues/4762

is some of you aware of any workaround?

anasqadrei:
@evilaliv3 My website is in an RTL language (Arabic) so I just hardcoded text align to be right somewhere in ui-bootstrap-custom-tpls.js file

anasqadrei:
[...]

PR #5444, opened 2016-02-08
feat(datepicker): implement auto position

RobJacobs:
Added popup-placement attribute that implements the auto position
feature that exists in the tooltip directive.  Accepts the same
placement options as the tooltip.

This [plunk](http://plnkr.co/edit/aXDt2IIE0J91JetH6iU9?p=preview) demos the feature.

I haven't added test for this as I'm not sure on what the best approach is for finding an elements position in a test and would like some guidance on that.  The 2 minor changes in the readme.md is a result of my editor trimming trailing whitespace.

If this approach is acceptable, I plan on doing the same thing to the typeahead directive.


icfantv:
Would it make sense to address #5414 with this commit as well?  That way you can work out the kinks before moving to typeahead and dropdown?

Linking in #4762 for tracking purposes.

Foxandxss:
Sounds good to me.

wesleycho:
This LGTM too

wesleycho:
Hm, I guess I forgot about the test part - how much testing does this need? We could perhaps test given certain CSS scenarios with regards to maybe display: block, position: relative/absolute, and floats.

Tooltip/popover would need them too.

RobJacobs:
@wesleycho I'm working on it still, will create another PR in the near future.

PR #5447, opened 2016-02-08
chore: lock angular version

Foxandxss:
This are the dev dependencies so it doesn't affect the users that wants to try the latest version.

Since travis will do a `npm install` after test, it will install latest angular (latest 1.x version, AKA any of them) and that gives us lot of headaches.

I think it is better to lock it and update it by hand each release with all the needed fixes.

WDYT?


RobJacobs:
Yes please.

PR #5451, opened 2016-02-09
fix(buttons): make uib-uncheckable work as expected and documented

### LINKED ISSUE #5412: Bug: uib-uncheckable not working on radio button directive
I found this using your demo page: https://angular-ui.github.io/bootstrap/#/buttons

The right button in the right group is never rendered uncheckable.  

Examining/tracing through the code, I think I can explain why it is not working and suggest a solution.
Lines 45-49 of buttons.js are intended to remove the `uncheckable` attribute from the element if `uib-uncheckable` evaluates to false.  However, removing an attribute using `attrs.$set(uncheckable, null) does not seem to work.  It looks like the angular function `$set()` leaves the attribute key/value of `uncheckable="uncheckable"` on the `attrs` object.  In line 37, a check is made to see if `uncheckable` is present using `angular.isDefined(attrs.uncheckable)`.  This seems to always returns true because `uncheckable="uncheckable"` evaluates to `uncheckable=null`.

You might need to remove the attribute from the element directly using `removeAttribute()` instead of using the angular function `$set()`.

Let me know how I can help.

rasenplanscher:
Closes #5412

### DIFF
--- a/src/buttons/test/buttons.spec.js
+++ b/src/buttons/test/buttons.spec.js
@@ -328,12 +328,38 @@ describe('buttons', function() {
         $scope.uncheckable = false;
         var btns = compileButtons('<button ng-model="model" uib-btn-radio="1">click1</button><button ng-model="model" uib-btn-radio="2" uib-uncheckable="uncheckable">click2</button>', $scope);
         expect(btns.eq(0).attr('uncheckable')).toBeUndefined();
-        expect(btns.eq(0).attr('uncheckable')).toBeUndefined();
+        expect(btns.eq(1).attr('uncheckable')).toBeUndefined();
+
+        expect($scope.model).toBeUndefined();
+
+        btns.eq(0).click();
+        expect($scope.model).toEqual(1);
+
+        btns.eq(0).click();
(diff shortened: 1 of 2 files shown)


src/buttons/test/buttons.spec.js
```diff
@@ -328,12 +328,38 @@ describe('buttons', function() {
...
         expect(btns.eq(0).attr('uncheckable')).toBeUndefined();
-        expect(btns.eq(0).attr('uncheckable')).toBeUndefined();
+        expect(btns.eq(1).attr('uncheckable')).toBeUndefined();
+
+        expect($scope.model).toBeUndefined();
+
+        btns.eq(0).click();
+        expect($scope.model).toEqual(1);
+
+        btns.eq(0).click();
```
wesleycho: Why does this need to be clicked twice in a row?

rasenplanscher: That is to assert that, with `uncheckable` not being set, clicking the same button twice is the same as clicking it once. Otherwise, it might just always be uncheckable irrespective of the state of `uncheckable`.

PR #5462, opened 2016-02-11
Docs: `error` should be `danger` in ProgressBar

camjc:
See:
http://getbootstrap.com/components/#progress
and also the examples.
Error doesn't do anything, so its still a blue progress bar.


wesleycho:
LGTM - thanks for the fix!

PR #5475, opened 2016-02-14
represent Persian Calendar in Datepicker

rkeshmir:
It could be a Persian Datepicker like the Gregorian one. All we need to add is a persian date-parser and a new option field to set the calendar (for ex. 'Persian' or 'Gregorian')


wesleycho:
The history looks busted on this - also this is against the wrong branch. Feel free to open another PR appropriately against `master`.

rkeshmir:
Can you help me about opening a PR against <code>maseter</code> ?

PR #5483, opened 2016-02-16
fix(modal): fix race condition with openedClass

tjokimie:
When `modal-window` entering animation runs after the modal has closed `openedClass` is left to `body` element.

This pull request fixes the issue by checking if modal closing has been scheduled after the animation has finished.


tjokimie:
This issue has only occurred on Firefox (v. 43.0.4) if modal is fired on application startup and closed almost immediately after.

wesleycho:
This LGTM - subtle, good catch!

PR #5488, opened 2016-02-16
fix(tabs): adding bootstrap 4 specific class

uxtx:
adding `nav-link` class to tab items to future-proof  for bootstrap 4: docs here: 
http://v4-alpha.getbootstrap.com/components/navs/#tabs

As far as I have looked, this class should not interfere or collide with any existing bootstrap 3 classes, and makes tabs usable for bootstrap 4.

### DIFF
--- a/src/tabs/test/tabs.spec.js
+++ b/src/tabs/test/tabs.spec.js
@@ -56,6 +56,9 @@ describe('tabs', function() {
     it('should pass class and other attributes on to tab template', function() {
       expect(elm).toHaveClass('hello');
       expect(elm.attr('data-pizza')).toBe('pepperoni');
+      //Ensure that we have bootstrap 4 link class so things are future proofed.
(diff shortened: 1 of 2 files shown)


src/tabs/test/tabs.spec.js
```diff
@@ -56,6 +56,10 @@ describe('tabs', function() {
     it('should pass class and other attributes on to tab template', function() {
       expect(elm).toHaveClass('hello');
       expect(elm.attr('data-pizza')).toBe('pepperoni');
+      // ensure that we have our bootstrap 4 link prefix so things are future
+      // proofed.
+      var link  = $(elm.find('a')[0]);
```
wesleycho: Remove extra space

uxtx: Thanks, On it.

PR #5425, opened 2016-02-06
refactor(tab): change to use centralized active state

wesleycho:
- Change to store active index on tabset controller
- Set index on tab by tab basis

BREAKING CHANGE: The tab API is changed - please refer to the
documentation on new usage of tabs

This implements the hardest part of #4836.

PR #5489, opened 2016-02-16
fix(tab): make active optional

RobJacobs:
This change makes the active binding added under #5425 optional.  [Plunk](http://plnkr.co/edit/lPwmPpFK0sLVGjsJsUge?p=preview) with updated code.


src/tabs/test/tabs.spec.js
```diff
@@ -150,6 +150,41 @@ describe('tabs', function() {
...
+      ].join('\n'))(scope);
+      scope.$apply();
+      return elm;
+    }));
+
+    it('should bind tabs content and set first tab active', function() {
+      expectContents(['first content is 1', 'second content is 2']);
+      expect(titles().eq(0)).toHaveClass('active');
+      expect(titles().eq(1)).not.toHaveClass('active');
+      expect(scope.active).toBe(1);
```
wesleycho: This is the wrong way to test this - you need to grab the controller via `elm.controller('uibTabset')` and check active on that

RobJacobs: DOH!  Sorry I missed that, I'm surprised it didn't fail when running the test in Chrome?

wesleycho:
LGTM

PR #5491, opened 2016-02-16
fix(position): getRawNode to handle select/ul elements

### LINKED ISSUE #5354: Position of select element not determined correctly
This behaviour exposed itself when i was using the popover functionality but i believe it is caused by the position module.

When i place a popover on an select element, the position of the popover is not determined correctly.
[See this demo:](http://plnkr.co/edit/tjXWPeH8gmLFYbdJvGEp?p=preview)
In chrome / IE: When you hover over the select element the popover is placed in the top left corner of the (absolute positioned) form element.
In FireFox: it is placed differently but still not in the correct position.

After some investigation it looks like this is caused by the position module. The [getRawNode function](https://github.com/angular-ui/bootstrap/blob/master/src/position/position.js#L35) tries to get the first array element of the parameter which works in most cases because if it is a jq object this will get the raw element but for (raw) select elements this will get the first option element within the select element. After that the position of this option element can not be determined correctly.

An easy fix would be to make the getRawNode function aware of this exception but since this is a function in the core of this framework which is called  quite often i can imagine there might be a better approach in solving this.

When i overwrite the getRawNode with a (naive) fix the position is determined as expected as can be seen in [this demo](http://plnkr.co/edit/eAXVMOQhGTrDfnZ366Cj?p=preview)

RobJacobs:
The getRawNode function was not returning the correct object
for collection elements (select, ul, etc...).

Fixes #5354


wesleycho:
LGTM

PR #5497, opened 2016-02-17
docs(progressbar): fix wrong type value

livelazily:


Foxandxss:
That doc is generated, you need to modify the doc under the module (progressbar in your case).

livelazily:
ok, I found that it was fixed in the doc under the module

PR #5500, opened 2016-02-17
chore(readme): Update installation notes

### LINKED ISSUE #5499: Doc, add ngTouch as optional dependency
Carousel is using `ng-swipe-right` and `ng-swipe-left` which are from ngTouch module
https://github.com/angular-ui/bootstrap/blob/master/template/carousel/carousel.html

https://docs.angularjs.org/api/ngTouch/directive/ngSwipeRight

Readme.md contains this under "Installation"
"Note: Since version 0.13.0, UI Bootstrap depends on ngAnimate for transitions and animations, such as the accordion, carousel, etc. Include ngAnimate in the module dependencies for your app in order to enable animation."

I think it should also mention that ngTouch is required if you need swipe to work with carousel.

Br,
Markku

deeg:
Note that the carousel requires ngTouch for swiping.

Closes #5499


Foxandxss:
LGTM

PR #5502, opened 2016-02-17
feat(tooltip): arrow position

### LINKED ISSUE #5464: Popover arrow in bottom-left or bottom-right format isn't centrally aligned to button
I have a 36 x 36px button and i've set up the popover to be auto bottom. When in a mobile view, i'm expecting the popover arrow to appear central to the button when presented in the bottom-left or bottom-right format. 

As you can see in the image below, the arrow isn't appearing centrally when bottom left or bottom right. Is this intended or is there a way to get it to display centrally?

image

image

RobJacobs:
- Move arrow position style overrides to a class instead of inline styles.  This will allow for consumers to position the arrow to suit their needs by setting margins on the arrow element.

- Optimize when the positionTooltip function calls the positionArrow function.  Now the arrow position will only be recalced when the tooltip placement changes.

Fixes #5464

[Plunk](http://plnkr.co/edit/LD4XHjwn1ZaVowSAeALv?p=preview) with changes.

### DIFF
--- a/src/position/position.js
+++ b/src/position/position.js
@@ -476,10 +476,17 @@ angular.module('ui.bootstrap.position', [])
           return;
         }
 
+        var arrowCss = {
+          top: '',
+          bottom: '',
+          left: '',
+          right: ''
+        };
+
         placement = this.parsePlacement(placement);
         if (placement[1] === 'center') {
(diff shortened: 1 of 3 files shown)


wesleycho:
LGTM

PR #5503, opened 2016-02-17
fix(typeahead): keep input bg-color in line with css

PerfectPixel:
With option typeahead-show-hint enabled, the underlying input element should not have its own custom background-color. Otherwise, the typeahead will always have a white background-color, regardless the general input appearance defined in stylesheets.

see: http://plnkr.co/edit/dbyQuGOrPZCXdrZxL6WZ?p=preview


PerfectPixel:
scrap that PR. There are more issues at hand with typeahead-show-hint. I'll create a new PR soon.

PR #5507, opened 2016-02-18
chore(docs): create new issue template

icfantv:
* create new template for UIBS issues


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+The issues forum is __NOT__ for support requests. It is for bugs and feature requests only.
+Please read https://github.com/angular-ui/bootstrap/blob/master/CONTRIBUTING.md and search
+existing issues (both open and closed) prior to opening any new issue and ensure you follow the instructions therein.
```
Foxandxss: I know you like this, but perhaps just the first line? There is a big warning about reading contributing and the main issue are the support request and you could maintain that.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
...
+
+### Version of Angular, UIBS, and Bootstrap:
+
+### Bug description:
+
+### Link to minimally-working plunker that reproduces the issue:
```
Foxandxss: I prefer putting the version of angular / uibs / bootstrap as the last step.

deeg: Maybe also add:
```
Angular: 
UIBS:
Bootstrap:
```
below header?

ISSUE_TEMPLATE.md
icfantv: Right, but no one reads it.  No one sees the yellow banner and the text to click - they are guaranteed to see the text in the box.  I know where you are coming from, but I'm darn near adamant here...please convince me otherwise.

ISSUE_TEMPLATE.md
icfantv: Sure, both are good.

deeg:
LGTM if all good with @Foxandxss

Foxandxss:
I still don't agree with that block of text, but I guess I have to pass on it.

That being said, insert it on a .github folder so we don't clutter the root.

icfantv:
I saw that as an option.  Wasn't sure how we wanted to do that.  I'll move, squash, and commit.

icfantv:
@Foxandxss thank you for humoring me here.

PR #5508, opened 2016-02-18
feat(position): add isScrollable function

RobJacobs:
Added a function to check whether an element is scrollable.  This is useful in some positioning operations where the element needs to be checked if it's scrollable before going after the scrollParent.


deeg:
This LGTM

RobJacobs:
I think this change will have unintended side affects, hold off on merging.

wesleycho:
LGTM

PR #5509, opened 2016-02-19
feat(modal): Call $onInit on controller if defined

### LINKED ISSUE #5472: $onInit hook from Angular 1.5.0 will not fire for ModalInstance controller
Of course I have in mind that, at this moment, Angular 1.5.0 is not officially supported:
> AngularJS (requires AngularJS 1.4.x, tested with 1.4.9)

But have I decided anyway to submit an issue about new `$onInit` hook problem. The problem is that it will not fire for modal instance controller (specified with `controller` and `controllerAs` options for `open` method). Here is plunker forked from your demo pego which shows an issue: [$onInit will not fire for modal instance controller](http://plnkr.co/edit/0kccyc2Qh6OLXHxfiJQl?p=preview). Some content has been removed, Angular CDN have been updated to use 1.5.0 version and controllerAs syntax has been used.

Update:
Of course problem does not exists with current supported Angular version (1.4.9), because hook is available in 1.5.0 version

dgsmith2:
After being instantiated, modal controllers which use ‘bindToController’ will have $onInit called, if defined.

Closes #5472

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -687,6 +687,9 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
                       ctrlInstance.$close = modalScope.$close;
                       ctrlInstance.$dismiss = modalScope.$dismiss;
                       angular.extend(ctrlInstance, providedScope);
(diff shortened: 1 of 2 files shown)


Foxandxss:
LGTM

PR #5511, opened 2016-02-19
fix(typeahead): reduces styling issues

PerfectPixel:
This fixes multiple issues:
1. Hint-Input background color is not always white. The user can style their input in a uniform manner now. (see http://plnkr.co/edit/dbyQuGOrPZCXdrZxL6WZ?p=preview)
2. Removes a background image from the input. Before a background image might just conceal the hint-input. Hint-input will carry background-color and -image.
3. hintInputElem is no longer a copy of the typeahead but a new input that carries the class/style/ng-class/ng-disabled attributes and is compiled with the parent scope. Before using ng-disabled on a typeahead did not disable the hint-input and breaking uniform styling (see above mentioned plnkr).

However, I would like some input on the third point as compiling the element might introduce some unwanted side effects.


src/typeahead/typeahead.js
```diff
@@ -127,10 +127,12 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
...
       element.after(inputsContainer);
-      hintInputElem = element.clone();
-      hintInputElem.attr('placeholder', '');
+      hintInputElem = angular.element('<input></input>');
       hintInputElem.attr('tabindex', '-1');
-      hintInputElem.val('');
+      hintInputElem.attr('class', element.attr('class'));
+      hintInputElem.attr('style', element.attr('style'));
+      hintInputElem.attr('ng-class', element.attr('ng-class'));
+      hintInputElem.attr('ng-disabled', element.attr('ng-disabled'));
```
wesleycho: This also needs to be done for the regular `disabled` attribute & use `.prop` in addition to this.

wesleycho:
I'm hesitant about this approach - this seems fragile, and prone to regressions. Is there no other way than to eliminate cloning of the element?

deeg:
Could [.clone](http://api.jquery.com/clone/) not be used? It is available on jqLite.

PerfectPixel:
@deeg originally .clone was used on the typeahead. However, if the cloned element is not compiled, ng-disabled and other directives won't work. If you compile the cloned element, you end up in a loop as typeahead-show-hint is compiled as well.

So what are the options:

1. Remove all attributes/properties that may cause a loop. You may remove typeahead-show-hint but what about the one you do not know. I feel in this case we have little control over what is actually happening and introduce a lot of overhead due to additional directives.
2. The approach I proposed. I thought about it as cherry-picking styling-relevant attributes/properties. Certainly we won't cover everything. But we have more control over the whole thing.
3. We leave it as it is: broken if you use ng-disable or ng-class (or have a different background-color than white or an opaque background-image - these are however fixable by adjusting the css only)

PR #5513, opened 2016-02-20
fix(tabs): adding 'active' ng-class for tab <a> tags for bootstrap 4

uxtx:
- adds 'active'/'disabled' classes for anchor links to affect proper active tab styling in bootstrap 4.

### DIFF
--- a/src/tabs/test/tabs.spec.js
+++ b/src/tabs/test/tabs.spec.js
@@ -8,6 +8,7 @@ describe('tabs', function() {
   function titles() {
     return elm.find('ul.nav-tabs li');
   }
+
   function contents() {
     return elm.find('div.tab-content div.tab-pane');
   }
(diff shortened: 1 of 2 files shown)


Foxandxss:
We are not giving BS 4 support at this moment.

icfantv:
@foxandxss, @wesleycho allowed a PR in last week or so.  IMO, if the change doesn't break anything and doesn't add bloat, I don't have an issue with it.  Thoughts, guys?

Foxandxss:
I don't have an issue as well, I just don't want to give the false impression that we are supporting BS4

uxtx:
I'll work to address any regressions that get reported.  To my knowledge the active toggle class on anchor tags, along with the 'nav-link' class on the anchor tag are all that are needed for bs4, and I didn't notice any real negatives to including these tags in existant bootstrap versions.

uxtx:
I do know bs4 makes more sense going forward for ng2, but my feeling (with tabs at least) was that this was low hanging fruit to get to a minimal level of functionality.

wesleycho:
This adds a $watch from ng-class - I do not think this is wise while support Bootstrap 3. The `nav-link` class is fine because it is a harmless change, but this has perf ramifications.

I'm inclined to reject this.

uxtx:
I suppose that makes sense. I could look at possibly using a different template for BS4 that would remove the duplicated watch, if that'd be preferred. but that may mean enlarging the scope of the pr beyond what the team would desire. Any suggestions?

wesleycho:
At this time, I think you should just override the default template with your own template in $templateCache.

When we investigate BS4 support, we can then make the modifications, but that won't be for a little while.

Issue #5423, opened 2016-02-06
1.2 release tasks


wesleycho:
- [x] Move datepickerMode usage support to datepickerOptions object
- [x] Move other attributes to datepickerOptions object & deprecate old attributes
- [x] Update documentation to use new datepickerOptions object
- [x] Document $onInit for modal
- [x] Fix #5470

Bonus

- [x] Fix tab API
- [x] Fix carousel API
- [ ] Fix modal issues
- [ ] Fix #5471

wesleycho:
Closing, as 1.2.0 is releasing soon.

PR #5526, opened 2016-02-23
feat(datepicker): move datepicker mode to options

wesleycho:
- Fully move datepickerMode support to options, deprecate attribute usage

This addresses the first task in #5423


RobJacobs:
LGTM

PR #5528, opened 2016-02-23
feat(datepicker): move attribute support to options

wesleycho:
- Move some attribute usage to datepickerOptions

This gets us most of the way with the second task in #5423


src/datepicker/datepicker.js
```diff
@@ -35,6 +35,7 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
 
   if (optionsUsed) {
     [
+      'dateDisabled',
```
deeg: Does customClass not need to go here?

src/datepicker/docs/readme.md
```diff
@@ -33,6 +33,8 @@ The datepicker has 3 modes:
 
   The supported options are:
 
+  - customClass
+  - dateDisabled
```
deeg: Looks like missing datepickerMode from previous changes?

src/datepicker/datepicker.js
wesleycho: Good catch, need to add it

src/datepicker/docs/readme.md
wesleycho: Yep, need to add it.

wesleycho:
Added updates, updated unit tests to use options.

deeg:
This LGTM

PR #5531, opened 2016-02-24
fix(modal): Open modal is shifting body content to the right on Chrom…

### LINKED ISSUE #2631: the background move a little when opening a modal.
when I click button to show a "modal", the background move a little then move back, which is not happened in bootstrap and uikit modal.

burlhock:
…e and Firefox

Port of issues fixes twbs/bootstrap#9855 and twbs/bootstrap#15353

I had to move the addClass on body before the animation.

Maybe it would have been better to set the padding on the _appendToElement_ instead of the body ?

Closes #2631

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -254,10 +254,11 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
     };
   })
 
-  .factory('$uibModalStack', ['$animate', '$animateCss', '$document',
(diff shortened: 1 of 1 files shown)


wesleycho:
There is already a helper function in the `uibPosition` service - @RobJacobs how would you recommend this be refactored?

RobJacobs:
I would like to see all this moved to the $uibPosition service, there is nothing here specific to the modal directive and this could potentially be useful somewhere else.  I'll put a PR together...

wesleycho:
Closing as I believe a fix is already in a release

PR #5533, opened 2016-02-24
Update collapse.js

johanneskrtek:
Use 'ease' timing function for the expanding and collapsing transition.


wesleycho:
So this is actually an ngAnimate bug - this might cause potential issues when this is fixed if users use custom easing functions.

johanneskrtek:
Alright. Is there an issue for the ngAnimate bug we could refer to?

Actually it should not be necessary to set the timing function to `ease` as it is the default of the CSS `transition-timing-function` property.

wesleycho:
https://github.com/angular/angular.js/issues/12656

PR #5540, opened 2016-02-25
chore(datepicker): change deprecation message for popup

Foxandxss:
The previous message was a bit confusing because popup attributes are not deprecated, just the inline ones.


icfantv:
LGTM

PR #5541, opened 2016-02-25
fix(datepicker): assign initDate correctly to the controller

Foxandxss:
This feature is THAT useless that no one noticed this.

Also not sure about "fix" or chore since this bug hasn't been released yet.


wesleycho:
LGTM

PR #5547, opened 2016-02-26
[Pagination] Add page-label attribute 

### LINKED ISSUE #2532: Pagination: add custom page indexes

aitboudad:
Closes #2532


wesleycho:
I like this - it is uninvasive.

PR #5552, opened 2016-02-29
fix(tooltip): ensure tooltip custom CSS is imported with webpack

mattlewis92:


wesleycho:
Does this play nice with Browserify?

mattlewis92:
If you have the right plugin installed it should, it's how the carousel, typeahead etc already function i.e they already require css files

wesleycho:
Looks like we already do this in the other components, this just got missed.

LGTM

mattlewis92:
Awesome, thanks!

RobJacobs:
@wesleycho @mattlewis92 Looks like this is missing for the position.css in the position index.js too. Popover index.js has a require on tooltip.js, is tooltip.css needed there as well?

Foxandxss:
If you require an index.js, all the requires inside are loaded as well.

wesleycho:
Position needs it still - not sure if we're missing it on any others.

PR #5554, opened 2016-02-29
docs(changelog): add missing breaking change

mattlewis92:
As a sidenote, it might be worth employing something like [semantic-release](https://github.com/semantic-release/semantic-release) for incrementing the version number of the project to take the human decision out of the equation. As you now support npm and the default is to install with `^1.2.1` which assumes no breaking changes in minor releases this will break a lot of projects when running `npm update` if they don't lock dependencies. Just my $0.02, keep up the awesome work! :smile:


Foxandxss:
I am not reaaaally fond of releasing after each push, but I agree that making this kind of BC on minor releases is wrong. We apologize for this.

mattlewis92:
You don't have to release on every push, you can make a new branch called `release` or w/e and only release when you merge into the branch so you still have full control over when you release a new version

icfantv:
closing as duplicate of #5555.  I didn't see this before.  thanks for the effort.

the onus is on us as developers to ensure we add BREAKING CHANGE to the commit message on a PR for an issue that has a breaking change.  this is the only way the changelog generator will pick up the breaking change and add to the appropriate section.

PR #5430, opened 2016-02-07
feat(tabs): Allow setting optional classes on each tab.

amahfouz:
Allow setting any set of optional CSS classes on a uib-tab element.

This is to allow specifying classes such as "btn-sm" for tab pills.

PR #5555, opened 2016-02-29
fix(docs): fix changelog for tab breaking change

### LINKED ISSUE #5551: active two-way data binding is broken in  1.2.0 and 1.2.1
### Bug description:

The 2-way data binding for the 'active' attribute in the tab component seems to be completely broken with the latest two versions. I'd give a plnkr, but plnkr is giving me a helluva time loading the latest js, in fact, so is jsfiddle. here's a working example: https://jsfiddle.net/willbdaniels/2mq3106a/14/ it just has the js I needed (1.2.0) embedded. you can change the raw js around to see the problem not being present in the version just before that

Angular: 1.4.8

Bootstrap: 1.2.1

**Edit to add in nice plunker**

http://plnkr.co/edit/GkyGI4V92I4DXp3CWYwp

icfantv:
* add breaking change note for 1.2.0 release for tab change in #5430.

Fixes #5551


wesleycho:
LGTM

Issue #5548, opened 2016-02-27
Unit tests in IE11 get progressively slower with angular-bootstrap >=1.1.0


mikedaly:
We have a fairly large angular app that utilizes ui-bootstrap. We also have a fairly extensive set of unit tests (~1900) that exercise as much of the system as we can in isolation. We run our unit tests using karma and jasmine against the latest version of Chrome and IE11.

During a recent upgrade from ui-bootstrap 0.14.3 to 1.2.0, I discovered that in IE11 each test in a run gets subsequently slower. Eventually, the tests slow down to the point that IE11 stops responding, and the test run fails. There's no one test that starts the sequence; each one just takes a little longer. At about 800ms for a single test, the run fails. Unit tests in Chrome behave the same through all versions.

I walked backwards through ui-bootstrap versions and discovered that the behavior was introduced in 1.1.0; 1.0.3 and all previous versions do not have this problem, and all subsequent versions do, including the latest (1.2.0).

I'll see if I can put together a set of simple unit tests that show the problem. It has to be a series of tests, though; running a single test works fine, and even running a small number of tests you might not notice the change.

Angular: 1.5.0
UIBS: >=1.1.0
Bootstrap: 3.3.6 (via bootstrap-sass)

icfantv:
Closing as duplicate of #5496.

Please extend us the courtesy of reading our project documentation and searching existing issues filed (both open and closed) before opening a new issue.

mikedaly:
I do not believe that this is a duplicate of the linked issue (which I did find before posting). As evidence I present the following:

* The bug I reported occurs in versions 1.1.0 and greater. It does NOT occur in 1.0.3 or earlier. (The linked bug was identified as being introduced in 1.0.2)
* The bug I reported only occurs in IE. It does not affect Chrome. (I have not yet tested other browsers.)
* The bug I reported isn't just a small performance issue that adds 1-2 seconds. Each spec in our suite takes longer than the previous spec, so eventually each spec is taking 800ms+ and the karma timeout kills the whole run. The entire test suit run takes 45 seconds in Chrome and 9 minutes in IE before it fails.

This bug prevents us from running our test suite in IE at all.

I have not looked at the commits between those two versions (1.0.3 and 1.1.0) when the bug was introduced. But I suspect that there is something added to the DOM that isn't being remove, thus accumulating over long test runs since the DOM is not recreated between each spec. I surmise that whatever is affecting our tests would also (over time) affect a production system.

[...]

PR #5557, opened 2016-03-01
fix(grunt): do not inject uib css by default (issue #5548)

pusherman:
Injecting CSS in the run block causes memory leaks when running jasmine tests

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -436,7 +436,7 @@ module.exports = function(grunt) {
       .replace(/\\/g, '\\\\')
       .replace(/'/g, "\\'")
       .replace(/\r?\n/g, '\\n');
-    js = `angular.module('ui.bootstrap.${moduleName}').run(function() {!angular.$$csp().noInlineStyle && angular.element(document).find('head').prepend('<style type="text/css">${css}</style>'); });`;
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -436,7 +436,7 @@ module.exports = function(grunt) {
       .replace(/\\/g, '\\\\')
       .replace(/'/g, "\\'")
       .replace(/\r?\n/g, '\\n');
-    js = `angular.module('ui.bootstrap.${moduleName}').run(function() {!angular.$$csp().noInlineStyle && angular.element(document).find('head').prepend('<style type="text/css">${css}</style>'); });`;
+    js = `angular.module('ui.bootstrap.${moduleName}').run(function() {!angular.$$csp().noInlineStyle && angular.$$uibInjectCss && angular.element(document).find('head').prepend('<style type="text/css">${css}</style>'); });`;
```
wesleycho: I would prefer this to just be called `$$uibCss`. In addition, it needs `!angular.$$uibCss` before the style tag injection and `&& angular.$$uibCss = true` at the end of the expression.

pusherman: @wesleycho updated

PR #5565, opened 2016-03-01
docs(changelog): add missing carousel breaking change

mattlewis92:
As a side note I think active could / should be made optional like the tabs API. Will try and find the time to send another PR tomorrow :smile:


wesleycho:
LGTM

PR #5569, opened 2016-03-02
fix(modal): fix bindToController bug

kuitos:
when set bindToController true,props not be bound to controller instance correctly before it initialize,this pointer not be init with $scope props

what I added in modal unit test
```js
var $scope = $rootScope.$new(true);
$scope.foo = 'bar';
open({
  template: '<div>{{test.fromCtrl}} {{test.closeDismissPresent()}} {{test.foo}}</div>',
  controller: function($uibModalInstance) {
  + expect(this.foo).toEqual($scope.foo);
    this.fromCtrl = 'Content from ctrl';
    this.closeDismissPresent = function() {
```
it will make the test fail
```shell
Expected undefined to equal 'bar'.
```
because the props not bound to the this instance of controller correctly before the controller constructor called.

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/bootstrap/5569)

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -671,7 +671,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
                   }
                 });
 
-                var ctrlInstance, ctrlLocals = {};
+                var ctrlInstance, ctrlInstantiate, ctrlLocals = {};
 
                 //controllers
                 if (modalOptions.controller) {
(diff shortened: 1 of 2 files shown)


src/modal/modal.js
```diff
@@ -681,18 +681,28 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
...
                       ctrlInstance.$dismiss = modalScope.$dismiss;
                       angular.extend(ctrlInstance, providedScope);
-                      if (angular.isFunction(ctrlInstance.$onInit)) {
-                        ctrlInstance.$onInit();
-                      }
+                    }
+
+                    ctrlInstance = ctrlInstantiate();
+
+                    if (modalOptions.bindToController && angular.isFunction(ctrlInstance.$onInit)) {
```
wesleycho: Why is `bindToController` being required here for $onInit usage?

src/modal/modal.js
```diff
@@ -681,18 +681,28 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap'])
                     ctrlLocals[key] = value;
                   });
 
-                  ctrlInstance = $controller(modalOptions.controller, ctrlLocals);
+                  // the third param will make the controller instantiate later,private api
+                  // @see https://github.com/angular/angular.js/blob/master/src/ng/controller.js#L126
+                  ctrlInstantiate = $controller(modalOptions.controller, ctrlLocals, true);
                   if (modalOptions.controllerAs) {
+
```
wesleycho: Remove extra space

src/modal/modal.js
icfantv: Per https://docs.angularjs.org/guide/component, in the Intercomponent Communication section, `$onInit` does NOT require `bindToController` being present.

src/modal/modal.js
kuitos: OK

src/modal/modal.js
kuitos: yes u are right,I had checked the source code of angular.js,it does not require `controllerAs` too.

PR #5573, opened 2016-03-02
fix(rating): fix usage of aria-valuetext

### LINKED ISSUE #5571: uib-rating: The WAI-ARIA property aria-valuetext is not supported on elements with WAI-ARIA role button

### Bug description:
The WAI-ARIA property aria-valuetext is not supported on elements with WAI-ARIA role button.
This is for uib-rating star element, and is related with accessibility. 
https://www.w3.org/TR/wai-aria/states_and_properties#aria-valuetext

<i class="glyphicon ng-scope glyphicon-star" **aria-valuetext**="one" ng-attr-title="{{r.title}}" ng-class="$index < value && (r.stateOn || 'glyphicon-star') || (r.stateOff || 'glyphicon-star-empty')" ng-click="rate($index + 1)" ng-mouseenter="enter($index + 1)" ng-repeat-end="" role="**button**" tabindex="0" title="one"></i>

Angular: 1.4.9

UIBS: 1.2.1

Bootstrap: 3.3.6

wesleycho:
- Fix usage to be on top level element

This fixes #5571.


src/rating/rating.js
```diff
@@ -8,7 +8,8 @@ angular.module('ui.bootstrap.rating', [])
 })
 
 .controller('UibRatingController', ['$scope', '$attrs', 'uibRatingConfig', function($scope, $attrs, ratingConfig) {
-  var ngModelCtrl = { $setViewValue: angular.noop };
+  var ngModelCtrl = { $setViewValue: angular.noop },
+    self = this;
```
deeg: Not that it really matters, but won't ```this.getTitle``` work in this situation, making self = this not needed?

deeg:
This LGTM

src/rating/rating.js
wesleycho: It actually doesn't - I tried that first :(

deeg: Hmmm, do you know why that does not work? Just curious.

wesleycho: Because it is called in the render function, which is called in $scope.rate is my suspicion.

PR #5575, opened 2016-03-02
fix(tabs): update doc and fix tab demo

### LINKED ISSUE #5551: active two-way data binding is broken in  1.2.0 and 1.2.1

icfantv:
* update tab documentation to clarify proper use and intended behavior of the
  `active` attribute on both the `tabset` and `tab` directives.
* fix the tab select buttons in the tab demo.

Fixes #5551


wesleycho:
LGTM - that's a pretty bad mistake I did in the demo template.

PR #5577, opened 2016-03-03
Allow string as Tab index value

vadim-p:
Simple situation - when iterating hash, so hash key plays well as index.


icfantv:
@vadim-p, i'm ok with this, but this is going to need tests and a documentation update as well.

icfantv:
@vadim-p, this is a good feature, but we are unable to use this PR and will close it if you don't address the concerns previously mentioned.  Thanks.

RobJacobs:
We are also doing the isNumber check in the select and addTab functions.  I'd like to get an opinion from @wesleycho on whether the index needs to be a number or we can just use isDefined instead?

wesleycho:
I assumed an index should be a number - I'm not hugely opinionated on whether it should be though, and am fine with changing that. @icfantv is right in what this PR needs though.

vadim-p:
Hello. Sorry for silence from my side - I'm traveling. I'll try to get back to you with updates in 3-4 days. Thank you.

icfantv:
@vadim-p, awesome.  Thanks.

icfantv:
@vadim-p do you still intend on completing this?  If not, we're going to close it.

icfantv:
Closing due to lack of response.  Replaced by #5713.

PR #5580, opened 2016-03-03
feat(datepicker): add helpers for styling

wesleycho:
- Add helper variables available for custom templates to hook into for styling


deeg:
Did we want any tests for this?

deeg:
This LGTM

wesleycho:
I thought about it, but we don't even have tests for current excepting that it adds the btn-info class appropriately - I made no changes to that logic though.

icfantv:
I'm ok with this.

PR #5582, opened 2016-03-04
fix(position): auto position append to body

### LINKED ISSUE #5576: Auto positioning (new feature) not working when overflow: auto | scroll is applied (at least in datepicker)
### Bug description:
Auto positioning is not working when a containing elements has the property *overflow: auto | scroll* although datepicker-append-to-body is set to true.

https://plnkr.co/edit/uKGjGCX1cxBnjY6dN0ha

- Angular: 1.5.0
- UIBS: 1.2.1
- Bootstrap: 3.3.6

RobJacobs:
Use the body as the viewport to determine available size for auto position when the target element is append-to-body.

Fixes #5576

### DIFF
--- a/src/position/position.js
+++ b/src/position/position.js
@@ -387,7 +387,7 @@ angular.module('ui.bootstrap.position', [])
         var targetElemPos = {top: 0, left: 0, placement: ''};
 
         if (placement[2]) {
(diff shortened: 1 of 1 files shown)


wesleycho:
LGTM - super simple!

PR #5587, opened 2016-03-05
fix(typeahead): update isOpen on escape

### LINKED ISSUE #5579: Typeahead is-open does not update when typeahead closes from escape keydown
### Bug description: 
`typeahead-is-open` does not update when typeahead closes from escape keydown.  To reproduce this, simply type some text and the hit the escape key.  The typeahead menu will close itself, but the variable bound to `typeahead-is-open` does not update.  It updates twice after being reopened with new inputs.

Angular: 1.4.7, 1.5.0

UIBS: 1.2.1 (with templates)

Bootstrap: 3.3.6

deeg:
[Plunker with fix](http://plnkr.co/edit/IwanX1L5n7cxZpVoeIxX?p=preview)

Fixes #5579


wesleycho:
In this case, looks like `scope.$parent.$digest()` would be sufficient, as seen [here](http://plnkr.co/edit/h0MdPOz2HHFycX8gTbyz?p=preview)

deeg:
Plunker and code updated.

wesleycho:
LGTM

PR #5588, opened 2016-03-05
fix(typeahead): change to original scope

### LINKED ISSUE #5467: Typeahead is-open change via document click does not trigger digest on original scope
When typeahead's drop-down is open, clicking outside of it closes the drop-down but variable bound to is-open does not change to false. Here's a minimal [test case] (http://plnkr.co/edit/fkl2wd?p=preview).

Changing 'scope.$digest()' to 'originalScope.$digest()' [here](https://github.com/angular-ui/bootstrap/blob/master/src/typeahead/typeahead.js#L452) seems to fix the problem.

wesleycho:
- Change to original scope for running $digest

This fixes #5467.


deeg:
Ah, sorry I didn't see that.

LGTM

PR #5593, opened 2016-03-06
fix(outsideClick): closing when event element is removed from DOM

onumossn:
Prevent popover and dropdowns that have outsideClick from closing when the element clicked is inside the popover or dropdown the click event causes the element to be removed from the DOM.

Bug Example: http://plnkr.co/edit/a7mCnSmMfrARKCfDxCTw?p=preview
Fix Example: http://plnkr.co/edit/TSAjXnlMMNRSnuneRZsT?p=preview

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -10,7 +10,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
 
   this.open = function(dropdownScope) {
     if (!openScope) {
-      $document.on('click', closeDropdown);
+      document.body.addEventListener('click', closeDropdown, true);
       $document.on('keydown', keybindFilter);
     }
 
@@ -24,7 +24,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
   this.close = function(dropdownScope) {
     if (openScope === dropdownScope) {
       openScope = null;
-      $document.off('click', closeDropdown);
+      document.body.removeEventListener('click', closeDropdown, true);
       $document.off('keydown', keybindFilter);
     }
   };
--- a/src/dropdown/test/dropdown.spec.js
+++ b/src/dropdown/test/dropdown.spec.js
@@ -54,14 +54,14 @@ describe('uib-dropdown', function() {
       expect(element).toHaveClass(dropdownConfig.openClass);
 
       var optionEl = element.find('ul > li').eq(0).find('a').eq(0);
-      optionEl.click();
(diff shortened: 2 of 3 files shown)


src/dropdown/dropdown.js
```diff
@@ -10,7 +10,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
 
   this.open = function(dropdownScope) {
     if (!openScope) {
-      $document.on('click', closeDropdown);
+      document.body.addEventListener('click', closeDropdown, true);
```
deeg: Would rather see: `$document[0].body`

wesleycho: I think the whole reason he/she switched was to pass in the third parameter to addEventListener...but I could be wrong.

wesleycho:
I don't like this solution - why does it need to be changed from jqLite to raw DOM event listeners?

src/dropdown/dropdown.js
deeg: I thought $document[0].body === document.body?

onumossn:
jqLite doesn't support the 3rd param.

src/dropdown/dropdown.js
wesleycho: Ah, I missed that. If that is necessary, I'd prefer `$document.find('body').on(...)`

onumossn: I can update that if this solution is acceptable.

wesleycho:
It sounds wrong to use the `useCapture` parameter of addEventListener - this seems to go against typical event bubbling.

Switching from `$document` to `$document.find('body')` sounds legitimate though.

onumossn:
Anything else I am imagining right now seems significantly worse or complicated to me. They would require watching for changes on all the elements that compose a popover content and adding them to a list of elements to be ignored, so detached nodes can still be considered part of the popover, or walking through the popover and wrapping or reordering all events created in the popover.

RobJacobs:
This can be avoided by just using stopPropagation() on the element, example[plunk](http://plnkr.co/edit/l14b3vO2JP2YzGzLUrIq?p=preview)

onumossn:
That breaks outsideClick if you have nested components using outsideSide click. If you click on the outer element, the trigger for the inner element won't be reached. However, that isn't my personal use case and stopPropagation does work for my personal use case for now.

wesleycho:
Going to close this as bypassing the event delegation mechanism is a major breaking change that is not desired in the library. Any attempt to tackle this should not break expected behavior.

Issue #5368, opened 2016-01-30
Support conditional update on panel-class


rohdef:
I have a set of options that I group in the accordion, I'd like to do something like this:

<uib-accordion-group ng-repeat="setting in mySettings" is-open="setting.isVisible"
                         panel-class="{{setting.isValid ? 'panel-danger' : ''}}">

thing is this works on load, but does not update when the isValid setting is changed :(

Foxandxss:
There is no support for that at this moment. `panel-class` is set once.

Not hard to do, but every watch is a performance hit on the app and we need to consider if that feature is worthwhile or not for that.

wesleycho:
We could support an optional @ binding. I think I would be fine with that, seems reasonable - the workaround is to do something like `class="{{foo ? 'panel-danger'}}"`, which potentially adds likely two watchers anyway. (1 interpolation * 2, 2 for the original element and for the element replacing it)

rohdef:
@Foxandxss that makes sense, I guess for the majority of the uses cases you're quite right. The suggestion by @wesleychy seemss quite reasonable to me :)

PR #5596, opened 2016-03-07
feat(accordion): panel-class now responds to change.

blueseph:
the panel-class attribute of accordion groups would never update even if the value changed. it now updates properly. useful for validations and stuff.


wesleycho:
Linking relevant issue, #5368.

This LGTM

src/accordion/accordion.js
```diff
@@ -64,6 +64,7 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
     },
     scope: {
       heading: '@',               // Interpolate the heading attribute onto this scope
+      panelClass: '@',           // Ditto with panelClass
```
wesleycho: Change to `@?` - should be optional

PR #5608, opened 2016-03-09
fix(tabs docs): remove uib-tab's "active" setting

ivanvoznyakovsky:
after #5425 `uib-tab` documentation was not updated properly

### DIFF
--- a/src/tabs/docs/readme.md
+++ b/src/tabs/docs/readme.md
@@ -27,12 +27,6 @@ AngularJS version of the tabs directive.
 
 ### uib-tab settings
 
-* `active`
-  <small class="badge">$</small>
-  <i class="glyphicon glyphicon-eye-open"></i>
-  _(Default: `false`)_ -
-  Whether tab is currently selected. Note that this is purely for CSS and will not physically set this tab to active.
-
 * `classes`
   <small class="badge">$</small> -
    An optional string of space-separated CSS classes.


wesleycho:
LGTM - thanks for the catch.

wesleycho:
Merged via f2004a9186e97cf40ac408e147bbcb508da072c0

icfantv:
@wesleycho wait a second.  we added that as part of the bug fix for moving the active state to the tabset.  This should not have been merged.  We updated the comment correctly.  This should be put back in.

RobJacobs:
Looks like the scope active variable is used internally to set the TWBS active class in the controller select function [here](https://github.com/angular-ui/bootstrap/blob/master/src/tabs/tabs.js#L14) and [here](https://github.com/angular-ui/bootstrap/blob/master/src/tabs/tabs.js#L20)  I don't think this should be documented as a publicly available attribute as I don't see how it could be used to report if the tab is active or not?

We should remove a reference to that in the docs [here](https://github.com/angular-ui/bootstrap/blob/master/src/tabs/docs/demo.html#L20)

icfantv:
I should have said it was _modified_ as part of the tabset changes for the 1.2.2 release.  @RobJacobs, it's used for CSS.  Previously, it was used to actually set the tab's state to active, but that was moved to the `tabset` element.  I still think it's a valid attribute for documentation.

RobJacobs:
@icfantv That's where I'm confused, I know the uibTab template uses uibTab scope.active variable to apply the TWBS active class, but how would a developer use that scope variable?  And if they can't, does it make sense to document it?

RobJacobs:
Is that what the documentation blurb that was removed is describing?  To me, it looks like it was saying you could use the active attribute to do something with the css, which I don't believe is correct?

icfantv:
My intention was to indicate that the attribute was for CSS purposes only - meaning that it was there for  CSS styling.

PR #5624, opened 2016-03-11
add a customOptions attribute to uibAccordionGroup

valmiRe:
This is useful if you want to pass custom attributes to a custom template. It needs a Json object as a value.

### DIFF
--- a/src/accordion/accordion.js
+++ b/src/accordion/accordion.js
@@ -78,6 +78,7 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
 
       scope.openClass = attrs.openClass || 'panel-open';
       scope.panelClass = attrs.panelClass || 'panel-default';
+      scope.customOptions = attrs.customOptions ? JSON.parse(attrs.customOptions) : '';
       scope.$watch('isOpen', function(value) {
         element.toggleClass(scope.openClass, !!value);
         if (value) {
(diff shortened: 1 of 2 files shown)


icfantv:
@valmiRe, i'm going to run this by the team so i'm not going to promise that this will be merged.  if we do, this will need documentation and tests.  also, please look at my comment on your proposed code.

src/accordion/accordion.js
```diff
@@ -78,6 +78,7 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse'])
 
       scope.openClass = attrs.openClass || 'panel-open';
       scope.panelClass = attrs.panelClass || 'panel-default';
+      scope.customOptions = attrs.customOptions ? JSON.parse(attrs.customOptions) : '';
```
icfantv: use `angular.parseJson` instead of `JSON.parse` and set the default to `{}` not an empty string - that's just bad ju-ju.

valmiRe:
hi @icfantv, just made the changes
no problem if you don't merge, I needed this and this was one line, so I just share

icfantv:
@valmiRe tests and documentation updates?  unfortunately, if you don't do the rest of it, we're not going to be able to use any of it.

wesleycho:
Can you explain what the purpose of this object is?

valmiRe:
like I said, this is useful if you want to pass custom attributes to a custom template. 
I close this, I won't be able to do the tests

Issue #5615, opened 2016-03-10
Document popover-enable


twwwt:
After realizing that the tooltip an popover have much in common in their implementation, I checked whether `popover-enable` works the same way as `tooltip-enable`. Yes, it does. Now, the former is not documented on the Web site while the latter is. So, please document it.

Version used: 1.2.4; Angular 1.5.0.

icfantv:
@twwwt, yes, popovers are essentially tooltips with some extra sugar.  The underlying code is shared.

Looks like we are missing the documentation for `popover-enable` and there's a typo in the popover documentation text:  `There are three versions of the popover: uib-popover and uib-popover-template, and uib-tooltip-html:` - that last one should be `uib-popover-html`.

PR #5626, opened 2016-03-13
Issue #5615 (Document popover-enable)

twwwt:
Extend doc to include popover-enable.

Issue #5615


icfantv:
@twwwt, can you please also fix the invalid `uib-` property in the sentence:  "There are three versions of the popover: `uib-popover` and `uib-popover-template`, and `uib-tooltip-html`?"

That last one should be `uib-popover-html` instead of `uib-tooltip-html`.  It's just a few lines up from what you added.  Thanks.

wesleycho:
Made the modification when merging @icfantv

icfantv:
Cool.  I was like....whaaaaa?

PR #5627, opened 2016-03-14
fix(modal): remove a trailing space in class attribute

shuhei:
- Remove a trailing space of `class="modal-dialog "`

It was introduced in #5224 with [a reason](https://github.com/angular-ui/bootstrap/pull/5224#discussion-diff-49548766) but we can remove it without using `ng-class`.

### DIFF
--- a/template/modal/window.html
+++ b/template/modal/window.html
@@ -2,5 +2,5 @@
     uib-modal-animation-class="fade"
     modal-in-class="in"
     ng-style="{'z-index': 1050 + index*10, display: 'block'}">
-    <div class="modal-dialog {{size ? 'modal-' + size : ''}}"><div class="modal-content" uib-modal-transclude></div></div>
+    <div class="modal-dialog{{size ? ' modal-' + size : ''}}"><div class="modal-content" uib-modal-transclude></div></div>
 </div>


wesleycho:
I'm not sure of the value of such a change - this hurts readability a bit, and would have minimal, if any, perf gains.

deeg:
I guess, I can go either way in terms of readability, I don't think this change makes it worse.

@shuhei, with all that being side, what was the reason for you wanting to make this change and remove the extra space? Was the space negatively effecting you in some way?

Foxandxss:
Unless this change fixes something we can't see, it is a no for me.

shuhei:
@wesleycho @deeg Thanks for your comments!

My case is that an XPath `//div[@class="modal-dialog"]` in my Selenium tests stopped working. It can be detoured with `//div[@class="modal-dialog "]` or [more complex XPath](http://stackoverflow.com/a/1604480/822317) but they both are awkward. If a future implementation changes it back to `class="modal-dialog"`, then I need to change `@class="modal-dialog "` to `@class="modal-dialog"` again.

wesleycho:
This seems like a poor way to select using selenium.

One should be using the appropriate CSS selector.

Closing this as an undesired change in the library.

PR #5630, opened 2016-03-14
fix(modal): dynamically fetch elements

### LINKED ISSUE #5050: Cannot tab to previously disabled button in modal
If you have a button that is disabled when the modal is opened and then becomes enabled later, you can't tab to the button.

Here's a plunk to reproduce it: http://plnkr.co/edit/8oCzQa3hPSloxnudxoG8?p=preview

1. Click the Open me! button
2. Enter a value in the Name field and tab to the Address field
3. Enter a value in the Address field and try to tab to the OK button

Notice that focus never goes to the OK button.

### LINKED ISSUE #5421: MODAL: Keyboard focus fail with dynamic content
**The problem:**

When the modal opens, it fires `clearFocusListCache` and as soon as a user presses the `tab` key, it calls `loadFocusElementList` to create a cache of all the potential tabbable elements. The problem arises when the modal contains dynamic content that wasn't initially available for the cacheing (ie: using `ng-if`).

http://plnkr.co/18B9eYZK94Wmq1l8cl8w

**Steps to reproduce:**

1. Open a modal and `tab` or `shift-tab` around. Everything works as expected.
2. Toggle the hidden content button.
3. Repeat step 1. You can now `shift-tab` focus behind modal window.

**Proposed fix:**

Add a couple more definitions to the `keydownListener` for the `spacebar` and `enter` / `return` keys, and when they're pressed, re-run `clearFocusListCache` and then `loadFocusElementList`.

Here's my working example of the new `keydownListener` (there's probably a cleaner way to do it, but it gets the job done):

```
function keydownListener(evt) {
  if (evt.isDefaultPrevented()) {
    return evt;
  }

  var modal = openedWindows.top();
  if (modal) {
    switch (evt.which) {
      case 27:
        {
          if (modal.value.keyboard) {
            evt.preventDefault();
            $rootScope.$apply(function() {
              $modalStack.dismiss(modal.key, 'escape key press');
            });
          }
          break;
        }
      case 9:
        {
          $modalStack.loadFocusElementList(modal);
          var focusChanged = false;
          if (evt.shiftKey) {
            if ($modalStack.isFocusInFirstItem(evt) || $modalStack.isModalFocused(evt, modal)) {
              focusChanged = $modalStack.focusLastFocusableElement();
            }
          } else {
            if ($modalStack.isFocusInLastItem(evt)) {
              focusChanged = $modalStack.focusFirstFocusableElement();
            }
          }

          if (focusChanged) {
            evt.preventDefault();
            evt.stopPropagation();
          }
          break;
        }
      case 13:
        { // keypress: enter/return
          $modalStack.clearFocusListCache();
          $timeout(function() {
            $modalStack.loadFocusElementList(modal);
          }, 200);
          break;
        }
      case 32:
        { // keypress: space
          $modalStack.clearFocusListCache();
          $timeout(function() {
            $modalStack.loadFocusElementList(modal);
          }, 200);
          break;
        }
    }
  }
}
```

wesleycho:
- Dynamically fetch elements for determination of which is the first/last to tab/shift+tab to

This fixes #5050 and #5421


RobJacobs:
LGTM - much more stable approach.

Issue #4288, opened 2015-08-27
TimePicker problem with hour column ( the first zero is hidden when you write on this column)


kevincaradant:
Hi

I need your help to fix that :).

When i wrote in the input "01" in the hour column , only "1" appears and when i leave the input the "0" appear. But i want keep the "0" in all case in the hours column . How can i do that ?

Thank you very much  :)

Kévin

wesleycho:
For UX, it has been preferred that the 0 be omitted - we could perhaps implement this behind a feature flag though.

If you would like to take a stab at it, feel free to open a PR.

kevincaradant:
oki and have you any idea, how i can show again the 0 even in the librairy if i have to change something ? 

otherwise i will wait a new feature ;)

Thank you

BobbieBarker:
"PRs plz," means "Pull Request Please," it is an invitation to users of this library to submit a pull request to fix the issue. This is an open source library and anyone an contribute to it. Including you, to fix your very own problem and help other people who use this library! Details on how to contribute to this library can be found [here](https://github.com/angular-ui/bootstrap/blob/master/CONTRIBUTING.md).

PR #5633, opened 2016-03-15
feat(timepicker): add pad-hours support

wesleycho:
- Add support for optionally padding hours

This implements #4288.


src/timepicker/test/timepicker.spec.js
```diff
@@ -986,6 +986,37 @@ describe('timepicker directive', function() {
     });
   });
 
+  describe('`pad-hours` attribute', function() {
+    function triggerInput(elem, val) {
+      elem.val(val);
+      elem.trigger('input');
+    }
+
+    it('should not pad the hours by default', function() {
```
icfantv: remove 'not' as it _should_ pad by default.

Issue #1797, opened 2014-02-13
Group By TypeAhead


conorw:
Is there anyway to group results by categories in the typeahead item, similar to the jquery autocomplete http://jqueryui.com/autocomplete/#categories? 
It would be nice to show the category name and list the items below it.

Issue #1691, opened 2014-01-29
Support grouping in typeahead


Urigo:
Hi, is there a way to support group inside the ui-bootstrap?

http://stackoverflow.com/questions/21430146/grouping-results-in-ui-bootstrap-typeahead

![screen shot 2014-01-29 at 1 41 34 pm](https://f.cloud.github.com/assets/1297448/2029135/a6256e0c-88dd-11e3-99e4-f138eae9e0fc.png)

kayhadrin:
I'm facing the same need on my current project.
So far, we've simulated the option group feature by adding dummy items in the array of matches, that have a special property (to know it's a group header), then we display them differently.
So that's... roughly possible. Now though, I need to make the option group element non-selectable.
I'll probably create a fork where we have the ability to disable some option elements.
Would that interest you?

brunowego:
+1

icfantv:
@brunowego, please do not just add a `+1` to issues as it adds no intrinsic value to the discussion and serves to only clutter up a thread.  You might be interested in the `ui-select` project for this functionality, but I will say that development has stalled somewhat on it - but it still works quite well.

brunowego:
@icfantv, you right, sorry :+1:

PR #5634, opened 2016-03-16
Extending typeahead-popup functionality

shank8:
Today, the `typeahead-popup-template-url` implementation only works if the array you wish to display in your new template is processed by a single `ng-repeat` directive. However, in the case where you may want to split up your initial array into "filtered" subsets and `ng-repeat` across all subsets, this implementation fails.

The reason it fails is because currently, the logic uses `$index` to track the array and therefore uses `$index` to "find" the selected item from the original list.

Let me explain with an example:

In my case, I have a list of 10 book objects. Lets say each book has a UID to identify it from any other book. Lets also say it has a `genre` property which could be 1 of 2 values ("fiction" or "non-fiction").  Then in my `typeahead-template-url` I set up the template to group each genre into their own "section" like this:

```
<div class="dropdown-menu" ng-style="{top: position().top+'px', left: position().left+'px'}" style="display: block;" ng-show="isOpen() && !moveInProgress" aria-hidden="{{!isOpen()}}">
    <div class="sk-dropdown-category">Fiction</div>
    <a class="dropdown-item" ng-repeat="match in matches | filter:{ model : { genre: 'fiction' } } | limitTo:5" ng-class="{active:isActive(match.model.id)}" ng-mouseenter="selectActive(match.model.id)" ng-click="selectMatch(match.model.id)" role="option" id="{{::match.id}}">
        <div uib-typeahead-match index="match.model.id" match="match" query="query" template-url="templateUrl"></div>
    </a>
    <div class="sk-dropdown-category">Non-Fiction</div>
    <a class="dropdown-item" ng-repeat="match in matches | filter:{ model : { genre: 'nonfiction' } } | limitTo:5" ng-class="{active:isActive(match.model.id)}" ng-mouseenter="selectActive(match.model.id)" ng-click="selectMatch(match.model.id)" role="option" id="{{::match.id}}">
        <div uib-typeahead-match index="match.model.id" match="match" query="query" template-url="templateUrl"></div>
    </a>
</div>
```

As you can see, I now have 2 separate `ng-repeat` directives in the template. This also means that when the controller references `$index`, there can actually be more than one of the same `$index` depending on how many `ng-repeat` directives you use.

Since the `uibTypeaheadController`'s `select()` function uses an index to select the correct item in the array, this causes issues.

Instead, I have added another attribute called `typeahead-popup-template-matcher` which simply takes a string that should map to a property name on your models in the initial array.  In most cases, this value will be your UID property (to ensure we find only 1 unique instance).

I know it is possible to have your UID property nested in your object which would break my implementation, but I was trying to go for the most simple solution at first. If you feel I should implement it in another way that allows nested UID matcher values then please let me know.

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -26,8 +26,8 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
     };
   }])
 
-  .controller('UibTypeaheadController', ['$scope', '$element', '$attrs', '$compile', '$parse', '$q', '$timeout', '$document', '$window', '$rootScope', '$$debounce', '$uibPosition', 'uibTypeaheadParser',
-    function(originalScope, element, attrs, $compile, $parse, $q, $timeout, $document, $window, $rootScope, $$debounce, $position, typeaheadParser) {
+  .controller('UibTypeaheadController', ['$scope', '$element', '$attrs', '$compile', '$parse', '$q', '$timeout', '$document', '$window', '$rootScope', '$$debounce', '$uibPosition', 'uibTypeaheadParser', 'filterFilter',
+    function(originalScope, element, attrs, $compile, $parse, $q, $timeout, $document, $window, $rootScope, $$debounce, $position, typeaheadParser, filterFilter) {
     var HOT_KEYS = [9, 13, 27, 38, 40];
     var eventDebounceTime = 200;
     var modelCtrl, ngModelOptions;
@@ -42,7 +42,7 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
     originalScope.$watch(attrs.typeaheadMinLength, function (newVal) {
         minLength = !newVal && newVal !== 0 ? 1 : newVal;
     });
-    
+
     //minimal wait time after last character typed before typeahead kicks-in
     var waitTime = originalScope.$eval(attrs.typeaheadWaitMs) || 0;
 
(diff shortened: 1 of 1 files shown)


icfantv:
there are potentially significant performance ramifications to consider here.  @wesleycho, thoughts?

shank8:
I agree, if you do choose to use the "matcher" implementation, then there are slight perf ramifications because your search goes from O(1) to O(N).  Can't beat that unless we have a sorted array to begin with. Then we could get O(log(N))

shank8:
Oh I also forgot to mention that I added a utility `$filter` method onto the `uib-typeahead-popup` scope because I was finding myself needing to run a filter in an expression inside of my template (used to hide/show the section headers if there are no results to display.. i left this code out of the example for simplicity sake). Not sure if you would have a better solution to that problem.

wesleycho:
This seems like a hack - a better solution IMO would be to add support for grouping in the typeahead component. This use case breaks a lot of assumptions of the underlying data model, and attempts to hack around that. Any support of that assumption most likely will cause lots of bugs to manifest inside the typeahead.

The better solution is to add support for grouping, which allows for more consistent model tracking & manipulation.

icfantv:
Grouping was talked about in #1691 and #1797.  We decided that the ui-select library (while abandoned) still works quite nicely.

shank8:
I understand your concerns, but I was trying to abstract the problem away from implementation details and allow the developer to have more freedom when constructing their custom template without forcing them to "re-learn" how to use the directive by adding some sort of `uib-typeahead-grouping` directive.  My way is the most simple, but forces assumptions to be made about how your custom template was implemented.

shank8:
Shall we close this?

wesleycho:
Closing as per discussion

PR #5636, opened 2016-03-16
feat(tabs): do nothing on middle click

### LINKED ISSUE #5332: Middle clicking on tabs has different behaviour in Firefox
Clicking a ui.bootstrap.tabs with the mouse wheel in Chrome or IE is the same as clicking normally, however in Firefox it opens a new browser tab of the same page. For instance, middle clicking the tabs on the [demo page](http://angular-ui.github.io/bootstrap/#/tabs) in Firefox will open the demo page again (and not select the clicked tab).

wesleycho:
- On middle click of a tab, do nothing

This closes #5332


src/tabs/tabs.js
```diff
@@ -6,6 +6,10 @@ angular.module('ui.bootstrap.tabs', [])
   ctrl.tabs = [];
 
   ctrl.select = function(index, evt) {
+    if (evt && evt.which === 2) {
```
icfantv: I'm confused.  According to this: https://developer.mozilla.org/en-US/docs/Web/Events/click, the middle button is 1.

icfantv: ...and apparently also 4?

wesleycho: Whoops...I think you're right - I need to look at this again

wesleycho: So `evt.which` is jQuery normalization - this should still be fine, but somewhat unreliable since this needs to be detected via `mousedown`/`mouseup` events.

icfantv:
Hmmm.  So what do you want to do here?

wesleycho:
Thinking about this, this is likely flawed, and I don't think a user would normally expect to open a tab in a new window - if one is caring about this situation, one should have a route associated with it in the UI.

Going to close this and the issue as won't fix.

Issue #5512, opened 2016-02-19
List of focusable elements in modal includes hidden ones


melaniesaraj:
### Bug description:

`$uibModalStack` queries for all of the focusable elements in the modal and uses that list to trap focus in the dialog.  Every time tab is pressed, it checks whether the currently focused element is the last focusable element in the dialog.  If it is, it moves focus to the first element in the dialog, rather than letting it escape into the rest of the page ([modal.js:386](https://github.com/angular-ui/bootstrap/blob/834975899b734369465c2737c4fc947a257c4b1c/src/modal/modal.js#L386)).  (For shift+tab, or tabbing backwards, switch "last" and "first".)

However, the selector used to get this list of focusable elements is incorrect in that it doesn't filter out hidden (`display: none`) elements ([modal.js:272](https://github.com/angular-ui/bootstrap/blob/834975899b734369465c2737c4fc947a257c4b1c/src/modal/modal.js#L272)): 
```
var tababbleSelector = 'a[href], area[href], input:not([disabled]), ' +
        'button:not([disabled]),select:not([disabled]), textarea:not([disabled]), ' +
        'iframe, object, embed, *[tabindex], *[contenteditable=true]';
```
So, in a dialog where the last element is hidden, focus will not properly be trapped, because we'll never end up focused on that hidden element.  Moreover, shift+tab won't be able to loop backwards to the last element, since that hidden element is not focusable, so it'll get stuck on the first element.  (For the shift+tab case, we have the same problems if the first element is hidden.)

Version: Current on master (commit linked above)

### Example:

Plunker: [http://plnkr.co/edit/hL3G8FUUiMnYKJcRkrFN?p=preview](http://plnkr.co/edit/hL3G8FUUiMnYKJcRkrFN?p=preview)

Here, we have a dialog that shows a 'Help' button as the last button in the dialog IFF there's a non-empty help link, using `ng-show` (which uses `display: hidden` to hide elements).  (Of course, this is a contrived example; `ng-if` would work fine in a situation like this.)  First, open the modal without generating a help link.  Tab through the modal.  Focus will escape from the modal and enter the address bar and the rest of the page, for the reason described above. Shift+tab through the modal.  Focus will get trapped on the first element.  Then, generate a help link and open it again.  Tabbing and shift+tabbing will properly keep focus in the model, because the last focusable element is... actually focusable.

### Proposed fix

[...]

PR #5644, opened 2016-03-17
feat(modal): exclude hidden elements from query

wesleycho:
- Modify element query to exclude hidden elements

This uses [jQuery's implementation](https://github.com/jquery/jquery/blob/0402963845be8d71c4e8ddf65e7c055014739b60/src/css/hiddenVisibleSelectors.js#L10) under the hood.

This implements #5512.


src/modal/test/modal.spec.js
```diff
@@ -761,6 +761,52 @@ describe('$uibModal', function() {
...
+      expect(document.activeElement.getAttribute('id')).toBe('tab-focus-link3');
+
+      $('#tab-focus-button').css('display', 'none');
+      triggerKeyDown(angular.element(document.activeElement), 9, false);
+      expect(document.activeElement.getAttribute('id')).toBe('tab-focus-link1');
+
+      initialPage.remove();
+    });
+
+    it('should change focus to next non-hidden element when shift+tab is pressed', function() {
```
Foxandxss: focus to previos non-hidden

Foxandxss:
Works for me.

src/modal/test/modal.spec.js
wesleycho: Will merge with change

PR #5663, opened 2016-03-21
chore: update to Angular 1.5.2

Foxandxss:
Let's bump angular version so we are sure that our new changes are compatible with the latest version.


wesleycho:
LGTM

PR #5666, opened 2016-03-21
fix example for $tooltipProvider.setTriggers()

adam187:
argument of setTriggers should be an object.


wesleycho:
LGTM - good catch!

PR #5667, opened 2016-03-21
Properly apply right alignment when appendTo is used

vasilak:
When appendTo is used and the dropdown is right-aligned the positioning is off
because it is calculated relative to the window instead of the specified container.

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -217,7 +217,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
         if (!rightalign) {
           css.left = pos.left - appendOffset.left + 'px';
         } else {
-          css.right = window.innerWidth -
+          css.right = appendTo.prop('offsetWidth') -
             (pos.left - appendOffset.left + $element.prop('offsetWidth')) + 'px';
         }
       }


icfantv:
@RobJacobs, can you review this please?  Thanks.

RobJacobs:
@vasilak Having a plunk demonstrating the problem is extremely helpful to us.  That being said, I created a demo plunk [here](http://plnkr.co/edit/PSpTVAuS0LWxjVYTeKGl?p=preview) that uncovers another issue with the append-to element alignment - the top alignment is off.  In the dropdown scope.$watch('isOpen' function we could do some refactoring to call the $position.positionElements better (with the proper alignment) to take care of this issue and potentially implement auto positioning.  I'll create a PR for this shortly.

icfantv:
Apologies - I should have asked @vasilak to adhere to the issue template.  In the future, please do not summarily delete the template because you don't want to go through the work required.  The template is there to make helping you faster.  As a result, @RobJacobs had to put forth the time to figure out what was going on here.

In the future, issues entered that do not adhere to the template that are believed by the enterer to be bugs will be closed without further action.

vasilak:
Sorry guys, you 're right.
I will make sure to adhere to the template in the future.

PR #5675, opened 2016-03-22
chore(dropdown): avoid IE11/Edge rendering bug

### LINKED ISSUE #5595: uib-dropdown-menu broken in IE 11 and Edge
Hi,
IE 11 / Edge is not applying "list-style: none;" when class "dropdown-menu" is added dynamically.  
uib-dropdown-menu

I encountered this issue today when testing my site today with Windows 10 and Edge. I was able to verify this issue with Windows 7 and IE 11 in your [Dropdown demo](https://angular-ui.github.io/bootstrap/#/dropdown)

I solved issue by just adding 'dropdown-menu' class for `ul` element and not letting 'uib-dropdown-menu' set it dynamically.

Angular: 1.4.9

UIBS: 1.2.2

Bootstrap: 3.3.6

wesleycho:
- Add `dropdown-menu` class to avoid rendering bug in IE11/Edge

This fixes #5595.


icfantv:
The drop-down-menu directive already adds this class, so do we want to remove it from the directive too?

wesleycho:
No, this is just for the demo HTML.

icfantv:
Ah, crap.  I missed that.  Sorry.  LGTM.

Issue #5574, opened 2016-03-02
Split off datepicker popup into separate component


wesleycho:
Currently the datepicker popup is bundled with the inline datepicker - it would be beneficial to split it off to increase modularity.

RobJacobs:
I was thinking it would be nice to have a generic popup directive that would just handle the open state, positioning and trigger events.  Then allow for template content which we could then use for a datepicker, popover, template tooltips, typeahead, etc...  Tooltip could be scaled back to serve just basic tooltips.

icfantv:
I like that idea @RobJacobs.

wesleycho:
That could separated into a different module - maybe this one can be renamed to `uib-datepicker-dropdown`

wesleycho:
Or...this one can still be `uib-datepicker-popup`, and the new one could be `uib-datepicker-popover`.

wesleycho:
This task is complete, looks like I forgot to mark it for closure - closing now.

PR #5676, opened 2016-03-22
feat(datepickerPopup): split off popup from datepicker

wesleycho:
- Split off datepicker popup into new datepickerPopup module

BREAKING CHANGE: The datepicker popup is no longer a part of the
datepicker module, but now a part of the datepickerPopup module. Please
change code accordingly if including specific modules

This implements #5574.


Foxandxss:
Sounds good to me.

PR #5677, opened 2016-03-22
feat(datepicker): watch for changes when falsy

### LINKED ISSUE #5672: minDate is ignored/not working in inline datepicker
I don't know if it's a bug or it's by design. But in the inline datepicker, if a mindate is passed per datepicker-options, the value is ignored. In the datepicker popup, everything works fine. You can reproduce that behaviour on the sample page. I copied it and produced a plunkr.
http://plnkr.co/edit/lMKk9yYkD41QgQvrXbwQ?p=preview

Angular: 1.5.0

UIBS: 1.2.5

Bootstrap: 3.3.6

Browser: latest Firefox

wesleycho:
- When initial value is falsy, continue to watch for changes to update

This implements #5672.


icfantv:
Do we need a doc update for this?  If not, LGTM.

wesleycho:
It's already marked as watched, this just makes it behave more naturally according to the docs.

icfantv:
Cool.  LGTM.

PR #5689, opened 2016-03-24
fix(tab): fix tab remove when used with ng-repeat

brentahiti:
- Fix tab remove when tabs are destroyed in the middle using filters with ng-repeat for example


wesleycho:
This is a good PR - thanks for the work!

src/tabs/tabs.js
```diff
@@ -55,9 +55,15 @@ angular.module('ui.bootstrap.tabs', [])
   };
 
   ctrl.removeTab = function removeTab(tab) {
-    var index = findTabIndex(tab.index);
+    var index;
+    for (var i = 0; i < ctrl.tabs.length; i++) {
```
icfantv: forEach?

PR #5692, opened 2016-03-25
test(dateparser): add literal tests

wesleycho:
- Add broken tests for dateparser literals

This adds failing tests for anyone to uncomment and use to test a fix.

I made a mistake merging the literals implementation without testing it thoroughly - the implementation appears flawed, so we may need a big refactor. I am adding these tests so that this knowledge/refactor isn't lost, and so anyone else may try their hand if they are relying on this feature.


src/dateparser/test/dateparser.spec.js
```diff
@@ -560,24 +560,48 @@ describe('date parser', function() {
   });
 
   describe('with value literals', function() {
-    it('should work with multiple literals', function() {
-      expect(dateParser.parse('29 de January de 2013', 'd \'de\' MMMM \'de\' y')).toEqual(new Date(2013, 0, 29));
-    });
+    // describe('filter', function() {
```
deeg: Do you want to keep commented out code in here?

deeg: Instead of commented out code should you use xit?

deeg:
LGTM

Issue #5485, opened 2016-02-16
Timepicker: DST issue


joh-klein:
On the fall day of DST the time picker can't go forward past 2 AM

http://plnkr.co/edit/f1kiq5J89eOhBLgtstRT?p=preview

wesleycho:
I'm not seeing this - can you give me the exact steps you are taking to see the issue?

joh-klein:
Ah, I think this might happen, if the day to switch back from DST in your time zone differs from mine (2016-10-30). Please change `$scope.mytime` accordingly.

wesleycho:
Got it, finally was able to reproduce in the Pacific timezone: http://plnkr.co/edit/aJSWySs4N458sERn7yum?p=preview

This is an interesting bug. We likely need to add a check in the increment method in the timepicker & special timezone handling - the decrement one seems to work properly.

sokla:
I really want to know hot to use arrowkeys in Timepicker.

icfantv:
Closing in favor of #5957 as this will result in a complete rewrite of the widget, addressing all the outstanding issues.

PR #5693, opened 2016-03-25
docs(timepicker): add note about DST

wesleycho:
- Add note about DST, and declare explicit intent to not support
  concurrent usage of the date object with the timepicker and datepicker

See #5485 to understand the issue leading to this declaration.


deeg:
LGTM

PR #5697, opened 2016-03-27
fix(tab): allow tab indices to play nicely with ng-repeat

brentahiti:
- Fix how UibTabsetController handles tab indices and remove cached indices as a result
- Add watcher on tab index in case ng-repeat updates it

@wesleycho, I had to remove the cached indices as ng-repeat + filters were causing too much trouble. Not sure if it was used as an API? Since it appears in tests..

I'd like to think that my use case is pretty legit though

Here's a plunkr: https://plnkr.co/edit/BxLOJ1jT79WChVLhRreV?p=preview
You can just play around with the filter search box to see the bug occuring

### DIFF
--- a/src/tabs/tabs.js
+++ b/src/tabs/tabs.js
@@ -159,9 +159,11 @@ angular.module('ui.bootstrap.tabs', [])
     },
     controllerAs: 'tab',
     link: function(scope, elm, attrs, tabsetCtrl, transclude) {
+      var disableWatcher, indexWatcher;
+
       scope.disabled = false;
       if (attrs.disable) {
-        scope.$parent.$watch($parse(attrs.disable), function(value) {
+        disableWatcher = scope.$parent.$watch($parse(attrs.disable), function(value) {
           scope.disabled = !! value;
         });
       }
@@ -174,7 +176,7 @@ angular.module('ui.bootstrap.tabs', [])
         }
       } else {
         // watch when index is ng-repeat $index
-        scope.$parent.$watch($parse(attrs.index), function() {
+        indexWatcher = scope.$parent.$watch($parse(attrs.index), function() {
           tabsetCtrl.updateActiveIndex(scope);
         });
       }
@@ -192,6 +194,14 @@ angular.module('ui.bootstrap.tabs', [])
       tabsetCtrl.addTab(scope);
       scope.$on('$destroy', function() {
         tabsetCtrl.removeTab(scope);
+        if (disableWatcher) {
+          disableWatcher();
+          disableWatcher = null;
+        }
+        if (indexWatcher) {
+          indexWatcher();
+          indexWatcher = null;
+        }
       });
 
       //We need to transclude later, once the content container is ready.
--- a/src/tabs/tabs.js
+++ b/src/tabs/tabs.js
@@ -7,24 +7,24 @@ angular.module('ui.bootstrap.tabs', [])
 
   ctrl.select = function(index, evt) {
     if (!destroyed) {
-      var previousIndex = findTabIndex(oldIndex);
+      var previousIndex = findTabByIndex(oldIndex);
       var previousSelected = ctrl.tabs[previousIndex];
       if (previousSelected) {
-        previousSelected.tab.onDeselect({
+        previousSelected.onDeselect({
           $event: evt
         });
         if (evt && evt.isDefaultPrevented()) {
           return;
         }
-        previousSelected.tab.active = false;
+        previousSelected.active = false;
       }
 
       var selected = ctrl.tabs[index];
       if (selected) {
-        selected.tab.onSelect({
+        selected.onSelect({
           $event: evt
         });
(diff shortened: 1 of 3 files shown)


src/tabs/tabs.js
```diff
@@ -152,6 +169,11 @@ angular.module('ui.bootstrap.tabs', [])
         } else {
           scope.index = 0;
         }
+      } else {
+        // watch when index is ng-repeat $index
+        scope.$parent.$watch($parse(attrs.index), function() {
```
wesleycho: This needs to garbage collect the watcher when the directive scope is destroyed.

wesleycho:
This mostly LGTM, minus that minor issue - I had originally implemented the cached index as I did not want the index to be dynamic. I'm still adamant that dynamically changing the index is a supremely bad idea, but the code change here is for the most part fine.

RobJacobs:
Why not just add an index property to your tab collection and use that instead of the Ng-repeat index? I would think that would be more meaningful in the select callback as well.

brentahiti:
The main purpose for this PR is for the use case I mentioned above, as I'm pretty sure other people are going to try doing the same thing. It's to make the tab work in any case and it wasn't clear to me that it was a design choice at first.
I tested it extensively on my side but if you guys think that's a super bad idea, then the docs should be updated at least

wesleycho:
I'm fine with this PR - it does a lot of cleanup I want to do.

RobJacobs:
The point I was trying to make is that it does not make sense to me to use the ng-repeat $index to track the active tab.  The way I understand the intended use, tab index is used to track the activeTab so you can respond to a tab selection and take action based on which tab is active/selected.  If the ng-repeat $index is used and the view collection is filtered, the ng-repeat $index no longer matches the position of the active tab in the controller tabs collection.  Here is the original [plunk](https://plnkr.co/edit/7OtTy7sMjYUSXm8JDBpC?p=preview) with this PR included.  Filter on 'o', select the 'Megatron' tab, in the activeTab1 watch where I try to use the activeTab1 to determine which tab is selected, it's based on the ng-repeat $index and say's that the active tab is 'Casablanca' with an index of 1.

So, if you care about tracking the active tab, you should use a property in your binding collection for the most predictable results.  If you don't care about tracking the active tab, leave the index attribute off the tab directive and the index attribute with a unique value will be created for you.

We talked about removing the number restriction on the tab index so folks could use any property from their binding collection (name) and not have to add a specific on for the index.

brentahiti:
Ok I didn't get this intended use of the activeTab from reading the docs, as it seemed that it would be only used to set the active tab and the index in each tab would allow the directive to identify it. And we can easily infer the selected tab from the select() binding.

As for the behavior in the plunk, the activeTab has an index of 1 relative to the array generated by ng-repeat+filter, so imo this is intended behavior

RobJacobs:
The select binding is on the tabset directive and only returns the $event, not the selected tab.  Relevant code: https://github.com/angular-ui/bootstrap/blob/master/src/tabs/tabs.js#L21

wesleycho:
This needs to be rebased.

src/tabs/tabs.js
```diff
@@ -72,9 +63,35 @@ angular.module('ui.bootstrap.tabs', [])
     ctrl.tabs.splice(index, 1);
   };
 
+  ctrl.updateActiveIndex = function updateActiveIndex(tab) {
+    if (tab.active) {
+      ctrl.active = tab.index;
+      oldIndex = tab.index;
+      ctrl.tabs.sort(function(t1, t2) {
```
wesleycho: What is the reason for sorting the tabs?

src/tabs/tabs.js
```diff
@@ -160,21 +184,19 @@ angular.module('ui.bootstrap.tabs', [])
...
-          tabsetCtrl.select(index, evt);
+          tabsetCtrl.select(tabsetCtrl.findTabIndex(scope), evt);
         }
       };
 
       tabsetCtrl.addTab(scope);
       scope.$on('$destroy', function() {
         tabsetCtrl.removeTab(scope);
+        if (disableWatcher) {
+          disableWatcher();
```
wesleycho: Need to set to `null` after executing to avoid holding references in memory (gc issue)

src/tabs/tabs.js
```diff
@@ -160,21 +184,19 @@ angular.module('ui.bootstrap.tabs', [])
...
+        }
+        if (indexWatcher) {
+          indexWatcher();
```
wesleycho: Ditto

src/tabs/tabs.js
brentahiti: In some cases, the tabs would get out of order. Although, the tabs don't necessarily need to get sorted at all in the first place but the tests assume an ordering of tabs

src/tabs/tabs.js
brentahiti: Alright changes made

PR #5698, opened 2016-03-27
fix(loading): "index-nocss" requires other "index-nocss" files

onlywei:
This prevents require()-ing of transitive css files by making sure
index-nocss files require other index-nocss files when available.
Modules: dropdown and popover, which had transitive css dependencies
now also gain "index-nocss" files of their own.


wesleycho:
LGTM - thanks for this work!

PR #5701, opened 2016-03-28
Documentation: Fix mixed content issue due to HTTP images in carousel demo

pjohnmeyer:
> I don't know if this is the "right" way to fix the documentation, so please let me know if I'm _Doing It Wrong_ -- pj

The live carousel documentation was emitting mixed content warnings
because the protocol was hard-coded to http. Changing the image paths
to be protocol-agnostic did not work because lorempixel.com's cert
has been expired for more than a year.

It seems to me that saving hand-chosen images in the asset path would be
a better long-term fix, but this should eliminate the warnings for
now.


wesleycho:
Can you also open a PR with the modification to `master` as well? The code [here](https://github.com/angular-ui/bootstrap/blob/master/src/carousel/docs/demo.js#L11)

pjohnmeyer:
@wesleycho done in #6144; sorry for the long delay in addressing your comment.

pjohnmeyer:
I've added newer versions to this pull. Is this necessary, or does the entire site including old versions get regenerated based on the changes to master?

PR #5707, opened 2016-03-29
feat(typeahead): always visible popup

netsurferj:
- Add functionality to not clear results when clicking away from editor
- This is useful when implemented as always visible result list (not a popup)

### DIFF
--- a/src/typeahead/docs/readme.md
+++ b/src/typeahead/docs/readme.md
@@ -91,6 +91,11 @@ This directive works with promises, meaning you can retrieve matches using the `
   _(Default: `false`)_ -
   On blur, select the currently highlighted match.
 
+* `typeahead-close-on-blur`
+  <small class="badge">$</small>
+  _(Default: `true`)_ -
+  When clicking away from editor, close results popup.
+
 * `typeahead-select-on-exact`
   <small class="badge">$</small>
   _(Default: `false`)_ -
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -824,6 +824,17 @@ describe('typeahead tests', function() {
 
   describe('non-regressions tests', function() {
 
+    it('feat 5707 - keeps matches popup open on click outside typeahead when close-on-blur is false', function() {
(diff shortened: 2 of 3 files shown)


wesleycho:
Couldn't one use `min-length="0"` to accomplish this?

netsurferj:
@wesleycho Unfortunately not. When displaying the results in an always visible list (not a popup) the results are cleared when clicking outside the text editor. The desired effect is to have the results always visible when clicking other components on the screen.

wesleycho:
What's the use case for always having it visible?

Also this needs documentation updates, as well as unit tests.

netsurferj:
- Added docs and unit test
- I have created a plunkr use case, where one might want the search results to always remain visible (and not to disappear when clicking other items on the screen): http://plnkr.co/edit/2kXtXE0D0vdn6pH7l2Az?p=preview

wesleycho:
Hmm question - is this not possible to do using `typeahead-is-open` and watch for it changing to false & just reset it to true?

netsurferj:
@wesleycho The issue lies in the dismissClickHandler (issue #3973) that calls resetMatches when clicking on other elements in the document. There is currently no way to disable this functionality. Perhaps the property name: CloseOnBlur is unclear. Do you have any other suggestions for the property name?

JeffSallans:
@wesleycho It is not possible to use the method you described with `typeahead-is-open`.  That property is only **set** by Angular UI Typeahead and it is not used to determine if the matches should show.  Below shows the only reference to `typeahead-is-open` in the code; you can see Angular UI creates a setter.  No getter or other use of attr.typeaheadIsOpen exists.

```javascript
    //binding to a variable that indicates if dropdown is open
    var isOpenSetter = $parse(attrs.typeaheadIsOpen).assign || angular.noop;
```

netsurferj:
I was able to bypass this functionality in Angular using an extended directive and monitoring the actual values assigned to the list:

if new list is empty and old list was not empty and there are results, re-assign scope.matches to the old list. I determine if there are results or not by extending the scope in my own controller and assigning bool values. A bit of a dirty approach, but avoids the limitation that typeahead assumes it is in a popup control and not a normal div.

PR #5708, opened 2016-03-29
feat(carousel): hide prev/next if not loop and indexes at bounds

plondon:


template/carousel/carousel.html
```diff
@@ -1,10 +1,10 @@
 <div ng-mouseenter="pause()" ng-mouseleave="play()" class="carousel" ng-swipe-right="prev()" ng-swipe-left="next()">
   <div class="carousel-inner" ng-transclude></div>
-  <a role="button" href class="left carousel-control" ng-click="prev()" ng-show="slides.length > 1">
+  <a role="button" href class="left carousel-control" ng-click="prev()" ng-show="slides.length > 1" ng-hide="active === 0 && noWrap()">
```
icfantv: these changes add two extra watches per slide - can you please investigate combining the ng-show/ng-hide?

RobJacobs:
I'd prefer ng-if over ng-hide/show.  We also need to consider, does it make for a jarring experience with elements being removed rather than just disabling them?

plondon:
Users can always override the display: none !important with a more specific display: block !important; opacity: 0.5 if they want.

plondon:
Also a fan of disable, that way users can choose the more jarring experience if they want with CSS. Since `disabled` is not a valid attribute for an anchor tag is it alright to swap `<a role="button">` for the regular `<button>`?

icfantv:
@plondon are there accessibility issues with doing that or is that the preferred mechanism?  IIRC, using the `btn-link` class had accessibility ramifications.

plondon:
@icfantv `<button>` has the aria `role="button"` by default, so it should act as `<a role="button">` from an accessibility perspective.

realalexhomer:
@plondon @icfantv Yes that is correct, screen readers prefer <button> over the a tag so it is better to use it when possible

plondon:
Since previous and next buttons are "disabled" by the noWrap logic I think we can leave them as anchor tag's.

Instead of `button disabled="true"` as we were discussing above I used `isPrevDisabled()` and `isNextDisabled()` to toggled a `disabled` class, leaving it up to the user to hide, show, fade, or whatever the button's him/herself.

I'll squash these commits if you all agree with this approach.

wesleycho:
This looks solid to me - anyone else have thoughts?

RobJacobs:
LGTM

PR #5711, opened 2016-03-30
fix(modal): body content shift

### LINKED ISSUE #2631: the background move a little when opening a modal.

RobJacobs:
Implements TWBS body padding fix to keep content in an element with a container class from shifting when the body overflow is set to hidden with the modal-open class.

Fixes #2631


deeg:
Thanks for doing this! We might want to add some tests for the new logic in position service?

RobJacobs:
Tests are difficult when it comes to DOM interaction.  You will notice most of the tests in the position module actually mock the DOM aspects (see offset test where it returns a mocked clientRect).  I don't know if testing against mocked values really does anything for us?

wesleycho:
We need to eventually modify the build system to run tests via SauceLabs/Browserstack and run positioning tests across browsers.

wesleycho:
This LGTM

PR #5713, opened 2016-03-31
feat(tab): allow active index string

### LINKED ISSUE #5687: ui-tab didn't displayed correctly when use string index
### Bug description:
I am refering the doc:
>`index` - Tab index. Must be unique

it's clearly work when use tab without index attribute, so i made the reproduction both.
Here the [reproduction](http://run.plnkr.co/plunks/cCUWgu/)

angular version `1.5.0`
uib `1.2.5`
bootstrap `3.3.6`
ngAnimate `1.4.2`

deeg:
Allow Tab index to be a string

Closes #5687
Closes #5577


deeg:
There are two other isNumber checks in the tabs component which did not have isString checks added from the pull request I started with.

@wesleycho, did we need to add some is string checks there as well?

src/tabs/tabs.js
```diff
@@ -85,7 +85,7 @@ angular.module('ui.bootstrap.tabs', [])
 
   function findTabIndex(index) {
     for (var i = 0; i < ctrl.tabs.length; i++) {
-      if (ctrl.tabs[i].index === index) {
+      if (ctrl.tabs[i].index === +index) {
```
deeg: This change was not in the pull request I started from, but it seemed needed.

deeg:
[Here is the plunker forked with the fix](https://plnkr.co/edit/4x2QDx?p=preview)

I haven't been following along really, but I'm not sure it is actually working. Was the previous PR ready to go besides test, or should I dig into this deeper?

RobJacobs:
@deeg I created a PR to demo my thoughts on how to approach this is it was easier than trying to write it out.  Here is a [plunk](http://plnkr.co/edit/drIzTGdobDkEc7bkjJ4p?p=preview) with the changes in action.

deeg:
@RobJacobs, sweet thanks for doing that!

deeg:
@wesleycho, @RobJacobs, is everyone good with Rob's solution in the commit referenced on this PR?

If so I can incorporate his changes into this PR to get this closed.

RobJacobs:
@deeg Like I mentioned, in the uibTab directive - link function where we try and add an index if none is defined needs to be re-worked.  Right now, it is assumed the indexes are numbers, but we may want to generate a random string (4 char?) and check that it is not already assigned to another tab index.  Or if you can think of a better way?

src/tabs/tabs.js
wesleycho: This seems like a strange change - do you know why this is as you have it?

icfantv:
@Foxandxss, @wesleycho, i don't want to lose momentum on this PR - do you guys have any thoughts to add here regarding dynamic ID generation?  We do this already in other components (datepicker and accordion).

mathielen:
+1

icfantv:
@mathielen please do not just +1 issues.  It adds no intrinsic value and only clutters up the thread.  If you would like to show your support for an issue please use GitHub's emoji feature.  It's the smiley icon next to the pencil in the upper right corner.

This is a work in progress and is nearly complete.

src/tabs/test/tabs.spec.js
```diff
@@ -168,6 +168,16 @@ describe('tabs', function() {
       expect(titles().length).toBe(scope.tabs.length);
       expectTabActive(scope.tabs[2]);
     });
+
+    it("should watch active state", function() {
```
wesleycho: Should use single quotes here

src/tabs/test/tabs.spec.js
```diff
@@ -168,6 +168,16 @@ describe('tabs', function() {
...
+      var controller = elm.controller('uibTabset');
+      spyOn(controller, "select");
```
wesleycho: Ditto

src/tabs/test/tabs.spec.js
```diff
@@ -168,6 +168,16 @@ describe('tabs', function() {
...
+      expect(titles().length).toBe(scope.tabs.length);
+      expectTabActive(scope.tabs[2]);
+      scope.active = "7";
```
wesleycho: Ditto

wesleycho:
@deeg can you make the needed adjustments and answer my question above?

With #5732, the 1.3 release work is complete, so if we want this in, it needs to be ready soon.

src/tabs/tabs.js
deeg: This was getting called from line 76 where we were now allowing the value to be a string.

The comparison wouldn't work for 1 === "1".

wesleycho: But why is the type getting mutated?

deeg:
@wesleycho, this PR wasn't working. I can make the changes you asked for and incorporate @RobJacobs's changes which are better and actually work, but there was still one outstanding question.

I believe it was about the auto generation of the index. Barring that answer, I don't mind cleaning all this up and getting it ready to go.

icfantv:
@deeg, were you still working on a fix for this or did we abandon this feature?

deeg:
@icfantv I thought I was still waiting on a decision. I'm happy to wrap this up, but @RobJacobs had a branch with a better fix and a question, and I thought we were waiting for feedback.

icfantv:
oh.  hmmm.  ok.  @RobJacobs do you recall what, if anything, we're waiting for?  thanks.

RobJacobs:
Just wanted some feedback on what would be the best approach for the indexes the directive auto generates.  Use case is if no index is defined on the tab, the tabset will never set it to active.  Previously, the auto generated tab index made the assumption the index would be a number and found the max number in the index collection and added 1.  Now with index as a string, that won't work...

deeg:
Sorry I didn't mean to close this, I deleted the branch by mistake. Will re-open once we decide on a path forward.

icfantv:
it might be too late for this, but in the ng2 port we're going to use the actual tab reference as the "index" - what do you think about that instead or in addition to?

PR #5714, opened 2016-03-31
Documentation : Datepicker afterTomorrow fixed

tteke:
After tomorrow shows the wrong date if the date is the last day of the month.


wesleycho:
This is confusing - `afterTomorrow` is two days after today, JS automatically does month adjustments. Could something else be going on here?

tteke:
take today for example when afterTomorrow first gets created from new Date() it is 31 March 2016. Then we set it's date to (tomorrow.getDate() + 1) which is 2 since javascript automatically corrected tomorrow(1 April 2016)'s date variable. However, the month variable in the afterTomorrow is still inherited from new Date(); which is still 31 March. So afterTomorrow is now equals 2 March 2016 instead of 2 April 2016.

wesleycho:
Ahh, now I see what was going on - this new change is good, thanks for this catch!

Issue #5719, opened 2016-03-31
Popover displays tooltip on itself and every child


ValentinH:
This is a duplicate of #5598 which was poorly described.

### Bug description:
When using the `popover-title` option, a `title` attribute is added to the popover element which results in a native browser tooltip being displayed on all the popup when the user stops moving his mouse. 
By default, this is pretty weird and it gets even worse when a `uib-tooltip` is used inside the popup:

![image](https://cloud.githubusercontent.com/assets/2678610/14180352/f0e0b6fe-f762-11e5-9af2-2802de2d0cf4.png)

### Link to minimally-working plunker that reproduces the issue:
http://plnkr.co/edit/DGz3BZ4zQwagCFbXGw4T?p=preview

### Version of Angular, UIBS, and Bootstrap

Angular: 1.5.0

UIBS: 1.2.5

Bootstrap: 3.3.6

RobJacobs:
@ValentinH Thanks for reporting this with all the relevant info, it makes it so much easier to diagnose issues.  Here is a [plunk](http://plnkr.co/edit/OXIGXlphNkkQmJ5LC1wq?p=preview) with a proposed change, I will be creating a PR shortly.

PR #5721, opened 2016-03-31
fix(popover): rename title attribute

RobJacobs:
Using an attribute of title on the tooltip template was causing the browser tooltip to display.

Fixes: #5719


wesleycho:
Needs docs updating.

ValentinH:
Thanks that was quick!

Issue #5716, opened 2016-03-31
uib-tab settings: click()


yasinkocak:
I want to have a uib-tab setting `click()` that called an optional expression without change the tabcontent.

There is already `select()` but this changes tab content.

icfantv:
I think what you may want is a way to call `event.preventDefault` to keep the tab from changing.  Would this work?

Currently this feature does not exist in the library but has been requested before and is certainly worth a conversation.

yasinkocak:
I have a table on activated tab 1

And i want a tabset that changes my records at the table when it's clicked without changing the tab.

If i use select() this change also the tab

icfantv:
Sorry, you want your table to change when _what_ is clicked?

yasinkocak:
```
<uib-tabset>
  <uib-tab heading="table-1">
    <table>...<table>
  </uib-tab>
  <uib-tab heading="table-2">
    <table>..</table>
  </uib-table>
  <uib-tab class="pull-right" select="$ctrl.changeRecords()">
    <uib-tab-heading>Change table records</uib-tab-heading>
  </uib-tab>
</uib-tabset>
```

As u can see i have 3 tab.
First 2 tab have a content with table. And the last one is clickable tab without content.

When i use select() the records from the table are changed. But the tab of "table-1" is disactivated and get blanco tab content of the last tab.

I want to click on the last tab to call an expression without disabling current tab.

wesleycho:
One temporary way one can do this I think is using ng-click to cache the prior tab value, and use $timeout to change it after.

I think @icfantv's suggestion would work for the library itself.

icfantv:
In the interim, @yasinkocak, you can override the tab template and have your `ng-click` supercede the original one.  Something like:
```
<li ng-class="[{active: active, disabled: disabled}, classes]" class="uib-tab nav-item">
  <a href ng-click="doSomething() && select($event)" class="nav-link" uib-tab-heading-transclude>{{heading}}</a>
</li>
```

And if `doSomething()` returns `false` the tabs won't change because `select(...)` won't be fired.  I give you that this is not ideal, but until we wire in the event propagation stoppage, this should keep you moving.

To override the template, simply specify a `template-url` property on your tab or you can override the tab template already in the `$templateCache`.  If you do the latter, the key you need to override is `uib/template/tabs/tab.html`.

I'm going to close this as a support issue but will open one to track the stoppage of the event propagation.

PR #5723, opened 2016-03-31
fix(tab): add support for tab deselect prevention

### LINKED ISSUE #5720: Update tab directive's select method to allow for event.stopPropagation()
The current `select` method on the `uibTab` directive does not allow the user to stop the selection process via `event.stopPropagation`.  We need to add an `if` block to check for this and not change tabs if the user's stopped the propagation.

Here's the relevant code:  https://github.com/angular-ui/bootstrap/blob/master/src/tabs/tabs.js#L161-L173

We need to update the documentation as well.

This addresses comments/concerns in #2715, #4836, and #5716.

icfantv:
* add ability for user to prevent currently selected tab's deselection by
calling `$event.preventDefault()` in tab's `deselect` callback.

Fixes #5720
Addresses #2715, #4836, and #5716.


wesleycho:
LGTM

PR #5733, opened 2016-04-04
fix(modal): ensure correct index is set for new modals

### LINKED ISSUE #5670: multiple modals & z-index issue
### Bug description:
Let's say multiple modals are opened at the same time - modals [A, B, C, D], opened in this order. If I close the first opened modal A, then open another modal E -> [B, C, D, E]. The newly opened modal E has the same index & z-index as the previous modal D. With the current reversed order of modals in DOM & same z-index, modal E will not be visible (will be covered by modal D).

[https://plnkr.co/edit/GNZtzlaxDM0X8STp8fTD](https://plnkr.co/edit/GNZtzlaxDM0X8STp8fTD)
In the plunkr, i set the limit to 5. So the issue is obvious after 5 opened modals...

Angular 1.4.9
UIBS 1.1.2
Bootstrap 3.3.6

josetaira:
Fix how the index for the modal and the backdrop are set, depending on
what index value the previous top modal has.

Closes #5670


src/modal/modal.js
```diff
@@ -278,6 +280,12 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
             topBackdropIndex = i;
           }
         }
+
+        // If any backdrop exist, ensure that it's index is always
+        // right below the top modal
+        if(topBackdropIndex > -1 && topBackdropIndex < topModalIndex) {
```
wesleycho: Add a space after `if`

src/modal/modal.js
```diff
@@ -293,6 +301,10 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
 
         //clean up the stack
         openedWindows.remove(modalInstance);
+        previousTopOpenedModal = openedWindows.top();
+        if(previousTopOpenedModal) {
```
wesleycho: Ditto

src/modal/modal.js
```diff
@@ -293,6 +301,10 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
...
+          topModalIndex = parseInt(previousTopOpenedModal.value.modalDomEl.attr("index"), 10);
```
wesleycho: Use single quotes.

wesleycho:
Can you add a unit test for this?

josetaira:
For the tests, no idea how grunt works, but I guess I can learn it. Do you have any templates or guidelines on how tests should be made?

icfantv:
@josetaira, you don't need to learn grunt, you need to look at the `model.spec.js` file and add your tests to that file and they'll run as part of the build.  there should be enough examples in there (and our other test files) to figure it out.

if you want to run a specific test or test suite, simply prefix the `it` or `describe` with an `f` (i.e. `fit` or `fdescribe`) and only that test will be run.  our linter will reject this as invalid syntax so you can run `karma start` to run the karma tests independently.

josetaira:
@icfantv Thanks for the info. Working on the unit tests now. Seems I did have to use grunt to create some templates (run `grunt default` before any tests). I'll add a commit for the test later on.

icfantv:
@josetaira good deal.

josetaira:
@icfantv @wesleycho tests committed.

wesleycho:
This LGTM - thanks for the effort!

PR #5734, opened 2016-04-04
feat(timepicker): Added support for having UTC time in the timepicker

filipkis:
Hi, 

I've implemented this small feature that adds another option to timepicker called `utcTime` which when set will display the time in the picker based on UTC value. In other words, no matter what timezone the date is created, the times will always be the same. 

This is a tiny change that should not break any of the existing code as it only takes effect when the attribute is set to true (false by default).

Hope this can make it to official build so I don't have to maintain my own :)


deeg:
I will let the rest of the team decide if they wanted to incorporate this feature, but I know unit tests will be needed before this PR could be merged.

icfantv:
i'm inclined to reject this as being pretty heavy-handed and something that the user can (and probably should) do on their end.

wesleycho:
I'm wary about this as well - this seems like it would bloat the component for marginal gains. One should think of a timepicker as a storage of hours and minutes information - we probably should never have used a date object and should have just used a POJO instead.

icfantv:
@wesleycho, I'm ok with a breaking change to switch to a pojo - with an interim support for both....maybe, but that would probably suck....

PR #5735, opened 2016-04-04
feat(dateparser): use 68 as the pivot year when year format is yy.

RobJacobs:
Use the same pivot year (68) for 2 digit date years as moment.js.  68 becomes 2068, 69 becomes 1969.


wesleycho:
Works for me, although this probably will become outdated in a few decades :P .

icfantv:
If this library is still valid in several decades, we have loads of other problems.... :-)

PR #5738, opened 2016-04-04
fix(tab): fix event handler call

### LINKED ISSUE #5720: Update tab directive's select method to allow for event.stopPropagation()

icfantv:
* switch event handler call to look at `event.isDefaultPrevented()` instead of
  `event.defaultPrevented` as angular has implemented that method on jqlite
  events.

Fixes #5720


wesleycho:
LGTM

deeg:
Did this not need an associate test change?

icfantv:
@deeg the test was already in there.  the issue is that we don't have separate karma files for testing with and without jquery.  i opened an issue to track that.  also see the thread in our slack for more details.

PR #5745, opened 2016-04-05
fix(datepickerPopup): add missing popup module.

aitboudad:


wesleycho:
Thanks for this - I'll release this immediately.

PR #5756, opened 2016-04-06
fix(site): Allow FastClick to be blocked

christopherthielen:
Some users' adblock and virus protection is accidentally blocking third party `fastclick.js`.

Put a guard around the line that invokes the FastClick api


icfantv:
Dude, this is brilliant.  LGTM.

PR #5761, opened 2016-04-06
fix(datepicker): ensure datepicker is focused

### LINKED ISSUE #5754: Datepicker keyboard navigation on Chrome doesn't work

wesleycho:
- When date is selected, ensure datepicker is selected in Chrome

Plunker with fix in action [here](http://plnkr.co/edit/1f5vkAXNgHYs1G3F3SFw?p=preview).

This fixes #5754


deeg:
This LGTM if you don't want a test added.

I wonder why it was working for me on an older version of OSX.

wesleycho:
Not going to bother right now since this requires infrastructure for cross-browser testing to even write a meaningful test.

frenkelor:
Hi @wesleycho 
This fix cause a regression with the focused element in uid-datepicker-popup which gives the wrong element the focus.

**way to reproduce:**
close-on-date-selection=true
datepicker-append-to-body=true

1. click on input to select a date
2. select a date from the list 

**Observe :**

On [line](https://github.com/angular-ui/bootstrap/blob/7e320e0e92fd4fffe82eb120c60dd0543e9a0bd8/src/datepickerPopup/popup.js#L218) the focus suppose to return to the input element but after your fix the focus is being set to the table which is being removed from  the DOM  after clicking on the button and the focus is passed to the parent element the body.

So i returned to the original [issue #5754 ](https://github.com/angular-ui/bootstrap/issues/5754) and tried to understand why ng-keydown is not being fired.
I realized that after clicking on date button the focused element is the button itself and because ng-keydown is being set to `self.element[0]` keydown event is not being invoked.

I managed to come with 2 solutions to this problem:
1. you can add to the element that holds the keydown a `contenteditable="true"` attribute and it's solving the problems because now ng-keydown is being invoked  on any of `self.element[0]` children's.
2. you can set the keydown with jquery selector that includes `self.element[0];` children's and the `self.element[0]` itself.

I think the first approach it's better then the other.
i created [this plunker](http://embed.plnkr.co/9zwrjx1iQCAIa7S5PwEd/) to illustrate the first approach.

Do you want me to create a pull request for that fix?

wesleycho:
Sure, file a PR with the first approach - I'm a little tired, so I'm a bit confused. It would be better to get a sense of what you mean in code.

frenkelor:
@wesleycho , i created a [PR](https://github.com/angular-ui/bootstrap/pull/6069)

udayraj90:
@frenkelor  can you please explain me exactly where to put this code in which file.

frenkelor:
after another iteration on that i decided  to leave this bug my fix is not working for IE 11

PR #5769, opened 2016-04-07
fix(typeahead): use $setViewValue on blur

### LINKED ISSUE #5694: typeahead-editable, typeahead-min-length and required validation are not syncing with each other

### Bug description: 
Setting the option typeahead-editable="false", typeahead-min-length="3" and required on an input text does not trigger the required error validation.  It's hard to explain but I have created a plunker to demonstrate it.  Please see link below.

1. When first loaded, enter "ben" for the input and tab out or click outside of input.  All is fine.  error object shows nothing.
2. Click on input and erase "ben".  All is fine, error shows {"required":true} as expected.
3. Enter "be" and tab out. The field is clear.  Here, the error object is blank and the field is valid.  This shouldn't be the case.  The field is blank therefore it should be in the invalid state.  This won't work for me since I need to display the error message.
4. Enter "be" again and tab out.  The field is clear again.  But now, the error object shows {"editable":true,"parse":true} and invalid is true.  The problem here is the error messages is based upon the required validation.  As you can see here, error object doesn't have {"required" : true} therefore the error message doesn't show up.  However, invalid is set to true.  I suppose one could just use myForm.myName.$invalid and display the error message.  This though won't work for me since I have more than one validation on the input field.  The field been invalid doesn't tell me the right error to display.  For simplicity, I only put one validation for demonstration only.  

Step 3 and 4 are both issues I need to resolve.  Greatly appreciated if someone can help me out here.

http://plnkr.co/edit/xQ6OoSd18xB1U4V7XUEy

Angular: 1.5.2

UIBS: 1.2.5

Bootstrap: 3.3.6

pdesgarets:
On blur, use $setViewValue instead of setting directly $viewValue to '', to make new value go through $validators.
Especially useful when required directive is used. 

May fix #5694 (I'm not sure I have understood what was your problem, @sv311 ); updated plnkr : http://plnkr.co/edit/tNrdiMzI9zCeMdrRSSS9?p=preview


wesleycho:
This LGTM - setting the view value directly may have been a result of some legacy code.

PR #5785, opened 2016-04-12
fix(tooltip): arrow position

### LINKED ISSUE #5779: Improve tooltip "on placement" functionality
Came across what I think might be a slight improvement to the uib-tooltip module. I noticed on the Popup demo that when someone selects a new placement to use, the new one doesn't display correctly if there was a popup already visible. That is, the little arrow indicator doesn't seem to update correctly in any of the cases. It isn't shown or is shown within the body of the tooltip. In other words, the popover isn't updating automatically (correctly) when the new selected placement is fired. Maybe it can be fixed, but maybe this will suffice. I just added this line of code to the tooltip.js file starting at line 450:

if (ttScope.isOpen) { hide(); } ...as so:
```
observers.push(
   attrs.$observe(prefix + 'Placement', function(val) {
       if (ttScope.isOpen) { hide(); }    // added this
       ttScope.placement = val ? val : options.placement;
       var placement = $position.parsePlacement(ttScope.placement);
      lastPlacement = placement[1] ? placement[0] + '-' + placement[1] : placement[0];
      if (ttScope.isOpen) {
          positionTooltip();
      }
  })
);
```
All it does is close the Popup Placement Selection Demo displayed popup so that when the popup demo button is clicked again, the indicator arrow shows correctly (as it does for any new click). 

Example: https://angular-ui.github.io/bootstrap/#/popover (at least in Google Chrome)

Angular: 1.5.3

UIBS: 1.3.1

Bootstrap: 3.3.6

RobJacobs:
- Ensure the arrow position gets recalculated if the tooltip position changes while open.

Fixes #5779


icfantv:
@RobJacobs is there a way to add a test for this?  If not, LGTM.

wesleycho:
LGTM

PR #5787, opened 2016-04-12
fix(dropdown): stop esc keydown event

### LINKED ISSUE #5778: Pressing ESC with a dropdown inside a modal dialog
### Bug description:
If I have a dropdown widget inside a modal dialog and the dropdown is open, and then I press Escape, it closes the dialog. I would expect pressing Escape to close the dropdown only and not the dialog as well. If the dropdown is closed, pressing Escape would then close the dialog. Could this be addressed by calling evt.stopPropagation() when handling the escape key in dropdown.js?

http://plnkr.co/edit/T2gDRud9wpg8XpbLu3Yc?p=preview

Angular: 1.5.3

UIBS: 1.3.1

Bootstrap: 3.3.6

deeg:
Fixes #5778

BREAKING CHANGE: Stops propagation of keydown event when escape key is pressed. Removes keydown event from the document and moves it to the dropdown element.

[Plunker with fix](http://plnkr.co/edit/DvljrbTp8uCpQWDgIvTW?p=preview)

### DIFF
--- a/src/dropdown/test/dropdown.spec.js
+++ b/src/dropdown/test/dropdown.spec.js
@@ -26,8 +26,11 @@ describe('uib-dropdown', function() {
 
   var triggerKeyDown = function (element, keyCode) {
     var e = $.Event('keydown');
+    spyOn(e, 'stopPropagation');
+    e.stopPropagation.and.callThrough();
     e.which = keyCode;
     element.trigger(e);
+    return e;
   };
 
   describe('basic', function() {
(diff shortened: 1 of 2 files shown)


deeg:
@wesleycho, not sure if switching the keydown event from document to element will have any ramifications I didn't think about.

The reason I did this was because the document event was catching after the modal escape key event had already fired so stopPropagation didn't do anything. Couldn't think of a nicer way to solve this than to move the event over to the element.

src/dropdown/test/dropdown.spec.js
```diff
@@ -26,8 +26,10 @@ describe('uib-dropdown', function() {
 
   var triggerKeyDown = function (element, keyCode) {
     var e = $.Event('keydown');
+    spyOn(e, "stopPropagation");
```
wesleycho: Use single quotes - also add `.and.callThrough()`

wesleycho:
I think that switch is ok, judging by the code and playing around with the Plunker.

I'm ok with this change.

deeg:
@wesleycho, sounds good, thanks.

I made the requested changes.

wesleycho:
LGTM if it passes the tests

PR #5790, opened 2016-04-12
fix(modal): scroll padding only added once

### LINKED ISSUE #5789: Body shifts to the right when opening a modal from within another modal
### Bug description:
When there's vertical scrolling on the body and we open a new modal, UIBS (since v. 1.3) hides the scrollbar and adds a right padding to the body to compensate (to avoid the annoying body shift issue).
This padding is removed when the modal is closed.

However, when opening a modal from another modal, **the padding is added twice** (incorrectly, since we only need to add the padding once), and removed only once when the second modal is closed. This leaves the body with a right padding set and the content shifted.

Please see the attached plunker for an example.

http://plnkr.co/edit/g3EqeQYlz97xCUrMVkD1?p=preview

To reproduce:
1. Click the open modal link (right aligned to better demonstrate the right padding on the body).
2. Click Open second modal to open another modal. Notice the right padding on the body was increased.
3. Close both modals. This leaves the padding that was added after opening the first modal.

Angular: 1.5.3

UIBS: 1.3.1

Bootstrap: 3.3.6

RobJacobs:
- Ensure the scroll padding meant to offset the modal-open class only gets added once when mulitple modal windows are open.

Fixes #5789

This is going to be difficult to add tests for as the amount of (or if) scroll padding is browser dependent.  [Plunk](http://plnkr.co/edit/5m3p3r9C6DWf5kkJYVig?p=preview) with fix in place.


wesleycho:
LGTM, assuming tests pass

PR #5793, opened 2016-04-12
feat(dropdown): improve key handling

### LINKED ISSUE #5778: Pressing ESC with a dropdown inside a modal dialog

asfordmatt:
Fixes #5778 Pressing ESC with a dropdown inside a modal dialog
- Move key handler from service to controller, and attach to element instead of $document, so that the key events are received before they are handled by the modal dialog. If ESC is pressed to close the dropdown, consume the event so that the modal dialog doesn't also close.
- Update the tests accordingly to generate key events on the element itself - this now matches the way the typeahead control is implemented and tested

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -10,8 +10,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
 
   this.open = function(dropdownScope) {
     if (!openScope) {
-      $document.on('click', closeDropdown);
-      $document.on('keydown', keybindFilter);
+      $document.on('click', this.closeDropdown);
     }
 
     if (openScope && openScope !== dropdownScope) {
@@ -24,12 +23,11 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
   this.close = function(dropdownScope) {
     if (openScope === dropdownScope) {
       openScope = null;
-      $document.off('click', closeDropdown);
-      $document.off('keydown', keybindFilter);
+      $document.off('click', this.closeDropdown);
     }
   };
 
-  var closeDropdown = function(evt) {
+  this.closeDropdown = function(evt) {
     // This method may still be called during the same mouse event that
     // unbound this event handler. So check openScope before proceeding.
     if (!openScope) { return; }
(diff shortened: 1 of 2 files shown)


wesleycho:
This should not have a merge commit - you will want to do `git reset --hard 4a32562 && git pull --rebase origin master` to resolve the conflicts.

asfordmatt:
@wesleycho sorry, bit of a git novice here! I've done that, do I need to create a new PR?

wesleycho:
Just do `git push -f` (force push) after - that's ok in this case because it is not in a mainline remote branch (in this case, `origin master`)

asfordmatt:
@wesleycho both steps done, does it look any better? I can't tell :/

wesleycho:
Hmm, looks like the merge commit is still there - 36dbb0f is problematic from a history standpoint.

asfordmatt:
@wesleycho looks like it's gone now, hopefully it's ok :)

deeg:
Didn't realize thus PR was out out when I did mine.

Probably only need one of this or #5787

RobJacobs:
@deeg The logs say your commit was 5 hours ahead of this one.

deeg:
@RobJacobs, I guess that would be why!

If it doesn't matter that the keydown event is only bound when the dropdown is open, then I guess I prefer this method. The code is a bit cleaner, but it makes another method on the service public and also has a keydown event binded regardless of open state.

I could go either way I guess.

src/dropdown/dropdown.js
```diff
@@ -115,6 +116,8 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
         self.dropdownMenu.remove();
       });
     }
+
+    $element.on('keydown', keybindFilter);
```
deeg: I believe this would need to get cleaned up on scope destroy of the directive if we chose to go with this implementation.

wesleycho: I believe on $scope destruction, the directive cleans up the component.

deeg: Ah, yes I guess you are right since the element would be getting removed from the DOM when the scope is being destroyed. Still interested on your thoughts about which PR implementation to use though.

deeg:
After talking with the team we are going to close this in favor of #5787

PR #5794, opened 2016-04-12
feat(typeahead): Support for a `typeahead-should-select` function

### LINKED ISSUE #5671: Typeahead: Allow controlling selection trigger keys
Currently, typeahead treats either a `Tab` or `Enter` key as a trigger to select the currently-highlighted option. This is probably a sensible default for many use cases.

I have a particular use case where I'd like to make it so that a tab does not trigger selection, but instead simply moves to the next form field. For my use case, a simple `typeahead-select-on-tab="false"` or something similar would suffice. 

A more general thought might be to allow arbitrary keys/combinations to trigger completion, or a callback function for determining this (something like `typeahead-is-trigger-key="vm.isTriggerKey($event)"`).

Before I go off generating a pull request...Is there some support for something like that that already exists? Any thoughts on the idea? Implementation/naming?

christor:
By default behaves as before. If specified, allows a custom function
to be defined to determine whether the keydown event should trigger
selection

Closes #5671

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -42,7 +42,7 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
     originalScope.$watch(attrs.typeaheadMinLength, function (newVal) {
         minLength = !newVal && newVal !== 0 ? 1 : newVal;
     });
-    
+
     //minimal wait time after last character typed before typeahead kicks-in
     var waitTime = originalScope.$eval(attrs.typeaheadWaitMs) || 0;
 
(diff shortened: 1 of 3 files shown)


src/typeahead/typeahead.js
```diff
@@ -385,36 +392,37 @@ angular.module('ui.bootstrap.typeahead', ['ui.bootstrap.debounce', 'ui.bootstrap
...
+              if (angular.isNumber(scope.debounceUpdate) || angular.isObject(scope.debounceUpdate)) {
+                $$debounce(function() {
+                  scope.select(scope.activeIdx, evt);
+                }, angular.isNumber(scope.debounceUpdate) ? scope.debounceUpdate : scope.debounceUpdate['default']);
+              } else {
+                scope.select(scope.activeIdx, evt);
+              }
+            });
+          }
+          break;
```
wesleycho: No need for `break;`

deeg:
Before this could get merged it would need docs added for the new attribute.

christor:
Ah, good point. I'll look into adding that now.

src/typeahead/typeahead.js
christor: True...what was there before had a `break;` at the end of each case, so I followed suit with the default case. I'll remove it.

christor:
Any thoughts on the name of the option? Does it fit with the others?

wesleycho:
I think `typeahead-should-select` would be a better name - thoughts?

deeg:
@wesleycho, that sounds good to me! @christor, let us know if you have any outstanding questions to get this finished.

christor:
No, no more questions...just keep getting busy :-) `typeahead-should-select` works for me. I'll probably have time later this week to return to this.

christor:
@deeg, updated with suggestions and basic documentation

christor:
@deeg, @wesleycho: please let me know if you would like any more changes to this.

christor:
@deeg , @wesleycho : Just following up to see if there's more work to be done here (don't mean to be a pest). Thanks.

deeg:
@christor, could you please squash this down to one commit?

christor:
@deeg All squashed up. Let me know if you need something more. Thanks.

Issue #4837, opened 2015-11-07
Support ng-model-options with datepicker


wesleycho:
We should support ng-model-options with the datepicker

In particular, we should support
- [x] `allowInvalid`
- [x] `updateOn`
- [x] `timezone`

sandeepthomas:
I would really like to set UTC timezone instead of local or without time/timezone when setting a date!

maxdeviant:
:+1:

icfantv:
@maxdeviant, please don't add `+1` to issue threads without providing further benefit to the conversation.  All it does is add clutter to the thread.  Thanks.

We know this is something users want and fully intend to implement it.  We have a lot on our plates at the moment trying to get 1.0 out the door.

icfantv:
I should also add that we welcome well-thought and well-developed PRs if you really want/need this feature sooner rather than later.

maxdeviant:
My apologies. It's a bit of a force of habit, at this point.

`timezone` support is what I am particularly interested in.

sandeepthomas:
I fixed my issue for now with this directive. Posting here in case any one needs.

https://gist.github.com/sandeepthomas/78f2de3f0f8f21869f91

coryasilva:
For firebase timestamps I am using this directive.
````js
  .directive('timestamp', function($parse, $filter) {
    return {
      restrict: 'A',
      require: ['ngModel'],
      link: function link(scope, element, attr, ctrls) {
        var ngModelController = ctrls[0];
        var format = attr.uibDatepickerPopup;
        if (format) {
          ngModelController.$parsers.push(function(viewValue) {
            if (!viewValue) {
              return undefined;
            }
            if (angular.isFunction(viewValue.getTime)) {
              viewValue.setMinutes(viewValue.getMinutes() - viewValue.getTimezoneOffset());
              return viewValue.getTime();
            } else {
              return null;
            }
          });
          ngModelController.$formatters.push(function(modelValue) {
            if (!modelValue) {
              return undefined;
            }
            return $filter('date')(modelValue, format, '+00:00');
          });
        }
      }
    };
  })
````

paulflo150:
@davious nice work sir, I just played with your timezone commit and I like it! There are a couple of issues I noticed: the calendar is ahead by one day, and the time moves ahead by one hour when you select a DST date (i.e.: March 7 vs March 14)

[...]

PR #5811, opened 2016-04-16
feat(datepickerPopup): support ng-model-options="{updateOn: 'blur'}"

czosel:
As discussed in #4837, datepicker-popup currently does not support `ng-model-options="{updateOn: 'blur'}"` (popup can't be closed). The PR fixes this by always setting `updateOn: 'default'` on the popup.

### DIFF
--- a/src/datepickerPopup/popup.js
+++ b/src/datepickerPopup/popup.js
@@ -89,10 +89,7 @@ function($scope, $element, $attrs, $compile, $log, $parse, $window, $document, $
       timezone = ngModelOptions.timezone;
       $scope.ngModelOptions = angular.copy(ngModelOptions);
       $scope.ngModelOptions.timezone = null;
-      if ($scope.ngModelOptions.updateOnDefault === true) {
-        $scope.ngModelOptions.updateOn = $scope.ngModelOptions.updateOn ?
-          $scope.ngModelOptions.updateOn + ' default' : 'default';
-      }
+      $scope.ngModelOptions.updateOn = 'default';
 
       popupEl.attr('ng-model-options', 'ngModelOptions');
     } else {
--- a/src/datepickerPopup/test/popup.spec.js
+++ b/src/datepickerPopup/test/popup.spec.js
@@ -573,6 +573,31 @@ describe('datepicker popup', function() {
     });
   });
 
+  describe('works with ngModelOptions updateOn : "blur"', function() {
+    var $timeout, wrapElement;
+
+    beforeEach(inject(function(_$document_, _$sniffer_, _$timeout_) {
+      $document = _$document_;
+      $timeout = _$timeout_;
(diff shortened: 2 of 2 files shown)


src/datepickerPopup/popup.js
```diff
@@ -89,10 +89,7 @@ function($scope, $element, $attrs, $compile, $log, $parse, $window, $document, $
       timezone = ngModelOptions.timezone;
       $scope.ngModelOptions = angular.copy(ngModelOptions);
       $scope.ngModelOptions.timezone = null;
-      if ($scope.ngModelOptions.updateOnDefault === true) {
-        $scope.ngModelOptions.updateOn = $scope.ngModelOptions.updateOn ?
-          $scope.ngModelOptions.updateOn + ' default' : 'default';
-      }
+      $scope.ngModelOptions.updateOn = 'default';
```
wesleycho: I hesitate on this change - one can potentially be binding to many different events, including custom ones.

czosel:
I'm not quite sure if i understand - what do you mean by custom events?

What i realize is that this is adding `updateOn: 'default'` even if `updateOn` is not even specified, which might not be ideal. Another approach would be to convert `updateOn: 'blur'` to `updateOn: 'default'` for the popup, which is not as intrusive and should still fix the current issue.

wesleycho:
See documentation for [ngModelOptions](https://docs.angularjs.org/api/ng/directive/ngModelOptions), in particular the `updateOn` section

czosel:
Right, is see the problem now. But again I wonder what the use-case for custom events on the _popup_ would be. What do you think about just removing/converting `blur` for the time being? If more events turn out to be an issue with the popup, they can also be removed/converted.

wesleycho:
Can you explain what exactly is your use case for using `blur`, just so I have a clearer understanding of the problem we need to solve?

czosel:
Of course - I use it for very custom date parsing when the date is entered manually. (Example in German locale: The input "010470" is parsed to 01.07.1970") 
This was impossible with the built-in parser and is performed onBlur.

wesleycho:
Would `{updateOn: 'default blur'}` work for you then?

czosel:
Unfortunately not, because the date model is then updated afer every keydown and not just on blur. I put together a little [plunker](http://plnkr.co/edit/W1hi0IqBuWpqn5Lyi0Lo?p=preview) (try the top datepicker)

wesleycho:
IMO, the correct fix is a breaking change - this is a hack on top of what I think is a broken implementation, and one that is also a breaking change.

ngModelOptions should probably be a configuration option for the inline datepicker, and ngModelOptions from the popup should not be passed through to the inline. This might require a bit more thought.

czosel:
I agree, that would be much better. I'll try to give this a shot and post some code as a base for further discussion.

Issue #3879, opened 2015-06-30
[datepicker] active state on every first day of the month


marco-gagliardi:
Don't know if that's the intended behavior anyway it seems weird to me that while navigating on and back through the months, day one is always marked as active (and couldn't find an option to disable this behavior). To see it open this page: http://plnkr.co/edit/F3cu3Fed6DhgjxJnxT0J?p=preview and click on the right arrow over and over...

wesleycho:
I would say this is a minor UX bug. PRs fixing this are welcome.

viros:
Found next code:
 $scope.move = function( direction ) {
    var year = self.activeDate.getFullYear() + direction * (self.step.years || 0),
        month = self.activeDate.getMonth() + direction * (self.step.months || 0);
    self.activeDate.setFullYear(year, month, 1);
    self.refreshView();
  };
It's really weird, why it's needed at all?

wesleycho:
I think part of the problem is that if it is to remain on the currently selected day when changing months/years, there is the possibility of certain days not existing on the new month, such as 31.

The activeDate code looks like it keeps it simple - I would be fine with a PR changing this behavior though, it should be pretty easy to fix if someone wants to file a PR with the appropriate test to cover all edge case dates.

johnnybaloney:
TL;DR
Both current implementation and the suggested PR don't seem right. The highlight (a.ka. *active day* which != *selected day*) should be scrapped altogether (with the exception of 'today'). All the problems with it are listed below.

----

@wesleycho It took me a while to realise what was the problem you were referring to. I think I know now but I don't understand why anyone would want the active day to follow you as you switch between months. What is the thinking behind this idea? 

I can see that some other pickers try to do it like that, e.g. [pickadate](http://amsul.ca/pickadate.js/), while some don't, e.g. [jQuery datepicker](https://jqueryui.com/datepicker/). From an **end user perspective** who uses a datepicker for a few seconds and moves on I find the pickadate (and current angular ui) behaviour confusing - why when I select 13th February does it highlight 13th March for me? Has the selection changed? Is it broken? And then if I indeed want to select 13th March I have to click on the cell that appears to have already been clicked.

Also with the current implementation you get strangeness like this. If you select 1st February 2017 and then go back to January both 1st Jan and 1st Feb appear selected. You could argue that normally styling solves the problem (for clarity - I don't think so at any rate) but here it actually is part of the problem as clearly 1st January looks selected but it isn't.

[...]

PR #5812, opened 2016-04-16
fix(feat): add activeDate to customClass for custom styling

jasonhargrove:
Proposed enhancement that passes the `activeDate` into the `customClass` function which decorates the `td`. Currently the `active class` is attached to the child button. The active date is a prominent feature of the calendar and thus feels important enough in the custom class context to justify this change. It's also an inexpensive addition that provides a non-invasive opportunity to address issues with the current `active class`.

In my case I'm using it as a non-breaking solution for the issue described in #3879 , where active class is added to the 1st of non-current months. This seems quite opinionated, but is an issue that would be breaking if fixed. Having the ability to wrap the class using the already existing `customClass` feature gives us an easy out.

```
custom-class="vm.calendar.getDayClass(date, mode, activeDate)"
```

```
vm.calendar.getDayClass = function (date, mode, activeDate) {
  var events = vm.calendar.events;

  // custom event classes
  for (var i = 0; i < events.length; i++) {
    var eventDate = new Date(events[i].date).setHours(0, 0, 0, 0);
    var calDate = new Date(date).setHours(0, 0, 0, 0);
    if (eventDate === calDate) {
      return events[i].status;
    }
  }

  // custom active class fix
  var activeMonth = new Date(activeDate).getMonth();
  var currentMonth = new Date().getMonth();
  if (activeMonth !== currentMonth) {
    return 'non-current-month';
  }

  // default
  return '';
};
```

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -243,7 +243,7 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
   };
 
   this.customClass = function(date) {
-    return $scope.customClass({date: date, mode: $scope.datepickerMode});
+    return $scope.customClass({date: date, mode: $scope.datepickerMode, activeDate: self.activeDate});
   };
 
   // Split array into smaller arrays


icfantv:
I could potentially see a use case whereby we keep the currently selected day/month when we navigate forward and backwards through the calendar rather than the current behavior.  That said, this should be done in code as there is a specific line that handles this.  I'm on mobile right now so I can't easily link it.  I will say that at first glance the solution above seems fairly involved for what sounds like a simple change.

All this said, if this is something we decide to pursue, we are going to need tests.

jasonhargrove:
Thanks for your thoughts @icfantv . I feel as though the current functionality could be broken out a bit further, for example, today's date should have `today` class attached to it. While `active` should be attached to a day the user has clicked.

What I've submitted is the ability for a custom override on what is supported today. If new features were to be released, I would also hope to see those passed into the `customClass` function for override.

Technically it's a new enhancement that doesn't address #3879 directly, but let's you workaround it, or do other things of your own design. Ideally this would be separate of fixing the `active` & `today` issues, which would be breaking changes.

Sure I can submit some tests, but perhaps we can get a bit more feedback on it first.

jasonhargrove:
Opened another PR with test coverage that addresses the `.today` issue that inspired this for me. #5814 

I still do believe that `activeDate` should be passed into `customClass`. Did not find any existing test coverage for the feature.

While working on `todayDate`, took a look at the `activeDate` issue and there's a lot of code referencing it. Willing to take a deeper look once there's more consensus for how it should work, or if I find more time in the next week +

icfantv:
@jasonhargrove, right now, the biggest hurdle is that the datepicker API is already way too bloated and we are extremely reticent to add any more configuration to it.

all this said, i've just done some quick scouring on the web and none of the jquery, angular material, or bootstrap datepicker implementations have this behavior.  if anything, removing the active class functionality entirely seems to be a better UX and more inline with what everyone else is doing.

i'm going to talk to the team and see what we want to do about this.

jasonhargrove:
@icfantv it is definitely bloated, and removing it would certainly solve the part of the problem where it's broken. There's a lot of code that seems to depend on this setting though, so removing it sounds like a job in and of itself.

I looked at jQuery datepicker just now and indeed there is no active support, though they do have support for `today` which is ultimately the thing we really needed; which `active` seems to be confused with. There's a very simple PR open for that here: #5813 

Thanks for taking a deeper look, please let me know if I can assist.

icfantv:
@wesleycho, what are your thoughts on this?

PR #5826, opened 2016-04-20
Add appendToLast option

norami:
Enable Cascaded Dialog with CSS below

@media (min-width: 992px) {
  .modal-wd{
    position:absolute;
    top:0;bottom:0;left:0;right:0;
    width: auto;
    height:auto;
  }
  .modal:nth-child(3n+1) .modal-wd{
    margin: 20px 40px 40px 20px;
  }
  .modal:nth-child(3n+2) .modal-wd{
    margin: 30px 30px 30px 30px;
  }
  .modal:nth-child(3n) .modal-wd{
    margin: 40px 20px 20px 40px;
  }
  .modal-wd .modal-content{
    position:absolute;
    top:0;bottom:0;left:0;right:0;
    width: auto;
    height:auto;
    margin:0;
  }
}

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -505,7 +505,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
         }
 
         appendToElement.addClass(modalBodyClass);
-        $animate.enter($compile(angularDomEl)(modal.scope), appendToElement);
+        $animate.enter($compile(angularDomEl)(modal.scope), appendToElement, modal.appendToLast ? previousTopOpenedModal && previousTopOpenedModal.value.modalDomEl : null);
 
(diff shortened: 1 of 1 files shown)


wesleycho:
This sounds fragile - a better solution would be to use an appropriate CSS class for each modal IMO.

norami:
Although thie pull request affect to the future (wrong) change 
depending on the order of elements,
the current implementation does not depend on the order of elements.

In other hand, it's not a elegant
to use an appropriate CSS class for each modal element.
An alternate plan should be the code below.

```javascript
$uibModal.open({
    className: function(index){
        return 'modal-' + index % 3;
    }
})
```

wesleycho:
I'm amenable to support a function there that passes in the index.

icfantv:
@norami, what's the status of this?  Shall I close this as something you do not intend to finish?  Thanks.

PR #5827, opened 2016-04-20
feat(tab): enable string as index

norami:
Enable string index for tab.

### DIFF
--- a/src/tabs/tabs.js
+++ b/src/tabs/tabs.js
@@ -77,7 +77,7 @@ angular.module('ui.bootstrap.tabs', [])
   };
 
   $scope.$watch('tabset.active', function(val) {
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
We need a unit tests before it can be considered for merging

RobJacobs:
This is more involved and will need additional changes.  See discussion in PR #5713, we should re-open that one.

sandeepthomas:
Waiting for this to be implemented. In the meantime, is there any workaround to add active class to the changed tab (ul li) on using string as index?

icfantv:
@deeg @RobJacobs what is the status of #5713 you guys were working on?  Is this something you guys can fix?  People have been asking for this and we need to get it implemented.  Thanks.

PR #5830, opened 2016-04-20
fix(dropdown): Fix alignment for dropdown-append-to-body with .dropdown-menu-right when there's a vertical scrollbar

### LINKED ISSUE #4317: Right alignment is off for dropdown-append-to-body .dropdown-menu-right when there's a vertical scrollbar

dolevd:
Fixes #4317 

This is going to be difficult to add tests for as the width of the scrollbar (that's used to calculate the alignment) is browser dependent.

[Plunk](http://plnkr.co/edit/bNk7aXxMyDRg0q6XK7sV?p=preview) with fix in place.

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -191,7 +191,8 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
     if (appendTo && self.dropdownMenu) {
       var pos = $position.positionElements($element, self.dropdownMenu, 'bottom-left', true),
         css,
-        rightalign;
+        rightalign,
+        scrollbarWidth;
 
       css = {
         top: pos.top + 'px',
@@ -204,7 +205,8 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
         css.right = 'auto';
       } else {
         css.left = 'auto';
-        css.right = window.innerWidth -
+        scrollbarWidth = $position.scrollbarWidth(true);
+        css.right = window.innerWidth - scrollbarWidth -
           (pos.left + $element.prop('offsetWidth')) + 'px';
       }
 


wesleycho:
This LGTM - @RobJacobs is this the right fix?

RobJacobs:
LGTM

achenxp:
When there is no vertical scroll bar, the position of the dropdown is now incorrect.

dolevd:
Can you please provide a plunkr that shows the problem?

achenxp:
Here is the [Plunkr](http://plnkr.co/edit/YjTQj4Ct62ubErLXlz3Z?p=preview).

dolevd:
Thanks. My code is missing a check for overflow.
I'll send another pull request to fix this.

achenxp:
Thanks for the quick action!

dolevd:
I have a fix in my fork: https://github.com/dolevd/bootstrap/commit/f838eb649ed59364fa7c19f4448e6e6ecae9f7f9

However, this might create a performance issue (since we need to calculate the styles of the append to element on every watch).
Any suggestions on how to proceed?

dolevd:
I've opened another issue to handle this (#5942), since I don't think anyone is looking at the comments here anymore.

Issue #5829, opened 2016-04-20
Example usage of accordion's custom template doesn't support transcluded header


michal-filip:
Current online docs show example usage of custom template for accordion group. This custom template doesn't support transcluded heading. Although it's just an example, I suspect that most people will copy-paste this into their code as a starter. Therefore I'd suggest updating this template by including the "uib-accordion-header" directive in this template.

Plunker:
http://plnkr.co/edit/dFbSQ1CNFvVofSlkPLbr

Angular: 1.5.3
UIBS: 1.3.2

wesleycho:
Funnily enough, I just ran into this at work yesterday - PRs welcome for this.

PR #5831, opened 2016-04-20
docs(accordion): add example with custom header

wesleycho:
- Add an example with custom header template

This addresses #5829.


pkozlowski-opensource:
Shouldn't you have `Fixes #5829` in the commit msg?

wesleycho:
I always modify when merging - if I ever have to rebase, it then pollutes the referenced issue with lots of messages.

deeg:
This LGTM

PR #5843, opened 2016-04-25
Performance improvments for daypicker.

Drag13:
Hi guys. Working with Daypicker i have faced some trubles with performance.
scope.rows = this.split(days, 7); caused to re render full list of 42 items. And this cause some troubles with Edge. 
So I want to offer some little fix. It is not great solution, but fixed my problems with performance. 
Before changing whole array of the rows, we are checking is it possible to avoid this.
If first and last day are the same, then we can nly update array insted of changing whole array and re render it.
If you have another solution, this will be great.


src/datepicker/datepicker.js
```diff
@@ -413,6 +414,28 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
...
+
+      if (!rows || newHash !== daysHash) {
+          daysHash = newHash;
+          return this.split(days, 7);
+      }
+
+      var splitted = this.split(days, 7);
+      for (var i = 0; i < splitted.length; i++) {
+          for (var k = 0; k < splitted[i].length; k++) {
+              angular.merge(rows[i][k], splitted[i][k]);
```
wesleycho: This is a potentially very expensive operation

src/datepicker/datepicker.js
```diff
@@ -413,6 +414,28 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
...
+          }
+      }
+
+      return rows;
+  }
```
wesleycho: Need semi-colon at the end of this line

Drag13:
Yeh, merge  operation is really expensive. I thought about this side of the fix also. But re-rendering 42 elements with creating new scopes also is not chip. Ill add semicolumn and think. Possibly I can use merge that already exists upper in the code.

Drag13:
12 tests faild. Not great.

icfantv:
@Drag13, do you have an update on this?  If not, I'm going to close it.  Thanks.

Drag13:
I tried some ideas but nothing good enough for me and test. So close. If any Ill ping. 
Have a nice day/

icfantv:
Closing per comment.

Issue #5841, opened 2016-04-25
`uib-dropdown-toggle` doesn't work on first half of split button


david-mohr:
### Bug description:

`uib-dropdown-toggle` doesn't work on first half of split button.

I have a split button which I would like to toggle using both halves of the split button.  If I add `uib-dropdown-toggle` to the first button in a `btn-group` (using the 'split button' example given here: https://angular-ui.github.io/bootstrap/#/dropdown), the menu will not open, however chrome devtools does indicate that aria-expanded is being altered although it remains set to false.

### Link to minimally-working plunker that reproduces the issue:

https://plnkr.co/edit/oo4zk5gd0vPJlBX6px9s?p=preview

There are three buttons on this plnkr that all trigger dropdowns. The first demonstrates the problem (clicking 'Action' will not open the menu, clicking the caret will). The second demonstrates a workaround (clicking 'Action' will open the menu). The third is a standard button with a dropdown.

### Version of Angular, UIBS, and Bootstrap

Angular: 1.4.9
Bootstrap: 3.3.6
UIBS: 1.3.2

icfantv:
@david-mohr marking this as needing investigation, but just a head's up that your `id` attributes are not unique - but that is not why this isn't working...

david-mohr:
Thanks @icfantv, I've updated the `id` attributes.

wesleycho:
The library only supports one uib-dropdown-toggle is the reason this does not work. I don't see a particular compelling reason why we should support more than 1, given that it will complicate the internal logic around checking the toggle elements and that this seems like an edge case.

I would recommend using the is-open binding to solve your problem.

david-mohr:
Thanks @wesleycho, I'll stick with the is-open work around.

Issue #5767, opened 2016-04-07
Allow dropdown to be positioned up or down


wesleycho:
So given the robust positioning present in some of the other components, we actually have a request at work for doing so, and automatically positioning the appropriate direction when too close to the bottom (position="auto" functionality).

This would be a useful feature for those using dropdowns as a fancier version of a select element.

I'll probably end up tackling this, but if anyone wants to attempt it, I welcome it.

deeg:
I had to do something like this at work also. I will try to make a PR if I have some time.

Wanted to put some code which might be helpful when building this feature:

```
menuPosition = $uibPosition.positionElements(
    element,
    controller.dropdownMenu,
    "bottom",
    true
);

//Check to see if it would go off bottom of screen
if ((menuPosition.top + controller.dropdownMenu[0].offsetHeight) >= ($window.pageYOffset + $window.innerHeight)) {
    menuPosition = $uibPosition.positionElements(
        element,
        controller.dropdownMenu,
        "top",
        true
    );
}
```

One thing to think about when building is at smaller window sizes it might go off bottom of screen, so you would position on top, but then could have an issue where it is going off top of screen or hidden by something like a navbar with a higher z-index.

wesleycho:
I think that's ok to a degree - simple checks would be enough IMO, and maybe just use the existing stuff built into the position service.

deeg:
Sounds good, and yeah agreed using the existing stuff in the position service!

wesleycho:
Also the position service has support for `auto` placement - it should work similarly to how the tooltip/popover placement works.

PR #5845, opened 2016-04-26
fix(dropdown): browser collision detection

SjorsH:
Fix for the dropdown issues #5841, #5767


src/dropdown/dropdown.js
```diff
@@ -187,17 +187,31 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
...
+    if (clientRect.top > dropdownHeight &&
+        window.innerHeight - clientRect.top < clientRect.height + dropdownHeight) {
+      return true;
+    }
+    return false;
+  }
+
   scope.$watch('isOpen', function(isOpen, wasOpen) {
     if (appendTo && self.dropdownMenu) {
       var pos = $position.positionElements($element, self.dropdownMenu, 'bottom-left', true),
```
RobJacobs: The placement parameter passed to the positionElements method should be changed to 'auto bottom-left' and leverage the work already being done there.

src/dropdown/dropdown.js
```diff
@@ -187,17 +187,30 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
...
+      var pos = $position.positionElements($element, self.dropdownMenu, 'auto bottom-left', true),
         rightalign,
-        scrollbarWidth;
-
-      css = {
-        top: pos.top + 'px',
-        display: isOpen ? 'block' : 'none'
-      };
+        scrollbarWidth,
+        css = {
```
wesleycho: Grouping the css setting of this object hurts readability - better to move it back to how it was prior.

src/dropdown/dropdown.js
```diff
@@ -187,17 +187,30 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
     }
   };
 
+  function toggleDropdownUp() {
```
wesleycho: This should be abstracted into a function into the $position service

SjorsH: Thanks, I'm going back to the drawing board and will let you know if I changed the position service.

RobJacobs: There should be no need to change the position service to accomplish this.  Calling the $position.positionElements with an auto placement should do what you need, just use the pos.top value returned.

SjorsH: @RobJacobs 
Indeed it works fine with placement set to auto, but I noticed that width and height of my targetElement inside the $position.positionElements are 0.

`// need to read from prop to support tests.
        var targetWidth = angular.isDefined(targetElem.offsetWidth) ? targetElem.offsetWidth : targetElem.prop('offsetWidth');
        var targetHeight = angular.isDefined(targetElem.offsetHeight) ? targetElem.offsetHeight : targetElem.prop('offsetHeight');`

Why not use the outerWidth and outerHeight?

RobJacobs: Because outerHieght and outerWidth are jQuery methods and we only support jQLite or native element methods and properties.  Besides, the outerHeight/width jQuery documentation states:

Get the current computed height for the first element in the set of matched elements, including padding, border, and optionally margin.

And the HtmlElement offsetHeight/width documentation from MDN states:

Typically, an element's offsetHeight is a measurement which includes the element borders, the element vertical padding, the element horizontal scrollbar (if present, if rendered) and the element CSS height.

So the 2 are the same unless you include the margin parameter.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "author": "https://github.com/angular-ui/bootstrap/graphs/contributors",
   "name": "angular-ui-bootstrap",
-  "version": "1.3.3-SNAPSHOT",
+  "version": "1.3.3",
```
wesleycho: This change needs to be undone

src/dropdown/dropdown.js
```diff
@@ -189,7 +189,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
 
   scope.$watch('isOpen', function(isOpen, wasOpen) {
     if (appendTo && self.dropdownMenu) {
-      var pos = $position.positionElements($element, self.dropdownMenu, 'bottom-left', true),
+      var pos = $position.positionElements($element, self.dropdownMenu, 'auto bottom-left', true),
```
wesleycho: What is the rationale for this change? This changes the default for all apps using this component.

src/position/position.js
```diff
@@ -436,8 +436,8 @@ angular.module('ui.bootstrap.position', [])
         targetElem = this.getRawNode(targetElem);
 
         // need to read from prop to support tests.
-        var targetWidth = angular.isDefined(targetElem.offsetWidth) ? targetElem.offsetWidth : targetElem.prop('offsetWidth');
-        var targetHeight = angular.isDefined(targetElem.offsetHeight) ? targetElem.offsetHeight : targetElem.prop('offsetHeight');
+        var targetWidth = angular.element(targetElem).outerWidth();
+        var targetHeight = angular.element(targetElem).outerHeight();
```
wesleycho: @RobJacobs how do you feel about this change here?

wesleycho: Actually, it looks like this change is breaking the tests

RobJacobs: Those are jQuery methods (outerWidth and outerHeight) and not native element or jQlite methods.  The note above the line of code removed says the prop lookups are there for a reason - to support the tests, so I'm not sure why they are removed.  The position test creates a mock element [here](https://github.com/angular-ui/bootstrap/blob/master/src/position/test/position.spec.js#L6) with those values as props.

wesleycho: Thanks, looks like this change needs to be undone.

wesleycho:
Going to close this due to fundamental issues - I'll take a stab at this soon at a proper implementation.

Issue #5808, opened 2016-04-14
Add $resolve to uibModal scope


radarfox:
Hello,

it would be super, if modal plugin would support new `$resolve` property same as angular-route does.

See example here (search for "$resolve") https://docs.angularjs.org/api/ngRoute/service/$route

It makes usage with components much easier, because there is no need for "modal controller", just the "component controller":

```
$uibModal.open({
	template: '<my-component data-foo="$resolve.bar"></my-component>',
	resolve: {
		bar: 'Hello world!'
	}
});
```

Then in component just add bindings, and everything is set:

```
var myComponent = {
	template: '<p data-ng-bind="$ctrl.foo"><p>',
	bindings: {foo: '<'}
};
```

All I had to do, was to add 2 new lines in your source:

```
var modalScope = providedScope.$new();
modalScope.$close = modalInstance.close;
modalScope.$dismiss = modalInstance.dismiss;
modalScope.$resolve = tplAndVars[1]; // here
```

and

```
if (modalOptions.bindToController) {
  ctrlInstance.$close = modalScope.$close;
  ctrlInstance.$dismiss = modalScope.$dismiss;
  ctrlInstance.$resolve = modalScope.$resolve; // here
  angular.extend(ctrlInstance, providedScope);
}
```

Plunker http://plnkr.co/edit/UtTBUZoztzVS8A5jcoQq

Example in angular-route https://docs.angularjs.org/guide/component#components-as-route-templates

deeg:
@radarfox, are you planning to fill out the issue description or was this created by mistake?

icfantv:
closing due to lack of content.

radarfox:
Sorry, i accidentaly pressed enter before entering description.

You is too fast for me :)

Now is all set, could you reopen?

icfantv:
@radarfox, i'm not seeing a $resolve property in the docs page you linked. regardless, if this is an angular 1.5.x thing, we are not yet ready to require angular 1.5+.

deeg:
@radarfox, is this somehow different than the scope property you can pass into open?

> scope (Type: $scope) - The parent scope instance to be used for the modal's content. Defaults to $rootScope.

radarfox:
@icfantv It is located near the end of page - Properties -> current

Well it is not 100% Angular 1.5. It is just adding resolved object to modal scope. It is not requiring any 1.5 functionality.

radarfox:
[...]

PR #5857, opened 2016-04-28
feat(modal): add resolve values to template

wesleycho:
- Expose resolve in template as $resolve for those modals opened with a controller

This implements #5808, but may feature a minor breaking change - this could be fine to hold off until 2.0 if there are concerns about introducing a breaking change.


deeg:
LGTM

deeg:
Also @wesleycho, don't forget to add a BREAKING CHANGE comment to the body of the commit when merging.

PR #5859, opened 2016-04-29
deal with the controllerAs option as the angular.js way

ruiann:
angular.js do as this:
var controllerInstance = $controller(controller, locals, true, directive.controllerAs);


wesleycho:
This needs the merge commit to be removed.

ruiann:
so i should redo a pull request?

PR #5861, opened 2016-04-29
deal with the controllerAs option as angular.js way

ruiann:
use the private api of $controller since the private api has already been used

angular do like this:
var controllerInstance = $controller(controller, locals, true, directive.controllerAs);


icfantv:
@wesleycho is this ok?  needs tests?  @ruiann, you need to squash your commits and these should not be merge commits because it jacks the history.

wesleycho:
Perhaps I can merge this locally and fix up this history since it seems like @ruiann is having some trouble here.

This LGTM though.

ruiann:
@icfantv yes angular.js deal with the controllerAs feature in $controller

ruiann:
@wesleycho i forked this repo and when try to pull a request i cannot clear the merge commits, i wonder how can i  do this?

wesleycho:
@ruiann it's merged in already as 5741be9 - in general for this situation though, you need to start with a fresh branch off of `master`, then cherry-pick the desired commits from the previous branch, then push up. If you rewrite history, then you need to force push.

ruiann:
@wesleycho thx. i've experience in git but i just don't know how to commit without merge histroy lol. Glad to make some contribution to such project.

PR #5872, opened 2016-05-03
fix(datepicker): add intermediary valid date check

### LINKED ISSUE #5865: uib-datepicker using new data() on "ngModelCtrl.$modelValue" in UibDatepickerController causes error in Internet Explorer
We use uib-datepicker with popup where we allow the user the enter a date manually.
The dateformat ('d MMMM y') has a year with a single y so a single-digit year can also be entered as a year.

On load we set today's date with a four-digit year. When you change the year to have only 2 digits the datepicker renders with all null and nan values, see attached screenshot. This happens only on Internet Explorer (tested on 11). Chrome and FireFox do not have this issue.

Plunker: http://plnkr.co/edit/sGbHK8N4C0EAmpyT9FB9?p=preview

Reproduce:

1. Open the plnkr in IE;
2. Place the cursor in the textbox at the end of the date;
3. Press backspace twice to create the 2-digit your;
4. Click on the icon to display the datepicker-popup.

- Angular: 1.5.5
- UIBS: 1.3.2
- Bootstrap: 3.3.6

datepickernan

In the "this._init" function in "UibDatepickerController" the following lines cause the error in IE.
```
this.activeDate = ngModelCtrl.$modelValue ?
  dateParser.fromTimezone(new Date(ngModelCtrl.$modelValue), ngModelOptions.timezone) :
  dateParser.fromTimezone(new Date(), ngModelOptions.timezone);
```
We have modified these like below to fix the issue.
```
if (ngModelCtrl.$modelValue) {
  var inputDate = new Date(ngModelCtrl.$viewValue),
      isInputValid = !isNaN(inputDate);
      
  if (isInputValid) {
    this.activeDate = dateParser.fromTimezone(inputDate, ngModelOptions.timezone);
  }
}
```

wesleycho:
- Add intermediary check for whether the date is valid due to an IE quirk of turning a valid date into an invalid date when using the date constructor

This fixes #5865.


RobJacobs:
LGTM

PR #5873, opened 2016-05-04
fix(datepicker popup): convert non-date values to date before parsing

### LINKED ISSUE #5871: TypeError: date.getTimezoneOffset is not a function
### Bug description:

`ng-model` as numeric does not work properly when using `datepicker-popup`. By looking at [this line](https://github.com/angular-ui/bootstrap/blob/bba3f27e3dd9f95463743bf36a15636a746ea3ec/src/datepickerPopup/popup.js#L136) from the source code, I would say that the check for numeric value should be done before and not after, i.e.:
````
$scope.date = dateParser.fromTimezone(value, timezone);

if (angular.isNumber($scope.date)) {
  $scope.date = new Date($scope.date);
}
````
Should be instead:
````
if (angular.isNumber(value)) {
  value = new Date(value);
}

$scope.date = dateParser.fromTimezone(value, timezone);
````
Because `dateParser.fromTimezone` expects a `Date` and not a `Number`, the error `TypeError: date.getTimezoneOffset is not a function` is thrown.

Angular: 1.5.5
UIBS: 1.3.2
Bootstrap: 3.3.5

mebibou:
fixes #5871

### DIFF
--- a/src/datepickerPopup/popup.js
+++ b/src/datepickerPopup/popup.js
@@ -133,11 +133,11 @@ function($scope, $element, $attrs, $compile, $log, $parse, $window, $document, $
           return value;
         }
 
-        $scope.date = dateParser.fromTimezone(value, timezone);
-
-        if (angular.isNumber($scope.date)) {
-          $scope.date = new Date($scope.date);
+        if (angular.isNumber(value)) {
+          value = new Date(value);
         }
+
+        $scope.date = dateParser.fromTimezone(value, timezone);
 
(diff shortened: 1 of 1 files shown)


chrislandeza:
@mebibou,  does this fix included in 1.3.3?

mebibou:
@chrislandeza no idea, havent followed up on this

pjmartorell:
@chrislandeza yes, it's included in 1.3.3

PR #5878, opened 2016-05-06
Update dropdown.js

matusszabo:
When using dropdown to show search suggestions, while hiding/showing dropdown menu and using key nav, it cause exception (cannot get property "focus" of undefined)

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -174,7 +174,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
         break;
       }
     }
-    elems[self.selectedOption].focus();
+    if(elems[self.selectedOption]) elems[self.selectedOption].focus();
   };
 
   scope.getDropdownElement = function() {


icfantv:
@matusszabo, your build failed because of the linter and you have no tests for this case.  that said, our demo is working fine, which makes me wonder if you're dynamically setting the values in the dropdown and causing this issue.

Please paste a working plunker that shows the issue.

matusszabo:
Hi @icfantv , yes,  only when loading options dynamically: https://plnkr.co/edit/WDqfQPJt5x52u3TDitFL?p=preview

icfantv:
Ok, I don't know if this is a use case we want to support, although the alleged fix seems innocuous enough at first glance.  Need to ask the team.  @wesleycho @Foxandxss?

wesleycho:
This needs a reproduction.

wesleycho:
Closing due to this needing to go through issues with proper reproduction first.

matusszabo:
@wesleycho, what kind of reproduction do you need, plunker is not good enough ?

PR #5880, opened 2016-05-07
chore(docs): update the docs

icfantv:
* update the datepicker and datepicker popup docs to let user know
  about using the `uibDateParser` service to convert a date string
  into a `Date` object.


wesleycho:
LGTM

icfantv:
Crap.  Need to drop the $ bit.  There's no scoping on a service.

icfantv:
And the eye icon....

PR #5889, opened 2016-05-10
feat(typeahead): add dynamic wait ms support

radotzki:
- Add support for dynamically changed wait ms

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -470,6 +470,28 @@ describe('typeahead tests', function() {
       };
       expect(prepareInvalidDir).toThrow();
     });
+
+    it('should support changing wait-ms', function() {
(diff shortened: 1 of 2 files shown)


wesleycho:
Can you explain your use case as to why one would want it to be dynamic?

radotzki:
Yes sure.
This is necessary in case you want to set the waitMs time based on the input length. 
For example if the user entered one letter I want the timeout to be 600ms, however if he entered more than one letter I want the timeout to be 300ms.

wesleycho:
Going to close this due to the wait setting being more about delaying the request from being fired off in a uniform amount of time - changing this to be dynamic has negative perf ramifications, and also encourages weird UX expectations.

I don't even think this setting really makes sense truthfully, especially given that we support `ng-model-options` `debounce` option.

PR #5890, opened 2016-05-10
chore(button): add known issue for tooltip/popover

icfantv:
* add known issue to button docs indicating a CSS issue when using a tooltip or
  a popover.


src/buttons/docs/readme.md
```diff
@@ -44,3 +44,7 @@ With the buttons directive, we can make a group of buttons behave like a set of
 * `toggleEvent`
   _(Default: `click`)_ -
   Event used to toggle the buttons.
+
+### Known issues
+
+To use tooltips or popovers on a button group, make sure you add the applicable tooltip or popover directive directly to your `btn-group` element and set `appendToBody` to `true`. This is due to Bootstrap CSS styling. See [here](http://getbootstrap.com/components/#btn-groups) for more information.
```
RobJacobs: The append-to-body is only necessary if adding the tooltip to an element inside the btn-group.

icfantv: Ah, ok.  I wasn't sure I understood this correctly.

icfantv:
@RobJacobs, fixed.  look good?

RobJacobs:
@icfantv I would remove the mention of adding the tooltip to the btn-group, so maybe something like:

make sure you set `appendToBody` to `true`

icfantv:
@RobJacobs ok, done.

Issue #3906, opened 2015-07-07
Close button for popovers


krnlde:
I had a few discussions and a lot of testing around with snippets from stackoverflow and the rest of the internet to accomplish the goal. The best solution I ended up with is the following:

directives.js
```js
app.directive('dismiss', ['$rootScope', '$timeout', function ($rootScope, $timeout) {
  return {
    restrict: 'A',
    link: function (scope, $element, attrs) {
      $element.on('click', function () {
        $timeout(function () {
          var $context = $element.closest('.' + attrs.dismiss);
          var $contextScope = $context.scope();
          $contextScope.$apply(function () {
            $contextScope.isOpen = false;
          });
        });
      });
    }
  };
}]);
```

popover-content.html
```html
<div id="my-content">
  <button type="button" class="close" data-dismiss="popover" aria-label="Close"><span aria-hidden="true">&times;</span></button>
  // ... stuff
</div>
```
It at least works, but it has one catch: When I closed it via the dismiss I'll need to click 2 times on the opener again since I don't have access to the closure logic of the tooltip (e.g. `hide()`, `hideTooltipBind()` or `removeTooltip()`).

Here's the thing: I really want to use bootstrap-ui - [all solutions](http://stackoverflow.com/a/28447475/622041) in the internet suggested a new directive which makes your popoverTemplate directive obsolete.

Any suggestions?

realityking:
This is closely related to #590. Unfortunately no solution there either.

krnlde:
It is and this [plunkr](http://plnkr.co/edit/iiP0t8?p=preview) seems to work so far. I will test it the next few days and give feedback.

krnlde:
Just an update: Right now this plunkr is restricted to `popover-template` and will not work with `popover`. Further updates will be discussed in the #590 directly.

wesleycho:
One cannot rely on `element.scope()` since it does not work if `$compileProvider.debugInfoEnabled(false)` is run.

Maybe exposing a function like `$scope.$close` in the tooltip service would be able to create support for this in the popup component.

icfantv:
@wesleycho it'd be nice to have an option as to whether or not to show the close button, but the problem is, with custom triggers, we won't know whether or not the trigger makes it so that the close button is even accessible.  i like the idea of exposing a function for closing - would something similar to $modal's close promise be worth it?

wesleycho:
[...]

PR #5896, opened 2016-05-15
handler to force close the popover/tooltip

jodinathan:
Hi,

My project uses the popover full window when it is mobile, so I needed a close button within the template and without changing the trigger default behaviour.
So, this helps the scopes inside the popup to $emit a 'close-popup' to hide it


src/tooltip/tooltip.js
```diff
@@ -537,6 +537,11 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
 
             appendToBody = angular.isDefined(appendToBodyVal) ? appendToBodyVal : appendToBody;
 
+            // let inner scopes to close the popup
```
wesleycho: Change to `// allow child scopes to close the popup`

wesleycho:
This needs tests.

RobJacobs:
This can be handled leveraging the popover/tooltip is-open attribute.  This [plunk](http://plnkr.co/edit/W1j4A3ZJh8Qw7wjnCRVO?p=info) demonstrates that approach.

jodinathan:
@RobJacobs , it is different. 
Lets say you want to insert a close button within all popovers and the popovers have different opening/closing triggers. With only is-open you would have to understand each popover to make it work. With my approach, I can dynamically add the close button and it will work with all popovers without much effort

wesleycho:
This is also less performant than using is-open - $scope eventing is generally to be avoided for that reason

RobJacobs:
@jodinathan The is-open attribute works independently of the triggers.

jodinathan:
Rob, could you give me an example of is-open with a close button that is dynamically added to the template please?
another question that would solve my problem: 
How could you close a popover if you only have access to the popover(.popover) element itself.

wesleycho:
One can create a wrapper directive around any element that uses the tooltip that adds a `tooltipClose` method on `$scope` and use it in the template - the tooltip/popover `$scope` inherits from the parent one.

jodinathan:
@wesleycho can you give an example please.

icfantv:
@jodinathan, have you attempted to create a wrapper directive to get the behavior you want?  adding event handlers for this type of feature/behavior is very poor.  This PR will be closed if we don't hear from your or don't see an attempt to resolve this based on the information provided.  The issues/PR forums are not for support related issues.  As we are all only volunteers, we simply do not have the time to appease every API request from people.

This issue has been addressed several times in our issue history.  See #3906 for more info.

jodinathan:
@icfantv, unfortunately I haven't successfully created the wrapper. I tried setting the isOpen within the tooltip's scope but it didn't work.
I will use the event handler for now as it works for me.

icfantv:
Closing per comment.

Issue #5899, opened 2016-05-16
Two way data binding for the active attribute on a tab


Theo-:
### Bug description:

When modifying the property "active" of a tab from the javascript side during a digest cycle, the tabset doesn't update the view to the current active tab.

In the plunker below, the data structure specifies that the second tab should be selected. However the first tab is selected at run time.

### Link to minimally-working plunker that reproduces the issue:

http://plnkr.co/edit/HY6UaOSTsW7YwEvTDlFm?p=preview

### Version of Angular, UIBS, and Bootstrap

Angular:
AngularJS v1.5.5

UIBS:
1.3.2

Bootstrap:
Bootstrap v3.3.6

Theo-:
Opened a pull request with the fix https://github.com/angular-ui/bootstrap/pull/5900

icfantv:
@Theo- why are you touching the `active` property of the tabs?  we manage and set these internally after rendering to ensure the right CSS class is set.

Theo-:
Why is it internal ? Let's say I have a data structure like this:

`
$scope.tabs = 
[
{
name: "tab1",
active: true
},
{
name: "tab2",
active: false
}
]
`
Let's set active='tab.active' in the ng-repeat. If I do `$scope.tabs[0] = false` and `$scope.tabs[1]=true` (for example when the user clicks on a button), the second tab will not get selected. If that an expected result ?

I understand it is supposed to be an internal property, but would't it be easier to just modify the current tab by change the active flag on the data structure instead of changing the "active" attribute on the tabset.

If not please close this issue.

icfantv:
Funny you should ask.  Until quite recently, we had it at the tab level.  We moved it to the tabset level because it made more sense to have the parent container manage its children and leave the children agnostic.

You can still set an active tab via its index - which can be a number or a string, but must be unique.  The latter of which is in `master`, but should be released this week.

I'm going to close this per this and your comments.

PR #5900, opened 2016-05-16
Fixing opened issue

Theo-:
See https://github.com/angular-ui/bootstrap/issues/5899 for more details


wesleycho:
This seems wrong - we already have logic for handling selection on active change. See [here](https://github.com/angular-ui/bootstrap/blob/master/src/tabs/tabs.js#L79-L83) - if something is wrong, we should fix whatever is wrong here.

Theo-:
The code you linked is about listening to the variable that points to the index of the active tab. But this variable is not set properly. There is no way of modifying the active tab just by modifying the data structure containing that the active attribute points to.

wesleycho:
There's a two-way binding defined [here](https://github.com/angular-ui/bootstrap/blob/master/src/tabs/tabs.js#L104-L107). This should work, and there are tests against this.

If it's broken, the broken piece should be fixed, not adding another watcher that isn't properly gc'd on $scope destruction.

Theo-:
See the plunker I provided in the bug description:
http://plnkr.co/edit/HY6UaOSTsW7YwEvTDlFm?p=preview

This two data binding is obviously not working. However I am not an angular pro, so I'll leave this to your judgment.

icfantv:
@Theo- the broken tests aside, please address my question in your issue report.

PR #5908, opened 2016-05-18
Role value should be changed.

zamankazi:
Currently role="dialog" causes an issue for screen readers like NVDA or Jaws to read the content that is inside of a modal. Proposing the value of role to be changed to role="contentinfo" so that applications like NVDA or JAWS will automatically start reading the contents of a modal once the modal opens.

### DIFF
--- a/template/modal/window.html
+++ b/template/modal/window.html
@@ -1,4 +1,4 @@
-<div modal-render="{{$isRendered}}" tabindex="-1" role="content" class="modal"
+<div modal-render="{{$isRendered}}" tabindex="-1" role="contentinfo" class="modal"
     uib-modal-animation-class="fade"
     modal-in-class="in"
     ng-style="{'z-index': 1050 + index*10, display: 'block'}">
--- a/template/modal/window.html
+++ b/template/modal/window.html
@@ -1,4 +1,4 @@
-<div modal-render="{{$isRendered}}" tabindex="-1" role="dialog" class="modal"
+<div modal-render="{{$isRendered}}" tabindex="-1" role="content" class="modal"
     uib-modal-animation-class="fade"
     modal-in-class="in"
     ng-style="{'z-index': 1050 + index*10, display: 'block'}">


zamankazi:
Currently role="dialog" causes an issue for screen readers like NVDA or Jaws to read the content that is inside of a modal. Proposing the value of role to be changed to role="contentinfo" so that applications like NVDA or JAWS will automatically start reading the contents of a modal once the modal opens.

wesleycho:
This sounds like a hack & incorrect usage - see https://www.w3.org/TR/wai-aria/roles#dialog and https://www.w3.org/TR/wai-aria/roles#contentinfo .

I am inclined to reject this as an invalid fix.

zamankazi:
We have come across this bug at my work, I can't reveal the company and the project info. However, in the effort to make our application be accessible via applications like NVDA and JAWS for visually blind the solution role="dialog" did NOT work. JAWS and NVDA will not read any content inside of a modal whether it's wrapped in a paragraph, label, div, span, section, article what ever it may be. These screen readers consider role="dialog" to be a dialog where the user is prompted to enter a value hence why it skips reading the content for any visually impaired users. We had to currently take the route of changing the role value in order these readers to be able to read through the content as our app is required by government regulations to be accessible for the disabled. I understand this is a hacky fix but I am sure other organization will face this issue sooner or later as there are growing numbers of disabled users using modern day web application through screen readers. There are bills on motion that many organizations WILL have to make their apps be accessible due to government regulations in the near future. I would request you to please reconsider.

icfantv:
@zamankazi my reading of [the official ARIA docs for `contentinfo`](https://www.w3.org/TR/wai-aria/roles#contentinfo) seems to indicate that your proposal is not only invalid, but incorrect.  If you feel differently, please feel free to convince me otherwise.

Otherwise, I'm inclined to agree with @wesleycho in rejecting this.

wesleycho:
Couldn't you just use a div wrapping the content and use `role="contentinfo"` there instead?

If you want a hack, you could just override the template locally, but this PR as is looks like a misuse of ARIA.

wesleycho:
Going to close this as an abuse of ARIA.

In addition, this appears to be invalid based on the research done [here](https://accessibility.oit.ncsu.edu/training/aria/modal-window/version-3/), also with a demonstration of an accessible modal using the role attribute as we do.

PR #5911, opened 2016-05-20
Fixes bug with backdrop click

no1lov3sme:
Fixes bug when selecting text and moving mouse away from modal container to the backdrop and releasing mouse button.

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -151,6 +151,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
           var modal = $modalStack.getTop();
           if (modal && modal.value.backdrop &&
             modal.value.backdrop !== 'static' &&
+            evt.target === evt.currentTarget &&
             scope.mouseDownEvt && scope.mouseUpEvt &&
             scope.mouseDownEvt.target === scope.mouseUpEvt.target) {
             evt.preventDefault();
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -145,12 +145,14 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
         element.addClass(attrs.windowClass || '');
         element.addClass(attrs.windowTopClass || '');
         scope.size = attrs.size;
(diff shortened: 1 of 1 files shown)


icfantv:
If you would like us to consider merging this, you will need to add tests and correctly fix the issue that caused the build to break.

wesleycho:
The implementation should be changed to something similar to [this](https://github.com/twbs/bootstrap/blob/9f2f51073cfbf3d2a4a16a901a60d184cbcfd4c4/js/modal.js#L70-L74) from Bootstrap

icfantv:
@no1lov3sme, do you intend on fixing the issues with this PR or shall I close it?  Thanks.

no1lov3sme:
@icfantv, please specify ehat should be fixed. I've already refactored the code in my last commit; tests are passing.

icfantv:
@no1lov3sme, see the comment from @wesleycho.

src/modal/test/modal.spec.js
```diff
@@ -427,7 +427,10 @@ describe('$uibModal', function() {
       var modal = open({template: '<div>Content</div>'});
       expect($document).toHaveModalsOpen(1);
 
-      $document.find('body > div.modal').click();
+      var selector = 'body > div.modal';
+      $document.find(selector).mousedown();
+      $document.find(selector).mouseup();
+      $document.find(selector).click();
```
wesleycho: This should be in a separate test - you could just copy/paste the test this is modifying, and revert the original test.

src/modal/test/modal.spec.js
```diff
@@ -427,7 +427,10 @@ describe('$uibModal', function() {
       var modal = open({template: '<div>Content</div>'});
       expect($document).toHaveModalsOpen(1);
 
-      $document.find('body > div.modal').click();
+      var selector = 'body > div.modal';
+      $document.find(selector).mousedown();
+      $document.find(selector).mouseup();
```
wesleycho: Shouldn't this be the backdrop?

alexandernst:
Any news on this?

pzpzpzp1:
there's a line 'element.one'... is that supposed to be element.on?

wesleycho:
`element.one` is correct - it is intended to only be bound once.

PR #5913, opened 2016-05-20
Fix IE9 compatibility by expliticly stating empty href urls

mrpmorris:
TL;DR: In IE9 it was not possible to give focus to accordion items or tabs via the keyboard. 

IE9 doesn't recognise a simple **href** attribute unless you give it a value **href=""**, and it seems to strip them out which results in the link no longer being a tab stop.


wesleycho:
Closing this as we actually did do this at one point, but this turns out to trigger routing issues in some instances.

Please open an issue with a reproduction if this is an issue.

PR #5922, opened 2016-05-23
fix(typeahead): change to select class

### LINKED ISSUE #5848: typeahead expects li in template
The code: 
          target = popUpEl.find('li')[scope.activeIdx];
          target.parentNode.scrollTop = target.offsetTop;
in typeahead.js in master fails (target is undefined) if the template doesn't contains -li- (we use -tr- instead of -li- and -table- instead of -ul- by typeahead-popup-template-url).

When the developper tool is not opened in the browser everything runs fine despite the error.

If using -li- is officialy required in custom templates, then this is a feature request to allow other html elements, and if -li- is already allowed today then it is to report the bug.

wesleycho:
- Change to select `.uib-typeahead-match` class instead of the `li` tag

BREAKING CHANGE: This changes the selector used so that it doesn't select for the `li` tag specifically, but the elements with the class `uib-typeahead-match` - if using a custom template, this class needs to be added to the `li` element in the typeahead popup template.

This fixes #5848


deeg:
This LGTM

PR #5927, opened 2016-05-25
removes empty href attribute from anchor tags in the accordion-group …

zonika:
…template, preventing the user from being able to open a new window with the same url of the current page, which seems counterintuitive and useless given that there is no value given to the href in the default template. Remains HTML safe because it uses tabindex, so I don't see a reason for there to be an empty href attribute on the tag.


wesleycho:
It actually affects CSS

zonika:
How? (just curious as I'm using it on a project and see no difference)

RobJacobs:
The blank href is there for a reason, I can't remember the specific details but if I remember correctly it is necessary to support the anchor element receiving focus correctly.  If you do a search for 'href' on closed issues you can find the history.

PR #5930, opened 2016-05-27
Update timepicker tabindices

johanneskrtek:
Prevent focus on increment/decrement link buttons.


wesleycho:
This has negative accessibility ramifications - what is the rationale for changing this?

johanneskrtek:
The intention of the proposed change is actually an improvement of accessibility: a native user agent input control of the type `datetime-local` or `number` does not expose the increment/decrement buttons to the accessibility API, those buttons do not receive the focus, that is.

While the increment/decrement buttons do have their point when the timepicker is used by a mouse, it is more convenient (and closer to HTML5 user agent controls) to skip them when the timepicker is used by a keyboard, avoiding unnecessary keystrokes in order to focus on the next input control. Using a keyboard, it seems more intuitive to simply enter numbers in an input control than to use the increment/decrement buttons.

icfantv:
@wesleycho, was @johanneskrtek's reasoning sound?  if so, this will need tests before merging.

wesleycho:
Hm, I think this is actually ok - as long as the inputs are tabbable, this is ok.

It actually turns out that this is not properly testable just due to how tabindex works, so this is fine to merge as is.

PR #5938, opened 2016-05-29
feat(tooltip): add expression support

### LINKED ISSUE #5221: Tooltip-trigger "input" hides the tooltip on every second keystroke
Hey people.

I'd like to validate a text input field and show a tooltip as soon as the input is evaluated as valid (e.g. through Angular's form property `myForm.myField.$valid`). In order to catch the input change events I set `tooltip-trigger="input"`.

But on every second change to the input the field, the tooltip would disappear, thus confusing the user (and me :P).

Here's a demo: http://plnkr.co/edit/6kIeoHSyFpIAvvTpgVFt?p=preview

Happens with Angular 1.4.8 and Angular-Bootstrap 1.0.3.

wesleycho:
- Add expression support for tooltip-trigger
- Add support for object event bindings for showing/hiding tooltips

BREAKING CHANGE: This removes support for plain strings and interpolations in tooltip-trigger and popover-trigger - please change these appropriately. See test changes in this commit for reference

This closes #5221.


icfantv:
LGTM

PR #5943, opened 2016-05-30
fix(dropdown): align position with or without vertical scrollbar

### LINKED ISSUE #5942: Right alignment is off for dropdown-append-to-body .dropdown-menu-right when there's no vertical scrollbar
### Bug description:
As pointed out by @achenxp, my fix for issue #4317 is incorrect, and causes the dropdown to receive the wrong position when there's no overflow (and the vertical scrollbar is hidden).
This is because the fix always adds the scrollbar width to the position, without checking if the parent (body) element overflows.

**Note**: I have a working fix at my fork here: [dolevd@f838eb6](https://github.com/dolevd/bootstrap/commit/f838eb649ed59364fa7c19f4448e6e6ecae9f7f9)
This fix checks for overflow whenever the dropdown is open and is right aligned relative to the body, and calculates the position accordingly.
I'll submit this fix as a pull request shortly.

http://plnkr.co/edit/dIXoXxcH7RoeyHYIB4qw?p=preview

Notice that the first dropdown is incorrectly aligned to the right.

Angular: 1.5.5

UIBS: 1.3.3

Bootstrap: 3.3.6

dolevd:
Fixes #5942.

Plunkr showing the fix:
http://plnkr.co/edit/jlBJjMgpDnR133ByUATF?p=preview

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ If you would prefer not to load your css through your JavaScript file loader/bun
 * datepicker
 * datepickerPopup
 * dropdown
+* modal
 * popover
 * position
 * timepicker


icfantv:
@dolevd your fix lacks tests and failed the eslinter which is why the build failed.  you should try and do the build locally before actually committing so you can ensure your tests pass and the build succeeds.  otherwise, you may just be wasting your time.

please correct the issues, squash your commits, and we will go from there.

icfantv:
@dolevd do you intend on fixing this or shall i close this?  thanks.

dolevd:
Sorry for the radio silence, I'll fix this patch and resubmit.

PR #5945, opened 2016-05-31
fix(datepickerPopup): clear date with unit test

rcholic:
Fixed a bug in clearing a previously selected date in the `datepickerPopup` and added a unit test for it. The committed code include the bug fix, which passes the unit test


wesleycho:
LGTM - thanks for this!

PR #5954, opened 2016-06-03
fix(datepickerPopup): datepicker.html loading error

cherry-geqi:
fix 
```
Error: [$compile:tpload] Failed to load template: uib/template/datepicker/datepicker.html (HTTP status: 404 Not Found)
```
when

```js
require('angular-ui-bootstrap/src/datepickerPopup');
```


wesleycho:
Yikes, this looks like a good catch - thanks for the PR!

Issue #5800, opened 2016-04-14
Mandatory semver standartization


Jurgiss:
### Bug description:
Current version model does not follow semver pattern and is causing problems with packages. According to it, every time breaking changes happen, MAJOR package version MUST be increased. As it is now, the team is increasing MINOR version even though breaking changes happen.

While this might look nice from subjective perspective, problems start happening when you employ bower/npm package files. These two tools use semver explicitly and if you configure package.json and bower.json going by that standard, you can usually include all minor releases because they should not break anything. In case of your package, it very often breaks entire apps.

**tl;dr**: please use semver - MAJOR version for breaking changes, MINOR version for non-breaking new features; PATCH version for bug fixes. More information http://semver.org/

Foxandxss:
I agree with what you say. We will discuss what we can do.

On the other hand, it is always recommended to check the changelog before bumping a version.

Still, I see the value on that.

Jurgiss:
It is not always possible to control version change. For example when working on bigger projects, if bower is configured, let's say "angular-ui": "^0.2.x", when building solution latest version, say 0.4.2, will be download, as bower folder is usually excluded from git repositories.
While this is a fault of person who initially configured package version, it could still be prevented by keeping semver logic strictly.

Foxandxss:
Also prevented by removing the `^` from it :P

But yes, we will talk about this.

wesleycho:
We just talked about this, we intend to follow semver starting at 2.0.

This will likely start with the removal of the deprecated datepicker literal usage code, and whatever else makes it into that release.

We need to document our versioning though so people are aware. Pre 2.0 will not follow semver, so advise to take a look at the changelog is important when considering any upgrade. 2.0 and onwards will support semver as much as possible.

wesleycho:
Tagging as a doc issue - we need to update the docs to detail that from 2.0.0 onwards, we will be following semver as best as possible (we may make mistakes out of accident, but that can't be wholly prevented).

PR #5958, opened 2016-06-03
docs(readme): add note about versioning policy

wesleycho:
- Add information on versioning policy

This should address #5800.


icfantv:
Lgtm

PR #5959, opened 2016-06-04
fix(tooltip): reposition for placement top

### LINKED ISSUE #5914: Indeterminant tooltip position with element near screen edge
### Bug description:

UIB tooltip alternates between bad position and desired position. Specifically, with elements near the edge of the screen. In this case, a float-right anchor. The tooltip has default position (top).

You'll notice that changing the element tag has no affect, and neither does changing the positioning to, for example, absolute with right: 0. The same result: an indeterminant tooltip position. I think it must therefore be related to the screen edge.

I explored whether this was a caching issue. I do not believe so. I am precaching locally and it does not resolve the issue. In the plunkr, I cache the template, and have verified it is in my chrome cache by inspecting `chrome://cache`.

Plunker is [here](http://plnkr.co/edit/eMulLM04mYtyr7yhAhl6?p=preview).

### Version Info
Angular: 1.5.3

UIBS: 1.3.2

Bootstrap: 3.3.6

wesleycho:
- Reposition the tooltip when the placement option contains top in order to account for resizing of tooltip from right side

I did some research and encountered this in Bootstrap's tooltip service: https://github.com/twbs/bootstrap/blob/master/js/tooltip.js#L282-L284

I replicated this logic, and it can be seen in action [here](http://plnkr.co/edit/J5j4JHUQH2WdpmFdGajy?p=preview)

This correctly fixes #5914

cc @RobJacobs


RobJacobs:
Nice! LGTM

wesleycho:
Updated the top class check - take another look again when you get the chance.

I want to abstract this logic away into a nice helper method on $uibposition I think, even if it is small. What do you think about that?

RobJacobs:
I'm all for abstracting logic for re-use.  I made a couple of tweaks that I'd like to run by you, see this [plunk](http://plnkr.co/edit/ynH9IwWz8mdRrhoy9aCv?p=info)  Biggest changes are:

Moved the height change check to the positionTimeout resolve rather than create another timeout.

When adjusting the css top value, it should be the difference of the initialHeight and currentHeight as the initialHeight of the tooltip is already factored into the css top value.

wesleycho:
Interesting, I like this except for the leaking of data. What are your thoughts on it? Maybe we should merge this and you can file a PR on top of it?

wesleycho:
Going to merge this, we can iterate from here.

PR #5983, opened 2016-06-11
refactor(position): abstract out common logic

wesleycho:
- Abstracting out conditional to position service adjustTop method to increase readability and to prepare it for use in dropdown/typeahead


icfantv:
LGTM

Issue #5712, opened 2016-03-30
Investigate ability to remove replace: true in components


icfantv:
Bootstrap 3's CSS class structure is too tightly coupled to the DOM for our current system of using `replace: true` in certain directives.

We need to research whether or not we can move the `uib-*` directives from an E to an A on `require` and use our templates as normal or with slight tweaks.

- [ ] Convert accordion directive
- [ ] Convert alert directive
- [ ] Convert carousel directive
- [ ] Convert datepicker directive
- [ ] Convert modal directive
- [ ] Convert pager directive
- [ ] Convert pagination directive
- [ ] Convert popover directive
- [ ] Convert progressbar directive
- [ ] Convert rating directive
- [ ] Convert tabs directive
- [ ] Convert timepicker directive
- [ ] Convert tooltip directive
- [ ] Convert typeahead directive

PR #5986, opened 2016-06-12
feat(alert): remove replace: true usage

wesleycho:

BREAKING CHANGE: This removes the `replace: true` usage, so this has an effect on how one uses the directive in the template - see documentation for reference

This assists in #5712.


src/alert/docs/demo.html
```diff
@@ -1,11 +1,9 @@
...
   <script type="text/ng-template" id="alert.html">
-    <div class="alert" style="background-color:#fa39c3;color:white" role="alert">
-      <div ng-transclude></div>
-    </div>
+    <div ng-transclude></div>
   </script>
 
-  <uib-alert ng-repeat="alert in alerts" type="{{alert.type}}" close="closeAlert($index)">{{alert.msg}}</uib-alert>
-  <uib-alert template-url="alert.html">A happy alert!</uib-alert>
+  <div uib-alert ng-repeat="alert in alerts" ng-class="'alert-' + (alert.type || 'warning')" close="closeAlert($index)">{{alert.msg}}</div>
```
icfantv: You removed the `type` atteibute from the docs but still reference it here.  Is that just because it's our model def and we can call it whatever we want?

icfantv:
LGTM

src/alert/docs/demo.html
wesleycho: Yep

aciccarello:
I wish the changelog comment had more information about what needed to be updated. For future reference, it wasn't clear to me what had changed so I missed the necessary changes to defining the alert type in my code. :rose:

aciccarello:
I wish the changelog comment had more information about what needed to be updated. For future reference, it wasn't clear to me what had changed so I missed the necessary changes to defining the alert type in my code. :rose:

PR #5988, opened 2016-06-12
feat(datepicker): remove replace: true usage

wesleycho:

BREAKING CHANGE: As a result of removal of `replace: true`, there is the potential that this may break some CSS layout due to the slightly different HTML. Refer to the documentation examples to see the new structure.

Implements part of the work for #5712.


icfantv:
LGTM

PR #5989, opened 2016-06-13
feat(modal): remove replace usage

wesleycho:

BREAKING CHANGE: This removes `replace: true` usage, causing some structural changes to the HTML - the major part here is that there is no more backdrop template, and the top level elements for the window/backdrop elements lose a little flexibility. See documentation examples for new structure.


TechCraver1:
@wesleycho Milestone 2.0.1 for this.

wesleycho:
? This is going to be in the upcoming 2.0.0 release.

src/modal/modal.js
```diff
@@ -493,13 +497,22 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
         topModalIndex = previousTopOpenedModal ? parseInt(previousTopOpenedModal.value.modalDomEl.attr('index'), 10) + 1 : 0;
         var angularDomEl = angular.element('<div uib-modal-window="modal-window"></div>');
         angularDomEl.attr({
+          'class': 'modal',
           'template-url': modal.windowTemplateUrl,
-          'window-class': modal.windowClass,
```
icfantv: do we need to document that this attribute is removed as part of the breaking changes?

wesleycho: The class is meant for adding to element the class, but that is now happening without needing a binding.

PR #5991, opened 2016-06-13
feat(pager): remove replace usage

wesleycho:
- Remove `replace: true usage

BREAKING CHANGE: This removes `replace: true` usage from the pager, which causes a slight usage change - see documentation examples for new usage.

This addresses #5712 partially.


icfantv:
LGTM

PR #5992, opened 2016-06-13
feat(pagination): remove replace usage

wesleycho:

BREAKING CHANGE: Due to the removal of `replace: true`, this affects the HTML structure slightly - see documentation examples to see new usage.


icfantv:
LGTM

PR #5993, opened 2016-06-13
feat(datepickerPopup): remove replace usage

wesleycho:

BREAKING CHANGE: Due to the nature of `replace: true`, this has a slight structural HTML change in the popup as a result - see documentation examples for the change in action.


icfantv:
LGTM

PR #5994, opened 2016-06-13
feat(tooltip): remove replace usage

wesleycho:

BREAKING CHANGE: The template structure changed slightly due to the removal of `replace: true` - see documentation examples in action to see differences in practice.


icfantv:
We need to fully document this change in the release notes as this is going to break a lot of people's code (and upset a lot of people probably...).  Also need to look at the demos and make sure nothing needs to be tweaked.

Should investigate having demos to show the users how to still get the functionality we've removed:  `is-open` and `placement`.  Hopefully, we don't have to explain `popup-class`.

wesleycho:
Looks like this survived without a docs update - nothing needs to be changed in the demo pages.

icfantv:
I was thinking that since we're _removing_ core API features, we should give examples on how the user should be doing this on their own end.  Otherwise, we'll be getting issues for "how do I...?"  It would be nice to be able to point to something instead of just slapping them down w/ the standard issue forum use response.  But I'm also in a benevolent mood at the moment.  ;-)

icfantv:
LGTM

PR #5996, opened 2016-06-13
feat(pager): toggle tabindex when disabled

wesleycho:
- Add `tabindex="-1"` when disabled and remove it when not for accessibility


icfantv:
LGTM

PR #5997, opened 2016-06-13
feat(timepicker): remove replace usage

wesleycho:

BREAKING CHANGE: This removes `replace: true`, which changes the HTML structure slightly - see the documentation examples to see it in action.


icfantv:
LGTM

PR #5998, opened 2016-06-13
feat(rating): remove replace usage

wesleycho:

BREAKING CHANGE: Due to the removal of `replace: true`, this results in a slight change to the HTML structure - see the documentation examples to see it in action.


icfantv:
LGTM.

PR #6002, opened 2016-06-14
fix(tooltip): remove when body is clicked

naveed3855:


pkozlowski-opensource:
Huh? I assume this one was opened by accident...

Issue #6004, opened 2016-06-14
uibDropdownConfig should be a .value() instead of a .constant() to allow decorating


onlywei:
I think it's pretty useful to allow people to customize this object via `angular.module().decorator()`, because maybe they want to use their own CSS classes. It should be an easy change, as I don't see `uibDropdownConfig` being injected anywhere other than a controller.

wesleycho:
One can use `$provide.value` to override it, but I'm actually not sure why they are `.constant`s.

PRs welcome!

PR #6006, opened 2016-06-14
feat(dropdown): use .value() for uibDropdownConfig

onlywei:
This allows people to use .decorator() to customize this object.

#6004.

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -1,6 +1,6 @@
 angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
 
-.constant('uibDropdownConfig', {
+.value('uibDropdownConfig', {
   appendToOpenClass: 'uib-dropdown-open',
   openClass: 'open'
 })


onlywei:
#6004

wesleycho:
Note that the issue shouldn't be closed, as this should be done for every component.

wesleycho:
Unfortunately, I'm going to have to revert this since it turns out this is a breaking change (the constant becomes no longer injectable inside a config function).

onlywei:
@wesleycho why is that a breaking change? I grepped through the `src/` directory and found that it was not being injected into any config functions.

pkozlowski-opensource:
> I grepped through the src/ directory and found that it was not being injected into any config functions.

But users of this lib might be doing so.

onlywei:
I see. In that case, going forward, will the below be the recommended way for users of the lib to configure the dropdown library?
```
import angular from 'angular';
import uibDropdown from 'angular-ui-bootstrap/src/dropdown';

angular.module('MyApp', [uibDropdown])
  .config(function (uibDropdownConfig) {
    uibDropdownConfig.openClass = 'is-open';
  });
```

?

PR #6007, opened 2016-06-14
* fix(datepickerPopup): use value instead of viewValue

dlemstra:
This looks like a bug to me. There is a check to test if `value` is a string but `viewValue` is passed to parseDateString.


wesleycho:
Yep, smells like an issue - good catch!

PR #6008, opened 2016-06-14
fix(dropdown): align position with or without vertical scrollbar

### LINKED ISSUE #5942: Right alignment is off for dropdown-append-to-body .dropdown-menu-right when there's no vertical scrollbar

dolevd:
Fixes #5942.

Plunkr showing the fix:
http://plnkr.co/edit/jlBJjMgpDnR133ByUATF?p=preview

Resubmitted after fixing eslint issues.

Some notes regarding the added test:
1. I've added a test that verifies that dropdown-menu-right works correctly when append-to-body is used. This test requires a real browser (since it must calculate the trigger and menu positions), so you may not wish to include it... On the other hand, it might be better than nothing.
2. The test doesn't cover the case where a vertical scroll bar is present. As far as I can tell, this is very problematic to test since it requires a real environment where we can correctly calculate the width of the scrollbar. This isn't the case when running in Karma.

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -195,7 +195,8 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
       var pos = $position.positionElements($element, self.dropdownMenu, 'bottom-left', true),
         css,
         rightalign,
-        scrollbarWidth;
+        scrollbarPadding,
+        scrollbarWidth = 0;
 
       css = {
         top: pos.top + 'px',
@@ -208,7 +209,12 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
         css.right = 'auto';
       } else {
         css.left = 'auto';
-        scrollbarWidth = $position.scrollbarWidth(true);
+        scrollbarPadding = $position.scrollbarPadding(appendTo);
+
+        if (scrollbarPadding.heightOverflow && scrollbarPadding.scrollbarWidth) {
+          scrollbarWidth = scrollbarPadding.scrollbarWidth;
+        }
+
         css.right = window.innerWidth - scrollbarWidth -
           (pos.left + $element.prop('offsetWidth')) + 'px';
       }
--- a/src/dropdown/test/dropdown.spec.js
+++ b/src/dropdown/test/dropdown.spec.js
@@ -702,4 +702,33 @@ describe('uib-dropdown', function() {
       });
     });
   });
+
+  describe('using dropdown-append-to-body with dropdown-menu-right class', function() {
+    function dropdown() {
+      return $compile('<li style="float: right;" uib-dropdown dropdown-append-to-body><a href uib-dropdown-toggle>Toggle menu</a><ul uib-dropdown-menu class="dropdown-menu-right" id="dropdown-menu"><li><a href>Hello On Body</a></li></ul></li>')($rootScope);
+    }
+
+    beforeEach(function() {
+      element = dropdown();
+      $document.find('body').append(element);
+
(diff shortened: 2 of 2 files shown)


src/dropdown/dropdown.js
```diff
@@ -195,7 +195,8 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.position'])
       var pos = $position.positionElements($element, self.dropdownMenu, 'bottom-left', true),
         css,
         rightalign,
-        scrollbarWidth;
+        scrollbarPadding,
+        scrollbarWidth = 0;
```
wesleycho: This probably causes an eslint issue - define this in a separate var

dolevd: Actually, the eslint issue had to do with using tabs instead of spaces (I used Notepad++ which isn't the best tool for the job).
Eslint passes now, but I can change this if you wish.

wesleycho: Nah, not a big deal

wesleycho:
On the contrary, having a test is only a good thing!

This PR looks good to me, but can you explain what am I supposed to see in the Plunker from #5942 and this one?

dolevd:
In the Plunker from this PR you should hopefully see nothing out of the ordinary :).

The Plunker from #5942 shows that the dropdown with append-to-body and dropdown-menu-right is offset to the right by the width of the scrollbar when the scrollbar **isn't** visible.
In Edge and Safari, the scrollbar width is 0, so this doesn't have any effect.
However, in Chrome the normal scrollbar width is 17px, so you should see that the first dropdown is offset by 17px to the right (instead of being aligned to the right edge of the trigger element).

This happened because my original fix (that landed in 1.3.3) was very naive and simply assumed that the scrollbar is always present which obviously isn't the case.

wesleycho:
I actually am not able to reproduce the issue in Chrome (OS X) - what is your OS that you are seeing the issue on, Windows?

dolevd:
Yes windows.
This is a screenshot from my computer: http://imgur.com/tWBhIh1

dolevd:
I think OSX has scrollbars that are overlaid on top, so they probably don't change the window size and affect the right alignment.

wesleycho:
Alright, I'll verify when I get home a little later and likely merge then after verification

src/dropdown/test/dropdown.spec.js
```diff
@@ -702,4 +702,33 @@ describe('uib-dropdown', function() {
       });
     });
   });
+
+  describe('using dropdown-append-to-body with dropdown-menu-right class', function() {
```
wesleycho: In a comment before this issue, can you mention `// issue #5942`?

wesleycho:
Sorry for taking so long - been traveling the past two weekends.

This LGTM.

PR #6011, opened 2016-06-15
feat(accordion): use .value() for uibAccordionConfig

wesleycho:


pkozlowski-opensource:
What is a motivation for this / other changes like this?

wesleycho:
Being able to override via `.decorator()`

pkozlowski-opensource:
> Being able to override via .decorator()

The potential downside being that it is no longer injectable in providers, if this matters,,,,

wesleycho:
Hmm, so I guess it technically is a breaking change. Hmm...I feel like the main reason it is used is to override our defaults with the user specified ones, but I guess someone could be using it to configure app wide settings.

IMO it should have always been .value, but it seems it's too late now. Probably should close all my PRs then, and need to revert the commit of the dropdown one.

PR #6020, opened 2016-06-15
feat(rating): parameterize icon base class

roman-garcia:
As of today it is not possible to use a different source of icon because the glyphicon class is systematically set in the rating template.

This patch adds a parameter in the directive for the icon base class, setting it to 'glyphicon' by default.

### DIFF
--- a/src/rating/test/rating.spec.js
+++ b/src/rating/test/rating.spec.js
@@ -365,4 +365,25 @@ describe('rating directive', function() {
       expect(getTitles()).toEqual(['one', 'two', 'three', 'four', 'five']);
     });
   });
+
+  describe('Default icon base class', function() {
+    it('should return the default icon base class for each star', function() {
+      var stars = getStars();
+      for (var i = 0, n = stars.length; i < n; i++) {
+        expect(stars.eq(i).hasClass('glyphicon'));
+      }
+    });
+  });
+
+  describe('Custom icon base class', function() {
+    it('should return the custom icon base class for each star', function() {
+      $rootScope.iconBaseClass = 'foo';
+      element = $compile('<span uib-rating ng-model="rate" icon-base-class="iconBaseClass"></span>')($rootScope);
+      $rootScope.$digest();
+      var stars = getStars();
+      for (var i = 0, n = stars.length; i < n; i++) {
+        expect(stars.eq(i).hasClass('foo'));
+      }
+    });
+  });
 });
--- a/template/rating/rating.html
+++ b/template/rating/rating.html
@@ -1,4 +1,4 @@
 <span ng-mouseleave="reset()" ng-keydown="onKeydown($event)" tabindex="0" role="slider" aria-valuemin="0" aria-valuemax="{{range.length}}" aria-valuenow="{{value}}" aria-valuetext="{{title}}">
     <span ng-repeat-start="r in range track by $index" class="sr-only">({{ $index < value ? '*' : ' ' }})</span>
-    <i ng-repeat-end ng-mouseenter="enter($index + 1)" ng-click="rate($index + 1)" class="glyphicon" ng-class="$index < value && (r.stateOn || 'glyphicon-star') || (r.stateOff || 'glyphicon-star-empty')" ng-attr-title="{{r.title}}"></i>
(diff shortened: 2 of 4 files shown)


icfantv:
I'm ok with this.  @wesleycho, thoughts?

@roman-garcia if we proceed with this, it will need the appropriate tests.

wesleycho:
This needs documentation updates as well.

I'm also fine with this feature.

template/rating/rating.html
```diff
@@ -1,4 +1,4 @@
 <span ng-mouseleave="reset()" ng-keydown="onKeydown($event)" tabindex="0" role="slider" aria-valuemin="0" aria-valuemax="{{range.length}}" aria-valuenow="{{value}}" aria-valuetext="{{title}}">
     <span ng-repeat-start="r in range track by $index" class="sr-only">({{ $index < value ? '*' : ' ' }})</span>
-    <i ng-repeat-end ng-mouseenter="enter($index + 1)" ng-click="rate($index + 1)" class="glyphicon" ng-class="$index < value && (r.stateOn || 'glyphicon-star') || (r.stateOff || 'glyphicon-star-empty')" ng-attr-title="{{r.title}}"></i>
+    <i ng-repeat-end ng-mouseenter="enter($index + 1)" ng-click="rate($index + 1)" ng-class="(r.iconBaseClass && r.iconBaseClass + ' ' || '') + ($index < value && (r.stateOn || 'glyphicon-star') || (r.stateOff || 'glyphicon-star-empty'))" ng-attr-title="{{r.title}}"></i>
```
icfantv: what exactly is this?  what's the point of the string concatenation?

icfantv:
@wesleycho oooo yea, good call on the docs.  totally missed that.

icfantv:
@roman-garcia, i'd LOVE to get this in to the 2.0 milestone release (imminent) so if you can address these muy pronto, that would be SWEET.  Cheers.

template/rating/rating.html
roman-garcia: I'm sorry I don't quite understand the question. I moved the icon base class from class to ng-class so that it would be interpreted correctly. I had to manage the case of undefined icon base class.

roman-garcia:
I added tests and documentation as asked.

src/rating/test/rating.spec.js
```diff
@@ -365,4 +365,25 @@ describe('rating directive', function() {
...
+  });
+
+  describe('Custom icon base class', function() {
+    it('should return the custom icon base class for each star', function() {
+      $rootScope.iconBaseClass = 'foo';
+      element = $compile('<span uib-rating ng-model="rate" icon-base-class="iconBaseClass"></span>')($rootScope);
+      $rootScope.$digest();
+      var stars = getStars();
+      for (var i = 0, n = stars.length; i < n; i++) {
+        expect(stars.eq(i).hasClass('foo'));
```
icfantv: also make sure it does NOT have class `glyphicon`.

template/rating/rating.html
icfantv: the more i think about this, why would we only support just the base class?  why not support overriding the star class as well?

wesleycho:
Thinking about this further, I don't think this is a good idea to land in the library - a better option would be to override the template so that the glyphicon references are switched to the desired font icons and/or making use of the `state-on` and `state-off` attributes.

Going to close as something we won't support unfortunately - supporting this would cascade across our other components, and a suitable workaround for all is modifying the template used.

icfantv:
@wesleycho yea, after thinking about this, i agree.  i just realized it didn't make sense to only just do the base class but to also do the star class as well and at that point, why not just override the template?

template/rating/rating.html
reda-alaoui: What do you mean? You already support the star class via stateOff and stateOn.

reda-alaoui:
Hello everyone,

Since `state-on` and `state-off` are already settable, I think it is consistent to allow the icon base class configuration. 
If the answer is `Replace the template`, then why `state-on` and `state-off` can be modified by the user? Why the default template should force us to use glyphicon?

icfantv:
@reda-alaoui - that's not a valid reason to add another configuration option.  we have to balance API features with maintainability.  we've added the ability for developers to override our existing templates so user's can tweak the behavior as they deem appropriate.

Additionally, adding more configuration options has performance ramifications as (arguably unnecessary) watches are then created.

PR #6022, opened 2016-06-16
fix(datepicker): reset datepickerMode to original values on destroy

### LINKED ISSUE #5968: Carousel transition breaks when viewing another tab on the browser
### Bug description:

When a carousel has automatic sliding, if you are viewing a different tab in the same browser when the transition happens, the carousel gets stuck. The auto transition no longer happens, and the manual transition using the arrows doesn't work either.  
I've recreated it in Chrome and Edge.

Note - if you simply minimise the browser or look at another window the bug doesn't happen, it seems to be only when looking at a different tab on the same browser instance. 

This is the same issue as reported here: https://github.com/angular-ui/bootstrap/issues/5471

**I can create this on your demo page**.

Angular: 1.5.6

UIBS: 1.3.3

Bootstrap: 3.3.6

### LINKED ISSUE #5978: click open popup button in datepicker second time, the popup is NOT 'year' mode anymore.

### Bug description:
set datepicker's datepicker-mode to 'year', `<input type="text" datepicker-mode"'year'" is-open="isOpen" /><button ng-click="isOpen = true">Open</button>`, when click the open button, the popup is RIGHTLY opened with `year` mode. Manually clean the date inside the input box, click the button again. This time the popup is displayed but NOT in `year` mode.

Angular: 1.4.8

UIBS: 0.13.4 and 0.14.3

Bootstrap: 3.3.6

josetaira:
Add setMode call on destroy to reset the datepickerMode based on original value
from options.

Closes #5978

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -157,6 +157,12 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
     }
     return false;
   };
+  
+  $scope.$watch("selectedDt", function(dt) {
+    if(!dt || ngModelCtrl.$invalid) {
+      setMode(_datepickerModeCache);
+    }
+  });
 
   this.init = function(ngModelCtrl_) {
     ngModelCtrl = ngModelCtrl_;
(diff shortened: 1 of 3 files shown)


wesleycho:
This is a flawed approach. On destruction alone it should not simply reset it. It should only reset it if it becomes an invalid date or not a date object.

josetaira:
Just to clarify, no reset to be done on $scope.$destroy?

I'll add the part where it will reset on invalid date and not a date object.

josetaira:
Latest commit should accommodate the changes you mentioned. I assumed that the setMode is not to be done in $destroy

wesleycho:
This needs to also account for when `datepickerMode` is programmatically changed by the user.

PR #6029, opened 2016-06-20
feat(modal): Switched the modal transclude to use $animate

### LINKED ISSUE #6023: uibModalTransclude is not animatable
### Description:
Within the modal, the uibModalTransclusion directive does not trigger ng-enter and ng-enter-active classes when the modal is appended. This is because in the transclusion function it calls
```js
element.append(clone);
```
If this is replaced with
```js
$animate.enter(clone, element);
```
Then it will trigger the animation cycle and allow them to be hooked into.

Angular: Any

UIBS: 1.3.3

Bootstrap: Any

cleavera:
Modal transclude switched to use $animate.enter rather than element.append, this is so that the transcluded content can hook in the animation lifecycle. Also added unit tests for the uibModalTransclude directive as there were none before.

Closes #6023


wesleycho:
LGTM

PR #6041, opened 2016-06-21
fix(modal): check for overflow hidden

### LINKED ISSUE #6037: Modal + Navbar + Styling causes body padding to be added where there was not previously a scrollbar
Hello, again.

### Bug Description:
When there's a navbar, followed by a set of containers configured to control scrolling in a specific way and a modal opens, the body is shifted left by one scrollbar's width even though there was not previously a scrollbar in place for which to compensate. Here's a short GIF of this effect in action:

bootstrap-modal-shift-plunkr

If the `<nav>` element is removed, the issue no longer occurs. If the classes of `page-content-wrapper` and `wrapper` are combined, the issue no longer occurs (and the inner scrollbar is no longer present). If the `overflow: hidden` on the body is removed, the outer scrollbar appears, and the compensation hides it without shifting anything.

I don't believe this is the same thing as #5749. This is the actual issue I alluded to in #6035. My intuition says that something about this combination of elements and styles is not being detected, and the calculation to determine the rendered width of the scrollbar is happening even though it is not needed, as the scrollbar was already hidden.

Hopefully I didn't miss another issue already tracking this!

Thanks, again, for your time, and this library.

http://plnkr.co/edit/07fH45MYlEfTFobTkCvA

### Version of Angular, UIBS, and Bootstrap:
**Angular:** 1.5.3
**UIBS:** 1.3.1
**Bootstrap:** 3.3.6

-------
_The URL for the plunker in the GIF isn't the same as the one in the link as I kept hitting <kbd>CTRL</kbd>+<kbd>S</kbd> out of habit, while I was writing the plunker, and it created like 920378402395 versions. I created a new one with a clean version history._

### LINKED ISSUE #6307: Tooltip not shown for divs within ng-repeat
I'm trying to show a tooltip on some divs of a website using AngularUI. The goal is to show them already when the user enters the site. That's why tooltop-is-open is set to true right from the beginning. However, this does not work for elements within a repeating group (i.e. divs created via ng-repeat). 

Link to plunkr: [https://plnkr.co/edit/m4Gd1b9CCpzdrIIJoOw3?p=preview](https://plnkr.co/edit/m4Gd1b9CCpzdrIIJoOw3?p=preview)

On [SO](http://stackoverflow.com/questions/40338916/tooltip-of-angularui-not-shown-for-divs-within-ng-repeat/40341788#40341788), one made the suggestion to use uib-tooltip-html instead. However, this only worked in the modified plunkr. Locally, even though the tooltips are shown, they are empty.

Versions:
- Angular 1.5.8
- UIBS 2.2.0
- Bootstrap 3.3.7 (CSS only)

RobJacobs:
Check if appendToElement has scrolling disabled before adding
the scrollbar padding.

Fixes #6037

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -487,9 +487,11 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
           }
           $compile(backdropDomEl)(backdropScope);
           $animate.enter(backdropDomEl, appendToElement);
-          scrollbarPadding = $uibPosition.scrollbarPadding(appendToElement);
-          if (scrollbarPadding.heightOverflow && scrollbarPadding.scrollbarWidth) {
-            appendToElement.css({paddingRight: scrollbarPadding.right + 'px'});
+          if ($uibPosition.isScrollable(appendToElement)) {
+            scrollbarPadding = $uibPosition.scrollbarPadding(appendToElement);
+            if (scrollbarPadding.heightOverflow && scrollbarPadding.scrollbarWidth) {
+              appendToElement.css({paddingRight: scrollbarPadding.right + 'px'});
+            }
           }
         }
 


icfantv:
LGTM

phazei:
This "fix" actually breaks the padding added in every other case.  It only checks the elements parent is scrollable, but the way it does the check, via visibility, fails for the body tag, since overflow is always 'visible' but it still produces scrollbars.  I'm not familiar with any case where the element isn't appended to the body, so I don't know what happens there, but in the default case, this breaks generall use so should be rolled back.

PR #6056, opened 2016-06-27
fix(modal): remove window class after animation

### LINKED ISSUE #6051: Modal removes openClass before the modal close animation has finished
### Bug description:
When using an ng-leave animation on the modal dialog the modal open class is removed at the start of the animation rather than at the end of the animation. This results in issues where you're using the open class to hide overflow on the body etc which get removed before the modal is closed and end up with a nasty flickering of the scrollbar in this example. Instead the modal open class should be removed after the animation.

### Example
http://plnkr.co/edit/A8YZeacWFK4yP6zMIIjY?p=preview
In this example the background is red while the open class is added to the body, you'll notice that the open class is removed at the start of the leave animation instead of when the animation has completed.

Angular: 1.5.3

UIBS: 1.3.3

Bootstrap: N/A

cleavera:
Modal dialog removeAfterAnimate switched to wait until after all animations are finished to call the done callback

Fixes #6051

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -375,6 +375,10 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
           afterAnimating.done = true;
 
           $animate.leave(domEl).then(function() {
+            if (done) {
+              done();
+            }
+
             domEl.remove();
             if (closedDeferred) {
               closedDeferred.resolve();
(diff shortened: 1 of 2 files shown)


wesleycho:
This makes sense - could you write up a quick test for this?

cleavera:
I've appropriated an existing test. All good?

wesleycho:
LGTM, thanks!

Issue #6086, opened 2016-07-13
Can't use 'outsideClick' as a hide trigger.


synthecypher:
I've configured a trigger, to show the popover when the custom event `popoverShow` is detected.

I also want the hide event of this custom trigger to be `outsideClick`.

If I configure it like so it will not hide on `outsideClick`.

```javascript
$uibTooltipProvider.setTriggers({
  'popoverShow': 'outsideClick'
});
```

Reproduced here.

http://plnkr.co/edit/QEv4i6iMMAw65IOfAuSh?p=preview

I basically need to have the popover appear after completing some async behaviour, so I have my own click handler which does the async behaviour then if applicable triggers the `popoverShow` event for the handler.

synthecypher:
Fixed the problem see this pull request.

https://github.com/angular-ui/bootstrap/pull/6087

icfantv:
closing per comments in #6087.

heavymanners:
The comment in #6087 doesn't address the core need, which is to have the outsideClick functionality for hide triggers. The comment says that the functionality is covered by 'click,' but it isn't, as click requires the user to click on the element itself, while outsideClick handles clicks outside of the element.

panxq5:
+1 this is helpful

PR #6087, opened 2016-07-13
Fix for using outsideClick as a hide trigger.

synthecypher:
Currently the implementation of binding the event listeners for tooltips only allows you to use `outsideClick` if it is the show trigger. This small change lets you use it as the hide trigger as well.

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -533,7 +533,11 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
                     element.on(trigger, toggleTooltipBind);
                   } else if (trigger) {
                     element.on(trigger, showTooltipBind);
-                    element.on(triggers.hide[idx], hideTooltipBind);
+                    if (triggers.hide[idx] === 'outsideClick') {
+                      $document.on('click', bodyHideTooltipBind);
+                    } else {
+                      element.on(triggers.hide[idx], hideTooltipBind);
+                    }
                   }
 
                   element.on('keypress', function(e) {


RobJacobs:
outsideClick by it's nature is already a 'hide' trigger, why would we ever support an outsideClick as an 'open' trigger?

wesleycho:
One can already get this functionality by specifying `hide: 'click'`.

Closing as redundant.

synthecypher:
@wesleycho Specifying `hide: 'click'` where in `setTrigger`?

sdicken:
@wesleycho This is not redundant. It is a needed enhancement. Ran into a use case for this today. Very frustrating to see this merge request ignored.

KarelCemus:
I run into the same scenario today, I vote for reopening the issue #6086 and merging this PR. 

Personally, I don't understand, why hiding triggers are ignored during the registration phase. They are [registered only based on show triggers](https://github.com/angular-ui/bootstrap/blob/f5b357fd1152f5462089b680fcf4a624b498185e/src/tooltip/tooltip.js#L544), but 
[unregistered independently](https://github.com/angular-ui/bootstrap/blob/f5b357fd1152f5462089b680fcf4a624b498185e/src/tooltip/tooltip.js#L516) using the array of hiding triggers. This **inconsistency doesn't make sense** plus **prevents the users from using embedded hide triggers with custom show triggers**. The current implementation is very limiting. I run into it when I need to show the popover after some async request but want to close it as usual.

NevenLeung:
I want to set the tigger like this, but the closeTrigger 'outsideClick' doesn't work.

$uibTooltipProvider.setTriggers({'mouseenter': 'outsideClick'});

$uibTooltipProvider.setTriggers({'mouseenter': 'click'});

This one works, but I perfer click on the outside of the button to close the popover. Any advice?

NevenLeung:
I want to set the trigger like this, but the closeTrigger '**outsideClick**' doesn't work.

```
$uibTooltipProvider.setTriggers({'mouseenter': 'outsideClick'});

$uibTooltipProvider.setTriggers({'mouseenter': 'click'});
```

The second one works, but I perfer clicking on the **outside** of the button to close the popover. Any advice?

------------------------update----------------------

I use the [this pull request](https://github.com/angular-ui/bootstrap/pull/6087/commits/f7b9b20c414bd1e27980ebc8f2f96f906ceb1f1f) by **synthecypher** to replace the code in ui-bootstrap-tpls.js. Now, my usage above works. Thanks! @synthecypher

synthecypher:
@wesleycho just thought I'd point out I've updated my plunker to use the latest libraries and show that `popover-trigger="outsideClick"` still does not achieve the designed behaviour of showing the popover when clicking on the element, and hide when clicking on anything else.

http://plnkr.co/edit/QEv4i6iMMAw65IOfAuSh?p=preview

synthecypher:
@wesleycho this is the closest you can get still with the latest version of the library there is still no way to use `outsideClick` as the hide trigger independently of the show trigger.

`popover-trigger="outsideClick"`

Will toggle the popover when clicking on the element.

`popover-trigger="{'outsideClick':'outsideClick'}`

Will toggle the popover when clicking on the element, and hide it when clicking off of the element.

http://plnkr.co/edit/QEv4i6iMMAw65IOfAuSh?p=preview

If my desired behaviour is to show and continue to show the tooltip when clicking on the element repeatedly and only hide when clicking off of the element, I can still not achieve this when the latest version of this library.

NevenLeung:
I have another behaviour want to achieve. 

It is showing the popover when hovering on an element, and hiding the popover when the cursor is away from **the element** and **the popover**, **not just away from the element**. 

I want to use such an implementation to show some content in popover by using a html template.

Any advice is appreciated!!

PR #6111, opened 2016-07-21
chore(docs): fix specified sizes

### LINKED ISSUE #6110: Footprint still misleading
The minified/gzipped version of angular-ui-bootstrap has sat at 28KB for a long time now. The 20KB claim at the top of this repo should be updated and/or removed.

icfantv:
Closes #6110


wesleycho:
Eventually we should probably find a way around this, as the file size will keep changing.

PR #6113, opened 2016-07-22
fix(tab): pass the tabs real index to onDeselect

### LINKED ISSUE #6107: Wrong $selecteIndex value for deselect callback
### Bug description
The ```deselect``` callback can take the ```$selectedIndex``` as a parameter. But this is always the index of the tab in the internal array of the tabset. But it should be the index of the tab itself.

When the index attribute of a tab is specified I would assume that the $selectedIndex parameter of the deselect callback is the index specified on the tab.

http://plnkr.co/edit/n2oXnJ5TqDQqfJBz5ur0?p=preview

Angular: 1.5.7

UIBS: 2.0.0

Bootstrap: 3.3.6

furti:
The $selectedIndex passed to the onDeselect callback is the same as the
index given in the tab template. One can use that index as it is for the
active value of the tabset.

Fixes #6107

### DIFF
--- a/src/tabs/tabs.js
+++ b/src/tabs/tabs.js
@@ -9,10 +9,12 @@ angular.module('ui.bootstrap.tabs', [])
     if (!destroyed) {
       var previousIndex = findTabIndex(oldIndex);
       var previousSelected = ctrl.tabs[previousIndex];
+      var selected = ctrl.tabs[index];
+
       if (previousSelected) {
         previousSelected.tab.onDeselect({
           $event: evt,
-          $selectedIndex: index
+          $selectedIndex: selected ? selected.index : undefined
         });
         if (evt && evt.isDefaultPrevented()) {
           return;
@@ -20,7 +22,6 @@ angular.module('ui.bootstrap.tabs', [])
         previousSelected.tab.active = false;
       }
 
-      var selected = ctrl.tabs[index];
       if (selected) {
         selected.tab.onSelect({
           $event: evt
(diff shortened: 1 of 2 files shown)


furti:
I'm not sure if this is the right place for this but i have a question for a test case in ```tabs.spec.js```

The test case is as follows:

```javascript
it('should prevent tab deselection when $event.preventDefault() is called', function() {
      spyOn(scope, 'deselectThird');
      titles().eq(2).find('> a').click();
      expect(scope.active).toBe(3);
      titles().eq(1).find('> a').click();
      expect(scope.deselectThird).toHaveBeenCalled();
      expect(scope.active).not.toBe(1);
      expect(scope.active).toBe(2);
    });
```

1. create spy -> OK
2. select the third tab -> OK
3. the active tab should be the third one -> OK
4. Select the second tab -> OK
5. The deselectThird callback should have been called -> OK
6. the active tab should not be the first one -> Confusing. The first tab was not even involved in this whole scenario because we tried to switch from the third tab to the second one. Why is this checked here?
7. the active tab should be the second one -> Confusing. The deselectThird callback prevents the default action. So the active tab should still be the third one as the tab change got cancelled.

wesleycho:
Sorry for the long silence here - I have been on vacation for the past 2 weeks in Europe, and return home this upcoming weekend. I will look at this & the original issue in more detail next week.

wesleycho:
Scheduling this for 3.0.0 - this is a breaking change, so as per semver, we'll have to hold off.

PR #6104, opened 2016-07-20
fix(modal): restore broken stacked modals

fpipita:
Hi, 

with 2.0.0, stacked modals are not getting the `zIndex` property correctly set anymore because the `index` property needed by `ngStyle` to compute the `zIndex` property, is not available on the `ngStyle` scope (which is the modal scope).

This PR solves it by simply adding a new "private" property to the modal scope, `$$topModalIndex` which provides the actual modal index value.

Closes #6103

PR #6116, opened 2016-07-23
fix(modal): prevent doubleclick from closing multiple nested modals

### LINKED ISSUE #6061: Mutliple modals, double click closes two modals instead top.
When there are two nested modals, i.e. one modal is shown from another, double click on backdrop of the top modal results in closing of both modals instead only top.

Problem is reproduced even if double click is done when cursor is above first modal. As in the attached image from plunker, need to click in selected area.

image

http://plnkr.co/edit/Fidgh0aBW6LMi9nUaetg

jkoontz2010:
Hi,

In the normal Bootstrap library, double clicking on the backdrop of a nested modal results in just the nested modal closing. It waits for the modal-closing animation to complete, and once it is, the user can then click on the next backdrop to close the remaining modal. 

The current Angular UI code doesn't do this because every time _any_ backdrop is clicked, the top modal on the $modalStack is chosen for closing. 

This PR prevents this functionality by simply adding a boolean value backdropClosing that, when true, will prevent the code from executing another modal close. Returning backdropClosing to false is unnecessary and won't affect the next modal close.

Closes #6061

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -144,11 +144,15 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
         element.addClass(attrs.windowTopClass || '');
         scope.size = attrs.size;
 
+        // inserted to fix issue #6061
+        var backdropClosing = false;
         scope.close = function(evt) {
           var modal = $modalStack.getTop();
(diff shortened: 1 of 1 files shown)


icfantv:
Closing as duplicate of #6104

PR #6139, opened 2016-08-03
Updated demo file

gerardorn:
The definition that overrides the default template required a wrapping div with the panel class


icfantv:
LGTM.  Thanks.

PR #6142, opened 2016-08-05
Adding a note about datetimepicker

zhaber:


wesleycho:
Closing, as this is not really a change that makes sense in the documentation. It does not make a recommendation for any particular library, and any library supporting certain things are searchable.

PR #6165, opened 2016-08-16
docs(tooltip): add ngSanitize to Plunker

### LINKED ISSUE #6157: Tooltip example plunkr bug (fixed plunkr provided)
Hello,

The plunkr example for tooltip is off when it comes to uib-tooltip-html with a string.

`<a href="#" uib-tooltip-html="'static. {{dynamicTooltipText}}. <b>bold.</b>'">inline string</a>`

The plunkr doesn't provide safe html or sanitizing so we have the following error: 
` [$sce:unsafe] Attempting to use an unsafe value in a safe context.`

I have fixed the issue in this plunkr: http://plnkr.co/edit/V6xyTmtGhQI5md7x8OQA?p=preview
by adding ngSanitize to the example.

Thanks for the library and have a nice day!

wesleycho:
- Add ngSanitize to Plunkers for uib-tooltip-html and uib-popover-html Plunkers

This fixes #6157.


icfantv:
LGTM

Issue #3474, opened 2015-03-30
Improve modal accessibility


ndarilek:
I'm attempting to improve the accessibility of some angular-ui modals, and am wondering if more can be done out of the box to make them workable by default. In general, modal accessibility is a bit complex and app developers almost always get it wrong, so it'd be great if the default experience could be better. But I'm unfortunately not familiar with Angular internals, so I'm hitting a bit of a wall. I've built accessible modals outside of Angular, so let me spec out what is required and see what improvements might be made.

You're already setting role="dialog" which is a good start, but there's a lot more that can be done. Here is a good primer:

https://www.marcozehe.de/2015/02/05/advanced-aria-tip-2-accessible-modal-dialogs/

TLDR:

1. Modals need an aria-labelledby and aria-describedby set to IDs of their titles and their descriptions, respectively. It's almost always safe to set aria-labelledby to the title heading or div, but perhaps an optional directive could be added for cases when this algorithm is wrong.

2. aria-describedby is set on the dialog's text, if applicable. This is harder to get right algorithmically, so perhaps a directive would do the trick.

3. Focus needs to be set within the dialog to its default control. Here another directive would be useful, but another algorithmic fallback would be assigning focus to the first focusable dialog content.

4. Not in the article, but a gotcha that stumped me for hours: all other document content elements need aria-hidden=true, but the dialog parent needs aria-hidden=false. I did this by surrounding all my top-level, non-dialog page content in a separate div, then concatenating the dialogs outside that. I.e.

<body>
  <div id="chrome">
    ...page content
  </div>
<div class="alertDialog" role="dialog">
    ...
  </div>

When alertDialog is shown, chrome gets aria-hidden=true and focus is redirected to the dialog. When it vanishes, chrome gets aria-hidden=false. There seems to be a backdrop element added. If that encompasses all non-modal content, it might be enough to set that to aria-hidden.

This all involves populating dialog titles and descriptions with IDs, or perhaps referring to existing IDs if they're already set. Unfortunately I don't know enough about Angular internals to pull this off.

It'd be great if the out-of-the-box experience for your dialogs can be improved. I was just looking at another library that uses yours, so the trickle-down of this change would be pretty impressive. Happy to help in an advisory/testing capacity if needed.

Thanks!

RicardoRFaria:
Hi guys, Im implementing the feature to ensure that "tab" key only will change focus inside the modal, bootstrap makes this, but I found one problem.
[...]

PR #6166, opened 2016-08-16
Improve ARIA support for modals.

NickHeiner:
Add ARIA properties and tests.

I figured out what to add by looking at the following:

1. https://www.w3.org/TR/wai-aria/roles#dialog
2. https://github.com/angular-ui/bootstrap/issues/3474
3. https://www.marcozehe.de/2015/02/05/advanced-aria-tip-2-accessible-modal-dialogs/

## Demo steps
1. Use ChromeVox
2. Go to the demo page and click "Open Me":

demo page screeshot

**Before:** Entire modal contents is read, including the buttons, with no distinction between label and content.
**After:** Reader says "I'm a modal! dialog" and then reads the description, but does not read the buttons. I'm not an accessibility expert, but given that the `aria-describedby` rules state that it should refer to the text describing the modal contents and not the form elements, I think this is an improvement.

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -504,6 +504,8 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
           'template-url': modal.windowTemplateUrl,
           'window-top-class': modal.windowTopClass,
           'role': 'dialog',
+          'aria-labelledby': modal.ariaLabelledBy,
+          'aria-describedby': modal.ariaDescribedBy,
           'size': modal.size,
           'index': topModalIndex,
           'animate': 'animate',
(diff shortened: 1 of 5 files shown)


src/modal/modal.js
```diff
@@ -495,7 +495,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
             }
           }
         }
-
+
```
wesleycho: Remove indentation

wesleycho:
This change looks good - just add a test to check for the presence of the attributes when passed in and this should be good to go.

NickHeiner:
Tests added. Thanks for the quick response @wesleycho!

wesleycho:
Merged as 4a5e6a7

PR #6173, opened 2016-08-18
fix(datepickerPopup): apply timezone conversion

### LINKED ISSUE #6147: uib-datepicker-popup : *Today* button does not count timezone
To reproduce : 

1. Set local timezone to +8:00
2. Set timezone option of uib-datepicker-popup to be '-4:00'
3. Set this date to ng-model option of datepicker-popup 'Mon Aug 08 2016 12:00:00 GMT+0800 (China Standard Time)'
4. Datepicker shows Aug 07 as expected, and selects todays date as Aug 07 too in the popup. But when you click on *Today* button, it selects Aug 08 instead of Aug 07.

wesleycho:
- Apply timezone conversion to selecting today

This should fix #6147.


icfantv:
LGTM

Issue #6174, opened 2016-08-18
Typeahead does not dismiss no results dialog in dismissClickHandler


jgdovin:
Bug description: When there are no results the dialog is not able to be dismissed when clicking away

Link to minimally-working plunker that reproduces the issue: https://plnkr.co/edit/g8PPmMlTKGnlAF99qEXy?p=preview (make asynch show no results and try to click elsewhere on the page.

Angular: 1.5

UIBS: 2.0.2

Bootstrap: 3

icfantv:
I'm not seeing the issue here, can you please provide explicit instructions for how to reproduce (generally, a good idea)?

wesleycho:
Going to close this as this does not appear to be an issue - the no results section being shown is not a part of the typeahead in the Plunker.

This was a bit of a lazy issue - in the future, please make sure the Plunker is minimal and that it is a clear reproduction, as we are not mind readers.

jgdovin:
A new plunkr: https://plnkr.co/edit/g8PPmMlTKGnlAF99qEXy?p=preview
In the box marked "Locations loaded via $http"
please type in the word test, you will see a result, then click in an empty space on the page, notice how the pop up results dismiss themselves.
Now remove the word test from the input box and put the letters "zzzzzzzzzzzzzzzzz" notice the no results shows up? Now click in an empty space on the page and you will see the no results never disappears, this behavior should be standard between both pop ups.

Calling me lazy/lazy issue was a bit rude, simply stating as the previous person id that you would like me to expand upon the issue and give a more precise plunker would have been a way to properly state your dislike of the issue. Sometimes people are new to reporting issues, sometimes people are in a rush and can explain better the next day.

wesleycho:
It shouldn't be opened then if it isn't well-thought out. Issues add a lot of noise if not well-framed, and not doing due diligence drains on project resources that can be used with verified issues/feature requests. We are all volunteers, and throughout the project, as a team, we have had to wade through over 4000 issues - that alone is a massive time sink, some of it which could have been used on difficult feature requests.

This is why we added https://github.com/angular-ui/bootstrap/blob/master/CONTRIBUTING.md#you-think-youve-found-a-bug, which is linked to at the top of creating a new issue.

Will take a look later.

icfantv:
[...]

PR #6175, opened 2016-08-18
fix for issue #6174

jgdovin:
This is a partial fix, the no results variable will get reset on the second click, if anyone could help figure out the digest cycle that would be awesome!


jgdovin:
#6174

icfantv:
Awaiting info in original issue.  Please do not merge yet.

jgdovin:
#6178

Issue #6108, opened 2016-07-21
Tooltip/Popover close on escape broken


stuartforster:
There is an issue with Chrome and potentially other browsers with capturing 'escape' for a 'keypress' event. 

This is preventing the ui-bootstrap tooltip (and popovers) from closing on press of escape key.

This should be changed to keydown or keyup here: https://github.com/angular-ui/bootstrap/blob/master/src/tooltip/tooltip.js#L74

I will submit a PR with a fix shortly.

icfantv:
@stuartforster, i can confirm that the event is not fired when `keypress` is used.  i will also confirm that changing to `keyup` or `keydown` does not cause the tooltip/popover to be closed.

what trigger do you expect this to work with?  also, please note that any code change will require tests.

stuartforster:
Thanks @icfantv, trigger could be keyup or keydown, I think keyup probably makes most sense.

I'll investigate why the popover still doesn't close even when the event is firing.

I will include tests with any PR I submit.

icfantv:
@stuartforster, after a thorough investigation, it appears the tooltip reference inside the close handling sequence is not defined meaning when the actual close route is executed, nothing happens and the tooltip stays open.

[Here's a plunker](http://plnkr.co/edit/WVTd5prw08Mj7tF6vEI4?p=info) to get you started.  I've already set the event to `keyup`.  Set a breakpoint at line 4724 and trace through and you'll see that ultimately the reference to `tooltip` is not defined which results in 4988 not being executed.

PR #6177, opened 2016-08-18
fix(tooltip): fix tooltip and popover not closing on escape #6108

stuartforster:
This PR fixes issue https://github.com/angular-ui/bootstrap/issues/6108

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -71,10 +71,10 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
    */
   this.$get = ['$window', '$compile', '$timeout', '$document', '$uibPosition', '$interpolate', '$rootScope', '$parse', '$$stackedMap', function($window, $compile, $timeout, $document, $position, $interpolate, $rootScope, $parse, $$stackedMap) {
     var openedTooltips = $$stackedMap.createNew();
-    $document.on('keypress', keypressListener);
+    $document.on('keyup', keypressListener);
 
     $rootScope.$on('$destroy', function() {
-      $document.off('keypress', keypressListener);
+      $document.off('keyup', keypressListener);
     });
 
     function keypressListener(e) {
@@ -82,7 +82,6 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
         var last = openedTooltips.top();
         if (last) {
           last.value.close();
-          openedTooltips.removeTop();
           last = null;
         }
       }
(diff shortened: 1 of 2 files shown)


src/tooltip/tooltip.js
```diff
@@ -71,18 +71,17 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
    */
   this.$get = ['$window', '$compile', '$timeout', '$document', '$uibPosition', '$interpolate', '$rootScope', '$parse', '$$stackedMap', function($window, $compile, $timeout, $document, $position, $interpolate, $rootScope, $parse, $$stackedMap) {
     var openedTooltips = $$stackedMap.createNew();
-    $document.on('keypress', keypressListener);
+    $document.on('keyup', keypressListener);
```
wesleycho: Is this change necessary? Seems like it could still be kept on keypress and work fine.

wesleycho:
The crux of this change feels correct to me with the movement of where the new tooltip is added, but I'm not certain that the keypress listener needs to be changed to a keyup listener - if those changes are removed, I'd be inclined to merge it.

Many thanks for this PR! I have investigated this a few different times, and felt that there was something wrong with when the tooltip was added, but could not place my finger on exactly what was happening.

src/tooltip/tooltip.js
stuartforster: I've just run a quick test in chrome and keypress when pressing esc is not being captured, whereas keyup is.

If you goto http://plnkr.co/edit/WVTd5prw08Mj7tF6vEI4?p=info and place a breakpoint on uibs.js:4720 in chrome then open a popup and press escape. For me - that method is never fired when using keypress.

Comment here https://github.com/angular-ui/bootstrap/issues/6108#issuecomment-234306945

The below also comes to mind:
https://github.com/angular/angular.js/issues/10905
http://stackoverflow.com/questions/3901063/jquerys-keypress-doesnt-work-for-some-keys-in-chrome-how-to-work-around

wesleycho: That's really weird - is this only with jQuery loaded? I haven't seen this problem when it is just angular with no jQuery.

stuartforster: Not sure. Another issue with keypress was the problem and replacing with keydown solves: https://github.com/angular-ui/bootstrap/issues/5013

Have you tried with the plunker?

wesleycho: Yeah, your Plunker does seem to indicate it being an issue. I guess I'm ok with this change then.

wesleycho:
This LGTM - thanks a lot!

Issue #5683, opened 2016-03-23
`$modal.open` compatibility with component (`angular.component` method)


zdychacek:
Hello,

I want to ask if you're considering support of `angular.component` method which is new API for creating components.

I'd like to have the ability to call `$modal.open` method with the component definition. Now I have to specify `controller` and `template` options...

**My API proposal:**

```
angular.module('myApp').component('myModal', {
  bindings: {
    greeting: '<'
  },
  template: '<div>{{$ctrl.greeting}}</div>',
  controller: function () {},
})

// open the modal
$modal.open({
  component: 'myModal',
  bindings: {
    greeting: 'hello',
  },
});
```

Thanks for response.

O.

icfantv:
I believe this would force us to only support Angular 1.5+ and i'm not sure we're ready to go there  yet.

zdychacek:
@icfantv Is there any issue with coexistence of both variants? Could't proposed API coexist with current one?

icfantv:
@zdychacek, i don't know.  this would require investigation if this is, in fact, something we want to support.  but we need to have the discussion first.

wesleycho:
I'm not against this, but we'll have to determine at what stage we want to implement this. We could implement it now, and have code where this only applies to angular 1.5+, but that adds bloat for those using 1.4.

Olgagr:
I'm trying to use existing modal with component this way:

```
modalInstance = this.$uibModal.open({
      template: '<my-component message="{{$ctrl.message}}"></my-component>',
      size: 'sm',
      controllerAs: '$ctrl',
      controller: () => {
        this.message = 'some message';
    });
```

Then in my component:

```
let myComponent = {
  bindings: {
    message: '@'
  }
};
```

However, it doesn't work :( this.message in component controller is empty. I'm not sure why.

icfantv:
@Olgagr there should be nothing wrong with having an arrow function if your transpiler is working correctly.  Arrow functions are simply rewritten as you have defined above.  I would encourage you to look at your transpiled code with the arrow function and see what it's generating.

Olgagr:
@icfantv I think in this case it makes difference because $uibModal.open is run in controller class of other component. To show wider context:

```
class PublicationController {

    openModal() {
        modalInstance = this.$uibModal.open({
         template: '<my-component message="{{$ctrl.message}}"></my-component>',
         size: 'sm',
         controllerAs: '$ctrl',
         controller: function() {
            this.message = 'some message';
[...]

PR #6179, opened 2016-08-19
feat(modal): add component support

wesleycho:
- Add support for `component` option

This should implement #5683.


src/modal/docs/readme.md
```diff
@@ -35,15 +35,29 @@ The `$uibModal` service has only one method: `open(options)`.
   _(Type: `boolean`, Default: `false`)_ -
   When used with `controllerAs` & set to `true`, it will bind the $scope properties onto the controller.
 
+* `component`
+  _(Type: `string`, Example: `myComponent`)_ -
+  A string reference to the component to be rendered that is registered with Angular's compiler. If using a directive, the directive must have `restrict: 'E'` and a template or templateUrl set.
```
icfantv: I think this implies that this is a BC, no?

wesleycho: This is a completely new option - it isn't supported currently, so it isn't a BC at all.

src/modal/test/modal.spec.js
```diff
@@ -131,6 +131,16 @@ describe('$uibModal', function() {
           elem.focus();
         }
       };
+    }).component('fooBar', {
+      bindings: {
+        resolve: '<',
```
icfantv: do we need to mention in the docs that this and the modal instance are one-way bound?  realizing that it doesn't apply to object contents...

icfantv: gah, nm.  i just realized this was a test...

icfantv:
LGTM

PR #6180, opened 2016-08-19
fix(collapse): set overflow to hidden on transition

### LINKED ISSUE #5474: collapse plugin renders a rising scrollbar on bootstrap navbar
Plunker: https://plnkr.co/edit/2H4yZfl6zSM0WmVzaba8?p=preview

Make sure the right panel is wide enough so that the navbar does not collapse and you will see a horizontal scrollbar (on initial load / refresh) that floats upwards starting from the bottom of the navbar.

The scrollbar does not appear if the navbar content is added to the file directly instead of using ng-include. 

Removing the navbar-right content makes the scrollbar appear to go away but an element is still floating upward which can be detected as it passes over the 'Left' text.

Found this while upgrading from 0.13.4. It was introduced in 0.14.0.

(Edit: changed plunker URL to go straight to preview)
(Edit: mention that this comment is related to ng-include)

wesleycho:
- When collapse is transitioning, set `overflow: hidden` to avoid possibility of scrollbars changing size

This should fix #5474.


icfantv:
LGTM

PR #6182, opened 2016-08-20
fix(collapse): default to css

### LINKED ISSUE #6045: feat(collapse): expand to style-defined height
If you define a fixed height on a particular panel via the CSS `height` attribute in a class or id-based style, but then also want to make it collapsible by adding `uib-collapse` to it, then it expands to the content height, not the style height.  (In some prior versions it would initially start at the style height, then collapse as expected, but then re-expand to the content height.)

This is different behaviour from the Bootstrap collapse.

It's possible to make the angular-ui version behave as expected by replacing the target heights of the expansion with the empty string rather than `auto` or a specific value -- this makes it expand to the inherited height from the style as expected.

[Example plunker](http://embed.plnkr.co/jufKlBwhjy5p4Yte4Tev/).

Ideally the directive should just automatically behave like the bottom one -- this is consistent with Bootstrap, where it will automatically use a fixed height if one is provided or an auto height otherwise, as you'd expect.  (Try it by removing the `fixed-height` style or its `height` attribute.)  If there's some reason why an explicit `auto` is required (it doesn't seem to be in desktop Chrome at least) then perhaps there should be an option to select between the two behaviours.

The bottom one was based on the angular-ui code as of 1.3.3, changing only four lines (other than the directive name).

wesleycho:
- Default to value set from CSS for height/width

This implementation was taken from Bootstrap, as seen [here](https://github.com/twbs/bootstrap/blob/df08d95495cf3cd9278099dc91c630423eccc846/js/collapse.js#L82).

Fix in action [here](https://plnkr.co/edit/NGAe1e9pOcFgTYg1ENR8?p=preview).

This fixes #6045.


wesleycho:
Looks like this causes a regression, as seen [here](https://plnkr.co/edit/85t9Mwz39aV1G4MuLvkF?p=preview).

I need to research this a little more.

icfantv:
LGTM

Issue #6186, opened 2016-08-20
Modal - component modal only works with jQuery


navarroaxel:
### Bug description:
The $uibModal service fails when I use a component modal without jQuery. Using this call to the open method:
````javascript
$uibModal.open({component: 'myComponent'})
````
the rendered HTML, without jQuery, is
````HTML
<div class="modal-content" uib-modal-transclude=""
  <span class="ng-scope">[[object HTMLElement]]</span
</div>
````
with jQuery works perfectly.
### Link to minimally-working plunker that reproduces the issue:
https://plnkr.co/edit/C61yaolPJoKH0WCnWQMr?p=preview

### Version of Angular, UIBS, and Bootstrap

Angular: 1.5.8

UIBS: 2.1.0

Bootstrap: 3.3.7

jQuery: 3.0.0+

PR #6187, opened 2016-08-20
fix open component modal jqlite compatibility

navarroaxel:
fix issue #6186.
Works fine with legacy controller modals and component modals.


navarroaxel:
in jQuery the the .html() method calls .empty().append() internally.
in jqlite the .html() set the HTMLElement into the innerHTML attribute of the parent element, so the .toString() method is called and displayed into the rendered HTML.

The solution for this was call the .append() method directly and the .empty() is not necessary because the parent is new and empty.

wesleycho:
Good catch! Thanks for the PR!

Issue #6191, opened 2016-08-21
Modal - component modal bindings close and dismiss are not working


navarroaxel:
### Bug description:
The binding for close and dismiss methods are incorrect and it's not working:
````javascript
angular.module('ui.bootstrap.demo').component('modalComponent', {
  bindings: {
    resolve: '<',
    close: '&',
    dismiss: '&'
  },
  controller: function () {
    this.close({$value: 'foo'});
  }
}
````

when I call the method close the modal is not closing and the ``result`` promise is not resolved:
````javascript
$uibModal.open({component: 'modalComponent'}).result.then(function(value) {
    console.log(value);
})
````

### Link to minimally-working plunker that reproduces the issue:
https://plnkr.co/edit/Z2KXdH?p=preview
### Version of Angular, UIBS, and Bootstrap

Angular: 1.5.8

UIBS: 2.1.1

Bootstrap: 3.3.7

PR #6192, opened 2016-08-21
fix(modal): close and dismiss bindings on a component modal

navarroaxel:
fix issue #6191 

The modalScope has the methods $close and $dismiss.

/src/modal/modal.js line 736
````javascript
modalScope.$close = modalInstance.close;
modalScope.$dismiss = modalInstance.dismiss;
````


navarroaxel:
Also I have found an issue with this this expect:
````javascript
expect(modal.result).toBeResolvedWith('baz');
````

The matcher ``toBeResolvedWith`` only fails when the promise is called with a wrong value.

**The matcher ``toBeResolvedWith`` don't fail when the promise is not resolved!**

wesleycho:
Oof, those were some careless mistakes by me...

I see the issue with the matchers - I will fix those shortly. Thanks for these catches!

PR #6193, opened 2016-08-21
docs(modal): added a simple component modal demo

navarroaxel:
I added a button for show a modal using components. It's using the same HTML of the old modal demos.

Can we remove the $scope usage in these demos and using the $ctrl in both? Component and controller modals.


wesleycho:
I would be fine removing $scope usage wherever possible - are you volunteering to do that cleanup? :)

navarroaxel:
Sure, I can do that.

navarroaxel:
Done, removed $scope from demo code

PR #6195, opened 2016-08-21
test(modal): fix custom matchers

wesleycho:
- Trigger a test failure if resolved on rejected promise and rejected on resolved promise
- Trigger a test failure if neither resolved or rejected when expected resolved/rejected promise


src/modal/test/modal.spec.js
```diff
@@ -183,7 +185,7 @@ describe('$uibModal', function() {
             var result = {};
 
             promise.then(function() {
-
+              expect(true).toBe(false);
```
pkozlowski-opensource: Don't we have sth like `fail()` in Jasmine?

wesleycho: Did not know that - will change

icfantv:
LGTM

PR #6196, opened 2016-08-22
fix(collapse): revert change to transition css

### LINKED ISSUE #6194: Accordion animation stopped working

### Bug description:

just upgraded from 0.14.3 to  2.1.1.

on 0.14.3 when  i clicked an accordion group it would slide nicely down. now it is just hang for a second, and instantly opens (without animation). this happens also on the demo  page - https://angular-ui.github.io/bootstrap/#/accordion

just putting the demo plunker... 

https://plnkr.co/edit/exfn8D6swHm1zJY63Ier?p=preview

Angular:1.5.8

UIBS:2.1.1

Bootstrap:3.3.7

wesleycho:
- Revert change to `getScrollFromElement` function

Incorrectly changed one part prior, this reverses it to completely agree with [Bootstrap's implementation](https://github.com/twbs/bootstrap/blob/df08d95495cf3cd9278099dc91c630423eccc846/js/collapse.js#L90-L94).

This results in some jank for the situation in the situation in #6045, but it looks like Bootstrap also would have that same problem.

This fixes the accordion regression, with fix as seen [here](https://plnkr.co/edit/TcxJZ3wIFPyM0K4F8HUh?p=preview).

This should fix #6194


icfantv:
LGTM

Issue #6221, opened 2016-09-05
TypeError: Cannot read property 'offsetHeight' of null


rarkins:
### Bug description:

I am seeing regular occurrences of this exception in browsers. The sourcemapped stack trace appears to point to code inside the distributed angular-ui bootstrap module:

```
TypeError: Cannot read property 'offsetHeight' of null
1
File "node_modules/angular-ui-bootstrap/dist/ui-bootstrap-tpls.js" line 4886 col 59 in [anonymous]
var currentHeight = angular.isDefined(tooltip.offsetHeight) ? tooltip.offsetH...
2
File "node_modules/angular/angular.js" line 19612 col 31 in [anonymous]
deferred.resolve(fn.apply(null, args));
3
File "node_modules/angular/angular.js" line 5964 col 10 in o
fn.apply(null, sliceArgs(arguments, 1));
4
File "node_modules/angular/angular.js" line 6243 col 7 in [anonymous]
completeOutstandingRequest(fn);
```

It would appear that `tooltip` in the above is null, which corresponds to: https://github.com/angular-ui/bootstrap/blob/f5ff12c0a4d8fd71791feeb6602d9da7dba41594/src/tooltip/tooltip.js#L164

Of course, checking for angular being non null is the easy solution but that may be a bandaid to the real problem rather than the cure.

### Link to minimally-working plunker that reproduces the issue:

Sorry, I am yet to figure out exactly what causes this, other than a suspicion that it's rapidly hovering tooltips.

### Version of Angular, UIBS, and Bootstrap

Angular: 1.5.8

UIBS: 2.1.3

Bootstrap: 3.3.7

karensg:
Same problem here

Johnnyrook777:
I was able to replicate it using a menu that overlaps a button with a tool-tip, when you click the menu and change navigation (i.e. the button with tool-tip is no longer there), you get the error. Seems the button is briefly available and triggers the tool-tip show as the menu disappears, but then the item is removed, and the timeout isn't cancelled

tkenakka:
Same problem here. Reproducing randomly, but usually related to a tooltip in a button which a) changes the width of a div i.e. visible area b) triggers a state transition into a new state.

Rouche:
Same error here. Random.
Happen when i have a popover opened, then i resize to small so the UI goes in mobile mode. The popover is in an ```ng-if="ctrl.desktop"```, so i guess the ```$timeout``` gets executed on something no longer in DOM.

[...]

PR #6226, opened 2016-09-07
Fixed: TypeError: Cannot read property 'offsetHeight' of null

Johnnyrook777:
If the tooltip is hidden before the timeout resolves, an exception is created. #6221 

Fix is to cancel the timeout when the tooltip is hidden

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -144,6 +144,7 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
             var showTimeout;
             var hideTimeout;
             var positionTimeout;
+            var adjustmentTimeout;
             var appendToBody = angular.isDefined(options.appendToBody) ? options.appendToBody : false;
             var triggers = getTriggers(undefined);
             var hasEnableExp = angular.isDefined(attrs[prefix + 'Enable']);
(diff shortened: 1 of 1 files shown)


tkenakka:
Hi, any idea when this will be merged and distributed via a new bower package?
We're reproducing this problem randomly, causing problems for us.

rarkins:
Same here, but we use `npm`

Rouche:
@Johnnyrook777 : Hello, did you try your fix against this Plunk : http://embed.plnkr.co/P6bLrajSHzkTdftcMHTw/

Just to be sure there is no race conditions deeper with $destroy.

src/tooltip/tooltip.js
```diff
@@ -315,6 +317,11 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
                 $timeout.cancel(transitionTimeout);
                 transitionTimeout = null;
               }
+              
+              if (positionTimeout) {
```
wesleycho: Don't we want to keep it if the hide is canceled?

Johnnyrook777: This was something I added thinking there was may be a left over timeout and potential memory leak
However, it isn't actually part of the issue, and was more of a memory tweak. 

Although it didn't cause issues in my scenario, it probably requires more testing

Johnnyrook777: Actually. Thought about this more on the way home. I suspect this condition could never occur, and there are no reports of it happening. Don't change what isn't broke

I'll update this tomorrow and remove this

wesleycho:
Made the change locally, merging shortly.

PR #6227, opened 2016-09-07
feat(accordion): support events from collapse in accordion

### LINKED ISSUE #5776: Callbacks doesn't work for accordion
I'm trying to attach callbacks with accordion. As per the documentation it extends from the collapse component which seems to support callbacks such as `expanding`,  `collapsing` etc.

 So I tried adding the following:

````html
<uib-accordion-group heading="first" collapsing="alert('collapsing')">
      first
</uib-accordion-group>
````

[plnkr demo](http://plnkr.co/edit/DusNhWbTb0v56UcAUxr9?p=preview)

Which I expect to work since accordion is an expansion of collapse. Is this a bug or am I doing something wrong..?

Angular: 1.5.3

UIBS: 1.3.1

Bootstrap: 3.3.6

theurere:
- Enable usage of collapse events of collapsed, collapsing, expanded, and expanding in accordion

Closes #5776

### DIFF
--- a/src/accordion/accordion.js
+++ b/src/accordion/accordion.js
@@ -66,7 +66,11 @@ angular.module('ui.bootstrap.accordion', ['ui.bootstrap.collapse', 'ui.bootstrap
       heading: '@',               // Interpolate the heading attribute onto this scope
       panelClass: '@?',           // Ditto with panelClass
       isOpen: '=?',
-      isDisabled: '=?'
+      isDisabled: '=?',
+      collapsed: '&?',
+      collapsing: '&?',
+      expanded: '&?',
+      expanding: '&?'
     },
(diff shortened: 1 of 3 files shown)


wesleycho:
This technically is a breaking change, so I will slot this for 3.0.0

wesleycho:
This needs documentation updates as well. Does this behave well if the attributes are not specified?

theurere:
Added documentation. When the attribute is not there it behaves fine.

bys92:
@wesleycho why is this change a breaking change ? 
There is only some attributes added. When they not set (what they can't be currently) they will be ignored. 

So this change is fully compatible.


We also need the same callback function to make our ui working correctly.

PR #6231, opened 2016-09-08
added documentation for popover-class

ravi-123:
added documentation for popover-class, kindly review and accept changes.


wesleycho:
Thanks, landed as 7671bd6

PR #6247, opened 2016-09-19
fix(datepicker): add aria labels to arrow buttons

### LINKED ISSUE #6229: ADA violation - buttons without discernible text
Bug description:

In the Datepicker, the arrow buttons signify going forward or backward by a day, month, or year, but to a screen reader this can not be discerned since there is no inner text or label on these button elements that describe said buttons' functions.

Additionally, there exists a `<ul>` element with `ng-click=$event.stopPropagation()`, and ngAria by default will add a `role="button"` to every element with an `ng-click` attribute.  However, this `<ul>` isn't functioning as a button.

Reproduce the issue:

Use this FireFox plugin to test accessibility:
https://addons.mozilla.org/en-us/firefox/addon/axe-devtools/

Make sure the datepicker is visible when you analyze the page.

Solution:

Adding labels to these button elements (e.g., `aria-label="view next month"` or `aria-labelledby=...`) should resolve these issues.

Version of Angular, UIBS, and Bootstrap:

Angular: 1.5.2
UIBS: 2.0.2
Bootstrap: 3.3.1

bifodus:
Fixes #6229

### DIFF
--- a/template/datepicker/day.html
+++ b/template/datepicker/day.html
@@ -1,9 +1,9 @@
 <table role="grid" aria-labelledby="{{::uniqueId}}-title" aria-activedescendant="{{activeDateId}}">
   <thead>
     <tr>
-      <th><button type="button" class="btn btn-default btn-sm pull-left uib-left" ng-click="move(-1)" tabindex="-1"><i class="glyphicon glyphicon-chevron-left"></i></button></th>
+      <th><button type="button" class="btn btn-default btn-sm pull-left uib-left" ng-click="move(-1)" tabindex="-1"><i aria-hidden="true" class="glyphicon glyphicon-chevron-left"></i><span class="sr-only">previous</span></button></th>
       <th colspan="{{::5 + showWeeks}}"><button id="{{::uniqueId}}-title" role="heading" aria-live="assertive" aria-atomic="true" type="button" class="btn btn-default btn-sm uib-title" ng-click="toggleMode()" ng-disabled="datepickerMode === maxMode" tabindex="-1"><strong>{{title}}</strong></button></th>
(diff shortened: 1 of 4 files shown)


amitse:
Shouldn't the text be localized?

template/datepicker/day.html
```diff
@@ -1,9 +1,9 @@
 <table role="grid" aria-labelledby="{{::uniqueId}}-title" aria-activedescendant="{{activeDateId}}">
   <thead>
     <tr>
-      <th><button type="button" class="btn btn-default btn-sm pull-left uib-left" ng-click="move(-1)" tabindex="-1"><i class="glyphicon glyphicon-chevron-left"></i></button></th>
+      <th><button type="button" aria-label="Previous Month" class="btn btn-default btn-sm pull-left uib-left" ng-click="move(-1)" tabindex="-1"><i class="glyphicon glyphicon-chevron-left"></i></button></th>
```
amitse: Shouldn't this be localized?

bifodus: Good point.  It probably should.  I'll see if I can update this to be language agnostic (perhaps using aria-labelledby) so that we don't need to worry about localizing strings.

bifodus: After looking into it a bit more, I followed the pattern currently being used for the carousel's forward and back buttons, and used `aria-hidden=true` on the glyphicon, along with a `<span class="sr-only">screen-reader text</span>`.  It doesn't fix the issue with localization (which the carousel also suffers from), but widespread localization of strings is probably big enough for a separate issue.  Thoughts?

wesleycho:
Going to merge this, but leave the issue open as something in need of a solution still.

PR #6259, opened 2016-09-23
fix(timepicker): properly update the validation on the input models

### LINKED ISSUE #6230: Timepicker - Validation issue
### Bug description:

When a bad value is entered in the timepicker, the input displays as invalid (that is, the class `has-error` is applied to the parent `<td>`, with which it can be styled as an error), but the class `ng-invalid` isn't applied to the input element itself (it will remain as `ng-valid`).

There may be other problems with this, but in my use case this conflicts with ngAria.  ngAria will add the attribute `aria-invalid` to the element and set it to `true` or `false` based on the presence of the `ng-valid` or `ng-invalid` classes.

### Reproduce the issue:

1. View the Timepicker on the demo page
2. Enter the value 'z' into the Hours input in the Start Time
3. Inspect this input and notice the presence of `ng-valid`

### Version of Angular, UIBS, and Bootstrap:

Angular: 1.5.2
UIBS: 2.0.2
Bootstrap: 3.3.1

bifodus:
Fixes #6230

### DIFF
--- a/src/timepicker/test/timepicker.spec.js
+++ b/src/timepicker/test/timepicker.spec.js
@@ -1230,6 +1230,7 @@ describe('timepicker directive', function() {
       changeInputValueTo(el, 'pizza');
       expect($rootScope.time).toBe(null);
       expect(el.parent().hasClass('has-error')).toBe(true);
+      expect(el.hasClass('ng-invalid-hours'));
       expect(element.hasClass('ng-invalid-time')).toBe(true);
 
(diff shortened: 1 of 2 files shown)


bifodus:
Fixes #6230

wesleycho:
Thanks for this - looks like good work here. I'd ideally like these to be done via validators, but perhaps that can be done later when the whole validation lifecycle is fixed in the component.

Issue #6262, opened 2016-09-27
Pagination directive, IE11 issue with prev/next buttons


scolt:
### Bug description:
After click on disabled previous/next button application redirects user to home page.

**Steps to reproduce via plunker below**
1) Open home page in IE11 (Internet Explorer 11)
2) Observe that previous button is disabled
3) Click on previous button

**Actual result:** 
Home page is opened.

**Expected result:**
Nothing should be happen.

### Link to minimally-working plunker that reproduces the issue:
http://plnkr.co/edit/1QBjhagSpwAb5dvFtm6S?p=preview

### Version of Angular, UIBS, and Bootstrap

Angular: 1.4.7

UIBS: 2.1.4

Bootstrap: 3.3.3

roger2hk:
I am having the same issue and was about to create one.
The same issue happens in pager as well.
The fix is to remove those href in uib/template/pagination/pagination.html and uib/template/pager/pager.html.

I will now create a pull request.

roger2hk:
I just realize removing the empty `href` would break the CSS. My current workaround is to change the empty `href` to `href="/"`.

wesleycho:
I need to verify my hypothesis at work tomorrow on a Windows VM with IE11, but I suspect this is a usage problem - this demo is not quite a minimal reproduction indicating it is an issue specifically with this library as it is polluted with ngRoute, and I have strong suspicions that that is actually the cause of this issue.

Please slim down the reproduction to not contain ngRoute if possible - otherwise, this is highly likely not a valid library issue, and thus filed in the wrong repository.

roger2hk:
The same issue happens in ui-router as well. Hope my information helps.

roger2hk:
@scolt There is another workaround before this issue is fixed.

According to the issue raised in ui-router, the workaround is to add a customized CSS. However, this would break the `cursor: not-allowed;` impacting the user experience.
https://github.com/angular-ui/ui-router/issues/2957

```css
.pagination > .disabled > a {
  pointer-events: none;
}
```

wesleycho:
Created a minimal reproduction from the homepage [here](https://plnkr.co/edit/5wicVvf9bAMSZJqJIjde?p=preview) - the Plunker in the OP had other distractions/issues with it that would prevent proper fixing.

wesleycho:
Hm, I was able to reproduce this on the homepage itself, but looks like my reproduction doesn't illustrate this...there's something else going on here.

[...]

PR #6263, opened 2016-09-27
fix(pager & pagination): prevent page reload in IE when anchor tag is disabled

roger2hk:
This fix is to resolve issue #6262.


roger2hk:
@wesleycho according to your previous experience, would it be a problem to switch from `href` to `href="/"`?

wesleycho:
Yes, that would break pretty much everyone's applications.

roger2hk:
Could you please explain on what situation it would break the application?
It seems the `href` content has no effect on click.

Plunker that fixed the issue by using `href="/"`
http://plnkr.co/edit/6shfJl7yL5OcURpMSuo8?p=preview

Plunker showing pagination without ngRoute using `href="/"`
http://plnkr.co/edit/3PyB3cC6U2uSoQtImD2Q?p=preview

wesleycho:
This breaks accessibility completely, as it is a false route. It also has SEO ramifications, and is susceptible to routing issues.

Closing this as invalid, this approach is a breaking change and has large impact negative ramifications.

PR #6264, opened 2016-09-27
Dropdown: add “showOnHover” attribute 

Ohar:
It let turns on `uibDropdownToggle` opening by `hover` event


wesleycho:
This adds configuration to the library, which increases library bloat - it would be better if you make use of the `is-open` binding and create a custom directive to toggle the dropdown on mouseenter.

I think most users are not interested in this feature, so it would be extra code loaded that isn't used. As a result, I am closing this PR, especially given that the right extensibility point is present in the library already.

Issue #6269, opened 2016-09-29
Timepicker - Hitting minimum/maximum time disables input field


techmaster242:
When I set any time picker field to have either a minimum or maximum time, as soon as it hits that limit, the field gets disabled.  This then makes it impossible to tab into the field and type a new value into it.  It breaks functionality for people with "accessibility" needs.  There should be an option that can be passed into the field, to disable this functionality.

techmaster242:
Pull Request #6270 fixes this.

PR #6270, opened 2016-09-29
This should satisfy Issue #6269.

techmaster242:
This adds a parameter called preventDisable.

The parameter defaults to false, but if set true, should keep the timekeeper from disabling the input fields.  This will be helpful for anyone needing better keyboard support.

### DIFF
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -11,6 +11,7 @@ angular.module('ui.bootstrap.timepicker', [])
   mousewheel: true,
   arrowkeys: true,
   showSpinners: true,
+  preventDisable: false,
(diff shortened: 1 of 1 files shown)


wesleycho:
A better solution is to override the template - that way this does not add library configuration bloat for what seems to be a weird request (especially given that this feature is tied to accessibility if I remember correctly).

Closing as an invalid approach.

techmaster242:
The changes I submitted satisfy federal government requirements of Section 508 of the Rehabilitation Act of 1973.  This is not a weird request.

MGouaillier:
Hi !
Do we have any news on this feature ? I'm highly interested

techmaster242:
No, I was told it was a stupid feature and they have no interest in
accessibility features.

PR #6273, opened 2016-10-04
fix(position): correct scrollbar width calculations

JuliaUsanova:


JuliaUsanova:
[Here is the link to source with bug](https://github.com/angular-ui/bootstrap/blob/master/src/position/position.js#L133), scrollbarWidth requires only 1 parameter 'isBody', [scrollbarWidth method](https://github.com/angular-ui/bootstrap/blob/master/src/position/position.js#L89).

As a result the first argument in scrollbarWidth is object, so the [condition](https://github.com/angular-ui/bootstrap/blob/master/src/position/position.js#L90) always passes. 

@RobJacobs, please check the improvement, because current behavior causes other plugins which track content width change to fire unnecessary events and even break layout.

wesleycho:
This fix looks correct to me.

RobJacobs:
LGTM

Issue #6208, opened 2016-08-26
[dropdown] Not unbinding keydown handler properly when closing dropdown by opening another one


Agamnentzar:
Keydown handle for dropdowns is not being unbinded correctly when dropdown is closed by opening another dropdown.

http://plnkr.co/edit/wQ0j12hABSNz6pYEqRQL?p=preview

Steps to reproduce
1. Open plunker, open dev console
2. Click on first dropdown to open it
3. Click on second dropdown to open it (without closing the first one first)
4. Click on the background to close second dropdown
5. Start pressing keyboard buttons
6. dev console will show `Uncaught TypeError: Cannot read property 'getDropdownElement' of null` errors on each key press

Angular: 1.5.5
UIBS: 2.1.3
Bootstrap: 3.3.6

Probably introduced by 6bad7592b9d7838099d920c98cddc7cb13933c9c

almothafar:
+1
Angular: 1.5.8
Bootstrap: 3.3.7

leon3s:
Hello there ! 

I got the same error and that was my way to fix it.

The error pop on function keybindFilter

```javascript
  this.keybindFilter = function(evt) {
    var dropdownElement = openScope.getDropdownElement();
    var toggleElement = openScope.getToggleElement();
    var dropdownElementTargeted = dropdownElement && dropdownElement[0].contains(evt.target);
    var toggleElementTargeted = toggleElement && toggleElement[0].contains(evt.target);
    if (evt.which === 27) {
      evt.stopPropagation();
      openScope.focusToggleElement();
      closeDropdown();
    } else if (openScope.isKeynavEnabled() && [38, 40].indexOf(evt.which) !== -1
         && openScope.isOpen && (dropdownElementTargeted || toggleElementTargeted)) {
      evt.preventDefault();
      evt.stopPropagation();
      openScope.focusDropdownEntry(evt.which);
    }
  };
```

When we close dropDown this is called

```javascript
  this.close = function(dropdownScope, element) {
    if (openScope === dropdownScope) {
      openScope = null;
      $document.off('click', closeDropdown);
      $document.off('keydown', this.keybindFilter);
    }
  };
```

openScope is set to null before use it on this.keybindFilter WTF ???
I fixed it like that but it maybe not the best way

```javascript
  this.keybindFilter = function(evt) {
    if (!openScope) { return; }
    var dropdownElement = openScope.getDropdownElement();
    var toggleElement = openScope.getToggleElement();
[...]

PR #6278, opened 2016-10-07
fix(dropdowm): null check to to resolve issue \#6208

nmccready:


wesleycho:
LGTM - thanks for this

PR #6280, opened 2016-10-07
fix(dropdown): added null check to .focus & .contains

AnthonyCC:


wesleycho:
Going to go with #6278 as that looks like the better fix - thanks for the effort though!

PR #6281, opened 2016-10-10
Add standalone month support (LLLL) to dateparser

Jiia:
Standalone month format support was added to Angular in February. https://github.com/angular/angular.js/pull/14013


wesleycho:
This should add an update in the dateparser documentation for this option & that this is only supported in Angular 1.5.1+

Jiia:
Updated docs and tests

wesleycho:
LGTM - scheduling for 2.3.0

PR #6295, opened 2016-10-20
fix(modal): focus defaults to the modal itself, not the first element

ajpz:
I recently bumped version from 2.1.4 to 2.2.0. 

In the updated version, modals set default focus to the [firstFocusableElement](https://github.com/angular-ui/bootstrap/compare/2.1.4...2.2.0#diff-15d409302580c93c654a382e2da68396R649). This seems like unwanted default behavior (for example, input elements may have additional focus behavior that I'd like not to trigger by default). 

Also, modals [already](https://github.com/angular-ui/bootstrap/blob/master/src/modal/modal.js#L201) had a viable and acceptable default state -- focus would be applied to the modal itself when `autofocus` wasn't explicitly set on any element. 

This change is a quick example of the minimal changes required to revert to simpler, less opinionated, default behavior.

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -646,7 +646,6 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
 
       $modalStack.modalRendered = function(modalInstance) {
         var modalWindow = openedWindows.get(modalInstance);
-        $modalStack.focusFirstFocusableElement($modalStack.loadFocusElementList(modalWindow));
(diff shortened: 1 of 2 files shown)


Jaco-Pretorius:
👍

wesleycho:
Looking at the original commit that changed this, this looks like feature creep that should have been in a separate changeset.

This LGTM

flehannier:
I guy,

Your change can have an impact on the uib-typeahead because the focus does not work anymore ?

thx for your answer

Issue #6030, opened 2016-06-20
Dropdown: Append-to-body breaks dropup


shyamal890:
When `class="dropup"` and `dropdown-append-to-body` are used together. The dropdown no longer dropsup.

icfantv:
@shyamal890, please be respectful of our time and processes by adhering to the issue template.  As we are all volunteers, we simply do not have the time to chase every incomplete issue entered by users.

icfantv:
closing due to lack of response.

shlomil:
http://plnkr.co/edit/9Mc7WQ7qUbbAgQDlIpIv?p=preview

icfantv:
@shlomil thanks for the plunker.  i'm not seeing what the issue is.  can you please post steps to recreate?  thanks.

shlomil:
Sorry, I thought the plunkr was clear. There are 3 examples in the above plunkr. All use class "dropup" which means it should display the dropdown menu above the button. It works fine in normal use (3rd button) but if you try to use the dropdown-append-to-body directive, it will display the menu below the button, ignoring the dropup class.

icfantv:
@shlomil, ah, I understand now. Reopening.  In general, it's a good idea to explain exactly what the issue is and exact steps to show us the issue.  As we are all volunteers, we simply do not have the time to chase everything down.  Thanks.

wesleycho:
This is likely related to the $uibPosition service usage in the dropdown - it has no knowledge of the dropup class. We need to add logic to account for that, and position it appropriately in that situation.

moses-seeq:
Any word on when the fixes are going to be reviewed and, hopefully, accepted?

PR #6300, opened 2016-10-25
fix(dropdown): Adding support for dropUp class

bartkiewicz:
Fix for #6030. 

Was not sure if the comment in the original issue was suggesting the dropUp check be done in the uibPosition, but it made more sense to me to check in dropdown, since why would anything else care about a dropUp class.


hexaclock:
Any word on getting this merged in?

Issue #2678, opened 2014-09-09
Multiple Modals and Backdrop


Exagram:
If I open a modal within a modal, a new backdrop doesn't instantiate for the second modal. Does Angular UI support this use case? If not, is there an option to move the backdrop to the highest modal?

borislit:
+1
In my case, I have closed one modal (lets call it Modal A) and immedialtly opened another one (lets call it Modal B). Modal A has a backdropClass defined for it.
When modal B is opened, the backdropClass is not removed.

karianna:
plunkr to demonstrate please

borislit:
http://plnkr.co/edit/KCBVGVzirLlApozRrbjl

The flow is as follows: Once to open the first modal, clicking OK will close it and will open the second one right away.
Notice that the first modal add a "custom-backdrop" class which is not removed when the second modal opens.

rosskukulinski:
So we've run into this as well.  I'm still poking around, but it seems that there's a race condition from closing the first modal (and removing its backdrop) and setting the backdrop of the second modal.

Specifically, I'm seeing that backdropDomEl is non-null [right here](https://github.com/angular-ui/bootstrap/blob/158d267659097076a832672e282e54fd675fe692/src/modal/modal.js#L312), which means that the second modal does not get its backdrop applied.

PR #6303, opened 2016-10-27
fix(modal): backdrops for appendTo (modal in modal)

### LINKED ISSUE #6219: Backdrop bug when using multiple modals and appendTo
When multiple modals are created (at the same time) only a single backdrop will be created. My use case: Let's say we have a list of items, designed business cardish. I want to be able to delete those from my view by clicking a link. Than a confirmation modal appears with the backdrop only above that specific card. When opening a seconds modal dialog, while the first one is stil active no second backdrop will be created. Only the z-index will be updated.

https://run.plnkr.co/plunks/nwGf2MfgeQ51zom3w9aR/#

Shouldn't there a backdrop per modal panel be created instead?

Versions:
Angular: 1.4.8
UIBS: 2.1.3
Bootstrap: 3.3.5

malaupa:
Modals couldn't be used in modals (open a modal in a modal) and other containers. A new backdrop is now created on each appendTo element which is used in the app. If a backdrop exist on a element it is reused and the backdrop index is increase (old behavoir). To recognize which element has already a backdrop appended, a data-backtrop attribute is written on the related element. 

Closes #6219 and #2678 is also related

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -246,7 +246,8 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.stackedMap', 'ui.bootstrap.p
     function($animate, $animateCss, $document, $compile, $rootScope, $q, $$multiMap, $$stackedMap, $uibPosition) {
       var OPENED_MODAL_CLASS = 'modal-open';
 
-      var backdropDomEl, backdropScope;
+      var backdropCount = 0;
+      var backdrops = $$stackedMap.createNew();
       var openedWindows = $$stackedMap.createNew();
       var openedClasses = $$multiMap.createNew();
       var $modalStack = {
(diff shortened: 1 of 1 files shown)


wesleycho:
This needs to be configurable, as this would otherwise be a significant breaking change.

wesleycho:
This also needs unit tests

malaupa:
Could we use `backdropAppend` as the a new param for backdrop creation on each appendTo element? Do you prefer another parameter name? Values could be `true` or `false`?

I will create unit tests and commit them in combination with the changes related to the configuration param after your response.

wesleycho:
We could, but that should perhaps be a separate PR - multiple backdrops as siblings should be ok, as long as they are created in order. Appending the backdrop to the appropriate location then can follow as separate work.

malaupa:
Could you please tell me in detail, which work should be separated in different PRs?

At the moment it is only the following behaviour, which will be changed:
1. configuration option to append backdrops to the appropriate location (the appendTo element), if appendTo is configured
And this is the idea of this PR. If you agree with me, could this PR be merged, if I add tests and the configuration option?

wesleycho:
That would be sufficient if those changes are made plus documentation.

paneyi:
@malaupa, what's the status of this? I got to this answer while looking for a way to keep multiple modals opened on different "tabs" (achieved using ui-router-sticky-states and ui-router-future-states), each one with its own backdrop.
Right now, each modal is contained in the proper tab, but the backdrop remains below the first created modal, leaving the modals on the other tabs without a backdrop.

malaupa:
Currently I have no time to add tests and the configuration option, but this PR is still on my todo list and will be fixed.

For clarification: the idea of this PR is to create a modal inside a already opened modal with suitable backdrop limited to outer modal.

PR #6339, opened 2016-11-30
feat(position): add 'positionElementAt' method

farooqu:
Add a method to position an element at client x/y coordinates. Using
code from the positionElements method, which has been refactored into a
positionElement method with both positionElementAt and positionElements
use.


farooqu:
I'm creating a context menu directive, and was wanting to use the same positioning logic as all of the bootstrap elements we use.

wesleycho:
I'm not sure I see the value of this in the library itself - it doesn't appear like something anything in UI Bootstrap would use.

In addition, this appears to be wholly independent code - one could just create a wrapper around the UI Bootstrap $uibPosition service and add this method to the wrapper, or decorate it with $provide.decorator. There are lots of ways to place it in one's application that doesn't require shipping code that is unused for most users.

PR #6340, opened 2016-12-01
Fix Links for CHANGELOG.md

say25:
Fix Links for CHANGELOG.md


wesleycho:
LGTM - real fix needs to come from https://github.com/conventional-changelog/conventional-changelog-angular though.

PR #6366, opened 2016-12-13
(docs)

radotzki:
Add explanation about how to place the menu in the right side.


wesleycho:
This does not belong in the documentation - this is a Bootstrap CSS feature, and not api related to this library.

Closing as a result.

Issue #5967, opened 2016-06-06
Carousel prev/next slide incorrect after double clicking page control


Skawt2k:
### Bug description:

When double clicking the previous/next control or clicking the previous/next control again before the current page transition has completed causes the controls and looping behaviour of the carousel to be incorrect/inconsistent. 
### Steps to reproduce

1) Visit http://angular-ui.github.io/bootstrap/ . The active slide is 0.

2) Double click the carousel next control. The active slide is now 1 as expected.

3) Click the previous control. The active slide now becomes 2, the carousel should actually transition to the previous slide instead of the next slide. 

4) Move the pointer away from the carousel (out). The carousel will move to the next slide, this is not expected.

Note: This also happens after quick swipe successions with touch enabled. 
### Link that reproduces the issue:

This is an issue on the current angular-ui bootstrap demo site:
http://angular-ui.github.io/bootstrap/
### Version of Angular, UIBS, and Bootstrap

Angular: 1.5.3

UIBS: 1.3.3

Bootstrap: 3.6.6

icfantv:
@Skawt2k we're working on removing `replace: true` from our code to make `ng-class` directives work (as well as resolve other issues).  as such the code has changed.  Can you please try building against master and test this again?  Thanks.

wesleycho:
No need for that - this sounds like a potential issue with $animate not resolving correctly, it may be an Angular ngAnimate bug, but I need to investigate.

TechCraver1:
@wesleycho Add label needs: info.

icfantv:
@wesleycho, ok, sounds good.

stuartraines:
@wesleycho - Have you managed to investigate this issue at all? Is there a workaround that you know of?

tariqporter:
@stuartraines  Yes, for now you can go to line approx 541 and comment out:

```
else if (nextSlide && nextSlide.slide.index !== currentIndex && $scope.$currentTransition) {
    bufferedTransitions.push(slides[nextIndex]);
}
```

Skawt2k:
@tariqporter - Many thanks for the workaround. I can confirm that commenting out the lines specified above resolves the issue. Thx.

stuartraines:
@tariqporter - Thanks very much for taking the time to look into this and for the fix

kive323:
I have comment out that part, but the issue still happens, any other solution I can fix that.

kive323:
[...]

PR #6367, opened 2016-12-14
Remove Buffered Transitions array

tariqporter:
[#5967](https://github.com/angular-ui/bootstrap/issues/5967)

Creating a buffered transitions array to handle clicking too quickly is unnecessary.
The code

`if (nextSlide.slide.index !== currentIndex && !$scope.$currentTransition) {
    goNext(nextSlide.slide, nextIndex, direction);
}`

Already ensures that if a transition is occurring not to process another slide transition (As the vanilla bootstrap does)

Clicking several time quickly to the right, and then clicking one time left, will still shift it right:

[angular-ui carousel example](https://angular-ui.github.io/bootstrap/#/carousel)

whereas this issue does not exist in vanilla bootstrap

[w3schools carousel example](http://www.w3schools.com/bootstrap/bootstrap_carousel.asp)

Removing the buffered array fixes this.

### DIFF
--- a/src/carousel/carousel.js
+++ b/src/carousel/carousel.js
@@ -5,7 +5,7 @@ angular.module('ui.bootstrap.carousel', [])
     slides = self.slides = $scope.slides = [],
     SLIDE_DIRECTION = 'uib-slideDirection',
     currentIndex = $scope.active,
-    currentInterval, isPlaying, bufferedTransitions = [];
+    currentInterval, isPlaying;
 
   var destroyed = false;
   $element.addClass('carousel');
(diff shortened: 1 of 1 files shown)


tariqporter:
Anything anyone needs from me to help expedite this being merged? Currently myself and others I have spoken to have to make these changes each time they upgrade otherwise the carousel can't be deployed.

wesleycho:
I will merge this, but I may have to revert if it does cause regressions - in the past this was due to rapid clicks not playing well with ngAnimate, although it has been quite a while since I dealt in depth with this component.

PR #5934, opened 2016-05-28
feat(dropdown): focus toggle on close from click

wesleycho:
- Focus the toggle element when the click causes the dropdown menu to close

BREAKING CHANGE: This changes the focus behavior of the dropdown slightly, and potentially may break code built around current usage

Fix in action [here](http://plnkr.co/edit/PeGiQCoYIz3PLHS52DAn?p=preview).

This fixes #5782

Issue #5782, opened 2016-04-11
uib-dropdown breaks tab order


jesspam:
### Bug description:

If you open the list and make a selection the tab order is reset and hitting tab after the list closes moves the focus to the top of the page.  This can be seen in the demo page: https://angular-ui.github.io/bootstrap/  Opening and closing the list without clicking a list item does not interrupt the tab order.

### Link to minimally-working plunker that reproduces the issue: 
the behavior is seen on https://angular-ui.github.io/bootstrap/

### Version of Angular, UIBS, and Bootstrap 
Angular:1.5.3

UIBS:1.3.1

Bootstrap:3.3.6

icfantv:
@jesspam, i'm not seeing this, can you please provide exact steps to duplicate, including which dropdown on which you're trying this on?  also, it may be better to try and isolate this in a stand-alone plunker.  you can create one easily by clicking the applicable button below the demo.

jesspam:
I just verified that I see the issue on firefox and chrome.  To reproduce, go to the demo URL indicated:
click "Dropdown on Body" in the dropdown section, but don't choose any item from the list
click "Dropdown on Body" again to close the list
press Tab key to tab to the next few controls, this should work fine

now, click "Dropdown on Body" again, but this time choose an item
the list will close.  Now when you press the Tab key, you'll tab to the top of the page rather than the next control as before.

jesspam:
@icfantv  Was that helpful?

icfantv:
@jesspam please understand that we are unpaid volunteers on this OSS project and that we will get to the issues when we are able to allocate the appropriate amount of time.  continued pings are not necessary and considered rude.  please exercise some patience.

to answer your question, yes, that was helpful.  we will need to research what is involved here.  it may not be as simple as returning the focus to the dropdown when closed because the dropdown may no longer exist (e.g., if the dropdown selection results in a route change).

jesspam:
@icfantv I just wanted to be sure you didn't need any more from me.  Thanks for your attention and good luck!

wesleycho:
That's an interesting bug. On close, we should probably focus the dropdown element.

This could potentially have side-effects, but I think it's an ok change assuming no problems.

stumblingmages:
@jesspam Not sure if it will help or not but some of my problems with unwanted navigation got better when I changed the anchor tags in my drop down menu to `<a href="javascript:void(0)">` rather than `<a href="#">`.

PR #6369, opened 2016-12-14
Fix issue where focus is stolen from form elements after dropdown shown

mondayuk:
Discovered an issue where form elements were losing their focus. Discovered this was related to #5934 having added a feature to apply the focus to the toggle element, but as this event fires on all mouse clicks the focus was being set regardless of whether the dropdown was actually open at the time.

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -347,7 +347,7 @@ angular.module('ui.bootstrap.dropdown', ['ui.bootstrap.multiMap', 'ui.bootstrap.
 
       self.selectedOption = null;
       
-      if (wasOpen) { openScope.focusToggleElement(); }
+      if (wasOpen) { scope.focusToggleElement(); }
 
     }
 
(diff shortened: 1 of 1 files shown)


wesleycho:
I believe this was originally added for accessibility reasons - does this preserve the accessibility situation denoted in #5782?

mondayuk:
Yes, the functionality is preserved when the popup closes (which is the intention of the fix) but it resolves the issue that the focus continues to be reverted back to the toggle control even after the popup has been closed.

This issue can be seen on https://angular-ui.github.io/bootstrap/. Click to open the dropdown then close, then try and focus on something else on the page (for example, one of the text boxes next to the date pickers) focus is pulled back to the dropdown.

dougludlow:
FYI, @lukehutton added e07f9fd  to #6353 to fix the same issue.

mondayuk:
Looks like that was done after mine and (unless i'm mistaken) has no pull request yet. Functionally identical though.

mondayuk:
Oh actually, I see it is associated with a pull request. No preference on which makes it in.

wesleycho:
Going to close this as I merged #6382 instead as the proper fix - thanks for the attempt, the fix should be in the latest release.

Issue #6370, opened 2016-12-14
Feature Request: Ability to override dateParser formatCodeToRegex


dougludlow:
Hey, so I have an issue where I need to overwrite the way the dateParser parses 2 digit years. Currently that collection is private and I'd like to expose it and make it public. Would there be any interest in merging a PR for this request? I'm thinking it could be exposed as config that could override the original.

PR #6373, opened 2016-12-14
feat(dateparser): Allow overriding of parsers.

dougludlow:
Proof of concept for #6370.

### DIFF
--- a/src/dateparser/test/dateparser.spec.js
+++ b/src/dateparser/test/dateparser.spec.js
@@ -790,27 +790,34 @@ describe('date parser', function() {
         expect(toWestDate.getTime() - toEastDate.getTime()).toEqual(1000 * 60 * 60 * 12);
       });
     });
+  });
 
-    describe('overrideParser', function() {
-      it('overrides the parser', function() {
-        var original = dateParser.getParser('yy');
-        try {
-          expect(dateParser.parse('67', 'yy').getFullYear()).toEqual(2067);
-          expect(dateParser.parse('68', 'yy').getFullYear()).toEqual(2068);
+  describe('overrideParser', function() {
+    var twoDigitYearParser = function (value) {
+      this.year = +value + (+value > 30 ? 1900 : 2000);
+    };
 
-          dateParser.overrideParser('yy', function (value) {
-              this.year = +value + (+value > 30 ? 1900 : 2000);
-          });
+    it('overrides the parser', function() {
+      dateParser.overrideParser('yy', twoDigitYearParser);
+      expect(dateParser.parse('68', 'yy').getFullYear()).toEqual(1968);
+      expect(dateParser.parse('67', 'yy').getFullYear()).toEqual(1967);
+      expect(dateParser.parse('31', 'yy').getFullYear()).toEqual(1931);
+      expect(dateParser.parse('30', 'yy').getFullYear()).toEqual(2030);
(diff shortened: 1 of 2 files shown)


src/dateparser/test/dateparser.spec.js
```diff
@@ -790,5 +790,27 @@ describe('date parser', function() {
...
+          dateParser.overrideParser('yy', function (value) {
+              this.year = +value + (+value > 30 ? 1900 : 2000);
+          });
+
+          expect(dateParser.parse('68', 'yy').getFullYear()).toEqual(1968);
+          expect(dateParser.parse('67', 'yy').getFullYear()).toEqual(1967);
+          expect(dateParser.parse('31', 'yy').getFullYear()).toEqual(1931);
+          expect(dateParser.parse('30', 'yy').getFullYear()).toEqual(2030);
+        }
+        finally {
```
wesleycho: I don't think this is necessary to wrap in a `try-finally` - this factory gets reinstantiated with each test execution due to the whole beforeEach running of `module('ui.bootstrap.dateparser')`

src/dateparser/test/dateparser.spec.js
```diff
@@ -790,5 +790,27 @@ describe('date parser', function() {
         expect(toWestDate.getTime() - toEastDate.getTime()).toEqual(1000 * 60 * 60 * 12);
       });
     });
+
+    describe('overrideParser', function() {
+      it('overrides the parser', function() {
+        var original = dateParser.getParser('yy');
+        try {
+          expect(dateParser.parse('67', 'yy').getFullYear()).toEqual(2067);
+          expect(dateParser.parse('68', 'yy').getFullYear()).toEqual(2068);
```
wesleycho: These two lines probably aren't necessary, as they should already be covered by tests for the format

wesleycho:
This mostly LGTM, just needs a little cleanup.

src/dateparser/test/dateparser.spec.js
dougludlow: One reason I had these here was because the dateParser appears to cache a parser instance for a particular format once it's already been used.

dougludlow:
@wesleycho how's that? I split the original test out into a few to better articulate the intent.

wesleycho:
I think the test cases can be simplified - what is mainly of concern is whether you can grab the current parser (and thus the original one), and whether you can override the parser used. That way if a failure in the overriding is present, it is easily isolated by the tests.

dougludlow:
So basically remove the "clears cached parsers when overridden" and we're good as it is implied in "can set a parser back to the original"?

wesleycho:
The two specs should read
`it('should get the current parser'`
And
`it('should override the parser'`
Or something like that, and test those scenarios

dougludlow:
Ok, simplified. I still think it's important to have a test for making sure that the parsers are cleared when a parser is overridden. Basically guarding against this line (https://github.com/dougludlow/bootstrap/blob/b02815b446ad03df6eeb8a7ac054b13d9669f579/src/dateparser/dateparser.js#L258) being removed. If they aren't cleared the overridden parser will not be used.

dougludlow:
@wesleycho any other feedback here?

wesleycho:
LGTM, going to schedule for 2.4.0

Issue #6368, opened 2016-12-14
Bug in ui.bootstrap.modal


Zadvornyi:
Hello everyone, first of all I want to apologize for my English
I had project on Angular:1.5.7, UIBS:0.13.3, Bootstrap:3.3.6 , but after I had updated UIBS to 1.3.3. I had the bug with ui.bootstrap.modal. My code change default position for modal window. The bug is connected with modal window display, as after update the “callbacks $uibModal.open().rendered” function is fired after the template is displayed on the page. I also got blinking effect, which I did not have  with version 0.13.3.
before: 
![image](https://cloud.githubusercontent.com/assets/10829855/21181844/7f971e5e-c208-11e6-98c1-8a6857dac849.png)
- https://plnkr.co/edit/9FsLDyi0STRzlyXxRQF8?p=preview
- open console
- run code
after:
![image](https://cloud.githubusercontent.com/assets/10829855/21181645/8e688edc-c207-11e6-9f41-db06f2d00afa.png)
- https://plnkr.co/edit/l0tDbf3CGcxN1HWmbxQn?p=preview
- open console
- run code

#### Possible solutions
I investigated the code of both versions and I think I can offer possible solution. If I comment this row $q.when(animationPromise).then(function() { , it can resolve this problem

cc @wesleycho, @christopherthielen

icfantv:
Closing per comments in rejected PR.

Your upgrade is quite a big jump.  Did you work through the breaking changes sections in ALL the versions from your old one to your new one?

PR #6380, opened 2016-12-20
fix(modal): fix bug when a modal is rendered

Zadvornyi:
Issue #6368
The “callbacks $uibModal.open().rendered” function is fired after the template is displayed on the page. I also got blinking effect

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -119,7 +119,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.multiMap', 'ui.bootstrap.sta
           var animationPromise = null;
 
           if (attrs.modalInClass) {
-            animationPromise = $animateCss(element, {
+            $animateCss(element, {
               addClass: attrs.modalInClass
             }).start();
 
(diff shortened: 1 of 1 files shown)


src/modal/modal.js
```diff
@@ -130,36 +130,32 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.multiMap', 'ui.bootstrap.sta
               }).start().then(done);
             });
           }
+          // Notify {@link $modalStack} that modal is rendered.
+          var modal = $modalStack.getTop();
+          if (modal) {
+            $modalStack.modalRendered(modal.key);
```
wesleycho: This changes the business logic of `rendered`. `opened` is meant for what it appears `rendered` has been changed to in this PR. This is no good.

Zadvornyi: I want to change logic for rendered. I created Issue #6368. Please check it. I can't change position for modal window because DOM is not rendered while `opened` is firing.

Zadvornyi:
I want to change logic for rendered. I created Issue #6368. Please check it. I can't change position for modal window because DOM is not rendered while opened is firing.

wesleycho:
This is no good as it is, as it's a breaking change.

Closing as invalid.

PR #6384, opened 2016-12-21
iOSModal.css

nikleh:
This fixes the disappearing modal in iOS

### DIFF
--- a/iOSModal.css
+++ b/iOSModal.css
@@ -1,0 +1,37 @@
+.modal-dialog {
+  margin: 0px;
+  margin-top: 2%;
+  margin-left: 2%;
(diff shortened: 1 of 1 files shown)


nikleh:
Currently, the bootstrap modal disappears when iOS devices change from landscape to portrait or portrait to landscape display. The opaque background remains, but the modal disappears. This has to do with the margin/fixed width. 

These changes fix this issue.

wesleycho:
Going to close this as too much of a breaking change - this should be done on the user's side, as this is a level of styling that is guaranteed to break many people's apps. This is better fixed in Bootstrap itself, or on a user by user basis.

PR #6389, opened 2016-12-29
Fix tab index when using string index

wang0109:
I am trying to use string index and found this bug..


wang0109:
From my understanding, when the `index` property is not used, it generates integer indexes, otherwise, it is supposed to use `index` from user and insert into the list of tabs..

wang0109:
Tests are failing.. I have no idea..

PR #6393, opened 2017-01-01
Update LICENSE

Scottkao85:
Update year to 2017


wesleycho:
Thanks, seems like we have to do this every year :)

PR #6406, opened 2017-01-13
make matches available to the typeahead-on-select callback

gcziprusz:
Usage example:
```
typeahead-on-select="onSelect($item, $model, $label, $event, $matches)"
```


wesleycho:
Can you add documentation?

gcziprusz:
@wesleycho added docs for the new on-select $matches param

PR #6414, opened 2017-01-18
fix(dropdown): avoid focus toggle element after dropdown is closed

yanhao1991:
This PR fixed a bug: 

As you can see in this code, it will call "closeDropdown" every time a click event:
image
Add I added the selection code for this function to avoid useless call:
image

As you can see in line 108, it will always focus on the toggle element whenever you click if we do not check the open status.

Thanks,
Alex


wesleycho:
I didn't understand this when I first looked at this, but the explanation is very good - thank you for this PR, this does sound like a good fix.

PR #6424, opened 2017-01-24
feat(collapse): The CSS classes are now configurable via constant

chrismbarr:
Now this is configurable just like the other modules. Hard-coded CSS classes are now possible to be changed.


wesleycho:
This is a nice feature - good work!

src/collapse/test/collapse.spec.js
```diff
@@ -73,6 +77,9 @@ describe('collapse directive', function() {
     scope.$digest();
     $animate.flush();
     expect(element.height()).toBe(0);
+    expect(element).toHaveClass(collapseConfig.collapseClass);
```
wesleycho: I think this is not a useful enough set of expectations in the tests - we need to have a test case where the configuration value is unchanged, and a test case where the configuration value is changed. This needs to be done in two separate tests.

chrismbarr:
@wesleycho tests added as requested, this is much better now, good idea!

chrismbarr:
@wesleycho Any update on this and my other similar PR's #6425 & #6426 - it's been a few months and I'd love to just be able to get these from NPM instead of having to maintain my own fork of this project just to have it work for my needs.

ChrisMBarr:
@wesleycho bumping this again, it's now been more than a year since this PR was submitted.  There are still no conflicts and I've added the changes you requested

PR #6425, opened 2017-01-24
feat(modal): The CSS classes are now configurable via constant

chrismbarr:
Now this is configurable just like the other modules. Hard-coded CSS classes are now possible to be changed.

PR #6426, opened 2017-01-24
feat(tooltip): The CSS classes are now configurable via constant

chrismbarr:
Now this is configurable just like the other modules. Hard-coded CSS classes are now possible to be changed.


chrismbarr:
@wesleycho ok, tests added but again like the modals PR #6425 I'm not sure what's going on here.  It seems like this should work, but I must not be understanding something here.  

Please advise!

src/tooltip/tooltip.js
```diff
@@ -136,7 +136,7 @@ angular.module('ui.bootstrap.tooltip', ['ui.bootstrap.position', 'ui.bootstrap.s
           'class="uib-position-measure ' + prefix + '" ' +
           'tooltip-animation-class="' + animationClass +'"' +
           'uib-tooltip-classes ' +
-          'ng-class="{ ' + openClass +': isOpen }" ' +
+          'ng-class="{ \'' + openClass +'\': isOpen }" ' +
```
chrismbarr: This change was needed to allow the configurable class name to contain characters like dashes, etc.  Otherwise angular would throw errors since t was not enclosed in a string.

Issue #6360, opened 2016-12-12
Library does not play well with Angular 1.6 ngModelOptions


wesleycho:
If a parent component is using ngModel with ngModelOptions in Angular 1.6, then the options will leak to the child using ngModelOptions (i.e. could be a datepicker, datepicker popup, or typeahead), and cause it to be set as the default to be used instead of the fallbacks of the binding specified options or the global constant specified options.

Cross-linking issue in Angular here https://github.com/angular/angular.js/issues/15497

petebacondarwin:
@wesleycho - can you provide a concrete reproduction of the problem? Then I will try to help find a solution.

wesleycho:
There are two potential issues here that I'm not sure how to address.

The first one is that `ngModelOptions.$options` is always present, so https://github.com/angular-ui/bootstrap/blob/master/src/datepicker/datepicker.js#L164-L166 would always default to the `ngModelOptionsCtrl` version - this breaks existing behavior with falling back to the other provided ways of configuring ngModelOptions in the datepicker (for that example).

The other situation deals with the inheritance aspect of the new `ngModelOptions`. If I wrap a component - for example, at work, we wrap the datepicker component and use ng-model & ng-model-options. With this new behavior, `ngModelOptions.$options` will use the options from the parent ngModelOptions if present, and not the override version if I were to provide it via config object through directive bindings.

The solution for both these situations will likely be the same, but I'm a bit mystified as to what is a good way around this that handles Angular 1.5 & 1.6 amicably.

petebacondarwin:
Is the problem that in 1.6 there is *always* a `$options` object on the `ngModel` controller?

If what you want to do is know whether someone has actually put a real `ngModelOptions` above the current `ngModel` then you could explicitly `require` the `ngModelOptions` controller.

Would that help?

wesleycho:
Hm, I guess that'd work - guess I got some refactoring to do...

Thanks for the advise!

petebacondarwin:
@wesleycho - I believe that the new inheritance of `ngModelOptions` should not be a problem for you. The way that we have implemented it means that you must "opt-in" to inheritance via use of `"$inherit"` as a property value.

Even before 1.6 it was always the case that a `ngModel` directive would "inherit" from the nearest `ngModelOptions` ancestor. What has changed in 1.6 is that you can now also inherit properties for one `ngModelOptions` directive from an ancestor `ngModelOptions` (but only via opt-in).

[...]

PR #6427, opened 2017-01-24
Update to Angular 1.6

dgsmith2:
Updating code and tests to address #6360, making angular-ui/bootstrap compatible with Angular 1.6.  These are **BREAKING CHANGES**, incompatible with Angular 1.5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,9 +30,9 @@
     "url": "https://github.com/angular-ui/bootstrap.git"
   },
   "devDependencies": {
-    "angular": "1.5.8",
-    "angular-mocks": "1.5.8",
-    "angular-sanitize": "1.5.8",
+    "angular": "1.6.0",
+    "angular-mocks": "1.6.0",
+    "angular-sanitize": "1.6.0",
     "grunt": "^0.4.5",
     "grunt-cli": "^1.2.0",
     "grunt-contrib-concat": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -30,9 +30,9 @@
     "url": "https://github.com/angular-ui/bootstrap.git"
   },
   "devDependencies": {
-    "angular": "1.6.0",
-    "angular-mocks": "1.6.0",
-    "angular-sanitize": "1.6.0",
+    "angular": "1.6.1",
+    "angular-mocks": "1.6.1",
+    "angular-sanitize": "1.6.1",
     "grunt": "^0.4.5",
     "grunt-cli": "^1.2.0",
     "grunt-contrib-concat": "^1.0.0",
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -161,18 +161,7 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
 
   this.init = function(ngModelCtrl_) {
     ngModelCtrl = ngModelCtrl_;
-
-    // grab any timezone value based on precedence
-    var timezone = ngModelCtrl_.$options.getOption('timezone') ||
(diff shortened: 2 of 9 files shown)


dgsmith2:
@wesleycho, what's the timeline for this being merged and a new release being cut?  We are still using Bower for dependencies, so I'm in search of the path of least resistance for myself to get this code into our app.

package.json
```diff
@@ -30,9 +30,9 @@
     "url": "https://github.com/angular-ui/bootstrap.git"
   },
   "devDependencies": {
-    "angular": "1.5.8",
-    "angular-mocks": "1.5.8",
-    "angular-sanitize": "1.5.8",
+    "angular": "1.6.0",
+    "angular-mocks": "1.6.0",
+    "angular-sanitize": "1.6.0",
```
wesleycho: These can all probably be bumped to 1.6.1

src/datepicker/datepicker.js
```diff
@@ -161,14 +161,24 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
...
-      datepickerConfig.ngModelOptions;
+
+    // grab any timezone value based on precedence
+    var timezone = ngModelCtrl_.$options.getOption('timezone') ||
+      ($scope.datepickerOptions.ngModelOptions ? $scope.datepickerOptions.ngModelOptions.timezone : null) ||
+      (datepickerConfig.ngModelOptions ? datepickerConfig.ngModelOptions.timezone : null);
+
+    // need to start with existing ModelOptions but also give it greatest precedence minus its default of null timezone
+    ngModelOptions = ngModelCtrl_.$options
+      .createChild(datepickerConfig.ngModelOptions)
```
wesleycho: This breaks compatibility with 1.5 I believe - we cannot have that for a minor version bump. Is there no way to made this compatible with both versions?

src/datepicker/datepicker.js
```diff
@@ -161,14 +161,24 @@ angular.module('ui.bootstrap.datepicker', ['ui.bootstrap.dateparser', 'ui.bootst
...
+      .createChild($scope.datepickerOptions.ngModelOptions)
+      .createChild(ngModelCtrl_.$options)
+      .createChild({timezone: timezone});
+
     if ($scope.datepickerOptions.initDate) {
-      self.activeDate = dateParser.fromTimezone($scope.datepickerOptions.initDate, ngModelOptions.timezone) || new Date();
+      self.activeDate = dateParser.fromTimezone($scope.datepickerOptions.initDate, ngModelOptions.getOption('timezone')) || new Date();
```
wesleycho: Ditto with `getOption`

wesleycho:
This LGTM minus the compatibility break - if this is to be merged earlier, then we absolutely cannot break compatibility yet. A mechanism for handling the ngModel.$options differences needs to be created to handle differences in 1.5 and 1.6.

package.json
dgsmith2: done

src/datepicker/datepicker.js
dgsmith2: done

src/datepicker/datepicker.js
dgsmith2: done

dgsmith2:
@wesleycho I'm pretty confident things are good to do.  As you can see, the tests past.  Locally, I also reset my version of Angular back to 1.5 and re-ran the tests.  All tests in regards to $options still passed.  The only failure, which was expected, was in regards to 1.5 adding the extra span for transclusion--I was surprised the other updated transclusion tests didn't fail.

wesleycho:
Thanks - I will take another pass over the weekend (sleep deprived currently) and if things are good, I will merge & cut a release then.

PR #6432, opened 2017-01-27
Changes First Focus Behavior if Text is Entered

### LINKED ISSUE #6431: typeahead-focus-first should select first if user has partially entered a value
In the `getMatchesAsync` function, the `activeIdx` is set to `0` if `focusFirst` is true and always to `-1` if false. I would argue that if user has already entered a partial value and then hits tab, the user would want that value to be selected. Instead, hitting tab without first hitting the down arrow results in no value being selected.

For instance, given a company field with values `['ABC Corp', 'XYZ Corp']` and `typeahead-focus-first='false'`, if a user clicks into the field, both values appear but neither is selected. If the user types in `ABC`, I think `ABC Corp` should become the selected value. It seems like this would be more intuitive then forcing the user to hit the down arrow--they could just hit `Tab` after typing `ABC`.

Current:
`scope.activeIdx = focusFirst ? 0 : -1;`

Proposed:
`scope.activeIdx = focusFirst ? 0 : inputValue && inputValue.length ? 0 : -1;`

peder:
Changes behavior of `typeahead-focus-first='false'` to select the first value if the field is not empty.

Closes #6431


icfantv:
@peder your change broke our CI build because you didn't update the relevant tests.  This PR cannot be merged until you appropriately account for, and test, your changes.

PR #6437, opened 2017-01-27
feat(modal): Added default value for controllerAs

### LINKED ISSUE #6434: Modal controllerAs with default value "$ctrl"
### Bug description:

Not a bug but a nice thing to have to be consistant with the angular way.
When a modal is open with a `controller` and no `controllerAs` attribute the value of `controllerAs` should have `$ctrl` as default value like a component controller: https://docs.angularjs.org/guide/component#comparison-between-directive-definition-and-component-definition

https://plnkr.co/edit/DJWJYoSFh0W7PZijMzax?p=preview

Angular: 1.6.1

UIBS: 2.4.0

Bootstrap: 3.3.7

bastienmoulia:
Closes #6434

Add the default value `$ctrl` for controllerAs in modal.

Change the test to not use $scope anymore.


src/modal/modal.js
```diff
@@ -828,3 +828,11 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.multiMap', 'ui.bootstrap.sta
 
     return $modalProvider;
   });
+
+var CNTRL_REG = /^(\S+)(\s+as\s+([\w$]+))?$/;
+function identifierForController(controller) {
+  if (typeof controller === 'string') {
```
wesleycho: Should use `angular.isString` here

src/modal/modal.js
```diff
@@ -828,3 +828,11 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.multiMap', 'ui.bootstrap.sta
 
     return $modalProvider;
   });
+
+var CNTRL_REG = /^(\S+)(\s+as\s+([\w$]+))?$/;
+function identifierForController(controller) {
```
wesleycho: This function and variable should be inside the service it is being used in - this will expose these as globals I believe after the build process.

src/modal/test/modal.spec.js
```diff
@@ -1064,12 +1064,12 @@ describe('$uibModal', function() {
 
     describe('controller', function() {
       it('should accept controllers and inject modal instances', function() {
-        var TestCtrl = function($scope, $uibModalInstance) {
```
wesleycho: This test should not be repurposed - two separate tests should be added, one to test the default value and one to test that it can be overridden.

bastienmoulia: I don't understand your comment, this test is for the default value, and there is another test to override this value.
Before this test use the $scope but if we add a default value the $scope variables will not work anymore.

src/modal/modal.js
bastienmoulia: ok, i was not sure for this...

bastienmoulia:
I close this pull request because I don't succeed to rebase it and my git tree was a mess...
The new clean pull request is https://github.com/angular-ui/bootstrap/pull/6438

PR #6439, opened 2017-01-29
fix(datepicker): support string in the non-HTML5 formatter

byronigoe:
support string in the non-HTML5 formatter, e.g. JSON date

### DIFF
--- a/src/datepickerPopup/popup.js
+++ b/src/datepickerPopup/popup.js
@@ -119,7 +119,7 @@ function($scope, $element, $attrs, $compile, $log, $parse, $window, $document, $
           return value;
         }
 
-        if (angular.isNumber(value)) {
+        if (!angular.isObject(value)) {
           value = new Date(value);
         }
 


src/datepickerPopup/popup.js
```diff
@@ -119,7 +119,7 @@ function($scope, $element, $attrs, $compile, $log, $parse, $window, $document, $
           return value;
         }
 
-        if (angular.isNumber(value)) {
+        if (!angular.isObject(value)) {
```
wesleycho: I think it would be better to do `!angular.isDate` here

byronigoe: Agreed

byronigoe:
For my use case of using a JSON date, I also needed to enhance the validator:
`    if (angular.isString(value)) {`
`      return !isNaN(parseDateString(value)) || value.match(/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.000Z$/);`
`    }`

Issue #6413, opened 2017-01-17
Popovers are closed even if popover-trigger is 'none'


Rouche:
### Bug description:
Is it normal that even with ```popover-trigger="'none'"``` tooltips are closed on ESC key?

Reading the code, it seems impossible to cancel this behavior.

Can an option be added? Our use case is that the user needs to read the information and click a "Dismiss" button, where a preference is saved for it to never appear again.

### Version of Angular, UIBS, and Bootstrap

Angular: 1.6.1

UIBS: 2.2.0

Bootstrap: 3.3.7

wesleycho:
This sounds like an easy feature - a PR would be good

Rouche:
Do you want to use the option ```triggers.indexOf('none')``` ? If yes, this will change the current behavior..

wesleycho:
Ah sorry - I think `triggers === 'none'` might do the job.

shmuel-web:
use none with an apostrophe
like so
```
popover-trigger="'none'"
```

PR #6442, opened 2017-01-30
Fixes #6420, #6413 handling of trigger 'none'

### LINKED ISSUE #6420: triggers.show !== 'none' in case popover-trigger="'none'" will be always "true"
Hello, found some bug in version 2.4.0. When i try set popover-trigger="'none'" i figured out comparison: triggers.show !== 'none' in case popover-trigger="'none'" will be always "true"
screenshot
 and it's bad 

```
 } else {
                triggers = getTriggers(val);
              }

              **if (triggers.show !== 'none') {**
                triggers.show.forEach(function(trigger, idx) {
                  if (trigger === 'outsideClick') {
                    element.on('click', toggleTooltipBind);
                    $document.on('click', bodyHideTooltipBind);
```

Rouche:


qwq-qwq:
How long we should wait for merge ?

PR #6479, opened 2017-03-01
Tabs - Changing from ng-class + display:none to ng-if

picheli20:
When we are using same directive inside the tab (for exemple [angular-ellipsis)](https://github.com/dibari/angular-ellipsis) which needs to do same DOM manipulation, the `ng-class` with `display:none;` compiles the JS code without a HTML.

Changing to `ng-if` we make the "directives" be activated when the user enter in the tab, not in the page.


template/tabs/tabset.html
```diff
@@ -3,8 +3,9 @@
   <div class="tab-content">
     <div class="tab-pane"
```
mmarchini: It should be class="tab-pane active", otherwise the content won't show

PR #6569, opened 2017-06-09
feat($uibModal): bind resolve properties to controller when opening modal as component

### LINKED ISSUE #6540: Resolve not bound to component controller like with UI Router
The modal documentation states:

> resolve (Type: Object) - Members that will be resolved and passed to the controller as locals; it is equivalent of the resolve property in the router.

But it behaves differently than the resolve property in UI Router:

> Resolve data can be provided to the component via the component's bindings object (for 1.3+ directives, the bindToController is used; for other directives, the scope declaration is used). For each binding declared on the component, any resolve with the same name is set on the component's controller instance. The binding is provided to the component as a one-time-binding. In general, components should likewise declare their input bindings as one-way ("<").

UI Router binds resolve properties to the controller, while UI Bootstrap passes a resolve object that holds these properties. That couples the component to UI Bootstrap and makes it only usable inside a Modal. Resolve properties should be bound to the controller as bindings like UI Router does.

austinleroy:
Bind resolve properties to controller of component opened with $uibModal.
This mimics current uiRouter usage of resolve so any component can receive
input data with $uibModal.  Previous behavior was to force component to
expose a resolve binding.

Closes #6540

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -460,7 +460,7 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.multiMap', 'ui.bootstrap.sta
           Array.prototype.forEach.call(
             Object.keys(modal.scope.$resolve),
             function(property) {
-              content.attr(property, '$resolve.' + property);
+              content.attr(snake_case(property), '$resolve.' + property);
             }
           );
         } else {
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -457,6 +457,12 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.multiMap', 'ui.bootstrap.sta
             close: '$close($value)',
             dismiss: '$dismiss($value)'
           });
+          Array.prototype.forEach.call(
+            Object.keys(modal.scope.$resolve),
+            function(property) {
+              content.attr(property, '$resolve.' + property);
+            }
+          );
         } else {
           content = modal.content;
         }


eseliger:
Would love to see that merged, so we can reuse our modal components anywhere else in applications

AnimaLupi:
Indeed, it's a much needed merge to ensure reusability...

austinleroy:
I'd like to note that this PR is usable **only for binding data to the component**.  Binding of callback functions is still not available.  If that were ever to be supported, I believe it would not be a part of `resolve`.

That being said, this would still improve the reusability of any component opened with $uibModal.

bitflower:
In which version will this be included? I could patch but it's a big project with many devs on board...

src/modal/modal.js
```diff
@@ -457,6 +457,12 @@ angular.module('ui.bootstrap.modal', ['ui.bootstrap.multiMap', 'ui.bootstrap.sta
             close: '$close($value)',
             dismiss: '$dismiss($value)'
           });
+          Array.prototype.forEach.call(
+            Object.keys(modal.scope.$resolve),
+            function(property) {
+              content.attr(property, '$resolve.' + property);
```
bitflower: It should be something like
`content.attr(property.replace(/[A-Z]/g, function(c) { return '-' + c.toLowerCase(); }), '$resolve.' + property);`
to make it respect the angular "snake case" / dash-case for bindings with caps in their name (eg. mySuperDuperBinding -> my-super-duper-binding)

bitflower (changes requested):
Hey, I ran into issues with bindings that have the camelCase and need to convert to dash-case.

See my comment.

src/modal/modal.js
bitflower: Even better would be to call the angular-internal function for this conversion like angular.snakeCase() pseudo code....

bitflower: See code here: [https://github.com/angular/angular.js/blob/master/src/Angular.js#L1886](https://github.com/angular/angular.js/blob/master/src/Angular.js#L1886)

austinleroy: Great suggestion! I'll push up changes that should address that scenario.

bitflower:
@austinleroy  Do you understand the problem here? Would love to see this merged ;-)

austinleroy:
@bitflower It looks like the Travis build failed due to a timeout when running the karma task.  The build should work correctly if another one is kicked off - I can't do it myself because I don't have write access to this repo

austinleroy:
I read that if I close the PR and reopen it, a new Travis build will kick off...

eseliger:
For the future: you can also amend your commit and force push it

PR #6570, opened 2017-06-12
fix(datepicker): remove superfluous closing tag

romanzoller:
Remove superfluous </i> closing tag.


runderworld:
Duplicate fix of #6572, which properly creates a GH issue (https://github.com/angular-ui/bootstrap/issues/6571) and attempts to close it.

PR #6576, opened 2017-06-17
Update readme.md

xemexpress:
It should be 'disabled'? Please correct me if I misunderstood sth. Thanks


runderworld:
Unfortunately this is incorrect and can be closed. Per the documentation, note the correct attribute is indeed `disable`:
```
<uib-tab index="$index + 1" ng-repeat="tab in tabs" heading="{{tab.title}}" disable="tab.disabled">
```

Issue #6568, opened 2017-06-09
Does the "src" directory need to be included in the package?


simonua:
Hello,

The _src_ directory is included in the packages on npm all the way up to the current 2.5 version. Does it need to be included alongside the _dist_ directory? I can see reasons for ease of development, getting to the source code since there aren't any sourcemaps, etc. but it does cause a significant amount of space to be used (1 MB+ in 165 files).

If _src_ is needed and cannot be excluded from the package, would it be possible to also publish a "source-less"/lite package, please? 

Thank you!

PR #6583, opened 2017-06-29
Exclude src directory

simonua:
Excluding the _src_ directory when packing/publishing to save ~1MB / 150 files in the package. #6568


simonua:
Very sheepishly threw this PR out there just to see if there's any possibility for this or whether this is a completely dumb thought and a snowball's chance in hell. If it can't be done to prevent _src_ to go into the package, I would love to learn the rationale for it as it's not obvious to me beyond the convenience of having access to the src files.

wesleycho:
This doesn't make sense to me - users do consume the `src` directory when using `require` to require specific files to include the minimum JS in their applications, this would be a major breaking change.

simonua:
Ah, that makes sense. I was thinking about this from a completely different angle and just consumption out of the _dist_ directory itself. I see your point. Thank you, @wesleycho.

PR #6589, opened 2017-07-11
The component was sending the wrong integer number when  selected thr…

jairorudas:
The component was sending the wrong integer number when  selected through the key board.

When the options in the autocomplete are more one, the component send the next option selected.

Fixed the bug sending the position minus 1 when the list option is more than 1 option.


jairorudas:
Hi! Idon`t know to do the test. Somebody can helpme. thanks!

mkdewidar:
Hi @jairorudas, if you click on `details` for the build you will be able to see which tests have failed.

For testing they make use of the Jasmine framework, you will notice the tests don't exactly have ID's or anything, that's cause they are actually defined by their description. The sentence that describes each test is actually a concatenated string made from a bunch of split up strings in the test file...kinda like a hierarchy of tests. Then again, the errors do show the file and line number so you can figure out exactly where the error is through that.

To run the test locally you will need npm and grunt installed, from there you can just call `grunt` from the command line and that should build all the code and run the tests.

Since I don't actually work on this project I have no clue what it is you were working on so can't help you with the tests themselves though I hope I've pointed you in the right direction.

Thanks 😄 ,
Mohamed

PR #6660, opened 2017-10-20
fix(modal): support bindToController with class controller

### LINKED ISSUE #6659: $uibModal bindToController not working when controller is class
When class is used as controller in a modal the scope properties are not bind to the controller.
The bug can be seen here: plunker http://plnkr.co/edit/5S4uv22eHYkYhfHA0LMx?p=preview

vidyu:
- The scope is bound to the controller and accessible in $onInit

Fixes #6659


icfantv:

PR #6661, opened 2017-10-20
docs: add missing default config options for uib-popover and uib-tooltiop

NSkelsey:
Yes this project is no longer active, but an update to the docs does not constitute a change to the library. A merge + update of the docs would be appreciated! <3


icfantv:

PR #6672, opened 2017-12-05
Remove has-error class to match other input components

dk8996:
None of the other input components have ` ng-class="{'has-error': ....}"` display change when the input is invalid.  Removing it from timepicker.


icfantv:

PR #6680, opened 2018-01-03
remove background color overwrite

bgeschka:
this breaks disabled state with bootstrap

when trying to disable a typeahead, it won't be styled correctly by bootstrap
default greyed out will not apply on typeaheads


icfantv:

PR #6703, opened 2018-04-30
Fix am-pm in differents locales

malloc32:
ui.bootstrap.dateparser

Actually  assumes that value for regext 'a' is 'PM', it does not consider different locales.

Thank you.


icfantv:

